PR #6, opened 2025-02-20
Enhancement: decentralize live migration flows

awels:
This VEP proposes to decentralize the live migration flow into distinct flows, each handling
part of the live migration. This will allow for easier management of more elaborate live migration strategies such as migrating across namespaces or clusters.

Tracking issue: https://github.com/kubevirt/enhancements/issues/24


veps/sig-compute/split-migration/split_cross_cluster.png
enp0s3, line 1: The VMI sync channel needs to be routable as well?

enp0s3 (commented):
@awels Thank you for the proposal!
What I am missing here in general is:

* VMI sync channel design: is this about enhancing the roles of existing live migration actors? shouldn't we add another actor here? how would the communication protocol look like?
* Changes in VM controller, what is the impact of having two VMIs that represent the same VM. In both the cross-namespace and cross-cluster scenarios. 
* Detailed test plan, what is required from test infra in order to run the e2e.

veps/sig-compute/split-migration/split_cross_cluster.png
awels: Correct, that is why I want to use the same network as the migration channel.

veps/sig-compute/split-migration/split-migration-vep.md
```diff
@@ -0,0 +1,234 @@
...
+
+The `migrationState` is a copy of the [VirtualMachineInstanceMigrationState](https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/types.go#L736-L793)
+
+There are a lot of fields in that state specific to either source or target. For instance:
+- TargetNodeTopology
+- TargetNodeAddress
+- TargetPod
+- SourcePersistentStatePVCName
+- SourcePod
+- SourceNode
```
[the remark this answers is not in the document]
awels: It is not an exhaustive list, I was just illustrating there are a lot of duplicate fields that start with Target and Source, and I think it would make sense to split that out into two structs, one for source and one for target.

veps/sig-compute/split-migration/split-migration-vep.md
```diff
@@ -0,0 +1,234 @@
...
+Figure 2 describes the interaction with a source and target `VirtualMachineInstanceMigration` in the same namespace
+![figure 2](split_in_namespace.png)
+
+Figure 3 describes the interaction with a source and target `VirtualMachineInstanceMigration` in different namespaces.
+![figure 3](split_cross_namespace.png)
+
+Figure 4 describes the interaction with a source and target `VirtualMachineInstanceMigration` in different clusters
+![figure 4](split_cross_cluster.png)
+
+## API Examples
```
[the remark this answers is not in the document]
awels: The actual migration is not going to be any different from the current live migration, So I am not entirely sure what you are looking for.

awels:
> @awels Thank you for the proposal! What I am missing here in general is:
> 
>     * VMI sync channel design: is this about enhancing the roles of existing live migration actors? shouldn't we add another actor here? how would the communication protocol look like?

Added more detail on how the synchronization is supposed to work. I specifically left the protocol TBD as in this phase of the design it doesn't really matter.

> 
>     * Changes in VM controller, what is the impact of having two VMIs that represent the same VM. In both the cross-namespace and cross-cluster scenarios.

The thing is there will be two VMs, one source and one target. The spec might be the same, but they are distinct VMs, and thus there will not be two VMIs for one VM.

> 
>     * Detailed test plan, what is required from test infra in order to run the e2e.

As I am not actually changing the behavior of the live migration, all the existing functional tests will still be relevant, and will indeed help me determine if I am creating any regressions or not. I am explicitly NOT testing cross cluster migrations as conceptually they are the same as cross namespace migrations, with the only difference being the network connectivity. I will be adding a few cross namespace functional tests, and a basic decentralized migration.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+_Source VirtualMachineInstanceMigration_
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstanceMigration
+metadata:
+  name: vmim-source
+spec:
+  key: <unique id>
+  connectURL: "the url to connect the sync channel"
```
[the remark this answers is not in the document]
awels: Yeah that is the one thing I didn't really like, if you forget to add the connectURL it goes into target mode. Having it be more explicit is good. Not sure about the `type` though. It is already automatically tcp + tls, what other options would one have?

fabiand:
I learned the discrimitaors help with validation.
I'd include the type, because maybe one day we want to migrate to file? But
in that case, maybe we want to call it suspend.

I'd ask Vladik for input here.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+Proposal to decentralize live migration into several flows to allow for more flexible live migration. This includes live migration to a different VM in a different namespace or cluster.
+
+## Motivation
+
+Current live migration is limited in its flexiblity because the migration is centralized. This limits live migration to the same namespace and cluster. There are use cases where it would be beneficial to live migrate between namespaces and even between clusters.
```
fabiand, line 15: Do you mind naming those almost mythical use cases that you have in mind?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+In order to support those use cases, it is needed that live migration is decentralizated into separate flows:
+
+- synchronization
+- source
+- target
+
+The synchronization flow establishes a connection between the source and target and synchronizes the `VirtualMachineInstance` status between the two.
```
fabiand, line 23: Why do we want kubevirt to estbalish the connection.
Could it help to simply allow kubevirt to send and receive a stream fro msomewhere? This would decouple the responsibilities.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+The synchronization flow establishes a connection between the source and target and synchronizes the `VirtualMachineInstance` status between the two. 
+
+The target flow sets up the receiving `virt-launcher` pod, and notifies the source entity through the `VirtualMachineInstance` status, that the migration is ready to begin.
+
+The source flow waits for the target flow to be ready and initiates the migration. Before the migration begins it generates the migration parameters which include the name of the target and any translation of fields/path/names to the target.
+
+## Goals
+
+ - Break the `VirtualMachineInstanceMigration` into two entities. With a separate flow for each entity. A `source` and a `target` flow.
```
fabiand, line 31: It's still one CR, right? Thus also one entity. But it's api is extended in order to decouple send from receive?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+ - Break the `VirtualMachineInstanceMigration` into two entities. With a separate flow for each entity. A `source` and a `target` flow. 
+ - Create a synchronization flow to allow the synchronization of two `VirtualMachineInstance` statuses.
+ - Update the `VirtualMachineInstance` migrationState to support those separate entities and to allow the virt-controller, and virt-handler to identify which entity they are processing.
+
+## Non Goals
+
+Orchestration of the building blocks provided by this proposal. A higher level orchestrator is required to make usage easier for end users.
+
+Ensuring connectivity between clusters if live migrating between clusters.
```
fabiand, line 39: Maybe create a bulletted list?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+The reason virt-launcher pods do not communicate directly with each other, is that it is not guaranteed that virt-launchers will have network connectivity. The virt-handler pods also have access to a migration network if one is defined.
+
+The migration channel in figure 1 is either the pod network, or a specially configured migration network.
+
+Figure 1 describes the current interaction between various components:
+![figure 1](migration_current_state.png)
+
+#### Proposed state
+To start a live migration one creates two `VirtualMachineInstanceMigration` resources. A `target` resource, and a `source` resource.
```
fabiand, line 71: one now has to

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+The target cluster will have a synchronization controller running in the kubevirt namespace, that has access to the migration channel (either pod network, or migration network). This controller waits for incoming synchronization connections from a source. When a `source` migration resource is created in the cluster, this synchronization controller will connect to the target synchronization and start synchronizing the `VirtualMachineInstance` status between them.
```
fabiand, line 73: What exactly will the synchronizatoin controller do?

In my limited understanding I'd assume we launch
dst: libvirt listening
src: libvirt sending to dst

I'm missing what the ctrler does.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+
+If the source and the target are in the same cluster, the connection will be inside of the synchronization controller.
+
+Once synchronization has been established, each `virt-handler` pod will interact with their respective `VirtualMachineInstance` the same way as before. The major difference being that it is possible there are two separate `VirtualMachineInstance`s.
```
fabiand, line 77: Who has created the VMI on the dst?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+##### Reconcile loop
+![figure 6](sync_controller_reconcile.png)
+
+## API Examples
+
+Two major changes to the API.
+
+#### VirtualMachineInstanceMigration
+
+A new `key` field that can be any unique string. This is an optional field. If it is omited this means it standard migration just like it exists now. The suggested unique id, is either the UID of the source VM or the VM of the target VM.
```
fabiand, line 113: What is the purpos eof the key?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+Figure 3 describes the interaction with a source and target `VirtualMachineInstanceMigration` in different namespaces.
+
+![figure 3](decentralized_cross_namespace.png)
+
+Figure 4 describes the interaction with a source and target `VirtualMachineInstanceMigration` in different clusters
+
+![figure 4](decentralized_cross_cluster.png)
+
+#### Synchronization controller
+The synchronization controller allows one to synchronize two distinct and separate `VirtualMachineInstance`s. The synchronization controller is a separate controller that has access to modify `VirtualMachineInstance`s and can connect to the migration network.
```
fabiand, line 94: It sounds like - but I'm missing the desctription  - as if the ctler is creating the target VMI.

But the target VMI depends on many things - NADs, DVs possibly, secrets, etc - who is creating those?
I assume somebody.

Now, why can "somebody" not also create the VMI?

If "somebody" (I guess the orechstrator) is responsible for creating - effectively - a working VM on the dst, including it's dependencies, then there is a clear cut.

Orch to create everything on the cluster (API), and libvirt to really own the migratoin.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: The virt-controller, based on the specified VM. But instead of 'starting' the VM, we 'start' the VM in a special receiver mode (identical to what a normal live migration does now, essentially a paused start).

So the flow is:
1. Orchestrator creates the VM
2. Orchestrator creates the 'target' VMIM
3. virt-controller sees the VMIM and starts the VMI in receiver mode, but doesn't start a pod yet (don't know where we are going to schedule it)
4. The synchronization controller sees the VMIM an waits for synchronization to happen using that key.
5. Once the synchronization has happened, we have enough information to start the receiver pod on a particular node (cpu type/node affinity/etc)
6. Update VMI to indicate we are ready to receive
7. The synchronization happens the other direction and now the source knows it can start the migration.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: The key is so we can unique identify the migration. Assuming the synchronzition controller handles multiple migrations at once.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: Because that somebody (orchestrator) is creating the VM and associated resources. And yes it should be a working VM, it is just stopped. If you don't migrate you should be able to start the VM (except that the disks are blank so it can't start). It can't start the VMI because it doesn't have enough information to properly start the VMI. In particular the node selector/node anti affinity/things of that nature. This is where the synchronization comes in. The synchronization makes it so the effective migration is not any different from what exists now. But the synchronization happens over a network connection instead of directly to the VMI

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: The synchronization controller makes it so we don't have to change the logic of how migrations work. Right now as explained in this document, we pass information between the source and target through the VMI. But with this proposal we end up with two different VMIs, so we can't just directly write to a VMI and expect the other side to see it. Or can we? Well that is the purpose of the synchronization controller. To synchronize the VMIs so we can pass information back and forth like we do directly with the VMI now.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: I am not quite sure what you mean here. The migration happens deep in the bowels of kubevirt. And thus it should be the one creating the connections as it has all the needed information to do so. You want some outside entity creating these connections and somehow passing them to kubevirt? I am not sure I follow how that would be possible.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: Correct, but how the controller handles the entities will be two completely separate flows. They can be processed completely independent of each other and in parallel if needed. The flow right now is very intertwined.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
vladikr: @awels It's a valuable feedback. Maybe the proposal should explicitly highlight this brief flow explanation at the beginning so it will be easy to focus.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,279 @@
...
+      vmiUID: <virtual machine instance UID>
+      namespace: <namespace of virtual machine instance>
+      syncAddress: <address to connect for sync channel>
+```
+
+To keep backward compatibility if there is no `key`, it can be assumed it is both and it can follow both source and target flows in the `virt-controller`
+
+## Alternatives
+
+Completely rework the mechanism used for live migrations to be decentralized from the start. As there are quite a few issues with the current mechanism. This was rejected due to the amount of time and effort it would take to do a complete rewrite.
```
vladikr, line 227: Is there a well defined alternative approach? 
If such an approach exists perhaps we should at least describe it here briefly.

You could also mention here why some of the components, for example, the synchronizer live outside of core kubevirt.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: Also VMI creation is normally the job of the kubevirt controller. Except when creating a stand alone VMI without a VM. However you cannot do this live migration without a VM resource on the target because of how storage live migration works. For the elusive use cases we will need to do a storage live migration.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: I reworded it a bit, I like the decoupling verbage.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: I added some verbage earlier in the proposal to explain the role of the synchronization controller, before we get to the synchronization section.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: I am not aware of a well defined alternative approach. Besides maybe a complete redesign of the entire migration mechanism. But doing that alone is going to be a large undertaking. I can definitely explain the reasoning for having a separate migration controller vs having it part of lets say the virt-handler.

awels:
@vladikr @enp0s3 @iholder101 @jean-edouard I updated the VEP with the discussion we had about how to deal with the target virtual machine and how the virt-controller should create the receiver VMI.

jean-edouard (approved):

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
+# VEP #Decentralized live migration
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
```
iholder101, line 7: nit: you can mark this as done.

nit2: can you add a link to the tracking issue in the PR description?

iholder101 (commented):
Thank you very much @awels for this VEP. It's very detailed and it's clear you've put a lot of effort into it!

Although there are remaining open questions, from my side this is ready for an Alpha stage.

Another note: since this VEP is heavy and related to storage and network as well, I'd appreciate if a representative from those SIGs could LGTM before this gets merged.

/lgtm

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: I have updated the description to include a link to the tracking issue. For the first nit, I don't think I can do that until the VEP is merged, so I can then link the issue to the location. Maybe I am not understanding the process correctly.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+  - Keep the same migration resources unmodified. This will ensure there are no regressions during the untangling.
+  - Possibly split the `VirtualMachineInstance` migration state into two states if necessary. If it not necessary this becomes part of the second phase.
+- Second modify the migration resources to have the new fields as proposed. And if not modified already, split the `VirtualMachineInstance` migration state into two states.
+  - Add functional test that mimics a basic migration functional but with the separated migration resources.
+- Third implement the synchronization controller and associated logic. The controller is not need in the first two phases, but once migrating across namespaces some synchronization is required.
+- Fourth Implement the logic for cross namespace migration. This will require adding fields to the migration state to properly handle the flow.
+  - Add a functional test to do a basic migration across namespaces.
+- Fifth To simulate migration across clusters, add a functional test that switches network and storage when migrating between namespaces.
+
+## Feature lifecycle Phases
```
alicefr, line 296: Can we really put a feature gate on this? I mean, it will refactor the code so deeply that I'm not sure we can really separate it with a feature gate. Or do we plan to have a code duplication, like 2 files with the 2 version of the code? I know, I'm going into technical details but I really would like to understand how this can be feature gated

awels: So the feature gate would be to enable the 'synchronization-controller' and allow the split VirtualMachineInstanceMigration resources. Throughout this entire process we are aiming to keep existing migration flows as they are, and thus that would not be subject to the feature gate.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+
+The reason virt-launcher pods do not communicate directly with each other, is that it is not guaranteed that virt-launchers will have network connectivity. The virt-handler pods also have access to a migration network if one is defined.
+
+The migration channel in figure 1 is either the pod network, or a specially configured migration network.
+
+Figure 1 describes the current interaction between various components:
+![figure 1](migration_current_state.png)
+
+#### Proposed state
+To start a live migration one now has to create two `VirtualMachineInstanceMigration` resources. A `target` resource, and a `source` resource. Additionally in order to migrate a special `VirtualMachine` resource has to be created on the target which is copy of the source `VirtualMachine` but it has a special `runStrategy` indicating it is the receiving `VirtualMachine`. The virt-controller will create a special receiving `VirtualMachineInstance` that can be referenced by the `target` `VirtualMachineInstanceMigration`.
```
alicefr, line 73: I'm not a big fan of adding this to the runStrategy. Once, the migration has completed then you need to switch the runStrategy to something else, who is going to to this?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+
+The source `VirtualMachine` will reference some disks and networks. Whatever creates the target `VirtualMachine` will be responsible for creating blank receiver disks, and properly mapping the networks from the source `VirtualMachine` to the target `VirtualMachine`
+
+When migrating inside the same namespace, the additional `VirtualMachine` and associated resources are not needed as they already exist in the namespace.
+
+The `source` resource will contain a `connectionURL` that points to the synchronization controller at the target. The `source` resource will also contain a key which has to be unique across all migrations. The `target` resource will contain the same key. This will allow the synchronization controller to map the `VirtualMachineInstance` it receives on the synchronization channel to the appropriate `target` `VirtualMachineInstance`. Once the synchronization has been established, any status updates can be communicated between the `source` and the `target` just like in the current state. This means enough information has been exchanged to start the receiver pod, and once the receiver pod is ready, the migration can start.
```
alicefr, line 85: why do they need to be in the VMIM? Isn't this information hidden from the user?

alicefr: Can you please add the details what the key is for?

alicefr:
The proposal is great, what scares me is to put this into practice. The migration code has proven to be buggy already, adding additional layers and synchronizations can become quickly a nightmare. I belive we need to have a solid plan in order to implement this

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
alicefr: I will throw here the idea, even if it is pretty ugly. If we could have for a while some code duplication, and the feature gate makes the switch between the 2 code versions, and then we gradually integrate the new code into the old one when we are more secure.

I'm well aware that this double(or triple) the reviewer work.
@vladikr wdyt?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: The virt-controller, and we even know what the runStrategy needs to be, because we can take it from the original VMI.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: > why do they need to be in the VMIM? Isn't this information hidden from the user?
We do not anticipate users interacting with these low level building blocks directly. We have an orchestrator that the users will interact with. The orchestrator will be the one that creates all the resources, including the VMIM. And it can look up the synchronization channel endpoint. This is the one piece of information that needs to be provided to the source, so we can establish the synchronization channel.

> Can you please add the details what the key is for?
On line 153, I give some details. It is just an identifier so we can map the source to the target.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
awels: So regular migrations will not change from a user perspective. Only if we want to do the more elaborate migrations will we see a difference. We can have confidence that migrations still work because of the test suite catching anything that gets broken with the new code.

EdDev (commented):
Thank you for this detailed design.

My understanding is that the current proposal is based on the existing migration solution, which by itself is very complex.
In my view, these new requirement are an opportunity to re-examine the current solution and put it in comparison with the extensions proposed here.

My main concern is about the complexity of all the different scenarios we will need to support and consider when developing individual features which may need special treatment during migration.
IMO we need to build a better migration flow which takes into considerations the fact that the target is not assured to have the same characteristics as the source.
E.g. ref for the same NAD name that exists, not colliding with mac or IP addresses, etc.

As I raised in the inline comments, I also am a bit concerned with too much logic landing in the kubevirt core project when the expected usage is unknown and potentially very low. If we could extract most of the logic out of the core, it would be beneficial to the maintainers and the majority of users.

I probably missed some offline discussions on this, but I would be interested to have several meetings to brain storm about alternatives and/or ways to limit the permutations in the core project.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+
+The reason virt-launcher pods do not communicate directly with each other, is that it is not guaranteed that virt-launchers will have network connectivity. The virt-handler pods also have access to a migration network if one is defined.
+
+The migration channel in figure 1 is either the pod network, or a specially configured migration network.
+
+Figure 1 describes the current interaction between various components:
+![figure 1](migration_current_state.png)
+
+#### Proposed state
+To start a live migration one now has to create two `VirtualMachineInstanceMigration` resources. A `target` resource, and a `source` resource. Additionally in order to migrate a special `VirtualMachine` resource has to be created on the target which is copy of the source `VirtualMachine` but it has a special `runStrategy` indicating it is the receiving `VirtualMachine`. The virt-controller will create a special receiving `VirtualMachineInstance` that can be referenced by the `target` `VirtualMachineInstanceMigration`.
```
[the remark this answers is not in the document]
awels: > The suggestion to split VMIM into two raises the question of why have it at all. E.g. if we need two VMIMs and two VMIs, then we can just have the VMIs without the VMIM.

I am not sure I understand, a VMI is a representation of a running (or about to run) VM. VMIs are namespaced resources, and if we go across namespaces or cluster, we will need 2 of them. At the same time we need some mechanism to initiate a live migration. And we need a mechanism to tell the receiver end to be ready to receive an incoming migration. I don't see how you would do that without a VMIM.

> If we are to handle the migration between namespaces in the same cluster, we could define a cluster VMIM object and a single one would be suffice. The fact that it needs to be functional for inter-cluster scenarios, adds a different complexity.

That is why we need 2 VMIMs, that way there is no technical difference between inter namespace, and inter cluster migrations. The only difference is the networking connectivity between the source and target, everything else is identical. Making cluster scoped resources will not solve the inter cluster issue.

> It would be nice to handle each scenario separately, so each will make sense and at the same time for all scenarios to work in the same flow (including the current in-namespace migration).

Why would we make different flow for different scenarios when the same flow for all scenarios works just fine? It sounds much more complex to me to have different flows for different scenarios.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+      vmiUID: <virtual machine instance UID>
+      namespace: <namespace of virtual machine instance>
+      syncAddress: <address to connect for sync channel>
+```
+
+To keep backward compatibility if there is no `key`, it can be assumed it is both and it can follow both source and target flows in the `virt-controller`
+
+## Alternatives
+
+Completely rework the mechanism used for live migrations to be decentralized from the start. As there are quite a few issues with the current mechanism. This was rejected due to the amount of time and effort it would take to do a complete rewrite.
```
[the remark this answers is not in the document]
awels: > IMO this is a weak reason to defer investment on alternatives.
The complexity of the current suggested solution better be balanced by other alternatives to assure we are not going in the wrong direction.
It is only reasonable to re-evaluate the existing components once so "heavy" requirements have been presented.

I disagree, we could argue over the 'best' solution for years, while making no progress. We only get so much time to implement a solution for the problem, and we have discussed it at length and this is what we came up with. This VEP is purely about enhancing the current mechanism to allow decentralized migration. The goal is NOT to completely rework the migration mechanism.

> E.g. duplication of VMIM & VMI hints only may be suffice.
The different scenarios which need to be supported are a heavy burden on maintenance, for all SIGs. If we had one way to do the migration, no matter the target location, we could lower the maintenance cost.

This is EXACTLY what is presented here. I obviously didn't do a good job presenting that if that is not what you came away with when you read the VEP. The entire goal is to have a single flow for everything, regardless of where the target is. Now there is a need for some extra information if the target is a different namespace/cluster, and that is represented by the new fields in the VMIM.

> I am also unsure how common this scenario is in practice. E.g. if the need is focusing on 2% of the users, then I would expect to see most of the support be placed outside the core Kubevirt project. Placing in the core only the essentials to allow a migration to occur, taking all the rest out.

This VEP is only about the mechanism of migrating, not the orchestration. Which is specifically called out in the non goals. The mechanism assumes there is some outside orchestrator that configures all the per-requisites before attempting to do the migration. So I am not sure what you mean by the 'rest'?

awels:
> My understanding is that the current proposal is based on the existing migration solution, which by itself is very complex.
In my view, these new requirement are an opportunity to re-examine the current solution and put it in comparison with the extensions proposed here.
 My main concern is about the complexity of all the different scenarios we will need to support and consider when developing individual features which may need special treatment during migration.
IMO we need to build a better migration flow which takes into considerations the fact that the target is not assured to have the same characteristics as the source.
E.g. ref for the same NAD name that exists, not colliding with mac or IP addresses, etc.
As I raised in the inline comments, I also am a bit concerned with too much logic landing in the kubevirt core project when the expected usage is unknown and potentially very low. If we could extract most of the logic out of the core, it would be beneficial to the maintainers and the majority of users.
I probably missed some offline discussions on this, but I would be interested to have several meetings to brain storm about alternatives and/or ways to limit the permutations in the core project.

I think you misunderstand the scope of the VEP. We are purely talking about the low level mechanism of migrating, not the overall orchestration of a migration. That is explicitly left to a higher level entity. We are just providing the primitives that the orchestrator can use to do the migration.

As I have said in the comments, the entire goal is to have a single flow for all combinations of targets, with the existing flow just being a special case of the generic flow. It will still follow the same flow though.

As for the usage being low, this is very heavily requested feature, so I expect there to be a lot of use for this. In particular the cross cluster version, but I have also seen requests for cross namespace migrations.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
alicefr: The spec should the part that the users modify. I feel that this information of the state of the VMI should be in the status not in the spec

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
alicefr: Similar comment as for the runStrategy, why in the spec and not in the status? If you put in the spec it means that the key is generated by the overlay orchestra tor, not by KubeVirt. Is this the case?

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,308 @@
...
+
+If the source and the target are in the same cluster, the connection will be inside of the synchronization controller.
+
+Once synchronization has been established, each `virt-handler` pod will interact with their respective `VirtualMachineInstance` the same way as before. The major difference being that it is possible there are two separate `VirtualMachineInstance`s.
+
+TBD: How to handle the source `VirtualMachine` once the migration has completed. We can shut down the source `VirtualMachine`, delete it, or keep it running. For now a good starting point is to simply shut down the `VirtualMachine` and something else decides what to do with the `VirtualMachine`
+
+Figure 2 describes the interaction with a source and target `VirtualMachineInstanceMigration` in the same namespace
```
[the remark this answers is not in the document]
EdDev: I guess it is a matter of context. From the top level, I think it is a different flow, e.g. one needs to "clone" a VMIM/VM/VMI and not consider a secondary pod while another that does not.
IIUC the low-level implementation is following the same flow.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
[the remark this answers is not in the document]
EdDev: > This VEP is only about the mechanism of migrating, not the orchestration. Which is specifically called out in the non goals. The mechanism assumes there is some outside orchestrator that configures all the per-requisites before attempting to do the migration. So I am not sure what you mean by the 'rest'?

Yes, I see it as a challenge to review this separately from the overall solution (i.e. with the orchestration).
The mechanics of the migration is coupled with some assumptions which inter-space and inter-cluster break. As mentioned in this text, the VM references and depends on external resources that control the availability of a VM to migrate or be able to operate (recover) on the target.
If we can ignore these details and focus only on a specific layer, then I do not see any core issue. However, at the same time, I do not know if it will allow solving the original need e2e.

EdDev:
> I think you misunderstand the scope of the VEP. We are purely talking about the low level mechanism of migrating, not the overall orchestration of a migration. That is explicitly left to a higher level entity. We are just providing the primitives that the orchestrator can use to do the migration.

Yes, I was thinking on the overall goal and tried to understand how this proposal will help solve it.
I found it challenging when thinking about the network aspects. I guess that the reason is related to the fact that the network part needs also to consider the domain itself and not only the migration method. I.e. how the domain continues operating on the target.

Thank you for refocusing me here.

alicefr:
My questions have been answered. I'm fine with the approach. Separate the migration from the regular VM life-cycle will be a great improvement. I'm really looking forward to see this in place.
/lgtm

iholder101:
> Another note: since this VEP is heavy and related to storage and network as well, I'd appreciate if a representative from those SIGs could LGTM before this gets merged.
> 
> /lgtm

/hold

Let's ensure a representative from each SIG LGTMs this.
@EdDev PTAL
@mhenriks @alicefr @akalenyu @ShellyKa13, can one of you please review?

EdDev (commented):
Thank you for this comprehensive VEP.

From a network perspective, we can split it into two challenge classes:
1. The connectivity between the two synchronizers on the pod or migration network.
2. The guest network wiring and its ability to connect to the network and preserve connectivity.

The first is relevant to inter cluster connectivity. It is not dependent or related on Kubevirt core networking, but on external orchestration of the network connectivity.

The second part depends on the existing core network steps and on the orchestration that assures the relevant resources exist on the target. Per my understanding, because the underlying behavior at the virt-handler and virt-launcher do not change, any possible issues are reflected to the source and therefore will result with the same outcome as today.

Regarding the e2e test coverage: It makes sense IMO to pick some migration tests from the sig-network and execute them in the relevant scenario e.g. (inter-namespace).

Therefore, form a sig-network perspective, I think we are good.
/lgtm

---

From a higher level view of this proposal, I would be interested to see another concrete alternative, even if it may not be chosen due to some limitation (e.g. time, complexity, etc). I think it is important because 1) we need a backup plan in case unforeseen challenges are discovered and 2) this change is an opportunity to rethink several decisions taken in the project early days, which may merit an update.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,311 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstanceMigration
+metadata:
+  name: vmim-target
+spec:
+  receive:
+    key: <unique id>
+  vmName: vmi-name
+status:
+  syncEndpoint: <address:port>
```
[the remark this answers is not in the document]
awels: So the endpoint will be published somewhere (current thinking is maybe the KubeVirt CR) This endpoint will be the same per cluster. From that endpoint an actual new connection will be spawned very similar to any other client/server configuration.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,311 @@
...
+
+Upon creation of `VirtualMachineInstanceMigration` resources, the target `virt-controller` generates a specialized `VirtualMachineInstance`. This instance is not subject to typical interactions with the `virt-controller` or `virt-handler`, nor synchronized with its source counterpart. The synchronization controller assumes responsibility for synchronizing this receiver `VirtualMachineInstance` with others it receives over the synchronization channel, based on the shared keys in the `VirtualMachineInstanceMigration`s.
+
+The synchronization controller empowers the decentralization of two distinct and separate `VirtualMachineInstance`s. It operates as a standalone controller that can modify `VirtualMachineInstance` state and connects to the migration network.
+
+A single active synchronization controller per cluster is designated, listening on an established address port combination. TBD: how this address will be advertised (perhaps through a kubevirt CR?). Since this synchronization controller should connect to the migration network, it cannot be part of the `virt-controller` pod. This means a distinct synchronization pod.
```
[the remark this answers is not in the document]
awels: In theory we could add the migration network to the virt-controller pod. But honestly I don't think it makes much sense except to force the synchronization controller into the virt-controller pod. Sure it will save us from one more pod, but I don't think it is that terrible to have an extra pod.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,311 @@
...
+
+The target cluster will have a synchronization controller running in the kubevirt namespace, that has access to the migration channel (either pod network, or migration network). This controller waits for incoming synchronization connections from a source. When a `source` migration resource is created in the cluster, this synchronization controller will connect to the target synchronization and start synchronizing the `VirtualMachineInstance` status between them.
+
+If the source and the target are in the same cluster, the connection will be inside of the synchronization controller.
+
+Once synchronization has been established, each `virt-handler` pod will interact with their respective `VirtualMachineInstance` the same way as before. The major difference being that it is possible there are two separate `VirtualMachineInstance`s.
+
+TBD: How to handle the source `VirtualMachine` once the migration has completed. We can shut down the source `VirtualMachine`, delete it, or keep it running. For now a good starting point is to simply shut down the `VirtualMachine` and something else decides what to do with the `VirtualMachine`
```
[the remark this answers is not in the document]
awels: Right, we would shut down the source (gracefully), and then leave it up to the orchestrator or user to determine what to do with the source. I don't see an issue that someone could decide they want to start it again on the source, while a copy is running on the target. They are independent entities at that point.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,311 @@
...
+
+![figure 3](decentralized_cross_namespace.png)
+
+Figure 4 describes the interaction with a source and target `VirtualMachineInstanceMigration` in different clusters
+
+![figure 4](decentralized_cross_cluster.png)
+
+## Synchronization controller
+
+The synchronization controller facilitates the synchronization of migration state between source and target `VirtualMachineInstance`s. This enables the `virt-controller` to interact with its `VirtualMachineInstance` as if it were the sole instance, maintaining consistency in proposed migration flows from the `virt-controller`'s perspective.
```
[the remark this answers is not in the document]
awels: Right when I say 'virt-controller' I mean vmi controller.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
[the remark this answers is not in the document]
EdDev: I'm not against or in favor of this option, just note that another pod means another deployable component that the operator needs to know about and the build system needs to prepare and publish.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
```diff
@@ -0,0 +1,311 @@
...
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+#### Current state
+To start a live migration one creates a `VirtualMachineInstanceMigration` resource in the cluster, which references the `VirtualMachineInstance` you want to migrate. This then creates a _receiver_ pod in the same namespace, and starts the migration process.
+
+The actual migration channel is proxied through the `virt-handler` pods that are running on the source and target nodes. Any migration state information is passed through the `VirtualMachineInstance` migrationStatus. This is because `virt-handler` pods can only read/modify `VirtualMachineInstances`.
+
+The reason virt-launcher pods do not communicate directly with each other, is that it is not guaranteed that virt-launchers will have network connectivity. The virt-handler pods also have access to a migration network if one is defined.
```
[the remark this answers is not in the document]
awels: Right, we could definitely say that it only works if you use a particular network binding and at that point we can assume the virt-launcher can communicate with other virt-launchers to do migrations. Something to consider for an alternative.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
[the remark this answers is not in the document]
awels: Right, and if not defined, KV will generate a port to use and show it in the kubevirt CR status (that is the idea right now anyway)

awels:
I want to merge this PR, and then immediately create a follow up PR that addresses the comment by Ed. And will have some more time to debate things in that PR since we are tracking this VEP once this PR is merged.

veps/sig-compute/decentralized-migration/decentralized-migration-vep.md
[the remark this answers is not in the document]
EdDev: Well, it opens up a possibility that was not there before.
If that would be possible, we will have a live-clone, which may be an interesting feature by itself.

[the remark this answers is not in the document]
awels: Right I am not entirely sure how useful that would be, but it would be possible.

awels:
follow up PR #44

PR #8, opened 2025-02-25
Integrate NIC Hotplug with LiveUpdate Rollout Strategy

orelmisan:

Fixes #

**Checklist**

- [ ] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note
NONE
```

### DIFF
--- a/veps/sig-network/ns-admin-nichotplug.md
+++ b/veps/sig-network/ns-admin-nichotplug.md
@@ -1,0 +1,272 @@
+# VEP #29: Integrate NIC Hotplug with LiveUpdate Rollout Strategy
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+Currently, the NIC [hotplug / hotunplug feature](https://github.com/kubevirt/community/blob/a4eb40ce2c0372a49c39e8289fb3d18ca3f30512/design-proposals/nic-hotplug/nic-hotplug.md)
+depends on a user-initiated migration following each NIC hotplug / hotunplug operation.
+The migration is required in order to attach / detach secondary interfaces to / from the target virt-launcher pod.
+
+> [!NOTE]
+> KubeVirt also supports in-place hotplug / hotunplug operations (without performing a migration)
+> by integrating with the [dynamic-networks-controller](https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller).
+>
+> This method does not support hot-plugging SR-IOV NICs, as SR-IOV hotplug operation requires a migration in order for the
+> [SR-IOV device plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin) to be invoked when the
+> target pod is created.
+
+PR [kubevirt/kubevirt#13497](https://github.com/kubevirt/kubevirt/pull/13497) removed Namespace administrators'
+permissions to create VirtualMachineInstanceMigration objects - thus breaking the NIC hotplug / hotunplug feature
+for them in v1.5.
+
+## Motivation
+
+In previous KubeVirt versions, a namespace admin could hotplug and unplug NICs.
+We want to ensure that this feature will keep working in v1.5 and future versions and enhance it so the migration will be
+automatically performed - without user intervention.
+
+## Goals
+
+- Integrate the NIC hotplug / hotunplug with the LiveUpdate rollout strategy for consistent user experience.
+- Enable namespace admins to perform NIC hotplug / unplug operations independently - without involving the cluster-admin 
+for each hotplug / hotunplug operation.
+
+## Non Goals
+
+## Definition of Users
+
+- Namespace administrators
+
+## User Stories
+
+As a namespace admin I would like to be able to:
+- Hotplug / unplug a secondary interface without involving the cluster admin.
+- Hotplug a new secondary network interface that uses bridge binding.
+- Hotunplug a secondary network interface that uses bridge binding.
+- Hotplug a new secondary network interface of type SR-IOV.
+- Use the staging rollout strategy so any hotplug / hotunplug operation will take effect only after restarting a VM.
+
+## Repos
+
+- kubevirt/kubevirt
+
+## Design
+
+### Integrate with LiveUpdate Rollout Strategy
+
+The CPU and memory hotplug features use the `LiveUpdate` [rollout strategy](https://kubevirt.io/user-guide/user_workloads/vm_rollout_strategies/)
+which automatically initiates a migration of the VM when certain spec fields are changed.
+
+Extend the feature to automatically initiate a live migration in case of NIC hotplug / hotunplug request.
+
+The existing in-place hotplug / hotunplug logic using the dynamic-networks-controller will be preserved. 
+
+#### API Additions
+
+Currently, the CPU, memory and volume hotplug / hotunplug flows have a dedicated condition on the VMI object to mark the VMI
+for automatic migration:
+- HotVCPUChange
+- HotMemoryChange
+- VolumesChange
+
+Since we do not wish to introduce another resource-specific condition, a new generic allowed value will be added to `VirtualMachineInstanceConditionType`:
+
+```go
+VirtualMachineInstanceMigrationRequired VirtualMachineInstanceConditionType = "MigrationRequired"
+```
+
+For additional details, please see [Rational for Adding a Shared Condition](#appendix---rational-for-adding-a-shared-condition).
+
+#### Migration Based Hotplug / Hotunplug Flow
+
+1. The VM owner adds one or more new secondary vNICs and matching networks, or sets their state to `absent` to / on a running VM.
+2. The VM controller will:
+
+   - Not add the `RestartRequired` condition to the VM object.
+   - Patch the VMI object with the relevant vNICs and networks.
+3. The VMI controller will:
+   
+   - Patch the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`) - since the dynamic-networks-controller is not present, this will not have an affect.
+   - Update the VMI interface status.
+   - In case an SR-IOV NIC was hot-plugged - add the `MigrationRequired` condition to the VMI with its status set to `True` and a timestamp.
+   - Else - add the `MigrationRequired` condition to the VMI with its status set to `False` and a timestamp.
+4. On a future iteration of the VMI controller's reconciliation loop, the VMI controller will update the `MigrationRequired` condition status to `True` in case:
+
+   - The VMI has the `MigrationRequired` condition set with its status set to `False`.
+   - More than ten seconds had elapsed since the condition was set on the VMI (in order to allow the dynamic-networks-controller to operate).
+5. The Workload update controller will:
+   
+   - Identify the VMI as a candidate for automatic migration - since it has the `MigrationRequired` condition with its status set to `True`.
+   - Create a VMIM object for it.
+6. The Migration controller will remove the `MigrationRequired` condition once the migration is complete.
+7. The vNICs will be hot-plugged / hot-unplugged following the migration - same as today. 
+
+#### In-Place Hotplug / Hotunplug Flow
+
+1. The VM owner adds one or more new secondary vNICs with bridge binding and matching networks, or sets their state to `absent` to / on a running VM.
+2. The VM controller will:
+
+   - Not add the `RestartRequired` condition to the VM object.
+   - Patch the VMI object with the relevant vNICs and networks.
+3. The VMI controller will:
+
+    - Patch the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`).
+    - Update the VMI interface status.
+    - Add the `MigrationRequired` condition to the VMI with its status set to `False` and a timestamp.
+4. The dynamic-networks-controller will:
+    
+    - Observe the change to the Multus network attachment annotation on the virt-launcher pod.
+    - Invoke Multus.
+5. On a future iteration of the reconciliation loop, the VMI controller will:
+
+   - Observe a change in the VMI interface statuses.
+   - Remove the `MigrationRequired` condition.
+6. The virt-handler will set up / remove the necessary plumbing.
+7. The virt-launcher will attach / detach the relevant interfaces to / from the domain.
+
+> [!NOTE]
+> SR-IOV hotunplug is currently unsupported.
+
+> [!NOTE]
+> The time it takes for:
+> 1. The dynamic-networks-controller to observe the change and invoke Multus.
+> 2. Multus to invoke its delegate CNIs.
+> 3. Delegate CNIs to perform their job.
+> 4. Multus to report its status on the pod's annotations.
+> 5. VMI controller to observe and report the interface statuses.
+> 
+> Is unbounded and may be longer than the timeout defined before an automated migration is requested.
+> It may cause the VMI to migrate even if an in-place operation took place.
+
+
+#### Pros
+
+- Improves the current user experience
+- Will work for hotplugging interfaces using bridge and SR-IOV bindings
+- Keeps the integration with the dynamic-networks-controller
+
+#### Cons
+ - Will not work for non-migratable VMs when SR-IOV NICs are involved or when the dynamic-networks-controller is not deployed 
+ - Performs a migration for every hotplug / hotunplug when the dynamic-networks-controller is not deployed
+ - Requires a minimal API change
+ - Requires the cluster-admin to configure the RolloutStrategy to `LiveUpdate`
+
+## API Examples
+
+<!--
+Tangible API examples used for discussion
+-->
+
+## Alternatives
+
+### Using (only) the dynamic-networks-controller
+
+The dynamic-networks-controller is a controller, running as a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/),
+listening to changes on Pods and invoking the [Multus v4 CNI](https://github.com/k8snetworkplumbingwg/multus-cni).
+
+It enables NIC hotplug / hotunplug operations in place, e.g. without creating another pod.
+A known limitation of the controller above is that it cannot hotplug interfaces that require invocation of
+[Kubernetes device plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/),
+for example: SR-IOV network interfaces.
+
+#### Pros
+
+- Improved user experience
+- No need for migration when attaching/detaching interfaces using bridge binding
+- Works on non-migratable VMs
+- Does not require changes to kubevirt/kubevirt
+
+#### Cons
+
+- The dynamic-networks-controller is an external component that needs to be deployed and managed
+- Not applicable for hotplugging SR-IOV interfaces
+
+### Providing Namespace Admins the required RBAC Rules
+
+Cluster admins could provide Namespace admins the permissions to create VMIM objects.
+The documentation will be updated to reflect the fact that there are additional RBAC rules missing for namespace
+admins to be able to perform NIC hotplug / unplug.
+
+#### Pros
+
+- Does not require code changes
+
+#### Cons
+
+- Losses the benefits of limiting VMIM creation to cluster admins
+- Requires cluster-admins to take action
+
+## Scalability
+
+N/A.
+
+## Update/Rollback Compatibility
+
+Users will have to configure their KubeVirt configuration similarly to CPU and memory hotplug in order to enable the automatic
+migration.
+
+## Functional Testing Approach
+
+The migration-based hotplug / hotunplug tests will be adjusted to remove the manual VMIM object creation.
+The in-place hotplug / hotunplug tests will be adjusted to assert that a migration did not occur.
+
+## Implementation Phases
+
+1. Support automatic migration on NIC hotplug / hotunplug (without dynamic-networks-controller).
+2. Fix the redundant `RestartRequired` condition added to VMs when NIC hotplug / hotunplug are performed.
+
+## Feature lifecycle Phases
+
+### GA
+Planned for v1.6.
+Since the LiveUpdate rollout strategy and NIC hotplug / hotunplug features are already GAed, their integration does not
+pose a significant risk.
+
+This integration could also be considered as a bug fix to restore namespace admins' ability to perform NIC hotplug / hotunplug.
+
+# Appendix - Rational for Adding a Shared Condition
+
+Currently, each resource's hotplug flow (CPU. memory, volume) may add its own dedicated condition to the VMI object to 
+signal that the VMI requires an automatic migration.
+
+While this approach provides fine-grained visibility into the state of individual resources, it introduces several challenges:
+
+
+1. Maintenance burden:
+
+As more hotplugable resource types are introduced, the condition set will grow, requiring continual adjustments to any
+logic consuming it, which will increase the long term maintenance burden.
+
+2. Complexity in interpretation:
+
+KubeVirt's components must evaluate several conditions to determine whether a VMI should be automatically migrated.
+This logic can become complex, error-prone and inconsistent across components.
+
+3. Coordination overhead:
+
+When multiple resource conditions co-exist, it's unclear which one has precedence or whether they should block each-other.
+
+The proposed solution introduces a single condition `MigrationRequired` that reflects the requirement to automatically
+migrate the VMI.
+Individual resource logic can still detect their own limitations internally, but instead of setting multiple conditions,
+they would contribute to setting or clearing this unified condition.
+
+Benefits:
+
+1. Clear ownership:
+
+The responsibility for deciding whether an automatic migration is required will be centralized.
+
+2. Simplified consumption:
+
+KubeVirt's components can react to a single condition, reducing the overall logic complexity.
+
+3. Future-proofing:
+
+Adding new resource types or constraints will not require changes in multiple components - only the centralize logic will change.
+It will allow for existing and future in-place hotplug flows to evolve and co-exist with the migration-based flows. 


veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,166 @@
...
+### Integrating with LiveUpdate Rollout Strategy
+
+The CPU and memory hotplug features use the `LiveUpdate` `RolloutStrategy` which automatically initiates a migration
+of the VM when certain spec fields are changed.
+
+Extend the feature to automatically initiate a live migration in case of NIC hotplug / hotunplug request.
+
+#### Pros
+
+- Improves the current user experience
```
tiraboschi, line 61: I think it's not only about user experience and consistency.
Currently we have also conditions to track in progress hotplugs:
https://github.com/kubevirt/kubevirt/blob/eff99ed4b8490f728b9beccbe8a3d8fda998b4f0/pkg/virt-controller/watch/workload-updater/workload-updater.go#L319-L323

and they are consumed to serialize operations like in the workload-updater or in virt-contreller:
https://github.com/kubevirt/kubevirt/blob/eff99ed4b8490f728b9beccbe8a3d8fda998b4f0/pkg/virt-controller/watch/vm/vm.go#L599-L602

NIC hotplugs are probably now aliene objects bypassing all of those safety checks.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,166 @@
...
+
+## API Examples
+
+<!--
+Tangible API examples used for discussion
+-->
+
+## Alternatives
+
+### Revert the VMIM Hardening PR
```
tiraboschi, line 78: Reverting is not the only option.
At any point in time the cluster-admin could grant live migrate right to selected node admis or even consciously restore the previous behavior.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,166 @@
...
+administrators could retain their permissions to create VMIM objects.
+
+#### Pros
+
+- Relatively simple
+- Allows for additional time to consider the preferred solution
+
+#### Cons
+
+- The VMIM hardening will be postponed to a later version
```
tiraboschi, line 90: Clusters are still exposed to an easy DoS on node drains and upgrades.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: > I think it's not only about user experience and consistency. 

Agreed.
> Currently we have also conditions to track in progress hotplugs: https://github.com/kubevirt/kubevirt/blob/eff99ed4b8490f728b9beccbe8a3d8fda998b4f0/pkg/virt-controller/watch/workload-updater/workload-updater.go#L319-L323
> 
> and they are consumed to serialize operations like in the workload-updater or in virt-contreller: https://github.com/kubevirt/kubevirt/blob/eff99ed4b8490f728b9beccbe8a3d8fda998b4f0/pkg/virt-controller/watch/vm/vm.go#L599-L602

I am aware of them, and wonder why was there a need to have a condition per resource?
It does not scale well as additional flows will be integrated with this "framework".

> NIC hotplugs are probably now aliene objects bypassing all of those safety checks.

NIC hotplug / unplug predates CPU and memory hotplug.
These operations could be done migration or via using the dynamic-networks-controller which is an external component that could be deployed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,166 @@
...
+#### Pros
+
+- Improves the current user experience
+- Will work for hotplugging interfaces using bridge and SR-IOV bindings
+
+#### Cons
+ - Will not work for non-migratable VMs
+ - Performs a migration for every hotplug / hotunplug
+ - Requires an API change
+ - Complicated implementation
```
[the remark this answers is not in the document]
orelmisan: Could you please elaborate?

veps/sig-network/ns-admin-nichotplug.md
orelmisan: > Reverting is not the only option. 

The VEP lays out multiple alternatives, this is one of them.

> At any point in time the cluster-admin could grant live migrate right to selected node admis or even consciously restore the previous behavior.

This alternative is stated bellow.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: > Now at least he is allowed to decide.

Yes, but a GA'ed functionality is broken for namespace admins.
Following the release of v1.5, cluster-admins will have to take action in order for namespace admins to be able to hotplug/hotunplug NICs.

tiraboschi: We also have a "dirty" workaround that requires no cluster-admin intervention:
do the first step of the hotplug as before, then instead of directly creating a VMIM object, hotplug a CPU core or a small amount of memory, this will implicitly trigger a live migration completing the nic hotplug.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
tiraboschi: I think we should invest a bit of time to identify viable implementation shortcuts like for isntance having some of the actual code that that is handling the CPU and memory hotplug also detect and react to a NIC hotplug.

veps/sig-network/ns-admin-nichotplug.md
EdDev: > instead of directly creating a VMIM object, hotplug a CPU core or a small amount of memory, this will implicitly trigger a live migration completing the nic hotplug.

Good luck documenting that :)
This is a no-go, we cannot do such hacks.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
EdDev: I'm pretty sure this con was added after analyzing the codebase and logic.
When I hear shortcuts and workarounds in code, I start to worry. Rarely that brings anything good.

Anyway, the fact it is complicated, does not mean this is not the right thing to do... For the mid term solution, integrating into the rollout is a must, no question about it.
Can we do this quickly before the GA? I am not sure, sounds risky (both from time and logic perspectives).

EdDev (commented):
Thank you for laying out the options here.
Do you have a specific recommendation on how to handle this? i.e. what patch to choose and in which stages?

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: > Can we do this quickly before the GA? I am not sure, sounds risky (both from time and logic perspectives).

I don't think we can do it quickly and safely enough before the GA of v1.5 (as additional API is required and logic should be added and tested in multiple components).
IMO we will have to postpone it for v1.6.

orelmisan:
> Thank you for laying out the options here. Do you have a specific recommendation on how to handle this? i.e. what patch to choose and in which stages?

I think that for v1.5 we have two viable options:
1. Document that a cluster admin has to explicitly provide namespace admins the ability to create VMIM object in order to have NIC hotplug / unplug.
2. Revert the hardening PR.

In v1.6 we could integrate NIC hotplug with the LiveUpdate rollout strategy and re-introduce the hardening PR in case we will decide to revert it in v1.5.

EdDev:
> > Thank you for laying out the options here. Do you have a specific recommendation on how to handle this? i.e. what patch to choose and in which stages?
> 
> I think that for v1.5 we have two viable options:
> 
>     1. Document that a cluster admin has to explicitly provide namespace admins the ability to create VMIM object in order to have NIC hotplug / unplug.
> 
>     2. Revert the hardening PR.

Thanks!
The safest would be to choose option 2. However, attempting to be practical here, I would support option 1 if it is announced to to the community and no major negative convincing feedback is received.

> In v1.6 we could integrate NIC hotplug with the LiveUpdate rollout strategy and re-introduce the hardening PR in case we will decide to revert it in v1.5.

This is in practice a bug, a nasty one that is breaking a released feature.
Therefore, I think we can fix it also in v1.5  and release it as part of v1.5.x as soon as possible.

veps/sig-network/ns-admin-nichotplug.md
EdDev: When you put it like this, it makes total sense. It is not an alternative to the permanent solution, just an emergency step towards the coming GA.

tiraboschi:
> This is in practice a bug, a nasty one that is breaking a released feature.
> Therefore, I think we can fix it also in v1.5 and release it as part of v1.5.x as soon as possible.

Technically it's not an API change nor a bug but simply hardening according to the Principle of Least Privilege (PoLP).
Now you need an explicit permission for something that in the past you were freely granted to do.
But if your cluster admin grants you the right to do it, everything will work exactly as in the past.
So we cannot call it a bug, it's a choice!
If we say that we can't remove anything just because it has been granted in the past we are saying that we will never able to do hardening by PoLP.

Please notice that in k8s RBAC there are no deny rules, so if we decide to revert the VMIM hardening, a cluster admin has no way to deny CRUD on VMIM to his namespace admins and so we need also to accept that we explicitly preventing the hardening opportunity to our conscious cluster admins.
While, if we keep it like it is, a cluster admin can instead decide to revert it to previous behaviour with a single command.

tiraboschi:
> Yes, but a GA'ed functionality is broken for namespace admins.

This assumption is not correct either: it just now requires explicit permission.
Otherwise we can or should also say that a generic live migration (independent of a NIC hotplug), which is a GA'ed feature in itself, should be considered broken for namespace administrators because in the past they were freely allowed to activate it while now it requires explicit permission.
But this is not the case: the live migration for the namespace administrator is not broken but, as a design choice, now it simply requires explicit permission.

EdDev:
From my point of view, the instructions given to a user in regards to net hotplug will no longer work, making it a bug from the user point of view.

I understand that you observe it from a different angle, but I cannot ignore the implications on users.

tiraboschi:
> From my point of view, the instructions given to a user in regards to net hotplug will no longer work, making it a bug from the user point of view.

This is not true, the instructions works are expected *if you have the rights to execute them*.

As a counterexample, please take the instructions here:
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
And try to follow them applying a taint to a node on a cluster where you are namespace admin but not a cluster admin.
- Is `kubectl taint nodes node1 key1=value1:NoSchedule` going to work for you? I don't expect so.
- Can you call it a bug just because it's not working (with a clear error message) since you are not allowed to do it? I don't think so.

The only difference is that in KubeVirt 1.4 every namespace admin was always authorized to triggered a live migration (and there was now way to deny it), in KubeVirt 1.5 it works exactly as in the past but you need an explicit authorization for it.
This is exactly the definition of PoLP hardening.

EdDev:
> > From my point of view, the instructions given to a user in regards to net hotplug will no longer work, making it a bug from the user point of view.
> 
> This is not true, the instructions works as expected _if you have the rights to execute them_.

I understand your position and examples to back it up.
However, I do not agree with your interpretation for this case.

The feature to hotplug a network interface has been designed and implemented with the knowledge that a regular VM owner can migrate the VM. The design proposal has not specified that this is limited to a cluster-admin [1].

As I see it, if a feature was intended to a specific user type and now these users cannot utilize the feature because of a hardening introduced, the feature has been broken.
Asking the cluster-admin to take a step to mitigate it is ok, but that has two main issues:
1. He may not want to open the option to allow migration when the project recommended not to do so.
2. Upon upgrade, the cluster-admin needs to take an active step and adjust the RBAC. Until the admin does this, the feature is broken. It is a burden which we impose now to allow a GA feature operate, which is not user-friendly.

[1] https://github.com/kubevirt/community/blob/main/design-proposals/nic-hotplug/nic-hotplug.md#definition-of-users

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### API Additions
+
+In order to mark the VMI for automatic migration, a new allowed value will be introduced to `VirtualMachineInstanceConditionType`:
+
+```go
+VirtualMachineInstanceVNICChange VirtualMachineInstanceConditionType = "HotVNICChange"
+```
+
+In order to keep supporting in-place hotplug / hotunplug, a cluster-wide knob will be added to the `NetworkConfiguration` struct:
```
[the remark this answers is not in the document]
orelmisan: Added it as a note.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+```go
+type NetworkConfiguration struct {
+	...
+	EnableInPlaceVNICChange *bool `json:"enableInPlaceVNICChange,omitempty"`
+	...
+}
+```
+
+A user will set it to `true` in order to let KubeVirt know that the dynamic-networks-controller is deployed.
+KubeVirt will use this configuration to enable in-place hotplug / hotunplug for supported network bindings (currently `bridge`).
```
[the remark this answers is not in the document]
orelmisan: Yes.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+Doing so will enable NIC hotplug / unplug for non-migratable VMs and will provide a better user experience.
```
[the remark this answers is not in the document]
orelmisan: By enabling the knob, the user tells KubeVirt to behave in a certain way (and assume the dynamic-networks-controller is deployed).
It is the cluster-admin's responsibility to deploy the dynamic-networks-controller.
In case it is not deployed - KubeVirt will not know about it.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### VM Controller
+The VM controller is responsible for patching VMI objects as a result of a change in the VM object.
+The VM controller contains a dedicated network VM controller, responsible for NIC hotplug / hotunplug.
```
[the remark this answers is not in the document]
orelmisan: I don't think its necessary to include it in the text.

https://github.com/kubevirt/kubevirt/blob/9c92eaaaa3e09f265fb3e1cd33b4f572e6ee370f/pkg/network/controllers/vm.go#L1

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### Network VM controller
+The network VM controller will return early in the following cases:
+- The VMI has the `HotVNICChange` condition set to `true` - meaning a hotplug / hotunplug operation is already underway.
+- The VMI is currently migrating.
+
+#### VMI Controller
+The VMI controller is responsible for two flows:
+1. Updating the VMI.Status field.
+2. Patching the virt-launcher pod when it is ready with an updated Multus network attachment annotation (used for in-place hotplug).
```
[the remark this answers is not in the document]
orelmisan: Done.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+KubeVirt will use this configuration to enable in-place hotplug / hotunplug for supported network bindings (currently `bridge`).
+Doing so will enable NIC hotplug / unplug for non-migratable VMs and will provide a better user experience.
+
+#### VM Controller
+The VM controller is responsible for patching VMI objects as a result of a change in the VM object.
+The VM controller contains a dedicated network VM controller, responsible for NIC hotplug / hotunplug.
+The network VM controller is being executed independently of the LiveUpdate rollout strategy feature.
+
+The network VM controller sync will be invoked under the same conditions the CPU and memory handlers are being executed.
+As a result, the logic responsible for setting the `RestartRequired` condition on the VM will be updated.
```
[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/3efd884858d2c5004d8b47e7c2a9a11292430b9a..a6aa3ac33a8786aa967d6904261f7b7dced7e353): Partially addressed @nirdothan's review.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+A user will set it to `true` in order to let KubeVirt know that the dynamic-networks-controller is deployed.
+KubeVirt will use this configuration to enable in-place hotplug / hotunplug for supported network bindings (currently `bridge`).
+Doing so will enable NIC hotplug / unplug for non-migratable VMs and will provide a better user experience.
+
+#### VM Controller
+The VM controller is responsible for patching VMI objects as a result of a change in the VM object.
+The VM controller contains a dedicated network VM controller, responsible for NIC hotplug / hotunplug.
+The network VM controller is being executed independently of the LiveUpdate rollout strategy feature.
```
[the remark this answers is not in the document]
orelmisan: Removed the word "being".

> Please also consider emphasizing the change: i.e. is currently... will be enhanced/modified to...

This sentence describes the current situation.
The change is described bellow.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+The network VM controller sync will be invoked under the same conditions the CPU and memory handlers are being executed.
```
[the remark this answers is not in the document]
orelmisan: Adjusted the wording.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+2. Shall not be considered for migration abortion.
+
+#### Pros
+
+- Improves the current user experience
+- Will work for hotplugging interfaces using bridge and SR-IOV bindings
+- Keeps the integration with the dynamic-networks-controller
+
+#### Cons
+ - Will not work for non-migratable VMs
```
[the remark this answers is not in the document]
orelmisan: Without the dynamic-networks-controller is is not possible to hotplug / hotunplug a NIC.
It is the same as today with the manual hotplug.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+## API Examples
+
+<!--
+Tangible API examples used for discussion
+-->
+
+## Alternatives
+
+### Using the dynamic-networks-controller
```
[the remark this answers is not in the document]
orelmisan: > IMHO it's not an alternative solution, but rather a complementary one. I think that we recommend dynamic-networks-controller anyway. No?

At the moment, one could deploy the dynamic-networks-controller and enjoy in-place hotplug / hotunplug for NICs using bridge binding.
They will need to perform manual migration in order to attach an SR-IOV NIC.

The solution proposed above combines the advantages of both in-place and migration-based hotplugs, while also providing a working solution for users who are not interested in deploying another component.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+The documentation will be updated to reflect the fact that there are additional RBAC rules missing for namespace
+admins to be able to perform NIC hotplug / unplug.
+
+#### Pros
+
+- Does not require code changes
+
+#### Cons
+
+- Losses the benefits of limiting VMIM creation to cluster admins
```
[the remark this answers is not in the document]
orelmisan: I don't follow.
Could you please rephrase?

The PR that removed permissions from namespace admins is referenced in the overview section.
The following blog post was published alongside of KubeVirt v1.5:
https://kubevirt.io/2025/Hardening-VMIM.html

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/a6aa3ac33a8786aa967d6904261f7b7dced7e353..ae58425199eff6670c0fbe3945adc55a29d0f9d2): Addressed @nirdothan's comments.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/ae58425199eff6670c0fbe3945adc55a29d0f9d2..1411a3ca2fde0cfdf64861cd7013d1e46696e391): Adjusted implementation phases.

Dropped unused sections.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+> [!NOTE]
+> KubeVirt also supports in-place hotplug / hotunplug operations (without performing a migration) 
+> by integrating with the [dynamic-networks-controller](https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller).
+>
+> This method does not support SR-IOV NICs hotplug.
+
+## Motivation
+
+In previous KubeVirt versions, a project admin could hotplug / unplug NICs.
+We want to assure that this feature will continue working in v1.5 and future versions.
```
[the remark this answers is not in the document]
orelmisan: Thanks for the suggestion. I've reviewed the wording with two grammar checkers and it seems correct as is.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: The meaning of "in-place" hotplug / hotunplug is "without performing a migration" (as was explained in a note under the "Overview" section).

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,212 @@
+# VEP #29: NIC Hotplug / Hotunplug For Namespace Admins Following VMIM Hardening
```
EdDev, line 1: > Following VMIM Hardening

I suggest the name will focus on what it does, e.g. integration with rollout strategy mechanism.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+Currently, the NIC hotplug / hotunplug feature depends on a user-initiated migration following each NIC hotplug / hotunplug
```
EdDev, line 11: Perhaps it will be useful to ref the original net hotplug feature design.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,212 @@
...
+> target pod is created.
+
+PR [kubevirt/kubevirt#13497](https://github.com/kubevirt/kubevirt/pull/13497) removed Namespace administrators'
+permissions to create VirtualMachineInstanceMigration objects - thus breaking the NIC hotplug / hotunplug feature
+for them in v1.5.
+
+## Motivation
+
+In previous KubeVirt versions, a namespace admin could hotplug and unplug NICs.
+We want to ensure that this feature will continue working in v1.5 and future versions.
```
EdDev, line 30: ... and enhance it to perform the migration without manual intervention.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Goals
+
+Enable namespace admins to perform NIC hotplug / unplug operations independently - without involving the cluster-admin
```
EdDev, line 34: Please convert it to a bullet.

I think the main goal is to integrate the network hotplug to the existing rollout feature, for consistency.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,212 @@
...
+```go
+type NetworkConfiguration struct {
+	...
+	EnableInPlaceVNICChange *bool `json:"enableInPlaceVNICChange,omitempty"`
+	...
+}
+```
+
+> [!NOTE]
+> The `NetworkConfiguration` struct resides in the KubeVirt CRD under `Spec.Configuration.NetworkConfiguration`.
```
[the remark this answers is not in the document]
orelmisan: Used the field names instead of the struct names and added a link to the API reference document.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: Changed the wording a bit and added another reference to what is "in-place".

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: Changed the wording a bit.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+## Non Goals
+
+## Definition of Users
+
+- Namespace administrators
+
+## User Stories
+
+- As a namespace admin I would like to be able to hotplug / unplug a secondary interface without involving the cluster
+  admin.
```
EdDev, line 46: Please list the different scenarios:
- Hotplug a new secondary network interface that uses bridge binding.
- Hotunplug a secondary network interface that uses bridge binding.
- Hotplug a new secondary network interface of type SR-IOV.
- Disallow hotplug of a primary network interface. (or mark that it requires reset?)
- Disallow hotunplug of a primary network interface.  (or mark that it requires reset?)
- Allow using a staging strategy so any hotplug/hotunplug will take effect only after restarting a VM.

EdDev (commented):
Reviewed the first part: requirements.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: The wording was adjusted.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Added all of them but the following two:
- Disallow hotplug of a primary network interface. (or mark that it requires reset?)
- Disallow hotunplug of a primary network interface. (or mark that it requires reset?)

Currently, the NIC hotplug feature only supports secondary networks, but doesn't block the addition of primary interfaces.
Lets discuss it first, prior to adding it here.
We could also add SR-IOV NIC hotunplug, which is currently unsupported.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Done.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/6838e88094189dacd2ca18054363eb85c509e341..cd0dfa7fa4a55d78936849a562235ad4237c82c9): Addressed @EdDev's comments.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Hotplugging  / unplugging a primary NIC is also not covered by e2e tests.

iholder101 (commented):
Thank you @orelmisan!
In general the approach to automatically migrate the VM seems like the right direction to me.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+
+In order to keep supporting in-place hotplug / hotunplug, a cluster-wide knob will be added to the `NetworkConfiguration` struct:
+
+```go
+type NetworkConfiguration struct {
+	...
+	EnableInPlaceVNICChange *bool `json:"enableInPlaceVNICChange,omitempty"`
+	...
+}
+```
```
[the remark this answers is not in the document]
orelmisan: Currently, KubeVirt has two methods of hotplug / hotunplug:
1. Migration based (requires the user to manually initiate a migration).
2. In-place (without a migration), which requires the user to also deploy the dynamic-networks-controller.

The in-place hotplug of SR-IOV NICs is not supported (due to the requirement to invoke the SR-IOV device plugin).

At the moment, KubeVirt always performs all the actions that enable in-place hotplug / hotunplug.
If the migration will be automatic, we need a way to tell KubeVirt when not to perform an automatic migration.
@EdDev had suggested doing it based on some timeout, but IMO it is racefull.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
EdDev: @nirdothan , it is presented as an alternative with the relevant pros/cons.
The preferred alternative presented before includes this as a complementary functionality by using the KV config knob and deploying this add-on.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: https://github.com/k8snetworkplumbingwg/multus-dynamic-networks-controller

It watches pods, and if their network attachment annotation had changed - it invokes the Multus CNI to add / remove NICs from the pod.

> Can't we dynamically check if it's deployed?
I don't think its good for KubeVirt to be aware of its plugins / helper components as it makes the core dependent on the plugins and not the other way around.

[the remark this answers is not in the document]
iholder101: @orelmisan thanks for clarifying.

Two thoughts here:
1. Can't we just use the update strategy for that? Do we really need a different strategy for volumes, cpu/memory and network?
2. We have `VM Rollout Strategy` and `Volume Update Strategy`, but this knob looks different. If we're sure we want to keep it, can we re-brand it to `NetworkUpdateStrategy` or something similar?
3. Kubernetes handled in-place changes by introducing a dedicated subresource (`/resize`). That's also an alternative to the proposed API.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+1. Introduce new condition.
+2. Adjust network VMI controller logic to set the condition.
+3. Adjust the workload update controller to consider the new condition.
+4. Adjust the e2e tests to work without dynamic-networks-controller.
+5. Adjust the VM controller to invoke the network controller differently.
+6. Adjust the network VM controller.
+7. Introduce new knob to KubeVirt CRD.
+8. Adjust the VMI controller logic to consider the new knob.
+9. Adjust the e2e tests to work with the dynamic-networks-controller.
```
EdDev, line 213: The phases need to present a functional operation, even if limited.
Please re-arange these to provide a functional operation.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Feature lifecycle Phases
+
+### GA
+Planned for v1.6.
+Since the LiveUpdate rollout strategy and NIC hotplug / hotunplug features are already GAed, their integration does not
+pose a significant risk.
```
EdDev, line 220: We can also look at this as a bug fix.

In general, this proposal is enhancing the existing network hotplug feature and could as well be added to it.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/cd0dfa7fa4a55d78936849a562235ad4237c82c9..72813c21aa46fd780725ba7c68f0537615e47212): Addressed @jean-edouard's [comment](https://github.com/kubevirt/enhancements/pull/8#discussion_r2010248529).

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: > "will keep working" maybe then? :)

Done.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+which automatically initiates a migration of the VM when certain spec fields are changed.
+
+Extend the feature to automatically initiate a live migration in case of NIC hotplug / hotunplug request.
+
+#### API Additions
+
+In order to mark the VMI for automatic migration, a new allowed value will be introduced to `VirtualMachineInstanceConditionType`:
+
+```go
+VirtualMachineInstanceVNICChange VirtualMachineInstanceConditionType = "HotVNICChange"
```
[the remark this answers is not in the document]
orelmisan: AFAIK, there are currently at least three existing conditions used for the following:
- CPU hotplug
- Memory hotplug
- Volume hotplug 

CPU and memory hotplug use the condition for several flows (the list is not comprehensive):
- Marking the VMI for automatic migration [1]
- Affecting the resource allocation for the target pod [2][3]
- Enabling the hotplug flow at the virt-handler's level [4]

This proposal proposes to use the condition only in order to mark the VMI for automatic migration.
Generalizing the condition will be challenging and will require adjustments to existing flows.

WDYT?

[1] https://github.com/kubevirt/kubevirt/blob/a3b2471b4df11a24048d9bdaa84e3ec3539b8453/pkg/virt-controller/watch/workload-updater/workload-updater.go#L337
[2] https://github.com/kubevirt/kubevirt/blob/6f923df1d89a88eca96f88425ff79f2f6f8982eb/pkg/virt-controller/watch/migration/migration.go#L998
[3] https://github.com/kubevirt/kubevirt/blob/8ffa163e9a90ead79b15a45d5b2bbcf2efd55192/pkg/virt-handler/vm.go#L3613
[4] https://github.com/kubevirt/kubevirt/blob/8ffa163e9a90ead79b15a45d5b2bbcf2efd55192/pkg/virt-handler/vm.go#L3608

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+In order to keep supporting in-place hotplug / hotunplug, a cluster-wide knob will be added to the `NetworkConfiguration` struct:
```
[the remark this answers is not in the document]
orelmisan: Added the alternative you've suggested offline.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Done.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: The wording has been adjusted.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/72813c21aa46fd780725ba7c68f0537615e47212..4cef99f44f0886c8afa353844873722558f7cff0): Partly addressed @EdDev's comments.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+which automatically initiates a migration of the VM when certain spec fields are changed.
+
+Extend the feature to automatically initiate a live migration in case of NIC hotplug / hotunplug request.
+
+#### API Additions
+
+In order to mark the VMI for automatic migration, a new allowed value will be introduced to `VirtualMachineInstanceConditionType`:
+
+```go
+VirtualMachineInstanceVNICChange VirtualMachineInstanceConditionType = "HotVNICChange"
```
ormergi, line 72: What's "HotVNIChange" mean?

Which entity (human, controller?) that should manage this condition lifecycle, and which entity consume it? 

The spec should be used for describing the desired state,  what the status has to do with that?

orelmisan: > What's "HotVNIChange" mean?

Hot vNIC Change.
I tried to follow the naming of `HotVCPUChange` and `HotMemoryChange` conditions [1][2].
There is also `VolumesChange` [3].

> Which entity (human, controller?) that should manage this condition lifecycle, and which entity consume it?

The VMI controller will add and remove it.
Workload updater controller will create a VMIM for the VMI in case it has this condition. See the explanation bellow under "Design".

> The spec should be used for describing the desired state, what the status has to do with that?

I don't follow the question, could you please rephrase?

Please see the PoC PR https://github.com/kubevirt/kubevirt/pull/14259.

[1] https://github.com/kubevirt/kubevirt/blob/6aee7891e7dc993a003c886988852f0cc350c9fd/staging/src/kubevirt.io/api/core/v1/types.go#L570
[2] https://github.com/kubevirt/kubevirt/blob/6aee7891e7dc993a003c886988852f0cc350c9fd/staging/src/kubevirt.io/api/core/v1/types.go#L573
[3] http://github.com/kubevirt/kubevirt/blob/6aee7891e7dc993a003c886988852f0cc350c9fd/staging/src/kubevirt.io/api/core/v1/types.go#L576

ormergi: I wasn't aware of this mechanism, thanks for explaining.

What about the condition reason and message?
I think this doc should reflect them as well because they are mandatory by the convention.

I find the existing conditions type naming off, maybe it should have use past tense at least ("SomthingChanged").
I think that if the condition type name would emphasize a change is pending for action would describe the state better, e.g.:  "PendingNetworkChange".
Or use a generic condition type name and change the condition reason be concreate, for example: 
```
type: PendingSpecChange
reason: NetworkInterfaceAdded
```
I think that would resolve this comment https://github.com/kubevirt/enhancements/pull/8#discussion_r2010114195

If you dont want to follow any of the above, please extend the "VNI" part at least, or use the "SomthingChange" naming.
"VNI" doesnt say much and I have strong networking orientation :sweat_smile:.

veps/sig-network/ns-admin-nichotplug.md
EdDev: There is still a need to define what should happen when one attempts to perform actions on the primary.
You could define that they need to be rejected or to mark it for restart. Leaving it undefined is not nice.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Functional Testing Approach
+
+The migration-based hotplug / hotunplug tests will be adjusted to remove the manual VMIM object creation.
+The in-place hotplug / hotunplug tests will be adjusted to assert that a migration did not occur.
+
+## Implementation Phases
+
+1. Support automatic migration on NIC hotplug / hotunplug (without dynamic-networks-controller).
+2. Support in-place NIC hotplug / hotunplug (with dynamic-networks-controller).
```
EdDev, line 219: I do not understand this one. It is an existing capability, so how can you consider it as an addition when it is already there?

orelmisan:
[Change 1](https://github.com/kubevirt/enhancements/compare/4cef99f44f0886c8afa353844873722558f7cff0..a13ceb9cf5d8b68b5d999b7a1a3bd2fc21b90418), [Change 2](https://github.com/kubevirt/enhancements/compare/a13ceb9cf5d8b68b5d999b7a1a3bd2fc21b90418..24a65153af2335604b5e0215fb955921bbe66fd7): Addressed @EdDev's comments.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: This knob was removed from the document in favor of a time-based solution that was suggested by @EdDev and has @vladikr's blessing.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: The resource-specific condition was replaced with a generic one.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: The knob was removed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+> [!NOTE]
+> The `NetworkConfiguration` struct resides in the KubeVirt CRD under `spec.configuration.network`.
+> For additional details, please see the API reference: https://kubevirt.io/api-reference/v1.5.0/definitions.html#_v1_networkconfiguration
+
+A user can set it to `true` to inform KubeVirt that the dynamic-networks-controller is deployed.
+When the LiveUpdate rollout strategy is enabled, KubeVirt will use this additional configuration to enable in-place 
+hotplug and hotunplug (without migration) for supported network bindings (currently `bridge`).
+This will enable NIC hotplug / unplug for non-migratable VMs, thus improving the user experience.
+
+#### VM Controller
```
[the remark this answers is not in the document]
orelmisan: Done.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,214 @@
...
+The VM controller is responsible for patching VMI objects as a result of a change in the VM object.
+The VM controller contains a dedicated network VM controller, responsible for NIC hotplug / hotunplug.
+The network VM controller is executed independently of the LiveUpdate rollout strategy feature.
+
+The network VM controller sync will be invoked under the same conditions as the CPU and memory hotplug handlers.
+As a result, the logic responsible for setting the [RestartRequired](https://kubevirt.io/user-guide/user_workloads/vm_rollout_strategies/#restartrequired-condition)
+condition on the VM will be updated.
+
+#### Network VM controller
+The network VM controller will return early in the following cases:
```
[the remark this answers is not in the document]
orelmisan: This section and its siblings were removed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+> [!NOTE]
+> The `NetworkConfiguration` struct resides in the KubeVirt CRD under `spec.configuration.network`.
+> For additional details, please see the API reference: https://kubevirt.io/api-reference/v1.5.0/definitions.html#_v1_networkconfiguration
+
+A user can set it to `true` to inform KubeVirt that the dynamic-networks-controller is deployed.
+When the LiveUpdate rollout strategy is enabled, KubeVirt will use this additional configuration to enable in-place 
+hotplug and hotunplug (without migration) for supported network bindings (currently `bridge`).
+This will enable NIC hotplug / unplug for non-migratable VMs, thus improving the user experience.
```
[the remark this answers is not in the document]
orelmisan: The knob was removed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+
+An alternative to adding this knob could be to: 
+1. Set the condition on the VMI with the `False` state.
```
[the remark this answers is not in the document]
orelmisan: Added the recording of a timestamp.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+2. Give the dynamic-networks-controller a timeframe to perform its action (in case it is deployed).
+3. Watch the VMI's interface status to look for the expected status.
+4. In case the expected status is not present and the timeframe had expired - set the condition to `True`.
+
+Cons for this alternative:
+1. For every hotplug / hotunplug operation there will be a builtin delay.
+2. The dynamic-networks-controller, Multus CNI and the delegate CNIs are not guaranteed to complete their actions in the defined timeframe.
+3. The report Multus applies to the pod, and the status reports going up the controllers chain are not guaranteed to be available on time.
```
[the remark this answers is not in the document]
orelmisan: This section was removed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+hotplug and hotunplug (without migration) for supported network bindings (currently `bridge`).
+This will enable NIC hotplug / unplug for non-migratable VMs, thus improving the user experience.
+
+An alternative to adding this knob could be to: 
+1. Set the condition on the VMI with the `False` state.
+2. Give the dynamic-networks-controller a timeframe to perform its action (in case it is deployed).
+3. Watch the VMI's interface status to look for the expected status.
+4. In case the expected status is not present and the timeframe had expired - set the condition to `True`.
+
+Cons for this alternative:
```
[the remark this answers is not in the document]
orelmisan: This section was removed.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: This phase was removed.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/24a65153af2335604b5e0215fb955921bbe66fd7..cc8ded7324d1d836c05f3839630cc411782d571b): Added a note about the dynamic-networks-controller and its subordinates.
Added the migration controller's role.

Minor wording changes.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### API Additions
+
+Currently, the CPU, memory and volume hotplug / hotunplug flows have a dedicated condition on the VMI object to mark the VMI
+for automatic migration:
+- HotVCPUChange
+- HotMemoryChange
+- VolumesChange
+
+Since we do not wish to introduce another resource-specific condition, a new generic allowed value will be added to `VirtualMachineInstanceConditionType`:
```
EdDev, line 75: > Since we do not wish

I think this requires reasoning in order to convince reviewers.
You could point to an appendix in order to elaborate on this.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+   
+   - Patches the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`) - since the dynamic-networks-controller is not present, this will not have an affect.
+   - Updates the VMI interface status.
+   - Adds the `MigrationRequired` condition to the VMI with its status set to `False` and a timestamp.
+4. On a future iteration of the reconciliation loop, the VMI controller will update the `MigrationRequired` condition status to `True` (Following no change in the interface statuses for more than TBD seconds after the initial timestamp set on the condition).
+5. The Workload update controller will identify the VMI as a candidate for automatic migration and will create a VMIM object for it.
+6. The Migration controller will remove the `MigrationRequired` condition once the migration is complete.
+7. The vNICs will be hot-plugged following the migration - same as today. 
+
+#### Migration Based Hotunplug Flow
```
EdDev, line 97: The logic is almost identical to the hotplug flow, unless I missed something.
That is a good thing that we want to emphasize, so it looks odd to see it separated here.
I see it was combined in the "In-Place" case.

Was there a special reason to have this separation?

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### In-Place Hotplug / Hotunplug Flow
+
+1. The VM owner adds one or more new secondary vNICs with bridge binding and matching networks, or sets their state to `absent` to / on a running VM.
+2. The VM controller patches the VMI object with the additional relevant vNICs and networks.
+3. The VMI controller:
+
+    - Patches the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`).
+    - Updates the VMI interface status.
+    - Adds the `MigrationRequired` condition to the VMI with the `False` with its status set to `False` and a timestamp.
```
EdDev, line 119: Something went wrong with the sentence here (double `False`).

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+#### Migration Based Hotplug Flow
+
+1. The VM owner adds one or more new secondary vNICs and matching networks to a running VM.
+2. The VM controller patches the VMI object with the additional relevant vNICs and networks.
+3. The VMI controller:
+   
+   - Patches the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`) - since the dynamic-networks-controller is not present, this will not have an affect.
+   - Updates the VMI interface status.
+   - Adds the `MigrationRequired` condition to the VMI with its status set to `False` and a timestamp.
+4. On a future iteration of the reconciliation loop, the VMI controller will update the `MigrationRequired` condition status to `True` (Following no change in the interface statuses for more than TBD seconds after the initial timestamp set on the condition).
```
EdDev, line 92: The information in the brackets is important, perhaps add it as a bullet.
As this is a formal design proposal, we better provide an initial estimation to the backoff. A range may be provided and experimented on with different clusters.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
EdDev: Now the DNC is an optional part of the solution.
I guess, there are two options here: 1) Drop this con, reasoning that it presents no change to the current state. 2) Keep it but add specifics, e.g. in case of SR-IOV and lack of DNC.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### Pros
+
+- Improves the current user experience
+- Will work for hotplugging interfaces using bridge and SR-IOV bindings
+- Keeps the integration with the dynamic-networks-controller
+
+#### Cons
+ - Will not work for non-migratable VMs
+ - Performs a migration for every hotplug / hotunplug
```
EdDev, line 152: ... in case DNC is not deployed.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,225 @@
...
+ - Requires an API change
```
EdDev, line 153: If it is only about the new condition, I would add here "minimal" and mention it is only about the status condition.
After-all, the addition is only for controllers to monitor, not for operators who just want this hotplug to work.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/cc8ded7324d1d836c05f3839630cc411782d571b..610bd50622a51ec4c5d5c66e9b17ec527b5b6635): Partially addressed @EdDev's comments.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Done.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: > The information in the brackets is important, perhaps add it as a bullet.

The wording was adjusted.

> As this is a formal design proposal, we better provide an initial estimation to the backoff. A range may be provided and experimented on with different clusters.

Currently the text contains "ten seconds".
Lets not set it in stone.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Thanks.
Done.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: Thank you for the suggestion.

Done.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,228 @@
...
+2. Give the dynamic-networks-controller a timeframe to perform its action (in case it is deployed).
+3. Watch the VMI's interface status to look for the expected status.
+4. In case the expected status is not present and the timeframe had expired - set the condition to `True`.
+
+Cons for this alternative:
+1. For every hotplug / hotunplug operation there will be a builtin delay.
+2. The dynamic-networks-controller, Multus CNI and the delegate CNIs are not guaranteed to complete their actions in the defined timeframe.
+3. The report Multus applies to the pod, and the status reports going up the controllers chain are not guaranteed to be available on time.
+
+Which brings us to the conclusion that sometimes migrations will be triggered instead of in-place hotplug / hotunplug operations.
```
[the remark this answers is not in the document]
orelmisan: The explanation was added as a note.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: > Now the DNC is an optional part of the solution. I guess, there are two options here: 1) Drop this con, reasoning that it presents no change to the current state. 2) Keep it but add specifics, e.g. in case of SR-IOV and lack of DNC.

I've decided to go with the second approach:
> 2) Keep it but add specifics, e.g. in case of SR-IOV and lack of DNC.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
orelmisan: This whole section was removed in favor of the time-based solution.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/610bd50622a51ec4c5d5c66e9b17ec527b5b6635..82cd9dc9b5d0051fcaceebcce671071892a1150c): Added an implementation phase to fix the `RestartRequired` condition added on NIC hotplug / hotunplug.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/82cd9dc9b5d0051fcaceebcce671071892a1150c..0545e1873c6f133263c41bb803dda675a775e28c): Added reasoning for the `MigrationRequired` condition.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/0545e1873c6f133263c41bb803dda675a775e28c..2c06fd11f3321eda0f804b850b18aff331652359): Added missing colons.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+pose a significant risk.
+
+This integration could also be considered as a bug fix to restore namespace admins' ability to perform NIC hotplug / hotunplug.
+
+# Appendix - Rational for Adding a Shared Condition
+
+Currently, each resource's hotplug flow (CPU. memory, volume) may add its own dedicated condition to the VMI object to 
+signal that the VMI requires an automatic migration.
+
+While this approach provides fined grained disability into the state of individual resources, it introduces several challenges:
```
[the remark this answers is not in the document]
orelmisan: I misused the automatic spell checker.
Thanks

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/2c06fd11f3321eda0f804b850b18aff331652359..5981b420ba0f229dd67d5c88e07f4a5003081365): Addressed @nirdothan's [comment](https://github.com/kubevirt/enhancements/pull/8#discussion_r2034849568).

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+When multiple resource conditions co-exist, it's unclear which one has precedence or whether they should block each-other.
+
+The proposed solution introduces a single condition `MigrationRequired` that reflects the requirement to automatically
+migrate the VMI.
+Individual resource logic can still detect their own limitations internally, but instead of setting multiple conditions,
+they would contribute to setting or clearing this unified condition.
+
+Benefits:
+
+1. Clear ownership:
```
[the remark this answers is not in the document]
orelmisan: This section is about the benefit of having a centralized logic to set or clear the `MigrationRequired` condition.
The centralized logic owns the request of an automatic live migration.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+
+## API Examples
+
+<!--
+Tangible API examples used for discussion
+-->
+
+## Alternatives
+
+### Using the dynamic-networks-controller
```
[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/5981b420ba0f229dd67d5c88e07f4a5003081365..78a22f7d1855136feab26afc5773404b294b7ddc): Addressed @EdDev's [comment](https://github.com/kubevirt/enhancements/pull/8#discussion_r2035079798).

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+- HotMemoryChange
+- VolumesChange
+
+Since we do not wish to introduce another resource-specific condition, a new generic allowed value will be added to `VirtualMachineInstanceConditionType`:
+
+```go
+VirtualMachineInstanceMigrationRequired VirtualMachineInstanceConditionType = "MigrationRequired"
+```
+
+For additional details, please see [Rational for Adding a Shared Condition](#appendix---rational-for-adding-a-shared-condition).
```
lyarwood, line 83: Could we not handle this move to a generic condition type in a follow up enhancement that aligns all of the `*Change` conditions? As it stands this part of the enhancement feels like something that's going to continue to cause differences between the flows and just go unaddressed in sig-{compute,storage} backlogs.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+1. The VM owner adds one or more new secondary vNICs with bridge binding and matching networks, or sets their state to `absent` to / on a running VM.
+2. The VM controller will:
+
+   - Not add the `RestartRequired` condition to the VM object.
+   - Patch the VMI object with the relevant vNICs and networks.
+3. The VMI controller will:
+
+    - Patch the virt-launcher pod's Multus network attachment annotation (`k8s.v1.cni.cncf.io/networks`).
+    - Update the VMI interface status.
+    - Add the `MigrationRequired` condition to the VMI with its status set to `False` and a timestamp.
```
lyarwood, line 120: Would this be problematic in a future world where all of the hot plug implementations align to this single condition? Say you've also updated socket count at the same time which hot plug path wins setting `MigrationRequired` to their required value?

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Implementation Phases
+
+1. Support automatic migration on NIC hotplug / hotunplug (without dynamic-networks-controller).
+2. Fix the redundant `RestartRequired` condition added to VMs when NIC hotplug / hotunplug are performed.
+
+## Feature lifecycle Phases
+
+### GA
+Planned for v1.6.
```
lyarwood, line 226: I'm personally not against this but if folks are you could provide a configurable policy to switch between the existing manual and this new automatic behaviour, the latter also sitting behind a FG that we could graduate.

veps/sig-network/ns-admin-nichotplug.md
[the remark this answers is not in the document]
EdDev: > The rational is that even-though the algorithm is well defined, it is unpredictable in runtime:

The whole system is async & distributed, no one can claim any strong predictability in terms of timing except for one thing: Given a desired state, the system will try and reach it.

> In retrospect we may never know why a certain migration was triggered and whether DNC was properly deployed, configured and running in a certain cluster.

Yes, and that is a very good thing. At least, that is the whore purpuse of a system which reconciles to reach a desired state. We can go over many exampeles on how a pod or VM reaches running state, passing many steps which sometime slows it down (e.g. a resource it requires is not yet available, but becomes available later).

> It is also tricky to properly test and research flakes in retrospect, needing to collect DNC logs and make time calculations. I'd much rather have a clear boolean that indicates if migration should be triggered or not.

We can make exact observations on the state of the system, check events (and trigger them), wait on conditions, etc.

> Going back to @iholder101's [suggestion](https://github.com/kubevirt/enhancements/pull/8#discussion_r2010191583), We could try to look for a way to identify DNC's existence in the future though contribution to Multus (having DNC register itself with new pods) and get rid of the knob.

I will not support any knob unless a strong enough reason is provided. I find the one raised here good but weak.
I have also given previously reasons why a knob will be breaking existing users that use the in-place solution today. But my main reason I prefer it out is that it complicates users/admins and takes the responsibility to do the best thing possible from the system and passes it to the admin or some other component.

veps/sig-network/ns-admin-nichotplug.md
```diff
@@ -0,0 +1,272 @@
...
+> target pod is created.
+
+PR [kubevirt/kubevirt#13497](https://github.com/kubevirt/kubevirt/pull/13497) removed Namespace administrators'
+permissions to create VirtualMachineInstanceMigration objects - thus breaking the NIC hotplug / hotunplug feature
+for them in v1.5.
+
+## Motivation
+
+In previous KubeVirt versions, a namespace admin could hotplug and unplug NICs.
+We want to ensure that this feature will keep working in v1.5 and future versions and enhance it so the migration will be
```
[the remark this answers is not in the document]
orelmisan: The issue that broke the NIC hotplug feature was introduced quite late in v1.5 development cycle.
For KubeVirt v1.5 it was decided that the steps to provide namespace admins the necessary RBACs to keep the feature working will be documented.
For v1.6 onward we want to go with the proposed approach.
This enhancement was written before v1.5 was released.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: This VEP proposes a way to consolidate the automatic migration under a single condition with an invitation for other hotplug flows to join.

As you've said, currently there are multiple existing hotplug flows, each using its own condition.
It would be difficult IMO to create a VEP to align them all to use the shared condition, but I think it is reasonable to introduce it here and have follow-up VEPs for other hotplug flows to use the shared condition.
WDYT?

veps/sig-network/ns-admin-nichotplug.md
orelmisan: I think there should be a precedence mechanism to determinate that.
NIC hotplug will still work if a migration will be performed while waiting for the dynamic-networks-controller.

veps/sig-network/ns-admin-nichotplug.md
orelmisan: The goal of initiating automatic migration was (among others) to address the issue introduced by the VMIM hardening while not providing namespace admins the permissions to create VMIM.
It is basically a bug fix.

Additionally, both NIC hotplug and LiveUpdate rollout strategy are GAed, so it will be confusing IMO to have their integration as non GA.
Also, which code paths should be protected by a FG? in which configuration would the e2e test execute? etc.

vladikr: yes, I tend to agree that we can do without it.

vladikr:
@orelmisan Thank you!
From my point of view, everything looks good here.
/approve
/hold
@lyarwood please unhold when your concerns are addressed.

orelmisan:
[Change](https://github.com/kubevirt/enhancements/compare/78a22f7d1855136feab26afc5773404b294b7ddc..7a157394148ef4087295caaedd8b9f2330a1e9aa): Align commit subject with PR subject.

veps/sig-network/ns-admin-nichotplug.md
EdDev: @lyarwood , I do not see any conflict, but if you detect one, could you share what it may be?

Any current or future hotplug implementation should reflect the request/need to trigger a migration through this condition. Therefore, if socket count joins the party, it’s logic should look similar to this (assuming no backoff required like with network):
- If no such condition type already in place, add it with status True.
- If condition type exists, set status to True (override anything that is there).

No one should override a status from True to False.

EdDev:
> [Change](https://github.com/kubevirt/enhancements/compare/78a22f7d1855136feab26afc5773404b294b7ddc..7a157394148ef4087295caaedd8b9f2330a1e9aa): Align commit subject with PR subject.

/lgtm

veps/sig-network/ns-admin-nichotplug.md
lyarwood: I appreciate that but still think it's a mistake for a singular SIG to introduce a new shared condition like this as part of a VEP specific to functionality they own. That approach essentially caused the fragmented implementation of hot plug that we are now trying to fix after all.

I don't want to block this VEP so as others don't appear to agree I'll get out of the way but I honestly believe that a centralised shared VEP across sig-{compute,network,storage} introducing this new condition is a better approach here.

veps/sig-network/ns-admin-nichotplug.md
lyarwood: So I wasn't aware that False to True wouldn't break the `dynamic-networks-controller` flow. Still True to False would be problematic. It's details like this that make be believe that a centralised VEP where this is all spelt out and documented for the different implementations is important.

lyarwood:
/hold cancel

Apologies for the delay, as I've said in my latest replies I don't want to block this VEP but I still believe it's a mistake to introduce a new centralised condition as part of this VEP without discussion and wider co-ordination with the SIGs that own other hot plug implementations. Hopefully the other SIGs do follow up to align with this new condition and we don't see further implentation fragmentation with hot plug.

akalenyu:
> > /hold cancel
> > Apologies for the delay, as I've said in my latest replies I don't want to block this VEP but I still believe it's a mistake to introduce a new centralised condition as part of this VEP without discussion and wider co-ordination with the SIGs that own other hot plug implementations. Hopefully the other SIGs do follow up to align with this new condition and we don't see further implentation fragmentation with hot plug.
> 
> Thank you @lyarwood. SIG compute were involved with this VEP and gave it their blessing. SIG storage are welcome to chime in and share their opinion @akalenyu @ShellyKa13.

Actually, @mhenriks is working on declarative volume hotplug in https://github.com/kubevirt/kubevirt/pull/13847 but I don't think there's an intention to use a condition for that

PR #11, opened 2025-02-25
add design for supporting DRA devices in KubeVirt

alaypatel07:
VEP for Supporting DRA Devices in Kubevirt

Fixes #

**Checklist**

- [x] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [ ] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [ ] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [ ] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note
NONE
```

### DIFF
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -100,6 +100,7 @@ type VirtualMachineInstanceSpec struct {
 type GPU struct {
 	// Name of the GPU device as exposed by a device plugin
 	Name string                    `json:"name"`
+	// DeviceName string `json:"deviceName"`    <-- inlined into DeviceSource
 	DeviceSource DeviceSource      `json:",inline"`
 	VirtualGPUOptions *VGPUOptions `json:"virtualGPUOptions,omitempty"`
 	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
@@ -205,12 +206,12 @@ generate the domain xml.
 ### DRA API for reading device related information
 
 The examples below shows the APIs used to generate the vmi.status.deviceStatuses section:
-1. the pod status has reference to the resourceClaimName, pod.status.resourceClaimStatuses[*].resourceClaimName where
-   the name of the claim is same as vmi.spec.resourceClaims[*].Name
+1. the pod status has reference to the resourceClaimName, pod.status.resourceClaimStatuses[].resourceClaimName where
+   the name of the claim is same as vmi.spec.resourceClaims[].Name
 1. pod spec has node name, pod.spec.nodeName
 1. the resourceclaim status has device name and driver use for allocating the device,
-   resourceclaim.status.allocation.devices[*].deviceName and resourceclaim.status.allocation.devices[*].driver, where
-   resourceclaim.status.allocation.devices[*].request is same as vmi.spec.domain.devices[*].gpus[*].claim.request
+   resourceclaim.status.allocation.devices[].deviceName and resourceclaim.status.allocation.devices[].driver, where
+   resourceclaim.status.allocation.devices[].request is same as vmi.spec.domain.devices[].gpus[].claim.request
 1. Using node name and driver name, the resource slice for that node could be found. Using device name, the attributes
    of the device could be found
 
@@ -639,7 +640,7 @@ type MigrationState struct {
 
 ### Alternative 2
 
-In order to handle the live-migration usecase, the following changes are required to the API:
+In order to handle the live-migration usecase, the following changes are required to the VMI API:
 ```go
 // DeviceStatus has the information of all devices allocated spec.domain.devices
 type DeviceStatus struct {
@@ -737,12 +738,13 @@ Refer to https://github.com/kubevirt/community/blob/main/design-proposals/featur
 
 - Code changes behind feature gate
 - unit tests
+- mock e2e test
 
 ### Beta
-- 
+
 - evaluate user and driver authors experience
 - feature gate turned on by default
-- e2e tests
+- e2e tests with 1 real driver implementation
 - consider additional usecases if any
 
 ### GA
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -26,6 +26,7 @@ control of their devices using Virtual Machines and Containers.
 
 - Replace device-plugin support in KubeVirt
 - Align on what drivers KubeVirt will support in tree
+- Address any complexity issues when DRA APIs in KubeVirt
 
 ## Definition of Users
 
@@ -99,9 +100,11 @@ type VirtualMachineInstanceSpec struct {
 
 type GPU struct {
 	// Name of the GPU device as exposed by a device plugin
-	Name string                    `json:"name"`
+	Name string `json:"name"`
+	// DeviceSource is the name of the device provisioned either by device plugins
+	// or by DRA enabled device
 	// DeviceName string `json:"deviceName"`    <-- inlined into DeviceSource
-	DeviceSource DeviceSource      `json:",inline"`
+	DeviceSource      DeviceSource `json:",inline"`
 	VirtualGPUOptions *VGPUOptions `json:"virtualGPUOptions,omitempty"`
 	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
 	// +optional
@@ -110,6 +113,9 @@ type GPU struct {
 
 type HostDevice struct {
 	Name string               `json:"name"`
+	// DeviceSource is the name of the device provisioned either by device plugins
+	// or by DRA enabled device
+	// DeviceName string `json:"deviceName"`    <-- inlined into DeviceSource
 	DeviceSource DeviceSource `json:",inline"`
 	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
 	// +optional
@@ -119,8 +125,19 @@ type HostDevice struct {
 type DeviceSource struct {
 	// DeviceName is the name of the device provisioned by device-plugins
 	DeviceName *string `json:"deviceName,omitempty"`
-	// Claim is the name of the resource claim that is going to provision the DRA device
-	Claim *k8sv1.ResourceClaim `json:"claim,omitempty"`
+	// ClaimRequest provides the ClaimName from vmi.spec.resourceClaims[].name and
+	// requestName from resourceClaim.spec.devices.requests[].name
+	// this fields requires DRA feature gate enabled
+	ClaimRequest *ClaimRequest `json:",inline,omitempty"`
+}
+
+type ClaimRequest struct {
+	// ClaimName needs to be provided from the list vmi.spec.resourceClaims[].name where this
+	// device is allocated
+	ClaimName string `json:"claimName"`
+	// RequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
+	// device is requested
+	RequestName string `json:"requestName"`
 }
 
 type VirtualMachineInstanceStatus struct {
@@ -182,7 +199,7 @@ DomainSpec API.
 In v1alpha3 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.31.0/resource/v1alpha3#DeviceClaim), multiple drivers
 could potentially provision devices that are part of a single claim. For this reason, a separate list of claims required
 for the VMI (section 1) is needed instead of mentioning the resource claim in devices section
-([see Alternate Designs](#Alternative 1))
+[see Alternate Designs](#alternative-1)
 
 The second sections allows for the resource claim to be used in the spec.domain.devices section. The two uses cases
 currently handled by the design are:
@@ -206,12 +223,12 @@ generate the domain xml.
 ### DRA API for reading device related information
 
 The examples below shows the APIs used to generate the vmi.status.deviceStatuses section:
-1. the pod status has reference to the resourceClaimName, pod.status.resourceClaimStatuses[].resourceClaimName where
-   the name of the claim is same as vmi.spec.resourceClaims[].Name
-1. pod spec has node name, pod.spec.nodeName
+1. the pod status has reference to the resourceClaimName, `pod.status.resourceClaimStatuses[].resourceClaimName` where
+   the name of the claim is same as `vmi.spec.resourceClaims[].Name`
+1. pod spec has node name, `pod.spec.nodeName`
 1. the resourceclaim status has device name and driver use for allocating the device,
-   resourceclaim.status.allocation.devices[].deviceName and resourceclaim.status.allocation.devices[].driver, where
-   resourceclaim.status.allocation.devices[].request is same as vmi.spec.domain.devices[].gpus[].claim.request
+   `resourceclaim.status.allocation.devices[].deviceName` and `resourceclaim.status.allocation.devices[].driver`, where
+   `resourceclaim.status.allocation.devices[].request` is same as `vmi.spec.domain.devices[].gpus[].claim.request`
 1. Using node name and driver name, the resource slice for that node could be found. Using device name, the attributes
    of the device could be found
 
@@ -381,9 +398,8 @@ spec:
     devices:
       gpus:
       - name: pgpu
-        claim: 
-          name: pgpu-claim-name
-          request: pgpu-request-name
+        claimName: pgpu-claim-name
+        requestName: pgpu-request-name
 status:
   deviceStatus:
     gpuStatuses:
@@ -458,12 +474,12 @@ metadata:
 spec:
   config:
   - opaque:
-    driver: gpu.nvidia.com
-    parameters:
-      apiVersion: gpu.nvidia.com/v1alpha1
-      driverConfig:
-       driver: vfio-pci
-      kind: GpuConfig
+      driver: gpu.nvidia.com
+      parameters:
+        apiVersion: gpu.nvidia.com/v1alpha1
+        driverConfig:
+         driver: vfio-pci
+        kind: GpuConfig
   selectors:
   - cel:
       expression: device.driver == 'gpu.nvidia.com' && device.attributes['gpu.nvidia.com'].type == 'gpu'
@@ -483,9 +499,8 @@ spec:
     resourceClaimTemplateName: single-gpu
   domain:
     gpus:
-    - claim:
-         name: gpu-resource-claim
-         request: gpu
+    - claimName: gpu-resource-claim
+      requestName: gpu
       name: example-pgpu
 ```
 
@@ -504,9 +519,8 @@ spec:
     resourceClaimTemplateName: single-gpu
   domain:
     gpus:
-    - claim:
-         name: gpu-resource-claim
-         request: gpu
+    - claimName: gpu-resource-claim
+      requestName: gpu
       name: example-pgpu
 status:
   deviceStatus:
@@ -739,6 +753,11 @@ Refer to https://github.com/kubevirt/community/blob/main/design-proposals/featur
 - Code changes behind feature gate
 - unit tests
 - mock e2e test
+
+Note: there are multiple challenges in writing a real e2e test like, developing a driver implementing kubevirt devices,
+bringing in the driver in kubevirt CI. Before vendors get a chance to develop a driver, there needs to be an API that 
+they can develop against which will not be available until the alpha release. For this reason, writing an e2e test with 
+real driver is outside the scope of alpha and will be handled in beta.
 
 ### Beta
 
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -332,18 +332,19 @@ spec:
 ---
 ```
 
-### Web hook changes
+### Webhook changes
 
-1. Allow DRA devices to be requested only if the corresponding DRA feature gate is enabled in kubevirt configuration
-   Note: All the following sections will assume that DRA feature gate is enabled
+1. Allow the VMI only if vmi spec is correct, i.e. resource claims for requested devices are specified
+
+Note: The feature gate verification when VMI requesting DRA device will either be done in webhook or controller, TBD
 
 
 ### Virt controller changes
 
 1. If devices are requested using DRA, virt controller needs to render the virt-launcher manifest such that
-   pod.spec.resourceClaims and pod.spec.containers.resources.claim sections are filled out.
+   `pod.spec.resourceClaims` and `pod.spec.containers.resources.claim` sections are filled out.
 1. virt-controller needs a mechanism to watch for virt-launcher pods, resourceclaims and resourceslices to populate the
-   vmi.status.deviceStatus using the steps mentioned in above section that has all the attributes (for example the
+   `vmi.status.deviceStatus` using the steps mentioned in above section that has all the attributes (for example the
    pciAddress for the gpu device):
     1. The pod status has information about the allocated/reserved resourceClaim.
     1. The resourceClaim has information about the individual requests in the claim and their allocated device names.
@@ -539,7 +540,7 @@ status:
 
 ## Alternatives
 
-## Alternative 1
+### Alternative 1
 ```
 type GPU struct {
 	// Name of the GPU device as exposed by a device plugin
@@ -707,9 +708,10 @@ Steps to make live migration work:
 
 ## Update/Rollback Compatibility
 
-- Since the feature will be behind a feature gate updates will not be impacted
-- Since the feature will be behind a feature gate, rollback will continue to work as long as feature gate is disabled
-- If the feature is enabled, the VMIs that use DRA devices will have to be cleaned up before attempting rollback.
+- The changes in this design are upgrade compatible
+- Rollback will continue to work as long as feature gate is disabled
+- If the feature is enabled, the VMIs that use DRA devices will have to be deleted and feature gate disabled before 
+  attempting rollback.
 
 ## Functional Testing Approach
 
@@ -735,14 +737,6 @@ An overview on the approaches used to functional test this design)
 1. e2e tests
 1. upgrade downgrade tests
 1. scale tests
-
-Beta Phase:
-- e2e test coverage
-- additional usecases if any
-- updates to API fields if needed
-
-GA:
-- upgrade/downgrade testing
 
 ## Feature lifecycle Phases
 
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -63,7 +63,13 @@ control of their devices using Virtual Machines and Containers.
     1. In the future a standardization could be envisioned for a PCI device where the attributes are set
        automatically through the DRA framework. When this is achieved, any DRA device should be available in KubeVirt VMs
 2. NetworkDevices or Storage Devices which need a lifecycle of its own will not be supported through HostDevices in alpha
-   release
+   release.
+   - For storage, NVMe devices may need to persist beyond pod deletion (e.g., during VM pause). However, current K8s 
+     DRA implementation deletes resource claims when the virt-launcher pod is deleted. Managing claims beyond pod 
+     lifecycle (for example VM pause case) is out of scope for this design. 
+   - For network devices, there is already existing mechanism of configuring the network, any DRA network device will have
+     to work existing functionality. There is a need to explore ideas on how to best achieve it potential in sig-network.
+     Hence, it is out of scope, will have to be picked up as a separate VEP
 
 ## Repos
 
@@ -612,17 +618,17 @@ type VirtualMachineInstanceStatus struct {
 	ResourceClaimStatuses []PodResourceClaimStatus `json:"resourceClaimStatuses,omitempty"`
 }
 type PodResourceClaimStatus struct {
-// Name uniquely identifies this resource claim inside the pod.
-// This must match the name of an entry in pod.spec.resourceClaims,
-// which implies that the string must be a DNS_LABEL.
-Name string `json:"name" protobuf:"bytes,1,name=name"`
-// ResourceClaimName is the name of the ResourceClaim that was
-// generated for the Pod in the namespace of the Pod. If this is
-// unset, then generating a ResourceClaim was not necessary. The
-// pod.spec.resourceClaims entry can be ignored in this case.
-//
-// +optional
-ResourceClaimName *string `json:"resourceClaimName,omitempty" protobuf:"bytes,2,opt,name=resourceClaimName"`
+    // Name uniquely identifies this resource claim inside the pod.
+    // This must match the name of an entry in pod.spec.resourceClaims,
+    // which implies that the string must be a DNS_LABEL.
+    Name string `json:"name" protobuf:"bytes,1,name=name"`
+    // ResourceClaimName is the name of the ResourceClaim that was
+    // generated for the Pod in the namespace of the Pod. If this is
+    // unset, then generating a ResourceClaim was not necessary. The
+    // pod.spec.resourceClaims entry can be ignored in this case.
+    //
+    // +optional
+    ResourceClaimName *string `json:"resourceClaimName,omitempty" protobuf:"bytes,2,opt,name=resourceClaimName"`
 }
 ```
 
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -543,7 +543,7 @@ type HostDevice struct {
 	Tag string `json:"tag,omitempty"`
 	// If specified, the ResourceName of the host device will be provisioned using DRA driver . which will not require the deviceName field
 	//+optional
-	ResourceClaim *ResourceClaim `json:"resourceClaim,omitempty"`
+	ResourceClaim *k8sv1.ResourceClaim `json:"resourceClaim,omitempty"`
 }
 
 // ResourceClaim represents a resource claim to be used by the virtual machine
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -52,6 +52,10 @@ control of their devices using Virtual Machines and Containers.
     2. Mediated device uuid for vGPU.
 1. Devices have to either be of the type GPU or HostDevices.
 
+### Future Usecases
+1. NVMe devices https://github.com/kubevirt/community/issues/254
+1. Live migration of VMIs using DRA devices
+
 ### Unsupported Usecases
 
 1. Devices where the DRA driver does not set the attributes needed to configure libvirt dom XML for the devices will not
@@ -196,7 +200,7 @@ The first section vmi.spec.resourceClaims will have a list of devices needed to 
 available as a list will allow users to use the device from this list in GPU section or HostDevices section of the
 DomainSpec API.
 
-In v1alpha3 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.31.0/resource/v1alpha3#DeviceClaim), multiple drivers
+In v1beta1 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.32.0/resource/v1beta1#DeviceClaim), multiple drivers
 could potentially provision devices that are part of a single claim. For this reason, a separate list of claims required
 for the VMI (section 1) is needed instead of mentioning the resource claim in devices section
 [see Alternate Designs](#alternative-1)
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -136,7 +136,7 @@ type DeviceSource struct {
 	// DeviceName is the name of the device provisioned by device-plugins
 	DeviceName *string `json:"deviceName,omitempty"`
 	// ClaimRequest provides the ClaimName from vmi.spec.resourceClaims[].name and
-	// requestName from resourceClaim.spec.devices.requests[].name
+	// DeviceRequestName from resourceClaim.spec.devices.requests[].name
 	// this fields requires DRA feature gate enabled
 	ClaimRequest *ClaimRequest `json:",inline,omitempty"`
 }
@@ -145,9 +145,9 @@ type ClaimRequest struct {
 	// ClaimName needs to be provided from the list vmi.spec.resourceClaims[].name where this
 	// device is allocated
 	ClaimName string `json:"claimName"`
-	// RequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
+	// DeviceRequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
 	// device is requested
-	RequestName string `json:"requestName"`
+	DeviceRequestName string `json:"deviceRequestName"`
 }
 
 type VirtualMachineInstanceStatus struct {
@@ -410,7 +410,7 @@ spec:
       gpus:
       - name: pgpu
         claimName: pgpu-claim-name
-        requestName: pgpu-request-name
+        deviceRequestName: pgpu-request-name
 status:
   deviceStatus:
     gpuStatuses:
@@ -511,7 +511,7 @@ spec:
   domain:
     gpus:
     - claimName: gpu-resource-claim
-      requestName: gpu
+      deviceRequestName: gpu
       name: example-pgpu
 ```
 
@@ -531,7 +531,7 @@ spec:
   domain:
     gpus:
     - claimName: gpu-resource-claim
-      requestName: gpu
+      deviceRequestName: gpu
       name: example-pgpu
 status:
   deviceStatus:
@@ -769,6 +769,9 @@ real driver is outside the scope of alpha and will be handled in beta.
 - feature gate turned on by default
 - e2e tests with 1 real driver implementation
 - consider additional usecases if any
+- work with kubernetes community to:
+  - find a generic solution for supporting device-plugins styles strings
+  - support discoverable pcie address and mdev UUID attributes
 
 ### GA
 - upgrade/downgrade testing 
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -1,0 +1,760 @@
+# VEP #10: Support DRA devices in KubeVirt
+
+## Release Signoff Checklist
+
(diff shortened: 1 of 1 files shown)


alaypatel07:
along with current reviewer list adding @jean-edouard 

/assign @jean-edouard

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+	// +optional
+	MDevUUID *string `json:"mdevUUID,omitempty"`
+}
+```
+
+The first section vmi.spec.resourceClaims will have a list of devices needed to be allocated for the VM. Having this
+available as a list will allow users to use the device from this list in GPU section or HostDevices section of the
+DomainSpec API.
+
+In v1alpha3 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.31.0/resource/v1alpha3#DeviceClaim), multiple drivers
```
alaypatel07, line 181: v1alpha3 will be depracated soon, we should move to beta APIs, have it on my todo to change this

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+}
+
+type GPU struct {
+	// Name of the GPU device as exposed by a device plugin
+	Name string                    `json:"name"`
+	DeviceSource DeviceSource      `json:",inline"`
+	VirtualGPUOptions *VGPUOptions `json:"virtualGPUOptions,omitempty"`
+	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
+	// +optional
+	Tag string `json:"tag,omitempty"`
```
jean-edouard, line 107: Nit: it would be nice to make it obvious what's old and what's new, like:
```suggestion
	..
	// DeviceName string `json:"deviceName"`    <-- removed in favor of DeviceSource
	DeviceSource DeviceSource      `json:",inline"`
	..
```
same below

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
+}
+
+type GPU struct {
+	// Name of the GPU device as exposed by a device plugin
+	Name string                    `json:"name"`
+	DeviceSource DeviceSource      `json:",inline"`
```
jean-edouard, line 103: Don't we have to keep `DeviceName` around to avoid breaking existing use-cases?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+
+### DRA API for reading device related information
+
+The examples below shows the APIs used to generate the vmi.status.deviceStatuses section:
+1. the pod status has reference to the resourceClaimName, pod.status.resourceClaimStatuses[*].resourceClaimName where
+   the name of the claim is same as vmi.spec.resourceClaims[*].Name
+1. pod spec has node name, pod.spec.nodeName
+1. the resourceclaim status has device name and driver use for allocating the device,
+   resourceclaim.status.allocation.devices[*].deviceName and resourceclaim.status.allocation.devices[*].driver, where
+   resourceclaim.status.allocation.devices[*].request is same as vmi.spec.domain.devices[*].gpus[*].claim.request
```
jean-edouard, line 213: The stars above are interpreted as *italic* for some reason.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+  pool:
+    generation: 0
+    name: kind-1.31-dra-control-plane
+    resourceSliceCount: 1
+---
+```
+
+### Web hook changes
+
+1. Allow DRA devices to be requested only if the corresponding DRA feature gate is enabled in kubevirt configuration
```
jean-edouard, line 315: Not sure if it makes sense to check for the FG in a webhook. In my opinion, VMs with DRA stuff defined should instead fail to start if the FG is missing. That would also cover cases where the FG is set/unset after VM creation.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+type HostDevice struct {
+	Name string `json:"name"`
+	// DeviceName is the resource name of the host device exposed by a device plugin
+	DeviceName string `json:"deviceName"`
+	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
+	// +optional
+	Tag string `json:"tag,omitempty"`
+	// If specified, the ResourceName of the host device will be provisioned using DRA driver . which will not require the deviceName field
+	//+optional
+	ResourceClaim *ResourceClaim `json:"resourceClaim,omitempty"`
```
jean-edouard, line 545: Why not:
```suggestion
	ResourceClaim *k8sv1.ResourceClaim `json:"resourceClaim,omitempty"`
```
That would effectively be the same thing as the main proposal, just without the additional struct and legacy API to maintain.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+    // This is an optional field available only when DRA feature gate is enabled
+    // +optional
+    TargetDeviceStatus *DeviceStatus `json:"targetDeviceStatus,omitempty"`
+}
+```
+
+
+### Alternative 2
+
+In order to handle the live-migration usecase, the following changes are required to the API:
```
jean-edouard, line 642: The API of which object?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+
+Steps to make live migration work:
+1. DRA Status controller will include looking at target pods updating the above fields with device information
+2. Migration Controller currently, moves the migration from Scheduling to Scheduled when the target pod is running. At
+   this point the dra controller should have update the target pod information. Migration Controller should assert that
+   DRA status it sees is from the target pod by comparing the target pod name with podName in the status. The changes
+   should go here: https://github.com/kubevirt/kubevirt/blob/ada65cb7d99033baa3c096820027d08532e25c1e/pkg/virt-controller/watch/migration/migration.go#L624
+3. Once the above is asserted, the migration will continue to flow in the same way as it does today. When target
+   virt-handler will call SyncVMI on the virt-launcher, the virt-launcher will use the same code to convert the
+   deviceStatus into domxml.
```
jean-edouard, line 678: Please add pros and cons for each alternative

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+1. virt-controller will have an additional control loop, however this will use the same shared informers of virt-controller
+   resulting in 0 additional watch calls
+1. virt-controller watch events will be properly filtered such that only changes to relevant status sections of pod and
+   vmi will trigger reconcile loop. Additional metrics for the workqueue of this control loop will be exposed
+1. At max virt-controller should generate 2 additional patch calls to vmi status in a happy path
+
+## Update/Rollback Compatibility
+
+- Since the feature will be behind a feature gate updates will not be impacted
+- Since the feature will be behind a feature gate, rollback will continue to work as long as feature gate is disabled
```
jean-edouard, line 692: What about when the feature gate GA?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+An overview on the approaches used to functional test this design)
+-->
+- Unit tests, new code added will have optimum unit test coverage
+- An e2e test will be added that checks for the following:
+  - API fields are only populated if feature gate is enabled
+  - Once a VMI is created with DRA devices, the right spec values are set by virt-controller in virt-launcher pod
+  - Once the resourceclaims are created and allocated, the correct status values are set
+  - deletion of VMI will lead to release of resources
+  - If GPU clusters are available in CI then e2e tests will be real with a real driver installed
+  - If GPU clusters are not available the e2e tests will be mocked
```
jean-edouard, line 707: I think we need to make sure CI will be able to run at least one un-mocked test

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+Refer to https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#releases for more details
+-->
+
+### Alpha
+
+- Code changes behind feature gate
+- unit tests
+
+### Beta
+-
```
jean-edouard, line 742: empty

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,760 @@
...
+### Alpha
+
+- Code changes behind feature gate
+- unit tests
+
+### Beta
+- 
+- evaluate user and driver authors experience
+- feature gate turned on by default
+- e2e tests
```
jean-edouard, line 745: Why no e2e test in alpha?

jean-edouard (commented):
The main idea sounds good to me, especially since it will be first introduced behind a feature gate.
Added questions/comments/concerns in-line.
Thanks!

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: DeviceName is embedded in DeviceSource, see Line 121.

The verification will be one and only one of DeviceName or Claim fields being non-nil

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: It will be really challenging to support dynamic turning on and turning off of feature gates. 

If a user turns on an alpha feature gate, it is reasonable in my opinion to ask them to delete the resources they used with the feature gate before turning it off. This is the pattern followed by alpha DRA feature gates in k8s as well.

Hence we guard users at webhook level and then leave it up to them to handle turning off the gate with care.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: this is implementation detail, is it okay if we discuss it during code-review?

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: In the current version of this document, update/rollback compatibility is only filled out for alpha stage of the feature. When this feature is in GA, this section will be updated accordingly. 

Is that correct implementation of the VEP process? atleast that is how I have done it with couple of Kubernetes KEPs

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: I am unsure if CI environment with GPU, a driver implementing device allocations all of it will be available in alpha.

Once the rough API gets put up in alpha, vendors can implement in dra plugins accordingly.

Having said that we could potentially have mock e2e tests in alpha not requiring the entire machinery, is that acceptable?

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: @jean-edouard updated, PTAL and resolve this comment if it makes sense

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: removed the stars, kept `[]` to help readers understand its a list

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: is it okay to add pros and cons when live migration feature is implemented? With all the churn that is planned w.r.t. breaking out of live migration API, I am not sure if an elaborate section here will age well

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: removed

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: I answered this is other comment below, in alpha we can have mocked e2e test

Once a driver is available post alpha we could have an actual e2e test if it is acceptable

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated

alaypatel07:
@jean-edouard I have updated with some review comments and responded to other, please let me know if it makes sense, thanks!

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,762 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  labels:
+    kubevirt.io/vm: vm-cirros
+  name: vm-cirros
+spec:
+  resourceClaims:
+  - name: pgpu-claim-name
+    resourceClaimTemplateName: pgpu-claim-template
```
vladikr, line 379: @alaypatel07 this API blows my mind, to be honest. What should a user even request here?  
The `claim.request` in the `devices.gpus` map references something that is obscured and hidden inside the `ResourceClaimTemplate`.

As a user who simply wants to request `nvidia.com/TU104GL_Tesla_T4` what should I write here?

Can you please explain why the template is needed in the VM case? 
Are we planning this claim to be reusable?

Have you considered to reference the deviceClassName directly, either: 
```
spec:
  domain:
    devices:
      gpus:
      - name: gpu1
        resourceClass: nvidia.com/TU104GL_Tesla_T4
``` 
or with a claim: ```
spec:
  resourceClaims:
  - name: gpu-claim
    resourceClass: nvidia.com/TU104GL_Tesla_T4

  domain:
    devices:
      gpus:
      - name: gpu1
        claim:
          resourceClaimName: gpu-claim```

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,762 @@
...
+  domain:
+    devices:
+      gpus:
+      - name: pgpu
+        claim: 
+          name: pgpu-claim-name
+          request: pgpu-request-name
```
xpivarc, line 386: Why do we need to provide yet another name? Couldn't you simply do
```
gpu:
  - name: powerful
    deviceName:  pgpu-claim-name
```

xpivarc: I just noticed that is what Vladik is proposing as well just not reusing the `deviceName`

alaypatel07: > The  pgpu-claim-name would be resolved to claim which could have multiple sources or to old device plugin.

I am not sure if this is doable. If the claim have multiple sources, how do you know if its a pgpu/vgpu? Additionally for vgpus, there are other configurations in the API that has to be applied https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/schema.go#L595

So we have to be able to map a single entry in this list to a single device in the resource claim which is only available with `requestName`

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: synced with @jean-edouard on Slack, got an agreement that the thought process here is valid.

veps/sig-compute/10-dra-devices/vep.md
vladikr: That seems to be even better. 
```
resourceClaims: 
  - name: pgpu-claim-name
    resourceClass: nvidia.com/TU104GL_Tesla_T4
```

With that approach the `resourceClaim` may have multiple sources and would allow users either directly reference a `reosurceClass` (maybe it has a different name) or a `resourceClaimTemplate`

The idea is to mimic the approach we have with `Disks` and `Volumes` where the latter can reference different sources. This also separates virtual (Disks) and physical (Volumes) resources.

The fallback (when the `gpu.deviceName` doesn't reference a `resourceClaim` ) is to use a device plugin.

@xpivarc did I capture your idea correctly?

alaypatel07: sorry guys, I lost the proposed API here, can you please construct the entire spec for both device plugins and DRA devices?

alaypatel07: @vladikr are you suggesting something like this for DRA?
```
spec:
  resourceClaims: 
    - name: pgpu-claim-name
      resourceClass: nvidia.com/TU104GL_Tesla_T4  # Referencing resourceClass directly instead of requestName
  domain:
    devices:
      gpus:
        - name: pgpu
          claim:
            name: pgpu-claim-name  # Linking GPU to the claim
```
and the following for device plugins, (this is the current mechanism no change here)
```
apiVersion: kubevirt.io/v1
kind: VirtualMachineInstance
metadata:
  labels:
    kubevirt.io/vm: vm-cirros
  name: vm-cirros
spec:
  domain:
    devices:
      gpus:
        - name: pgpu
          deviceName: nvidia.com/TU104GL_Tesla_T4
```

varunrsekar:
```
        claim: 
          name: pgpu-claim-name
          request: pgpu-request-name
```
Folks, just to clarify:
A single resourceclaim can allocate multiple devices of different types. It does so with requests. And each request may be for different devices: ie 1 for a gpu, 1 for a nvme etc. And hence if we want to assign an allocated device from a resource claim, we need to know the resourceclaim name and then the request name inside it.
(See https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/resource#DeviceClaim)

vladikr: > @vladikr are you suggesting something like this for DRA?

No. What I'm suggesting is to keep the current `domain.devices.gpus` structure and let `deviceName` either reference a device (fallback to a device plugin scenario) or a `resourceClaim`. 

This API would mimic the `Disks` and `Volumes` API, where `Volumes` can reference multiple sources `PersistentVolumeClaim, ContainerDisk, HostDisk...` ([see](https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/schema.go#L783))

With this approach, we could claim a specific device directly:

```
spec:
  resourceClaims: 
    - name: pgpu-claim-name
      resourceClass: nvidia.com/TU104GL_Tesla_T4 

  domain:
    devices:
      gpus:
        - name: pgpu
          deviceName: pgpu-claim-name
```

OR reference a template as an expansion:

```
spec:
  resourceClaims: 
    - name: pgpu-claim-name
      resourceClaimTemplateName: pgpu-claim-template 

  domain:
    devices:
      gpus:
        - name: pgpu
          deviceName: pgpu-claim-name
```

OR a fallback to a device plugin when there is no claim present

```
spec:
  domain:
    devices:
      gpus:
        - name: pgpu
          deviceName: nvidia.com/TU104GL_Tesla_T4
```

alaypatel07: @vladikr The issue with both the above suggested approaches is that a ResourceClaim object could have multiple devices in its requests, as an example:
```
apiVersion: resource.k8s.io/v1alpha3
kind: ResourceClaimTemplate
metadata:
  namespace: vgpu-test
  name: mixed-gpus-template
spec:
  spec:
    devices:
      requests:
        - name: h100
          deviceClassName: h100.nvidia.com
        - name: a100
          deviceClassName: a100.nvidia.com
```

Now this is a simplistic usecase which could be satisfied by other means, the real benefit of the proposed API will be for performance computing in the future when sriov/rdma/nvme devices are supported. User could have a ResourseClaimTemplate that could request devices aligning on a specific numa or pcie, for example
```
apiVersion: resource.k8s.io/v1alpha3
kind: ResourceClaimTemplate
metadata:
  namespace: high-performance-gpu-nic
  name: pcie-aligned-gpu-nic-template
spec:
  spec:
    devices:
      requests:
        - name: gpu
          deviceClassName: gpu.nvidia.com    
        - name: nic
          deviceClassName: rdma.nvidia.com
      constraints:
      - requestNames: ["gpu", "nic"]
         matchAttribute: "k8s.io/pcieRoot"
``` 

The proposed API change where requestName is *not* mentioned in kubevirt VMI spec, will mean that these advanced usecases will be harder to achieve in the future.

alaypatel07: this was discussed in depth, summary here: https://docs.google.com/document/d/1bQdLoxwSC1ILvyIb4ljSm5RZmsVpvqfWKIbN0KOBKsw/edit?tab=t.0#heading=h.jrm4uta73jr6 

Everyone agreed on the following API:
```
spec:
  resourceClaims:
  - name: pgpu-claim-name
    resourceClaimTemplateName: pgpu-claim-template
  domain:
    devices:
      gpus:
      - name: pgpu
        claimName: pgpu-claim-name
        requestName: pgpu-request-name
```

I will reflect this in the VEP and resolve this thread

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,780 @@
...
+
+DRA adoption is important for KubeVirt so that vendors can expect the same
+control of their devices using Virtual Machines and Containers.
+
+## Goals
+
+- Align on the API changes needed to consume DRA enabled devices in KubeVirt
+- Align on how KubeVirt will consume devices by external DRA drivers
+
+## Non Goals
```
rthallisey, line 25: Add or reduce the DRA API complexity in KubeVirt

alaypatel07: Added

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated as per discussion on 03/05

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,781 @@
...
+- As a user, I would like to use my GPU dra driver with KubeVirt.
+- As a user, I would like to use KubeVirt's default driver.
+- As a user, in heterogeneous clusters, i.e. clusters made of nodes with different hardware managed through DRA drivers,
+  I should be able to easily identify what hardware was allocated to the VMI.
+- As a developer, I would like APIs to be extensible so I can develop drivers/webhooks/automation for custom use-cases.
+- As a device-plugin author, I would like to have a well documented way, intuitive way to support devices in KubeVirt.
+
+## Use Cases
+
+### Supported Usecases
```
xpivarc, line 48: While migration cannot be done today, let's have it as intention to support it in the future.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,781 @@
...
+	// DeviceName string `json:"deviceName"`    <-- inlined into DeviceSource
+	DeviceSource DeviceSource `json:",inline"`
+	// If specified, the virtual network interface address and its tag will be provided to the guest via config drive
+	// +optional
+	Tag string `json:"tag,omitempty"`
+}
+
+type DeviceSource struct {
+	// DeviceName is the name of the device provisioned by device-plugins
+	DeviceName *string `json:"deviceName,omitempty"`
```
xpivarc, line 127: This needs to be left untouched imho

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,781 @@
...
+}
+
+type ClaimRequest struct {
+	// ClaimName needs to be provided from the list vmi.spec.resourceClaims[].name where this
+	// device is allocated
+	ClaimName string `json:"claimName"`
+	// RequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
+	// device is requested
+	RequestName string `json:"requestName"`
+}
```
xpivarc, line 141: Do we have the API that was discussed on the last meeting?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,781 @@
...
+	// +optional
+	MDevUUID *string `json:"mdevUUID,omitempty"`
+}
+```
+
+The first section vmi.spec.resourceClaims will have a list of devices needed to be allocated for the VM. Having this
+available as a list will allow users to use the device from this list in GPU section or HostDevices section of the
+DomainSpec API.
+
+In v1alpha3 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.31.0/resource/v1alpha3#DeviceClaim), multiple drivers
```
xpivarc, line 199: We should build this on the beta that is present now as far I see

alaypatel07: +1, will update to beta.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: This was the agreed upon API from the meeting.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: From user perspective this is untouched. From a go client perspective, we are expressing an intent that the source either device-plugins or dra.

What is the preference? Have all the fields in GPU struct without the inling?

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: The volume API has this pattern, where the source of is specified by an inline struct `VolumeSource`, https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/schema.go#L776, this was the inspiration for introducing `DeviceSource`

veps/sig-compute/10-dra-devices/vep.md
jean-edouard: My point remains that the feature gate should be checked on VM start, not creation.
I agree that we don't want to fully support dynamic feature gates switches and all the edge cases that come with it, but I still disagree with having the check in a webhook, i.e. on creation. It should be just as simple, and much more elegant, to reject VM start when the necessary FG is not set.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,785 @@
...
+1. Allow DRA devices to be requested only if the corresponding DRA feature gate is enabled in kubevirt configuration
+   Note: All the following sections will assume that DRA feature gate is enabled
+
+
+### Virt controller changes
+
+1. If devices are requested using DRA, virt controller needs to render the virt-launcher manifest such that
+   pod.spec.resourceClaims and pod.spec.containers.resources.claim sections are filled out.
+1. virt-controller needs a mechanism to watch for virt-launcher pods, resourceclaims and resourceslices to populate the
+   vmi.status.deviceStatus using the steps mentioned in above section that has all the attributes (for example the
```
jean-edouard, line 346: Formatting nit (more below):
```suggestion
1. If devices are requested using DRA, virt controller needs to render the virt-launcher manifest such that
   `pod.spec.resourceClaims` and `pod.spec.containers.resources.claim` sections are filled out.
1. virt-controller needs a mechanism to watch for virt-launcher pods, resourceclaims and resourceslices to populate the
   `vmi.status.deviceStatus` using the steps mentioned in above section that has all the attributes (for example the
```

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,785 @@
...
+    gpuStatuses:
+    - name: example-pgpu
+      deviceResourceClaimStatus:
+        name: gpu-0
+        resourceClaimName: virt-launcher-vmi-fedora-hhzgn-gpu-resource-claim-c26kh
+        attributes:
+          pciAddress: 0000:01:00.0
+```
+
+## Alternatives
```
jean-edouard, line 540: I don't think we need those anymore, this VEP is already long enough IMO. Up to you though

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,785 @@
...
+
+## Alternative 1
```
jean-edouard, line 542:
```suggestion
### Alternative 1
```

veps/sig-compute/10-dra-devices/vep.md
jean-edouard: I think we should at least have a plan for when the feature gate gets graduated out.
Otherwise this entire document could be "don't worry about this, it's all behind a feature gate anyway!" :)
If any part of this design just doesn't possibly play well with upgrades, it'd be great to find out now and not later (even if API changes of alpha features are fair game).
This is why we expressed concerns about that string being moved to an inline struct for example.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,785 @@
...
+1. upgrade downgrade tests
+1. scale tests
+
+Beta Phase:
+- e2e test coverage
+- additional usecases if any
+- updates to API fields if needed
+
+GA:
+- upgrade/downgrade testing
```
jean-edouard, line 745:
```suggestion
```
Those are detailed below

jean-edouard (commented):
Almost there! Added a few comments in-line.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,785 @@
...
+This design misses the use-case where more than one DRA device is specified in the claim template, as each
+device will have its own template in the API.
+This design also assumes that the deviceName will be provided in the ClaimParameters, which requires the DRA drivers
+to have a ClaimParameters.spec.deviceName in their spec.
+
+## Alternative 2
+Asking the dra plugin authors to inject env variable to CDI spec.
+In order to uniquely identify the device required by the vmi spec, the follow env variable will have to be constructed:
+```
+PCI_RESOURCE_<RESOURCE-CLAIM-NAME>_<REQUEST-NAME>="0000:01:00.0"
```
vladikr, line 595: I don't think the RESOURCE-CLAIM-NAME is needed here. This can continue to be PCI_RESOURCE_<RESOURCE_NAME> or PCI_RESOURCE_<requestName> I don't see why a reference to claim is needed.

veps/sig-compute/10-dra-devices/vep.md
vladikr: VM start and VMIs creation?

jean-edouard: Good point yeah, and since a VM start is effectively a VMI creation, we could indeed make it all work just by adding to the VMI admission webhook.
However, we try not to add webhooks that depend on external components, which is why I think the VMI controller is a better location for the check.

alaypatel07: +1 VM start and VMIs creation works for me, I will add that to the document

veps/sig-compute/10-dra-devices/vep.md
jean-edouard: If it has been verified that moving a field into an inline struct doesn't cause any trouble on upgrade, which sounds like it's the case, with even a precedent (`VolumeSource`), I think we're good!

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: I am inclined to keep the alternatives, for future us to remember the long discussions that went into this design.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: 1. `PCI_RESOURCE_<RESOURCE_NAME>`

I assume by RESOURCE_NAME you mean the GPU name in the following spec:
```
spec:
  resourceClaims:
  ...
  domain:
    devices:
      gpus:
      - name: pgpu
```

so the proposed env var will be PCI_RESOURCE_PGPU here.

This will not work because vmi spec is not available in dra plugin, DRA plugin authors will not have any idea of what name VMI user has set for the device

2. `PCI_RESOURCE_ ` this will not work because there are multiple devices in the same resource claim, and each device could be from a different driver, how can virt-launcher know which device is PCI resource and which one is MDEV, even this will not work

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: removed

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated as requested, PTAL

jean-edouard (approved):

vladikr:
@alaypatel07 Thank you for all of this work!
We had many discussions about the API and I don't see any better ideas, at this point.

I still have two main concerns:

1. The API puts a lot of burden on the user. 
     Instead of being able to request a specific device, the user needs to be aware of all the added complexity associated with creating a resource claim template. 
     I would prefer that we have an API that puts the VM user first. Let the user express the desired state in the VM and allow KubeVirt to craft the template on the user's behalf.
     We should reevaluate this API before going to beta. Hopefully, some changes will occur in the DRA itself and we could readjust.
2. Having a list (metadata) of allocated devices inside the virt-launcher allowed it to configure the relevant devices without the need to synchronize these via the VMI object. 
     Relying on what has been allocated for a given container simplifies the creation and migration flows, without involving multiple components.

I really hope we can explore and address both of these issues before beta.

However, we need to start somewhere. 
/approve
/hold

@iholder101 feel free to remove the hold when you think it's ready.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,779 @@
...
+	// +optional
+	MDevUUID *string `json:"mdevUUID,omitempty"`
+}
+```
+
+The first section vmi.spec.resourceClaims will have a list of devices needed to be allocated for the VM. Having this
+available as a list will allow users to use the device from this list in GPU section or HostDevices section of the
+DomainSpec API.
+
+In v1beta1 version of [DRA API](https://pkg.go.dev/k8s.io/api@v0.32.0/resource/v1beta1#DeviceClaim), multiple drivers
```
alicefr, line 203: @alaypatel07 what actually means that multiple device could be part of the same claim? Does this grouping some how translate into different xml if the devices are part of the same claim or different ones?

alicefr:
@alaypatel07 do we plan to actually implement a dra driver in kubevirt or only the support for external dra drivers?

alaypatel07:
@alicefr the driver part of this conversation is explicit non-goal for this VEP.  

The current plan is to support external drivers that can publish attributes in a specific format, but we need to work with k8s wg-device-management to finalize the format of attributes like pcie address or mdev uuid.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: Yeah the devices will have unique names in status and their unique xml translations to either pair or mdev device.

veps/sig-compute/10-dra-devices/vep.md
iholder101: I'm sorry, but I'm not sure why @vladikr's [suggestion](https://github.com/kubevirt/enhancements/pull/11#discussion_r1980238167) is problematic. The google doc you've linked says `There was an agreement that we need a gully qualified resource claim, and request name for getting to the right device`, but it doesn't list any reasons for this decision.

 We already have the `deviceName` field, why not re-use it to pick one of the `ResourceClaim`'s devices, and fallback to the old mechanism if such a resource claim does not exist?

iholder101 (commented):
@alaypatel07 thank you very much for this proposal!
It is very detailed and well written, I hope this serves as an example for others!

Left a few small comments, let me know what you think. In general, I like this direction very much.

Thank you!

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,779 @@
...
+	ClaimRequest *ClaimRequest `json:",inline,omitempty"`
+}
+
+type ClaimRequest struct {
+	// ClaimName needs to be provided from the list vmi.spec.resourceClaims[].name where this
+	// device is allocated
+	ClaimName string `json:"claimName"`
+	// RequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
+	// device is requested
+	RequestName string `json:"requestName"`
```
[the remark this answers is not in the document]
iholder101: Please see my comment [here](https://github.com/kubevirt/enhancements/pull/11#discussion_r2030050205).

iholder101:
> @alaypatel07 Thank you for all of this work! We had many discussions about the API and I don't see any better ideas, at this point.
> 
> I still have two main concerns:
> 
> 1. The API puts a lot of burden on the user.
>    Instead of being able to request a specific device, the user needs to be aware of all the added complexity associated with creating a resource claim template.
>    I would prefer that we have an API that puts the VM user first. Let the user express the desired state in the VM and allow KubeVirt to craft the template on the user's behalf.
>    We should reevaluate this API before going to beta. Hopefully, some changes will occur in the DRA itself and we could readjust.
> 2. Having a list (metadata) of allocated devices inside the virt-launcher allowed it to configure the relevant devices without the need to synchronize these via the VMI object.
>    Relying on what has been allocated for a given container simplifies the creation and migration flows, without involving multiple components.
> 
> I really hope we can explore and address both of these issues before beta.
> 
> However, we need to start somewhere. /approve /hold
> 
> @iholder101 feel free to remove the hold when you think it's ready.

I share these concerns.
@alaypatel07 do you mind listing those as exploration points for the beta section please, so we can track it?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -0,0 +1,779 @@
...
+1. Live migration of VMIs using DRA devices
+
+### Unsupported Usecases
+
+1. Devices where the DRA driver does not set the attributes needed to configure libvirt dom XML for the devices will not
+   be supported.
+    1. In the future a standardization could be envisioned for a PCI device where the attributes are set
+       automatically through the DRA framework. When this is achieved, any DRA device should be available in KubeVirt VMs
+2. NetworkDevices or Storage Devices which need a lifecycle of its own will not be supported through HostDevices in alpha
+   release
```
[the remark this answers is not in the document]
varunrsekar: I believe the idea was that network/storage devices have their own complexities in getting made usable with kubevirt. Enabling DRA support for generic HostDevices might not be sufficient to work with these devices. We're just explicitly calling that we dont claim to support them in this proposal. Beyond alpha, there may be scope to explore this. But its non-goals for this document.

veps/sig-compute/10-dra-devices/vep.md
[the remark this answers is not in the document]
alaypatel07: @iholder101 you can find detailed discussion here: https://youtu.be/EHZIbSH6Knk

but I will summarize here as well. At the end of a long discussion we ended up arriving at two choices:
```
  domain:
    gpus:
    - claimName: gpu-resource-claim
      requestName: gpu
      name: example-pgpu
 ```
 or 
 ```
   domain:
    gpus:
    - claimName: gpu-resource-claim
      deviceName: gpu
      name: example-pgpu
 ```
 
You are recommending the second choice, but we ended up preferring the first because of the following reasons:
1. the DRA fields in VMI spec will be alpha fields, the DeviceName field is already associated with a different feature and it has a different lifecycle. I would rather keep the device plugins and DRA fields in vmi spec separate to manage DRA feature lifecycle independently. It will be very helpful with upgrades, downgrades and field evaluation with alpha, beta and GA phases.

2. At an implementation level both the API's look the same, its just the matter of re-using an existing field or introducing a new fields. Since first solution is adding a field that maps directly to DRA fields, users can continue to be in that context of DRA and how it will be translated to pod spec, instead of learning that an existing field will be re-used. This also gives us an advantage that anytime upstream k8s DRA will solve the problem of translating device plugins type strings into DRA spec, we will have many options of solving that in kubevirt. If we re-use the existing device-plugin field there will be limited opportunities to get the translation right.

Long story short, the approach we took in this design is to be as close to the DRA apis as possible and leave out options for making the user-experience better for beta releases when we have more information on how DRA in k8s is evolving.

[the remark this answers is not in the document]
jean-edouard: > I find the name `RequestName` extremely confusing.
> 
> As your comment suggests, this field represents one of `resourceClaim.spec.devices.requests[].name`, so IIUC this refers to a device name. Can we call it `DeviceName` or `ClaimDeviceName`?

Parsing this tells me it's the name of a device request, so I guess it's both. Would `deviceRequestName` work for everyone by any chance? :)
I personally have no objection against `requestName`.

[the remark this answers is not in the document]
iholder101: > Parsing this tells me it's the name of a device request, so I guess it's both. Would `deviceRequestName` work for everyone by any chance? :) I personally have no objection against `requestName`.

Yes, `deviceRequestName` sounds great to me.
@alaypatel07 it also does not go against the reasoning you've outlined above, as it's a new field that's only being used for DRA. It is also much clearer to understand what this field is referring to.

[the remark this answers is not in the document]
alaypatel07: updated to using `deviceRequestName`, PTAL

alaypatel07:
@iholder101 I have addressed your feedback in separate commits, please take another look, thanks!

iholder101:
> @iholder101 I have addressed your feedback in separate commits, please take another look, thanks!

@alaypatel07, thank you very much for this proposal. It is detailed and well-written and should serve as an example to other proposals. Thanks for the dedication and patience!

/lgtm
/hold cancel

PR #15, opened 2025-03-06
Enhancement: IBM Z Secure Execution

jschintag:
Add proposal for implementing Secure Execution

Fixes #

**Checklist**

- [x] Design: A [design document](https://github.com/kubevirt/community/tree/main/design-proposals) was considered and is present (link) or not required
- [x] PR: The PR description is expressive enough and will help future contributors
- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [x] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08.html)
- [x] Upgrade: Impact of this change on upgrade flows was considered and addressed if required
- [x] Testing: New code requires [new unit tests](https://github.com/kubevirt/kubevirt/blob/main/docs/reviewer-guide.md#when-is-a-pr-good-enough). New features and bug fixes require at least on e2e test
- [x] Community: Announcement to [kubevirt-dev](https://groups.google.com/g/kubevirt-dev/) was considered

**Release note**:

```release-note
NONE
```

### DIFF
--- a/veps/sig-compute/secure-execution.md
+++ b/veps/sig-compute/secure-execution.md
@@ -1,0 +1,61 @@
+# Overview
+Implement support for IBM Z Secure Execution VMs in kubevirt
+
+## Motivation
+To enable customers to deploy secure workloads on their IBM Z machines using kubevirt.
+
+## Goals
+To enable kubevirt to run Secure Execution VMs on s390x
+
+## Non Goals
+Changing how existing VMs run.
+
+## Definition of Users
+- IBM Z users wanting to secure their new VMs using Secure Execution
+- IBM Z users wanting to migrate existing Secure Execution VMs to kubevirt
+
+## User Stories
+(list of user stories this design aims to solve)
+
+## Repos
+https://github.com/kubevirt/kubevirt
+https://github.com/kubevirt/api
+
+# Design
+
+The feature should be behind a Feature Gate at first, similiar to AMD SEV.
+For implementing this feature, the api behaviour would need to be changed to allow for setting an empty `launchSecurity: {}`.
+
+For launching Secure Execution VMs, all devices need to have `iommu=on` set to allow them to be visible to the guest, as in Secure Execution Mode the hypervisor is prevented from accessing the guest memory. Otherwise Secure Execution is initiaded from the guest image during boot, not the hypervisor.
+
+To ensure that VMs are scheduled only on nodes supporting Secure Execution, node-labeller should add a label to these nodes. This can be done by parsing the output of `virsh domcapabilities`.
+When a Secure Execution VM is created, the SE label will then be added to the virt-launcher pod so that it will be scheduled on a node which has the feature enabled.
+
+Due to the access restriction of the guest memory, live migration is not possible. Therefore the `NonMigratableCondition` will be added to the VMs.
+
+## API Examples
+```
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  ...
+spec:
+  domain:
+    launchSecurity: {}
+  ...
+```
+
+## Scalability
+It should scale like normal VMs.
+The only restriction would be for Secure Execution VMs, as they ´can't run on IBM Z machines that they have not been encrypted with.
+To move a Secure Execution VM to a new IBM Z machine, it's public key would first need to be added to the keys used to encrypt the kernel+initramfs+parmfile.
+
+## Security
+This design improves Security by allowing users to protect their IBM Z Workloads with Secure Execution.
+
+## Update/Rollback Compatibility
+This feature would not affect existing/new VMs as long as they do not depent on it.
+
+## Functional Testing Approach
+- As much as possible should the feature be tested in unit-tests
+- Upstream tests would be difficult and likely require additional hardware, as Secure Execution does not work with nested virtualization and requires being run on LPAR (Bare Metal).


vladikr:
@jschintag can you please open an issue to track this?

jschintag:
> @jschintag can you please open an issue to track this?

I have created https://github.com/kubevirt/enhancements/issues/19

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,64 @@
...
+## API Examples
+```
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  ...
+spec:
+  domain:
+    launchSecurity:
+      secureExecution: {}
```
vladikr, line 40: I'm worried about an accumulation of CPU/Arch specific fields here.
I wonder if instead of using multiple dedicated fields we could simplify it by using a generic term, such as `secureExecution`. 
Then we could have a generic "SecureExecution" device plugin that would have value on any node that has support for either one of the technologies. Once the vmipod is scheduled to the node it would pass the correct device. For migrations, it could work the same way as it works for host-model cpu.

@alicefr what do you think?

vladikr:
@jschintag Thank you for this proposal.

 On one hand, I wish we could find a way to generalize the `secureExecution` field, instead of pilling up fields that represent various technologies. Users could simply express that they want a secure execution and KubeVirt could be smart enough to choose the correct technology based on the CPU/Arch. Have you given any thought to something like this?
I've discussed this with @jean-edouard recently.

On the other hand, we could think about generalization in the next steps.

jschintag:
> @jschintag Thank you for this proposal.
> 
> On one hand, I wish we could find a way to generalize the `secureExecution` field, instead of pilling up fields that represent various technologies. Users could simply express that they want a secure execution and KubeVirt could be smart enough to choose the correct technology based on the CPU/Arch. Have you given any thought to something like this? I've discussed this with @jean-edouard recently.
> 
> On the other hand, we could think about generalization in the next steps.

I have to admit i haven't thought to deep about it, but i already see 2 problems.
The first is admittedly solvable, it is just the way API versioning should be done, changing things to a generic `launchSecurity` would mean we would likely need to remove or deprecate the `SEV` field. This would require a new API Version. Adding a new field on the other can be done without changing the API version, as long as the default behavior does not break older implementations.
The second issue i see is that AMD SEV seems to have configuration options, while Secure Execution does not really need them.

veps/sig-compute/secure-execution.md
jschintag: I would add that the generic should if we choose to do it be `launchSecurity`, since that is what libvirt is using.
Also while i don't know about AMD SEV, but IBM Z Secure Execution is a CPU feature, so while it needs special permissions in the virt-launcher Pod, it does not really need a device plugin.

vladikr:
> > @jschintag Thank you for this proposal.
> > On one hand, I wish we could find a way to generalize the `secureExecution` field, instead of pilling up fields that represent various technologies. Users could simply express that they want a secure execution and KubeVirt could be smart enough to choose the correct technology based on the CPU/Arch. Have you given any thought to something like this? I've discussed this with @jean-edouard recently.
> > On the other hand, we could think about generalization in the next steps.
> 
> I have to admit i haven't thought to deep about it, but i already see 2 problems. The first is admittedly solvable, it is just the way API versioning should be done, changing things to a generic `launchSecurity` would mean we would likely need to remove or deprecate the `SEV` field. This would require a new API Version. Adding a new field on the other can be done without changing the API version, as long as the default behavior does not break older implementations. The second issue i see is that AMD SEV seems to have configuration options, while Secure Execution does not really need them.

Thanks!

KubeVirt didn't commit to this API, yet. SEV is in the Alpha stage, so it is not a stable API. We should be able to change it if we have a better one.

Regarding specific configuration, I'm not sure about this point. I wonder if it make sense to have a generic configuration and the ability to expand it to a specific one for those who need it.

```
spec:
  domain:
    launchSecurity:
      secureExecution: { specificConfig: Here }
```

jschintag:
> KubeVirt didn't commit to this API, yet. SEV is in the [Alpha](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go#L85) stage, so it is not a stable API. We should be able to change it if we have a better one.

I guess than that works.

> Regarding specific configuration, I'm not sure about this point. I wonder if it make sense to have a generic configuration and the ability to expand it to a specific one for those who need it.

I think one option for a more generic one would be to make it close to libvirt xml?
See: https://libvirt.org/formatdomain.html#launch-security

jschintag:
Just wanted to say I'm going on vacation next week.
@vamsikrishna-siddu and @basavaraju-g will be available for discussions on this topic and can make changes to my PRs.

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,64 @@
...
+Rolling back to a version without Secure Execution would prevent Secure Execution VMs from working.
+
+## Functional Testing Approach
+- As much as possible should the feature be tested in unit-tests
+- As part of the downstream Openshift Virtualization process it would be regulary tested
+- Upstream tests would be difficult and likely require additional hardware, as Secure Execution does not work with nested virtualization and requires being run on LPAR (Bare Metal).
+
+# Implementation Phases
+1. Implementation with Feature Gate: https://github.com/kubevirt/kubevirt/pull/14040
+2. Potentially graduating in future versions.
```
iholder101, line 64: This section is confusing :) it actually refers to implementations that have already happened.
I've created a PR to clarify this: https://github.com/kubevirt/enhancements/pull/30.

So if you don't mind please remove this section, and later update it with implementation worth noting.

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Update/Rollback Compatibility
+As the feature would need to be enabled in a VM before it takes effect, it would not affect existing or new VMs that do not use it.
+
+Rolling back to a version without Secure Execution would prevent Secure Execution VMs from working.
+
+## Functional Testing Approach
+- As much as possible should the feature be tested in unit-tests
+- As part of the downstream Openshift Virtualization process it would be regulary tested
+- Upstream tests would be difficult and likely require additional hardware, as Secure Execution does not work with nested virtualization and requires being run on LPAR (Bare Metal).
```
iholder101, line 60: I wonder if testing it properly upstream should be a requirement for this feature, at least at the Beta stage.

veps/sig-compute/secure-execution.md
jschintag: I have the removed the Section for now.

veps/sig-compute/secure-execution.md
jschintag: > 2\. I wonder how different the parameters are for the different technologies. IOW how different the configs for AMV SEV, Intel TDX and IBM secure execution are.

I would naively just map it 1:1 to what libvirt uses in the [xml](https://libvirt.org/formatdomain.html#launch-security). That way the differences should not matter.  I mean IBM Secure Execution does not use any of the parameters except type anyway to my knowledge.

vladikr:
> > KubeVirt didn't commit to this API, yet. SEV is in the [Alpha](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/featuregate/active.go#L85) stage, so it is not a stable API. We should be able to change it if we have a better one.
> 
> I guess than that works.
> 
> > Regarding specific configuration, I'm not sure about this point. I wonder if it make sense to have a generic configuration and the ability to expand it to a specific one for those who need it.
> 
> I think one option for a more generic one would be to make it close to libvirt xml? See: https://libvirt.org/formatdomain.html#launch-security

@lyarwood I wonder if you have any thoughts on this?

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,59 @@
...
+  ...
+```
+
+## Scalability
+It should scale like normal VMs.
+The only restriction would be for Secure Execution VMs, as they ´can't run on IBM Z machines that they have not been encrypted with.
+To move a Secure Execution VM to a new IBM Z machine, it's public key would first need to be added to the keys used to encrypt the kernel+initramfs+parmfile.
+
+## Security
+The virt-launcher pod would currently need to be run as privileged for Secure Execution VMs. Libvirt checks for `/sys/firmware/uv` when verifying that the Host supports Secure Execution VMs. Unfortunately `/sys/firmware` is intentionally empty in unprivileged pods and kubernetes prevents mounting anything under this path. There are potentially Feature Gates in kubelet to allow this, however i have not yet been able to get this to work. From what i gathered you would need to at least enable `ProcMountType` and `UserNamespacesSupport` to be able to launch a Secure Execution VM inside of an unprivileged pod.
```
alicefr, line 50: This seems to be a huge no-go. I think we have dropped the root support for VMs

alicefr: We should not touch or require any unsecure kubernetes feature gate or option. You could try to implement a device plugging for secure execution which mounts those directory in the pod. If it isn't a DP then a similar mechanism

veps/sig-compute/secure-execution.md
alicefr: If we manage somehow to have some sort of plugging, we could skip the upstream functional testing, and this will be entirely supported by IBM. @jschintag can you please take a look to the network pluggin interface and see if you could do something similar?
For example the VM could request secure-excution as resource, your device plugin mounts the needed directories and env variables in the pod, and then the sidecar adds the extra xml for secure execution.

veps/sig-compute/secure-execution.md
jschintag: It is still in there for ppc64le

veps/sig-compute/secure-execution.md
jschintag: Unfortunately that would still not solve the current issue of the pod needing to be run as privileged for access to `/sys/firmware` -> https://github.com/kubevirt/enhancements/pull/15#discussion_r2026975195

veps/sig-compute/secure-execution.md
jschintag: We could maybe do something more generic by using maps:
```
type LaunchSecurity struct {
    Type string
    Properties map[string]string
}
```

veps/sig-compute/secure-execution.md
alicefr: you can do the bindmounts from the privileged DP. Not very kubernetes friendly, but I'm pretty sure you can do it. 
For device pluging you can specify the hostpath and the container path. So, at least, I would like to have those 2 ways explored. 
As I said, running VMs as privileged makes this proposal not acceptable, so we need to find an alternative to it

veps/sig-compute/secure-execution.md
vladikr: @alicefr Yes, I understand your point. It concerned me that we will have multiple of these structs. The number will obviously grow in the future. In many cases, the users won't care specifically about the technology that is being used to achieve this functionality. 
On the other hand, we need to increase the number of validations we do, to ensure that various combinations are sensible.
I was hoping we could somehow generalize the API so we could have some kind of an envelope for a generic API and it could remain empty for users who don't care about specific configuration but also can be expanded for those who care. 
A pointer to an external CRD is an interesting idea.

veps/sig-compute/secure-execution.md
alicefr: Device plugin works since we are using it for mounting a socket for example. See [here](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/device-manager/socket_device.go#L169). Bind mounts we use them for container and hotplug disks. The right tool to take a look at is [virt-chroot](https://github.com/kubevirt/kubevirt/tree/main/pkg/virt-handler/virt-chroot). You can see as example, the [container disks](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/container-disk/mount.go) and [hotplug disk](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/container-disk/mount.go) in the handler

jschintag:
I have updated the proposal to avoid using privileged pods. @alicefr would you mind taking a look to see if this works for you?

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,60 @@
...
+## Repos
+https://github.com/kubevirt/kubevirt
+https://github.com/kubevirt/api
+
+# Design
+For implementing this feature, the api would need to be expanded to have a Secure Execution option under Launch Security next to AMD SEV.
+Similiar to SEV it would be first behind a Feature Gate.
+
+Afterwards kubevirt would need to be expanded to handle setting the correct Launch Security type for Secure Execution.
+To achieve this, the virt-launcher pod would currently need to be run as privileged for Secure Execution VMs, as Libvirt checks for `/sys/firmware/uv` when verifying that the Host supports Secure Execution VMs. Unfortunately `/sys/firmware` is intentionally empty in unprivileged pods and kubernetes prevents mounting anything under this path. There are potentially Feature Gates in kubelet to allow this, however i have not yet been able to get this to work. From what i gathered you would need to at least enable `ProcMountType` and `UserNamespacesSupport` to be able to launch a Secure Execution VM inside of an unprivileged pod.
```
alicefr, line 29: Is this required or not? I don't really understand from the sentence

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,60 @@
...
+To avoid using privileged pods, the first implementation should explicitly not add launchSecurity to the xml file, but instead do the needed changes in the xml directly. This ensures that libvirt does not check if Secure Execution is supported, enabling virt-launcher to run unprivileged. The drawback of this is of course, that kubevirt needs to ensure all devices use iommu, otherwise they will not be visible to the VM, leading to potentially hard to debug, unkown behaviour and errors.
```
alicefr, line 30: Is this using sidecars for changing the XML?
In any case, it sounds like a bug in libvirt. If libvirt is running in session mode, it should assume that the environment has already been prepared

veps/sig-compute/secure-execution.md
alicefr: What was the outcome of the device plugin investigation?

veps/sig-compute/secure-execution.md
jschintag: It is not required, i explain the workaround in the next paragraph. Maybe i should switch it around a bit, first explaining how i am doing it and then why it needs to be done that way?

veps/sig-compute/secure-execution.md
jschintag: It is not using sidecars for changing the xml, as what is needed is setting `iommu=on` for all devices. This is already done in kubevirt when using SEV, so the code is already mostly in place.

veps/sig-compute/secure-execution.md
alicefr: If this is not needed, then you can simply drop it

veps/sig-compute/secure-execution.md
jschintag: I'm marking this as resolved as we have a separate mail thread now and i managed to drop the requirement for api changes.

veps/sig-compute/secure-execution.md
jschintag: I haven't investigated it, since i found a simpler approach.

veps/sig-compute/secure-execution.md
jschintag: I have simplified this passage and dropped mentions of launch security type and privileged pods as they are not needed.

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,58 @@
...
+https://github.com/kubevirt/api
+
+# Design
+
+The feature should be behind a Feature Gate at first, similiar to AMD SEV.
+For implementing this feature, the api behaviour would need to be changed to allow for setting an empty `launchSecurity: {}`.
+
+For launching Secure Execution VMs, all devices need to have `iommu=on` set to allow them to be visible to the guest, as in Secure Execution Mode the hypervisor is prevented from accessing the guest memory. Otherwise Secure Execution is initiaded from the guest image during boot, not the hypervisor.
+
+To ensure that VMs are scheduled only on nodes supporting Secure Execution, node-labeller should add a label to these nodes. This can be done by parsing the output of `virsh domcapabilities`.
```
alicefr, line 31: You can add that the node will be label with the SE label and then the label will be added to the virt-launcher pod to ensure that it will be scheduled on node which support the feature.

alicefr:
@jschintag last comment about the scheduling and then the rest looks good to me
/approve
/hold 
to let you address the last point.
@vladikr PTAL

veps/sig-compute/secure-execution.md
jschintag: I have added a sentence to that effect

jschintag:
> @jschintag last comment about the scheduling and then the rest looks good to me

Thank you for your help @alicefr

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,59 @@
...
+spec:
+  domain:
+    launchSecurity: {}
+  ...
+```
+
+## Scalability
+It should scale like normal VMs.
+The only restriction would be for Secure Execution VMs, as they ´can't run on IBM Z machines that they have not been encrypted with.
+To move a Secure Execution VM to a new IBM Z machine, it's public key would first need to be added to the keys used to encrypt the kernel+initramfs+parmfile.
```
vladikr, line 49: Does it mean that migration is not possible by default?

veps/sig-compute/secure-execution.md
```diff
@@ -0,0 +1,59 @@
...
+## Repos
+https://github.com/kubevirt/kubevirt
+https://github.com/kubevirt/api
+
+# Design
+
+The feature should be behind a Feature Gate at first, similiar to AMD SEV.
+For implementing this feature, the api behaviour would need to be changed to allow for setting an empty `launchSecurity: {}`.
+
+For launching Secure Execution VMs, all devices need to have `iommu=on` set to allow them to be visible to the guest, as in Secure Execution Mode the hypervisor is prevented from accessing the guest memory. Otherwise Secure Execution is initiaded from the guest image during boot, not the hypervisor.
```
vladikr, line 29: Would this depend on the architecture, only for IBM Z?

vladikr (commented):
Thanks, I have some questions inline.

veps/sig-compute/secure-execution.md
jschintag: Migration is possible, but is limited by what the Guest Supports. Though the limit here is physical machines. This can be a bit confusing since usually a Machine is split into multiple LPAR (Logical Partitions) which is the closest to bare metal you can get on s390x.
The best thing here is to advice users to enable their VMs to run on all physical Machines they have in the cluster that support Secure Execution.

veps/sig-compute/secure-execution.md
jschintag: Currently AMD SEV is already enabling iommu, so currently it is not. This might change in the future though, as other technologies might have other requirements.

veps/sig-compute/secure-execution.md
jschintag: To clarify, i mean offline migration. Live migration is not possible due to the restrictions on accessing the guest memory.

vladikr: Ok, I was worried about the live migration. Can you please specify that NonMigratableCondition will be added to these VMs?

veps/sig-compute/secure-execution.md
vladikr: Yes, SEV has its own API..
I mean, the IOMMU is being added only if `.launchSecurity.SEV!=nil` right?

jschintag: Yes that is currently the case, but i would change it to `.launchsecurity != nil` since it applies to all current implemented cases.

veps/sig-compute/secure-execution.md
jschintag: I have added it to the design section.

veps/sig-compute/secure-execution.md
vladikr: Ok, we can figure this out on the PR.

vladikr:
/approve
Thank @jschintag, looks good to me for Alpha. We can work on the details in the PR and make adjustments for Beta.

jschintag:
/unhold

Can i also get a LGTM or is that not required for VEP?

PR #30, opened 2025-03-24
Improve the VEP template: implementation history and graduation requirements

iholder101:
### What this PR does
This PR clarifies two points that I've noticed that people seem to get wrong:
1. `Implementation phases` is now changed to `Implementation history`. Following k8s KEP process, this section is supposed to document the major implementations that have already happened, as a proof for why the graduation requirements are fulfilled (see next point).
Example of misusing the field is writing something like "we will first implement the main logic and then add tests".

2. `Feature lifecycle Phases` is changed to `Graduation Requirements`. Complementary to the above, this section is supposed to outline the requirements for graduation.
Example of misusing the field is writing something like "when the feature is stable enough it would move to beta".


veps/NNNN-vep-template/vep.md
```diff
@@ -92,10 +92,21 @@ For example:
...
 <!--
-How and when will the feature progress through the Alpha, Beta and GA lifecycle phases
+The requirements for graduating to each stage.
+Example:
+### Alpha
+- [ ] Feature gate guards all code changes
+- [ ] Initial implementation supporting only X and Y use-cases
+
+### Beta
+- [ ] Feature gate enabled by default
```
lyarwood, line 105: As much as I want this to be the policy it isn't at the moment :)

lyarwood (commented):
lgtm but beta FGs are not enabled by default at the moment even if I wish that wasn't the case

veps/NNNN-vep-template/vep.md
iholder101: This is surprising and demands a different discussion.
I share your wish. :)

iholder101: Anyways, I've removed it from this PR. PTAL

iholder101:
FYI @vladikr, do you mind having a quick look?

iholder101:
@xpivarc @vladikr

iholder101:
Should we get this in? @xpivarc @vladikr @lyarwood

lyarwood (commented):
/lgtm

vladikr:
:facepalm: how did I miss this... 
/approve

PR #32, opened 2025-03-26
Enhancement: Allow to disable the 64-bit PCI hole

0xFelix:
### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/33
**SIG label**:  /sig compute

### What this PR does

This proposal focuses on allowing to disable the 64-bit PCI hole in `VirtualMachineInstances`, so that legacy OSes like Windows XP or Server 2k3 can run on KubeVirt.

### Release note

```release-note
None
```

### DIFF
--- a/veps/sig-compute/pcihole64-proposal.md
+++ b/veps/sig-compute/pcihole64-proposal.md
@@ -1,0 +1,306 @@
+# VEP #33: Allow to disable the 64-bit PCI hole
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone /
+release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir
+  in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+This proposal focuses on allowing to disable the 64-bit PCI hole in
+`VirtualMachineInstances` (VM), so that legacy operating systems (OS) like
+Windows XP or Server 2k3 can run on KubeVirt.
+
+An explanation of the 64-bit PCI hole can be found
+[here](https://en.wikipedia.org/wiki/PCI_hole).
+
+## Motivation
+
+At the moment legacy OSes like Windows XP or Server 2k3 do not work out of the
+box on KubeVirt. This prevents big organizations running many VMs of all kinds
+from adopting KubeVirt. To allow greater adoption of KubeVirt, this proposal
+aims to make these legacy OSes work on KubeVirt.
+
+For more information on the topic
+see [this thread on kubevirt-dev](https://groups.google.com/g/kubevirt-dev/c/1Ue4brPoc1g)
+and [this issue](https://bugzilla.redhat.com/show_bug.cgi?id=990418)
+
+## Goals
+
+- A way of expressing that the 64-Bit PCI hole should be disabled is added to
+  the KubeVirt API in the most OS-independent way possible.
+- Legacy OSes like Windows XP or Server 2k3 can run on KubeVirt in a
+  supportable way.
+
+## Non Goals
+
+- Committing to a list of supported OSes in KubeVirt.
+- Adding OS specific configurables to the KubeVirt API.
+
+## Definition of Users
+
+- VM owners
+- Namespace owners
+- Cluster admins
+
+## User Stories
+
+- As a VM owner, I want to run legacy OSes like Windows XP or Server 2k3 on
+  KubeVirt.
+- As a VM owner, I want to migrate existing VMs with legacy OSes to KubeVirt
+  and to continue running them with the least possible amount of required
+  modifications.
+- As a namespace owner or cluster admin, I want to provide
+  `VirtualMachinePreferences` or `VirtualMachineClusterPreferences` for legacy
+  OSes like Windows XP or Server 2k3.
+
+## Repos
+
+- [https://github.com/kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+Users that want to run legacy OSes do not necessarily need to know what `64-Bit
+PCI hole` means (in this case setting its size to zero). Instead, they
+should be able to express that they want to run a legacy OSes without worrying
+about the details. Annotations provide a simple way of enabling configuration
+tweaks or compatibility behavior while avoiding modelling specific fields in the
+core API and becoming too technical for users.
+
+This proposal suggests to add the `kubevirt.io/disablePCIHole` annotation
+to `VirtualMachineInstance` objects, which will trigger KubeVirt to make the
+required configuration changes to allow running legacy OSes by disabling the
+64-Bit PCI hole. By using an annotation the core API is not changed and there is
+room for future changes to this feature independently of the core API. For
+example the implementation could be moved from KubeVirt into a
+[trusted or certified plugin](#extending-kubevirt-to-allow-using-trusted-or-certified-plugins)
+while retaining the same behavior.
+
+### Disabling the 64-Bit PCI hole on the libvirt domain level
+
+libvirt supports the configuration of the 64-Bit PCI hole by adding a
+`pcihole64` node below the `pcie-root` controller node in a domain XML.
+Setting the value of `pcihole64` to `0` disables the 64-Bit PCI hole.
+
+```xml
+<controller type='pci' index='0' model='pcie-root'>
+    <pcihole64 unit='KiB'>0</pcihole64>
+</controller>
+```
+
+When the feature is enabled, `virtwrap/converter` used by `virt-launcher` should
+configure the `pcihole64` on the `pcie-root` controller of an `api.Domain`
+object. The required configuration can be added as part of the conversion in
+`converter.Convert_v1_VirtualMachineInstance_To_api_Domain`.
+
+Since `converter.Convert_v1_VirtualMachineInstance_To_api_Domain` does not yet
+create `pcie-root` controllers, it must also be extended to add a basic
+`pcie-root` controller to `domain.Spec.Devices.Controllers` before configuring
+the `pcihole64` on it. This can be done before or after adding other controllers
+such as usb or scsi.
+
+To allow configuring the `pcihole64` on a `pcie-root` controller the feature
+requires an addition to `virtwrap/api/schema.go`.
+
+```golang
+type Controller struct {
+    Type      string            `xml:"type,attr"`
+    Index     string            `xml:"index,attr"`
+    Model     string            `xml:"model,attr,omitempty"`
+    Driver    *ControllerDriver `xml:"driver,omitempty"`
+    Alias     *Alias            `xml:"alias,omitempty"`
+    Address   *Address          `xml:"address,omitempty"`
+    PCIHole64 *PCIHole64        `xml:"pcihole64,omitempty"`
+}
+
+type PCIHole64 struct {
+    Value uint   `xml:",chardata"`
+    Unit  string `xml:"unit,attr,omitempty"`
+}
+```
+
+## API Examples
+
+By setting the `kubevirt.io/disablePCIHole` annotation on a
+`VirtualMachineInstance` object to `"true"` the feature is enabled. Any other
+value than `"true"` will leave the feature disabled. By default, the feature
+will also be disabled.
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-legacy-vm
+  annotations:
+    kubevirt.io/disablePCIHole: "true"
+spec:
+  domain:
+    devices: {}
+    memory:
+      guest: 512Mi
+  terminationGracePeriodSeconds: 180
+  volumes:
+    - containerDisk:
+        image: my.registry/images/legacy-os
+      name: legacy-os
+```
+
+## Alternatives
+
+### Adding a specific field to the core API
+
+The feature could be configured by adding a specific field
+`spec.domain.devices.disablePCIHole64` of type `boolean` to the core API.
+
+One advantage of this approach would be the improved discoverability of the
+feature, as it would become part of the core API and its documentation.
+
+The disadvantage of this approach, however that the feature cannot be changed
+independently of the core API. Adding this field to the core API could also be
+interpreted as KubeVirt caring about OS specific configuration, which this
+proposal tries to avoid.
+
+### Extending KubeVirt to allow using trusted or certified plugins
+
+Another way to implement this feature is to extend KubeVirt with a plugin
+mechanism that would allow vendors to ship extensions to KubeVirt in a
+supportable manner, while keeping the changes required for this or other
+features outside the KubeVirt codebase.
+
+Today KubeVirt already has a feature called `Sidecar`s, which among other things
+allows to change the libvirt domain specification of a `VirtualMachineInstance`
+before it is executed. A [proof of concept](#sidecar-proof-of-concept) for this
+proposal was created with the `Sidecar` feature. However, currently this
+feature is not supportable because, as can be seen in the PoC, it basically
+allows the execution of arbitrary code in the context of the `virt-launcher`
+pod.
+
+In order to make `Sidecar`s supportable, they would at least have to be changed
+so that only trusted code can be run as part of a hook. This could be realized
+for example by allowing to run only a single container image that contains all
+trusted hooks provided by the cluster admin. Before any hooks can be run, the
+cluster admin must deploy and configure this image in the cluster.
+
+The advantage of this approach would be that the KubeVirt codebase and API can
+be kept stable and clean, while vendors are still able to develop plugins for
+KubeVirt in a supportable way and with the least amount of effort required to
+maintain these plugins. This approach would allow third parties to experiment
+with new APIs before they potentially become part of the KubeVirt core API.
+
+The disadvantage of hooks is that they bypass KubeVirt's management, control
+and validation mechanisms, which is very likely to result in unexpected behavior
+and compatibility issues. Enabling multiple hook plugins in production at the
+same time is likely to lead to unmanageable interactions which ultimately means
+that it is difficult for vendors to support these hook plugins.
+
+### Using a `Sidecar` hook
+
+Although it was already dismissed in the previous alternative, in theory it
+would be possible to achieve the goal of this proposal by using a `Sidecar`
+hook. However, and as stated above, this feature is not supportable because,
+it allows the execution of arbitrary code in the context of the `virt-launcher`
+pod.
+
+#### `Sidecar` proof of concept
+
+This is a proof of concept that is using a `Sidecar` hook. It achieves the goal
+of being able to run legacy OSes on KubeVirt but it is not supportable.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pcihole64
+data:
+  pcihole64.py: |
+    #!/usr/bin/env python3
+    """
+    This module can be used as an onDefineDomain sidecar hook in KubeVirt to
+    ensure compatibility with Windows XP when using the q35 machine type.
+    """
+
+    import xml.etree.ElementTree as ET
+    import sys
+
+
+    def main(domain: str):
+        """
+        This function parses the domain XML passed in the domain argument, adds a
+        pcihole64 element with value 0 to every pcie-root controller and then
+        prints the modified XML to stdout.
+        """
+
+        xml = ET.ElementTree(ET.fromstring(domain))
+
+        controllers = xml.findall("./devices/controller[@model='pcie-root']")
+        for controller in controllers:
+            element = ET.Element("pcihole64", {"unit": "KiB"})
+            element.text = "0"
+            controller.insert(0, element)
+
+        ET.indent(xml)
+        xml.write(sys.stdout, encoding="unicode")
+
+
+    if __name__ == "__main__":
+        main(sys.argv[4])
+```
+
+### Utilizing the `pc-i440fx` machine type
+
+KubeVirt uses the Q35 machine type by default, which triggers the 64-bit PCI
+hole incompatibility with legacy OSes.
+
+A solution could be to enable the `pc-i440fx` machine type on the cluster
+and to use it instead of Q35 in the definitions of `VirtualMachineInstances`
+that boot legacy OSes.
+
+The advantage of this approach is that it requires no code changes to
+KubeVirt. All that is required is a configuration change to KubeVirt.
+
+The disadvantage of this approach is that the `pc-i440fx` machine type is
+deprecated and could be removed in the future, which means that a solution will
+have to be found again. In addition, this solution can only be enabled by
+cluster admins. A VM owner alone cannot enable the `pc-i440fx` machine type on a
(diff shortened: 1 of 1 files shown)


veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
+# VEP #33: Allow to disable the 64-bit PCI hole
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone /
+release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir
```
lyarwood, line 8: You can mark this as completed now.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+This should be brief and concise. We want just enough to get the point across
+-->
+
+Users that want to run legacy OSes do not necessarily need to know what `64-Bit
+PCI hole` means. Instead, they should be able to express that they want to run
+a legacy OSes without worrying about the details. Annotations provide a simple
+way of enabling configuration tweaks or compatibility behavior while avoiding
+modelling specific fields in the core API and becoming too technical for users.
+
+This proposal suggests to add the `kubevirt.io/legacyWindowsCompat` annotation
```
lyarwood, line 101: I'm obviously bias here but this would essentially duplicate the role of a preference and could easily set of precedent for more `Compat` or workload supporting annotations to be introduced.

I would much rather have the required configurable in the core spec and wrap it in a workload (xp, 2k3) preference that we can ship.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+  `Sidecar` hooks.
+
+## Non Goals
+
+<!--
+Limitations to the scope of the design
+-->
+
+- Committing to a list of supported OSes in KubeVirt.
+- Adding OS specific configurables to the KubeVirt API.
```
lyarwood, line 54: Is a `legacyWindowsCompat` annotation that magically disables PCI hole on the underlying domain not an OS specific configurable?

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+The feature could be configured by adding a specific field
+`spec.domain.devices.disablePCIHole64` of type `boolean` to the core API.
+
+Instead of enabling a broader compatibility behavior, this would express the
+configuration change to a `VirtualMachine` exactly.
+
+One advantage of this approach would be the improved discoverability of the
+feature, as it would become part of the core API and its documentation.
+
+The disadvantage of this approach, however, is that users must explicitly know
```
lyarwood, line 200: As you suggested above, I'd argue that the configurable is more discoverable than an annotation and especially so if we then wrap it in a preference.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+that they need to disable the 64-bit PCI hole, and that the feature cannot be
+changed independently of the core API. Adding this field to the core API could
+also be interpreted as KubeVirt caring about OS specific configuration, which
```
lyarwood, line 203: I thought that it was agreed that this isn't an OS specific configurable?

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir
+  in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+<!--
+Provide a brief overview of the topic
+-->
```
lyarwood, line 15: supernit - you can remove all of these comments btw

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+
+- As a VM owner, I want to run legacy OSes like Windows XP or Server 2k3 on
+  KubeVirt using the Q35 machine type.
+- As a VM owner, I want to migrate existing VMs with legacy OSes to KubeVirt
+  and to continue running them with the least possible amount of required
+  modifications.
+- As a namespace owner or cluster admin, I want to provide
+  `VirtualMachinePreferences` or `VirtualMachineClusterPreferences` for legacy
+  OSes like Windows XP or Server 2k3.
```
lyarwood, line 79: We can do this through an annotation but as the maintainer of instance types and preferences I'd much rather have access to the broken out configurables within the VMI spec instead of a singular annotation.

lyarwood (commented):
I'm really not a fan of the annotation approach as it feels like we are duplicating the work of a preference. I would still much rather have the underlying configurable expressed in the core VMI spec that would then allow us to provide specific XP and 2k3 preferences within KubeVirt.

veps/sig-compute/pcihole64-proposal.md
dominikholler: just a nit: Would there be a name that reflects better then `legacyWindowsCompat` what the annotation is doing, e.g. `disablePCIHole` ?

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+  in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+<!--
+Provide a brief overview of the topic
+-->
+
+This proposal focuses on allowing to disable the 64-bit PCI hole in
+`VirtualMachines`, so that legacy OSes like Windows XP or Server 2k3 can run on
```
jcanocan, line 18:
```suggestion
`VirtualMachines` (VM), so that legacy operating systems (OS) like Windows XP or Server 2k3 can run on
```

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+- [ ] (R) Enhancement issue created, which links to VEP dir
+  in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+<!--
+Provide a brief overview of the topic
+-->
+
+This proposal focuses on allowing to disable the 64-bit PCI hole in
```
jcanocan, line 17: Feel free to ignore, but could you please include a brief explanation about what 64-bit PCI hole is? Hopefully this will help to understand a bit better the importance of having a way to disable it.

veps/sig-compute/pcihole64-proposal.md
jcanocan: Do you know another linux OS example that will benefit from this?

veps/sig-compute/pcihole64-proposal.md
jcanocan: I agree with Lee, IMHO, this feels less OS specific than the annotation approach.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+so that only trusted code can be run as part of a hook. This could be realized
+for example by allowing to run only a single container image that contains all
+trusted hooks provided by the cluster admin. Before any hooks can be run, the
+cluster admin must deploy and configure this image in the cluster.
+
+The advantage of this approach would be that the KubeVirt codebase and API can
+be kept stable and clean, while vendors are still able to develop plugins for
+KubeVirt in a supportable way and with the least amount of effort required to
+maintain these plugins.
+
```
dominikholler, line 231: This would also allow third parties to experiment with new APIs before introducing them to core KubeVirt.

dominikholler (changes requested):
I am not familar with the expected structure of such an enhancement document. In my naive way, I would expect in the design section x a structure like:
x.1 Alternative a
  - idea
  - advantages
  - disadvantages

x.2. Alternative b
  - idea
  - advantages
  - disadvantages
  
x.3 . Alternative c
  - idea
  - advantages
  - disadvantages
  
x.4 decision
Approach z is used, because ...

veps/sig-compute/pcihole64-proposal.md
0xFelix: Somewhat, yes. Not entirely OS specific, but I agree it is OS family specific.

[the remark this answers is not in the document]
0xFelix: The API should not state that it is e.g. enabling Windows XP to run. But it can state features which are enabled by it. (in your example hyperv extensions)

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+<!--
+List of user stories this design aims to solve
+-->
+
+- As a VM owner, I want to run legacy OSes like Windows XP or Server 2k3 on
+  KubeVirt using the Q35 machine type.
+- As a VM owner, I want to migrate existing VMs with legacy OSes to KubeVirt
+  and to continue running them with the least possible amount of required
+  modifications.
+- As a namespace owner or cluster admin, I want to provide
```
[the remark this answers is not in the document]
0xFelix: I added the cluster admin here because you need cluster admin permissions to add cluster scoped objects like `VirtualMachineClusterPreferences`.

veps/sig-compute/pcihole64-proposal.md
0xFelix: @vladikr Can you comment on this? I followed your suggestion on kubevirt-dev with the current state of the design. But maybe we can put together a "broader" behavior using a preference instead?

veps/sig-compute/pcihole64-proposal.md
0xFelix: I'm somewhat summarizing @vladikr's concerns here. I agree that `spec.domain.devices.disablePCIHole64` does not look to OS specific to me.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+
+One advantage of this approach would be the improved discoverability of the
+feature, as it would become part of the core API and its documentation.
+
+The disadvantage of this approach, however, is that users must explicitly know
+that they need to disable the 64-bit PCI hole, and that the feature cannot be
+changed independently of the core API. Adding this field to the core API could
+also be interpreted as KubeVirt caring about OS specific configuration, which
+this proposal tries to avoid.
+
```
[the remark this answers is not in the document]
0xFelix: That is what `Extending KubeVirt with a hook plugin mechanism` is already suggesting, no?

veps/sig-compute/pcihole64-proposal.md
0xFelix: Going after @vladikr's suggestion here. Maybe we should make the annotation more specific and put together "broader" behavior by using preferences?

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+            controller.insert(0, element)
+
+        ET.indent(xml)
+        xml.write(sys.stdout, encoding="unicode")
+
+
+    if __name__ == "__main__":
+        main(sys.argv[4])
+```
+
```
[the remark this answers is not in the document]
0xFelix: Added an alternative for it

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+- Cluster admins
+
+## User Stories
+
+<!--
+List of user stories this design aims to solve
+-->
+
+- As a VM owner, I want to run legacy OSes like Windows XP or Server 2k3 on
+  KubeVirt using the Q35 machine type.
```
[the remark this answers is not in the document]
0xFelix: Added an alternative for `pc-i440fx`

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,323 @@
...
+## Definition of Users
+
+<!--
+Who is this feature set intended for
+-->
+
+- VM owners
+- Namespace owners
+- Cluster admins
+
```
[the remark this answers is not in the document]
dominikholler: Sure, but I see the perissions of a cluster-admin as a super-set of the vm owners permission, so an design working for the subset of VM owner permissions would avoid admin work.

veps/sig-compute/pcihole64-proposal.md
dominikholler: sounds like a good separation of responsibilities between the layers.

veps/sig-compute/pcihole64-proposal.md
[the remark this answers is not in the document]
dominikholler: No.  "Extending KubeVirt to allow using trusted or certified plugins trusted or certified plugins" is about allowing specific plugins, while the sidecar approach requires the cluster to allow any sidecar.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+`spec.domain.devices.disablePCIHole64` of type `boolean` to the core API.
+
+Instead of enabling a broader compatibility behavior, this would express the
+configuration change to a `VirtualMachine` exactly.
+
+One advantage of this approach would be the improved discoverability of the
+feature, as it would become part of the core API and its documentation.
+
+The disadvantage of this approach, however, is that users must explicitly know
+that they need to disable the 64-bit PCI hole, and that the feature cannot be
```
dominikholler, line 164: The sentence does not hurt, but makes no sense to me: The user can use the preference to express the desire to run a specific guest, so the user does not have to care on the how the preference enables the guest utilizing the kubevirt API.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+
+        ET.indent(xml)
+        xml.write(sys.stdout, encoding="unicode")
+
+
+    if __name__ == "__main__":
+        main(sys.argv[4])
+```
+
+### Re-enabling the `pc-i440fx` machine type
```
dominikholler, line 243:
```suggestion
### Utilizing the `pc-i440fx` machine type
```

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+
+KubeVirt uses the Q35 machine type by default, which triggers the 64-bit PCI
+hole incompatibility with legacy OSes.
+
+A solution could be to re-enable the `pc-i440fx` machine type on the cluster
```
dominikholler, line 248:
```suggestion
A solution could be to allowing the `pc-i440fx` machine type on the cluster
```

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+Users that want to run legacy OSes do not necessarily need to know what `64-Bit
+PCI hole` means. Instead, they should be able to express that they want to run
+a legacy OSes without worrying about the details. Annotations provide a simple
+way of enabling configuration tweaks or compatibility behavior while avoiding
+modelling specific fields in the core API and becoming too technical for users.
+
+This proposal suggests to add the `kubevirt.io/legacyWindowsCompat` annotation
+to `VirtualMachine` objects, which will trigger KubeVirt to make the required
+configuration changes to allow running legacy OSes. By using an annotation the
+impact to the core API is kept to a minimum and there is room for future changes
```
dominikholler, line 75:
```suggestion
impact to the core API is not modified and there is room for future changes
```

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,296 @@
...
+to this feature independently of the core API.
```
dominikholler, line 76: `For example the implementation could be moved from kubevirt into a  trusted or certified plugin.`

veps/sig-compute/pcihole64-proposal.md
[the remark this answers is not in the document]
dominikholler: I see. WinXP is just an example, I expect most 32 bit guests to require the feature.

veps/sig-compute/pcihole64-proposal.md
dominikholler: No. "Extending KubeVirt to allow using trusted or certified plugins trusted or certified plugins" is about allowing specific plugins, while the sidecar approach requires the cluster to allow any sidecar.

veps/sig-compute/pcihole64-proposal.md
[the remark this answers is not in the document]
dominikholler: My intention was to avoid requiring cluster scope changes at runtime (while they are still possible and useful, but not required.).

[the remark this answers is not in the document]
0xFelix: Cluster scope changes are not required at all to use the feature. But to me it's a possible user story to provide preferences with that enable this feature.

veps/sig-compute/pcihole64-proposal.md
0xFelix: Added a sentence mentioning it.

veps/sig-compute/pcihole64-proposal.md
0xFelix: That brings us back to the question if abstractions like this should simply live in preferences? @vladikr

veps/sig-compute/pcihole64-proposal.md
vladikr: From my point of view, the VM/VMI API shouldn't become as expressive as the libvirt API.
These are not equal. 

The VMI API should stay international. The VMI user should have a simple way to express their intention and KubeVirt itself should be smart enough to configure the VM in the right way, considering different architectures.

Annotations are less ideal than a formal API struct. However, a formal core VMI API would require a model of how this structure will be expanded in the future and a story of what it represents. I don't think we have a good enough story for this field.
At the moment we are trying to push a minor configuration tweak as a core API.

Users who want to enable legacy Windows support don't necessarily need to know what "PCI Hole 64" means and why we need to disable it.
The `kubevirt.io/legacyWindowsCompat` annotation feels like a good compromise to enable the functionality and avoid modeling it into the core API.

Preferences and instance types are a layer on top of the VM API - as much as I am in favor of this functionality, not everyone is using it. It also doesn't mean that the VMI core API should become a collection of unclear APIs.

I guess, something like  `kubevirt.io/disablePCIHole` would also be acceptable for me, as long as it remains an annotation.

veps/sig-compute/pcihole64-proposal.md
vladikr: Yes, I agree this is not OS specific. However, disabling this is only needed for supporting Windows XP and such [only when QEMU and Seabios are recent enough] - sounds a bit windows specific, and qemu/seabios specific

There are no other examples of this functionality being needed elsewhere, but we are rushing to put this in the core API.

Libvirt had a good story for it, placing it under the root controller; however, KubeVirt delegates the controller's creation to Libvirt and doesn't model it.

I would opt for an annotation at this point. However, if you insist, `spec.domain.features.pciHole64` would be a better place for it (allowing for its disabling and potentially expanding it to allow for size setting), not devices.

veps/sig-compute/pcihole64-proposal.md
```diff
@@ -0,0 +1,312 @@
...
+One advantage of this approach would be the improved discoverability of the
+feature, as it would become part of the core API and its documentation.
+
+The disadvantage of this approach, however, is that users must explicitly know
+that they need to disable the 64-bit PCI hole, and that the feature cannot be
+changed independently of the core API. Adding this field to the core API could
+also be interpreted as KubeVirt caring about OS specific configuration, which
+this proposal tries to avoid.
+
+### Extending KubeVirt to allow using trusted or certified plugins
```
dankenigsberg, line 172: Even if not selected for the PCI hole functionality, a trusted sidecar mechanism is important for the growth of KubeVirt. I'd like something like this to be implemented regardless, so that cluster admins can modify KubeVirt in a controlled way.

veps/sig-compute/pcihole64-proposal.md
lyarwood: > I would opt for an annotation at this point. However, if you insist, `spec.domain.features.pciHole64` would be a better place for it (allowing for its disabling and potentially expanding it to allow for size setting), not devices.

`features` makes sense as we don't model the PCI controllers etc under `devices` at the moment. I'd prefer the field to explicitly disable PCIHole64 but anything is better than an annotation IMHO.

veps/sig-compute/pcihole64-proposal.md
lyarwood: > From my point of view, the VM/VMI API shouldn't become as expressive as the libvirt API. These are not equal.

Yup agreed.

> The VMI API should stay international.

Assuming you mean intentional I agree, that's why I'm pushing for the specific configurable in our core API instead of a generic annotation.

> The VMI user should have a simple way to express their intention and KubeVirt itself should be smart enough to configure the VM in the right way, considering different architectures.

I don't agree at the VMI level, at the VM level sure they can use presets (for now) or preferences to express their intentions in a super simplistic way.

> Annotations are less ideal than a formal API struct. However, a formal core VMI API would require a model of how this structure will be expanded in the future and a story of what it represents. I don't think we have a good enough story for this field. At the moment we are trying to push a minor configuration tweak as a core API.

Why does this need to be expressed as a full struct when the only cases we care about are disabling (size=0) and inheriting the default from libvirt?
 
> Users who want to enable legacy Windows support don't necessarily need to know what "PCI Hole 64" means and why we need to disable it. The `kubevirt.io/legacyWindowsCompat` annotation feels like a good compromise to enable the functionality and avoid modeling it into the core API.

This is the use case templates, presets and now preferences have all attempted to target. Introducing yet another mechanism that groups configuration behind something (especially a generic `legacyWindowsCompat` annotation) outside of our core API is a mistake IMHO.

> Preferences and instance types are a layer on top of the VM API - as much as I am in favor of this functionality, not everyone is using it. It also doesn't mean that the VMI core API should become a collection of unclear APIs.

I don't understand what is unclear about `spec.domain.features.disablePCIHole64` tbh.

> I guess, something like `kubevirt.io/disablePCIHole` would also be acceptable for me, as long as it remains an annotation.

If we end up with an annotation my only request would be that we use something specific like this to avoid folks grouping any more configurables behind a generic annotation in the future.

0xFelix: I agree that `kubevirt.io/legacyWindowsCompat` somewhat duplicates grouping configuration on the VMI level, which we can easily do with preferences on the VM level already. But I also agree that with the annotation approach we retain the most flexibility for now. Therefore I've updated the proposal to keep the annotation approach but renamed it to `kubevirt.io/disablePCIHole`.

veps/sig-compute/pcihole64-proposal.md
0xFelix: Right, but flexibility of the API looks to be more important at this point, therefore going to keep the annotation.

vladikr:
/approve
Thank you. I think `kubevirt.io/disablePCIHole` is a good compromise at this point.

Issue #35, opened 2025-03-30
VEP 35: Add a Video configuration for VMs


dasionov:
**Primary contact (assignee)**:

/assign @dasionov

**Current Feature Stage**:
Alpha

**Feature Gate**:
VideoConfig

**Responsible SIGs**:
Primary SIG:
/sig compute

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/36

**Timeline**:

* Alpha release target (x.y): 1.6
* Beta release target (x.y): 1.7
* Stable release target (x.y): 1.8

* [x] Alpha
  * [x] VEP PRs:
    * https://github.com/kubevirt/enhancements/pull/36
  * [x] Code PRs:
    * https://github.com/kubevirt/kubevirt/pull/14673
  * [x] Docs PRs: https://github.com/kubevirt/user-guide/pull/928

* [x] Beta
  * [x] VEP PRs: #87
  * [x] Code PRs: N/A
  * [x] Docs PRs: https://github.com/kubevirt/user-guide/pull/928

* [ ] GA
  * [ ] VEP PRs: N/A
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/16599
  * [ ] Docs PRs: N/A

**Additional context**:
- https://github.com/kubevirt/kubevirt/pull/14752 Adding `virtio-vga` support for `amd64` arch VMs

> [!IMPORTANT]
> Please keep this description up to date.

fossedihelm:
@dasionov 
Can you update the timeline with the correct Alpha/Beta/GA versions? Thank you

dasionov:
> [@dasionov](https://github.com/dasionov) Can you update the timeline with the correct Alpha/Beta/GA versions? Thank you

Done 👍

vladikr:
@dasionov are there any beta related PRs out there? What are we tracking for beta?

dasionov:
> [@dasionov](https://github.com/dasionov) are there any beta related PRs out there? What are we tracking for beta?

https://github.com/kubevirt/enhancements/pull/87

https://github.com/kubevirt/user-guide/pull/928

mainly documentation updates.

enp0s3:
Hi @vladikr 
I see that the doc PR got merged. Therefore all the needed PRs for beta are ready. Can we close this issue as completed?

vladikr:
@xpivarc @dasionov @enp0s3 do we have everything we need on this issue ? Can we close?

dominikholler:
[...]

PR #36, opened 2025-03-30
Enhancements: Add Video configuration field for VMs

dasionov:
### VEP Metadata

**Tracking issue**: #35 
**SIG label**:  /sig compute

### What this PR does
We want to allow VM-owners to explicitly set the video device type when needed.

it's a follow-up to this [design](https://github.com/kubevirt/community/pull/371)

### Release note
```release-note
none
```

### DIFF
--- a/veps/sig-compute/vm_video_configuration_field.md
+++ b/veps/sig-compute/vm_video_configuration_field.md
@@ -1,0 +1,151 @@
+# VEP #35: Add a Video configuration field for VMs
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+- We want to allow VM-owners to explicitly set the video device type when needed.
+
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the `virtio` video device.
+
+In other words, the video device that is chosen for VMs is implicit and unclear.
+Different video devices have pros and cons and fit to different use-cases.
+For example, a guest with `virtio` drivers installed could enjoy from a `virtio` device which is generally faster and optimized for virtualization.
+Legacy guests, or guests with no `virtio` drivers, can use VGA for compatibility, and so on.
+
+With this proposal, the VM creator could adapt the video deceive to their needs.
+
+## Goals
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners who need to select a specific video device for their virtual machines to match their workload requirements.
+
+## User Stories
+- As an VM owner, I want my VM to provide high-resolution display, so that I can use high resolution
+  when I access the VM display via an in-guest graphical remote display mirroring access method like an in guest VNC server
+
+- As a Windows 11 on ARM user, I like to install my VM using the ramfb display adapter,
+  and replace it by the `virtio` adapter after I installed the `virtio` driver
+
+- As a VM creator with legacy guest or guest with no `virtio` drivers, I want to rely on VGA.
+- As a VM creator running a guest with `virtio` drivers, I want to enjoy from an optimized virtualization-aware video device which is generally faster and lighter on resources.
+
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+- [common-templates](https://github.com/kubevirt/common-templates)
+- [common-instancetypes](https://github.com/kubevirt/common-instancetypes)
+
+
+## Design
+
+### Preferred Option: Add `Video` Configuration Field to VM API [This approach was chosen to implement]
+
+#### Description
+Introduce a structured Video field under spec.template.spec.domain.devices with the following optional attributes:
+```yaml
+video:
+  type: virtio         # Optional: virtio, bochs, vga, etc.
+```
+This approach extends the API to provide declarative control over video settings.
+
+#### Integration with Existing Behavior:
+* If `autoattachGraphicsDevice` is set to `false`, no video devices will be attached, and the video field must not be present.
+* If `video.type` is provided, it will override the default video device model.
+* Webhook will validate that `video` is only set if `autoattachGraphicsDevice` is true or unset, and is valid with the relevant arch.
+* Architecture-specific logic will continue to provide sensible defaults unless overridden.
+
+The same hierarchy will apply for other architectures.
+
+#### Pros and Cons
+
+**Pros:**
+- **Granular Control:** Enables precise configuration for each VM.
+
+**Cons:**
+- **New API:** Introducing additional fields that need maintenance and testing.
+
+## Alternatives Considered:
+
+### Option 1: Default to Virtio Instead of Bochs for UEFI VMs - [Ruled out due to possible compatibility issues]
+#### Description
+Instead of relying on `Bochs` as the default video device for `AMD64` architectures, 
+KubeVirt would transition to `virtio` as the new default across all architectures.
+This change would provide higher resolutions, and improved compatibility with modern operating systems, especially Windows VMs.
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we need to adjust a fall-back logic to maintain full backward-computability
+- **pre-requisite driver** `virtio` driver must be installed for `virtio` video to enable the benefits over bochs
+- The hard coded `virtio` would not enable the windows on arm user story
+- 
+### Option 2: Annotation-Based Configuration
+#### Description
+This approach proposes introducing an annotation, kubevirt.io/video-device,
+which allows users to specify the video device type at the VM level without modifying the core API.
+
+**Pros:**
+- **No API Changes:** Avoids modifying the VM schema.
+- **Flexible:** Can be easily removed or changed in future releases without API versioning concerns.
+
+**Cons:**
+- **Lack of Visibility and Transparency:** Using an annotation for domain-level configuration makes it harder for users 
+to discover and understand its impact. 
+Unlike structured API fields, annotations are not well-documented in API references or validation mechanisms.
+
+- **Potential for Unintended Overrides:** Since annotations are loosely enforced,
+- they could create inconsistencies if users unknowingly override expected defaults.
+
+### API Examples
+
+#### Video Device Configured Explicitly with an API
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+   name: vm
+spec:
+   template:
+      spec:
+         domain:
+            devices:
+               video:
+                  type: virtio
+```
+
+#### Video Device Configured With Annotation
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+   name: vm
+annotations:
+  kubevirt.io/video-device: "virtio"
+```
+
+## Scalability
+Overhead of the virt-launcher/qemu will be measured and adjusted accordingly.
+
+## Update/Rollback Compatibility
+Currently, the virt-launcher arch-converter manages the default video type.
+We only provide a field for the user to specify explicitly if they want something else like virtio, so backward compatibility isn't affected.
+
+## Functional Testing Approach
+* Create a VM with video.Type set to `virtio` and expect launch successfully.
+
+## Feature Stage
+We are targeting this feature to be introduced in **Alpha**.
+
+- During Alpha:
+   - The `video` field will be **optional**.
+   - Users can opt into the new functionality without affecting existing behavior.
+
+Depending on adoption and feedback, we aim to promote the feature to **Beta** in a subsequent release, followed by **GA** once it is stable and widely used.


veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+The current behavior relies on the `autoattachGraphicsDevice` field, where:
+1. If `autoattachGraphicsDevice` is not specified or is set to `true`, the current logic in the virt-launcher determines the video device type.
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+
+The proposed change ensures that users can explicitly set the `type` for the video device **only if `autoattachGraphicsDevice` is not explicitly set to `false`**. This constraint will be enforced via the validation webhook and ensures that the new field does not conflict with existing configurations.
+
+The architecture-specific logic will continue to manage the video device type configurations.
+
+#### Implementation Logic
```
fabiand, line 50: How many new VMs do we expect to use the virtio device?

fabiand, line 50: Does this proposal also cover to include the video type in preferences?

fabiand, line 50: Do we have any plan to change the default video device in future?

dasionov: Fair question—I’d like others to weigh in on this.

In the past, there were concerns about changing the defaulting logic since RHEL 5/6 versions weren’t compatible with virtio (if I recall correctly). However, I’m not sure if we’re required to keep this default forever.

dasionov: Not quite—I prefer a step-by-step approach. First, we should agree on adding this to the API. After that, based on current demand, we can decide whether to open a PR to add it to preferences or instance types.

dasionov: Do numbers matter? I believe that if even a single user needs this configuration, and we can add it without breaking existing workloads or introducing backward compatibility concerns, we should expand our API to support as many users as possible—especially since the changes are not overly complex or difficult to maintain. What do you think?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+We want to allow VM-owners to explicitly set the video device type when needed.
+
+## Motivation
+Currently, the default video device type for AMD64 in KubeVirt is **VGA** or **Bochs** in some cases.
```
dominikholler, line 13: The current implementation is already using virtio for ARM [1] and S390x [2].
Only x64 [3] is using bochs for efi, and vga for legacy bios.

[1]
  https://github.com/kubevirt/kubevirt/blob/00418cab40405b8f9aa07238febea4bea942742a/pkg/virt-launcher/virtwrap/converter/arch/arm64.go#L41

[2]
 https://github.com/kubevirt/kubevirt/blob/00418cab40405b8f9aa07238febea4bea942742a/pkg/virt-launcher/virtwrap/converter/arch/s390x.go#L36

[3]
  https://github.com/kubevirt/kubevirt/blob/00418cab40405b8f9aa07238febea4bea942742a/pkg/virt-launcher/virtwrap/converter/arch/amd64.go#L38

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+Using `vga` video type can limit the video functionality available to virtual machines.
+For example, VGA restricts display resolution settings for Windows VMs and may not support modern workloads effectively.
+However, it is necessary to maintain compatibility for older guests that depend on VGA (e.g., RHEL 5/6).
+To address this, we propose introducing a new configuration field to allow users to set the video device type.
```
dominikholler, line 17: To address this, kubevirt is using vga for legacy bios workloads already.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Goals
+The VM spec template resource should have the ability to set the video device type.
```
dominikholler, line 20: Why?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Non Goals
+Change the defaulting logic
```
dominikholler, line 23: Why not change the default logic?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Definition of Users
+VM-owners who require specific video configurations for VMIs in their environments.
```
dominikholler, line 26: Which video adapter configs might required?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+## User Stories
+- As a VM owner, I want to be able to set a better video device like `virtio` instead of the standard `vga`.
```
dominikholler, line 29: 1. The user story needs to be platform dependent, because it might be different on ARM, s390x and x64
2. The user story should contain the "so that" or "because" part, see https://en.wikipedia.org/wiki/User_story#Common_templates

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
```
dominikholler, line 32: Even in a step by step approach, it might be worth to highlight that common-temapltes and common-instancetypes repos might be aftereffect.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,104 @@
...
+At the `addGraphicsDevice` method:
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+3. If `Video` is explicitly specified in the VM spec, the video type from the spec is used.
+4. If `autoattachGraphicsDevice` is not specified or is `true`:
+   - Use Bochs for EFI guests on AMD64.
+   - Use VGA for BIOS guests on AMD64.
+
+The same hierarchy will apply for other architectures.
+
+#### Pros and Cons
```
dominikholler, line 62: Pros and Cons related to what?
Aren't there the same alternatives like in https://github.com/kubevirt/enhancements/pull/32 ?
[ In my naive way, I would expect in the design section x a structure like:](https://github.com/kubevirt/enhancements/pull/32#pullrequestreview-2724674671)
x.1 Option/Alternative a

    idea
    advantages
    disadvantages

x.2. Alternative b

    idea
    advantages
    disadvantages

x.3 . Alternative c

    idea
    advantages
    disadvantages

x.4 decision
Approach z is used, because ...

dasionov: I adjusted the proposal, and added the option for defaulting the logic to set `virtio` video device over `vga/bochs`

Although I am not familiar enough with rhel 5/6 machines to determine weather the the backward compatibility concerns are valid or not.

veps/sig-compute/vm_video_configuration_field.md
dasionov: adjusted the proposal accordingly

veps/sig-compute/vm_video_configuration_field.md
dasionov: fixed

veps/sig-compute/vm_video_configuration_field.md
dasionov: added this option as an alternative to a dedicated video field

veps/sig-compute/vm_video_configuration_field.md
dasionov: adjusted this to
`VM owners who require specific video resolutions that are unavailable when using VGA or Bochs as the video device.
`

veps/sig-compute/vm_video_configuration_field.md
dasionov: adjusted the story to address `amd64` sepcifically, not sure what is needed for arm, and s390x since `virtio` is already enabled there by default.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,118 @@
...
+To balance modernization with backward compatibility, KubeVirt continues to use VGA for legacy BIOS-based workloads.
+
+## Goals
+Ensuring VM-owners won't have a limited resolution experience with their VMs
+
+## Definition of Users
+VM owners who require specific video resolutions that are unavailable when using VGA or Bochs as the video device.
+
+## User Stories
+As an AMD64 VM owner, I want my VM to provide high-resolution display settings when accessed via RDP, VNC,
```
dominikholler, line 26: RDP has its own virtual display adapter.
```suggestion
As an AMD64 VM owner, I want my VM to provide high-resolution display settings when accessed via VNC
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+- [common-templates](https://github.com/kubevirt/common-templates)
+- [common-instancetypes](https://github.com/kubevirt/common-instancetypes)
+
+
+## Design
+
+### Option 1: Default to Virtio Instead of VGA/Bochs
```
dominikholler, line 37: Technically, this is correct.
for prgamtic reasons, I propose to focus this option only on uefi:
```suggestion
### Option 1: Default to Virtio Instead of Bochs for UEFI VMs
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,118 @@
...
+Instead of relying on VGA/Bochs as the default video device for AMD64 architectures, 
+KubeVirt would transition to virtio as the new default across all architectures.
+This change would provide better video performance, higher resolutions,
+and improved compatibility with modern operating systems,especially Windows VMs.
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we might need to adjust a fall-back logic to maintain full backward-computability
```
dominikholler, line 48: This is a very valid point, and we should think twice about it.
I noticed some difference in bochs to virtio-vga, e.g. 8mb vs 16mb mem, the way the default resolution form OVMF is applied. But I did not notice anything what worries me, but it does not mean that there is no issue I did not recognize.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,118 @@
...
+   name: vm
+spec:
+   template:
+      spec:
+         domain:
+            devices:
+               video:
+                  type: virtio
+```
+
```
dominikholler, line 100: Option 3: Use an inoffical API like an annotation to incluence the behavior of kubevirt

This proposal suggests to add the kubevirt.io/virtio-vga annotation to VirtualMachine objects, which will trigger KubeVirt to make the required configuration changes to enable virtio graphics adapter for the VM. By using an annotation the impact to the core API is kept to a minimum and there is room for future changes to this feature independently of the core API.
Other way round, an annotation like ` kubevirt.io/bochs` could be used to ensure backwards compatibility in case the change to virtio intruduced regressions. In a view releases, the anotation could be removed, without conflicting with the API versioning rules.

veps/sig-compute/vm_video_configuration_field.md
dominikholler: > I adjusted the proposal, and added the option for defaulting the logic to set virtio video device over vga/bochs

Thanks, it improved the readability to me significantly!

> Although I am not familiar enough with rhel 5/6 machines to determine weather the the backward compatibility concerns are valid or not.

These old legacy guests are using BIOS, and I would vote for not touching VGA for BIOS default config.

dominikholler (commented):
@dasionov thanks for the update!

veps/sig-compute/vm_video_configuration_field.md
dasionov: Added this alternative

veps/sig-compute/vm_video_configuration_field.md
dasionov: fixed

veps/sig-compute/vm_video_configuration_field.md
dasionov: done

dominikholler:
Thanks for the updates @dasionov . 
@vladikr are you looking for more information, or does the enhancement contains all information required to decide?

veps/sig-compute/vm_video_configuration_field.md
0xFelix: It could be a breaking change for existing VMs though. Windows guests could be using hardware details to create unique ids for licensing purposes. Would such ids change then?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,145 @@
...
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the virtio video device.
+The use of VGA/Bochs can impose limitations on video functionality, such as restricted display resolution settings, particularly for Windows virtual machines.
+However, maintaining compatibility with older guest operating systems (e.g., RHEL 5/6) that rely on VGA/Bochs remains important.
+To balance modernization with backward compatibility, KubeVirt continues to use VGA for legacy BIOS-based workloads.
+
+## Goals
+Ensuring VM-owners won't have a limited resolution experience with their VMs
```
0xFelix, line 20: Is it all about resolution? The next issue could be that the performance of emulated video devices is too slow for VDI purposes.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Definition of Users
+VM owners who require specific video resolutions that are unavailable when using VGA or Bochs as the video device.
```
0xFelix, line 23: Why are they unavailable? Missing device drivers?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,145 @@
...
+
+## User Stories
+As an AMD64 VM owner, I want my VM to provide high-resolution display settings when accessed via VNC
+or other remote access methods, so that I have a better user experience when working with graphical applications.
```
0xFelix, line 27: See comment above, will it really improve the user experience for VDI purposes?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,145 @@
...
+Instead of relying on `Bochs` as the default video device for `AMD64` architectures, 
+KubeVirt would transition to `virtio` as the new default across all architectures.
+This change would provide better video performance, higher resolutions,
+and improved compatibility with modern operating systems,especially Windows VMs.
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we might need to adjust a fall-back logic to maintain full backward-computability
```
0xFelix, line 48: See my comment about unique ids derived from the hardware.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### Description
+Add a `Video` struct under `spec.template.spec.domain.devices` in the VM template schema.
+This struct will include a single field, `type`, to specify the desired video device type.
+
+The current behavior relies on the `autoattachGraphicsDevice` field, where:
+1. If `autoattachGraphicsDevice` is not specified or is set to `true`, the current logic in the virt-launcher determines the video device type.
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+
+The proposed change ensures that users can explicitly set the `type` for the video device **only if `autoattachGraphicsDevice` is not explicitly set to `false`**. This constraint will be enforced via the validation webhook and ensures that the new field does not conflict with existing configurations.
```
0xFelix, line 60: Will it allow to set `spec.template.spec.domain.devices` if `autoattachGraphicsDevice` is set to `true`?

dasionov: yes, IIRC the code in the converter won't add the graphic device only when `autoattachGraphicsDevice` is != nil and `autoattachGraphicsDevice` == false
 
 so when it's true it should be ok.

veps/sig-compute/vm_video_configuration_field.md
dasionov: could you elaborate on VDI?

veps/sig-compute/vm_video_configuration_field.md
0xFelix: I am not 100% certain that it would actually change a calculated ID. But it could be possible.

See https://support.microsoft.com/en-us/windows/reactivating-windows-after-a-hardware-change-2c0e962a-f04c-145b-6ead-fb3fc72b6665

veps/sig-compute/vm_video_configuration_field.md
0xFelix: If the aim of this proposal is to encourage the use of KubeVirt as a VDI solution then the next issue a user might encounter is that the performance of an emulated graphics device could be too slow for use with certain applications that rely on hardware acceleration.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,149 @@
...
+
+The architecture-specific logic will continue to manage the video device type configurations.
+
+#### Implementation Logic
+1. If `autoattachGraphicsDevice` is set to `false`, the `Video` field must not exist. This will be validated via the webhook.
+
+At the `addGraphicsDevice` method:
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+3. If `Video.Type` is explicitly specified in the VM spec, the video type from the spec is used.
+4. If `Video.Resolution` is explicitly specified the video resolution from the spec is used.
```
dominikholler, line 72: Just a comment: There might be dependencies in both directions between `Video.Type` and `Video.Resolution`, not sure if this can be avoided.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,149 @@
...
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we might need to adjust a fall-back logic to maintain full backward-computability
+
+### Option 2 VM-Level Video Configuration Field
+
+#### Description
+Introduce a Video struct under spec.template.spec.domain.devices in the VM template schema.
```
dominikholler, line 53: Is there the assumption that there can be only a single `Video` ? Should the API prepare for having multiple ones?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,149 @@
...
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+3. If `Video.Type` is explicitly specified in the VM spec, the video type from the spec is used.
+4. If `Video.Resolution` is explicitly specified the video resolution from the spec is used.
+
+The same hierarchy will apply for other architectures.
+
+#### Pros and Cons
+
+**Pros:**
+- **Granular Control:** Enables precise configuration for each VM.
```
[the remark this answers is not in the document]
dasionov: The user story highlights the limited resolution issue that Windows users are facing.
By adding a resolution setting to the API, we will be able to address the problem described in the user story.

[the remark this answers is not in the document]
dominikholler: The problem would be already addressed if the default would change to virtio.

veps/sig-compute/vm_video_configuration_field.md
dasionov: It's because the resolution is being set to a default value 1280X800, and we don't offer api for adjusting the resolution

veps/sig-compute/vm_video_configuration_field.md
dasionov: yes, you are correct, if a user will provide a video type which supports only resolution for 1920X1080, and configure resolution for 4K it should result an error

veps/sig-compute/vm_video_configuration_field.md
dasionov: at this moment i would prefer to have a single configuration just for primary video, if the use-cases in the feature rises to support more then the primary configuration it can be added later wdyt?

veps/sig-compute/vm_video_configuration_field.md
dasionov: Definitely

veps/sig-compute/vm_video_configuration_field.md
dasionov: Didn't find a source that say replacing video device will result uuid calculation to change. but it doesn't mean it's not true.. also if we go with api configuration rather then switching the default, on cases where the uuid calc does break, it will only affect the specific users who want to have virtio video over the default one, in such case we can throw a warning or let them know the uuid might break

veps/sig-compute/vm_video_configuration_field.md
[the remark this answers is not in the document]
dasionov: maybe, but we can't switch the default... I guess even for EFI VMs since virtio require driver to be opt-in.. we cant afford to break efi vms without prepared drivers.
@xpivarc @jean-edouard i initially thought it was possible to request virtio video device, which is processed by libvirt as `virtio-gpu` and if the drivers are not exist we will fallback to `virtio-vga`
but i am not sure this is possible anymore

wdyt?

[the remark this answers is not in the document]
jean-edouard: IIRC, @dasionov you tried booting a diskless EFI VM with virtio graphics and didn't see the firmware screen(s). That on its own is problematic enough to avoid switching to virtio by default IMO

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+To balance modernization with backward compatibility, KubeVirt continues to use VGA for legacy BIOS-based workloads.
+
+## Goals
+Ensuring VM-owners won't have a limited resolution experience with their VMs
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+In these cases, the guest operating system defaults to a limited resolution during initialization due to the constraints of the default VGA or Bochs video devices.
+
+- Users running graphical applications that benefit from higher resolution and better performance,
```
0xFelix, line 26: To me that's a problematic use case. virtio GPUs are emulated on the CPU and not very well suited for high performance applications.

veps/sig-compute/vm_video_configuration_field.md
0xFelix: You could test it with a Windows VM to be sure.

veps/sig-compute/vm_video_configuration_field.md
0xFelix: TBH I'm not sure whether such an API will benefit us in the long term. Maybe an annotation to force virtio GPUs is enough for now? What do you think @vladikr?

veps/sig-compute/vm_video_configuration_field.md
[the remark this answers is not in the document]
dominikholler: sure, there might be questions to investigate, e.g. if virtio-vga was used in this case, or if something like in https://www.kraxel.org/blog/2019/02/ramfb-display-in-qemu/ should be used.

veps/sig-compute/vm_video_configuration_field.md
dominikholler: I would vote for the other way round, using virtio by default, especially virtio-vga on x64. If this results in trouble, an annotation to fall back on bochs would give us time to understand why the backward compatibility issue happened and if an explicit API is required at all.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+We want to allow VM-owners to explicitly set the video device type when needed.
+
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the virtio video device.
+The use of VGA/Bochs can impose limitations on video functionality, such as restricted display resolution settings, particularly for Windows virtual machines.
```
xpivarc, line 15: I believe this sentence is not true.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+However, maintaining compatibility with older guest operating systems (e.g., RHEL 5/6) that rely on VGA/Bochs remains important.
+To balance modernization with backward compatibility, KubeVirt continues to use VGA for legacy BIOS-based workloads.
```
xpivarc, line 17: The latter part of the sentence is probably still tide to the "defaulting" VEP. This should be removed as it is not relevant imo

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
+# VEP #35: Add a Video configuration field for VMs
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+We want to allow VM-owners to explicitly set the video device type when needed.
```
xpivarc, line 10: Don't we want 2 things? Resolution and video type?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the virtio video device.
+The use of VGA/Bochs can impose limitations on video functionality, such as restricted display resolution settings, particularly for Windows virtual machines.
+However, maintaining compatibility with older guest operating systems (e.g., RHEL 5/6) that rely on VGA/Bochs remains important.
+To balance modernization with backward compatibility, KubeVirt continues to use VGA for legacy BIOS-based workloads.
+
+## Goals
+Ensuring VM-owners won't have a limited resolution experience with their VMs
```
xpivarc, line 20: Do you want to add: User have option to choose virtio?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+In these cases, the guest operating system defaults to a limited resolution during initialization due to the constraints of the default VGA or Bochs video devices.
```
xpivarc, line 24: What are the constraints?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Users running graphical applications that benefit from higher resolution and better performance, 
+particularly those relying on hardware-accelerated graphics. 
+The current Bochs configuration lacks the necessary support for such use cases, especially in modern desktop environments or remote access scenarios like VNC.
+
+## User Stories
```
xpivarc, line 30: Another one could be:
Application requiring specific resolution in order to work. Might not be for access/VDI, it could be for testing some application and others...

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+- [common-templates](https://github.com/kubevirt/common-templates)
+- [common-instancetypes](https://github.com/kubevirt/common-instancetypes)
+
+
+## Design
+
+### Option 1: Default to Virtio Instead of Bochs for UEFI VMs
```
xpivarc, line 42: We need to have a design here, not options and alternatives (and why your or after review, those options are alternatives)

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+Instead of relying on `Bochs` as the default video device for `AMD64` architectures, 
+KubeVirt would transition to `virtio` as the new default across all architectures.
+This change would provide better video performance, higher resolutions,
+and improved compatibility with modern operating systems,especially Windows VMs.
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we might need to adjust a fall-back logic to maintain full backward-computability
```
xpivarc, line 53: `we might need` -> `we need`

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Option 2 VM-Level Video Configuration Field
+
+#### Description
+Introduce a Video struct under spec.template.spec.domain.devices in the VM template schema.
+The struct includes:
+* type: optionally to specifies the video device type (e.g., virtio, vga, bochs).
+* resolution: An optional sub-struct with x and y fields to define the display resolution (e.g., x: 1920, y: 1080).
```
xpivarc, line 61: What would be valid options?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+
+The current behavior relies on the `autoattachGraphicsDevice` field, where:
+1. If `autoattachGraphicsDevice` is not specified or is set to `true`, the current logic in the virt-launcher determines the video device type.
+2. If `autoattachGraphicsDevice` is set to `false`, no graphics or video devices are attached.
+
+The proposed change ensures that users can explicitly set the `type` for the video device **only if `autoattachGraphicsDevice` is not explicitly set to `false`**. This constraint will be enforced via the validation webhook and ensures that the new field does not conflict with existing configurations.
+
+The architecture-specific logic will continue to manage the video device type configurations.
```
xpivarc, line 69: The validation will also needs to take care of architectures

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,154 @@
...
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+   name: vm
+annotations:
+  kubevirt.io/video-device: "virtio"
+```
+
+## Scalability
+Overhead of the virt-launcher/qemu may be impacted.
```
xpivarc, line 137: We need to explicitly state that this will be measured and the overhead will be adjusted

xpivarc (commented):
Please explicitly mention what stage do you target: Alpha/Beta/GA

veps/sig-compute/vm_video_configuration_field.md
dominikholler: > not options and alternatives
Why not?
Can't we list the options, and let the approver provide a subsection in the end which says "option x is choosen because of ...." , because the approver decides, not the author of the PR.

veps/sig-compute/vm_video_configuration_field.md
dasionov: IMO either asking for virtio video model, or asking for a specific resolution with the default video device

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed this sentence

veps/sig-compute/vm_video_configuration_field.md
dasionov: yes, removed it

veps/sig-compute/vm_video_configuration_field.md
dasionov: True, fixed

veps/sig-compute/vm_video_configuration_field.md
dasionov: Added, Thanks

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed this part.

veps/sig-compute/vm_video_configuration_field.md
dasionov: added this use case

veps/sig-compute/vm_video_configuration_field.md
dasionov: done

veps/sig-compute/vm_video_configuration_field.md
dasionov: yes, fixed

veps/sig-compute/vm_video_configuration_field.md
dasionov: done

veps/sig-compute/vm_video_configuration_field.md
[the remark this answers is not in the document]
0xFelix: > IIRC, @dasionov you tried booting a diskless EFI VM with virtio graphics and didn't see the firmware screen(s). That on its own is problematic enough to avoid switching to virtio by default IMO

That means option one is the preferred choice?

veps/sig-compute/vm_video_configuration_field.md
0xFelix: I'm missing the preferred option in this VEP right now. This doc should contain that piece of information. We can list alternatives, but there has to be a chosen option. Documenting it only as a comment on the PR is not enough in my opinion.

veps/sig-compute/vm_video_configuration_field.md
0xFelix: I don't think we can change the default for now, see the other comment threads.

veps/sig-compute/vm_video_configuration_field.md
0xFelix: How do we know that?

0xFelix (commented):
The current tendency is to chose option one and keep the defaults as is, it that correct? If so could you please update the doc to make that clearer?

veps/sig-compute/vm_video_configuration_field.md
dasionov: refactored the design to clarify the preferred option, which is to extand the api to allow specific configuration for video device and resolution

added a section under it for the alternative options

dasionov:
Hi @ksimon1, Thank you for sharing this!

> Hello! I've been doing some Windows testing on ARM machines, and I've found that we'll need the ability to specify the video device within the VM definition. Currently, when users attempt to create Windows VMs via Kubevirt on ARM, they encounter a few initial hurdles. These include a missing machine type label and secure boot being disabled. While these can be overridden, the primary challenge lies in the video device.

The machine type issue will be fixed after https://github.com/kubevirt/kubevirt/pull/14604 is merged

> My workaround involved the following steps:
> 
> 1. Install Windows using the `ramfb` video device.
> 2. Install the `viogpudo` driver from the `virtio-win` ISO.
> 3. Shut down the VM.
> 4. Change the video device from `ramfb` to `virtio-gpu-pci`.
> 5. The VM then booted successfully with the `virtio-gpu-pci` device.
> 
> This experience highlights the necessity for users to be able to define the video device in the VM specification, particularly for initial Windows installations on ARM.

I'm surprised we experience issues installing windows on `arm64` actually, this VEP was originally raised because windows vms on `amd64` were using `bocsh` resulting a limited default resolution

veps/sig-compute/vm_video_configuration_field.md
dasionov: I believe we should have minimal validation regarding this, let the errors be reported from libvirt in case of configuration miss match, like unsupported resolution per specific video. 

so we start with validating:
`devices.video` != nil -> `autoattachGraphicsDevice != false`
and since resolution only supported for `vga`, `qxl`, `bochs`, `gop`, and `virtio`
we can validate resolution != nil if video != nil and video.type in  (`vga`, `qxl`, `bochs`, `gop`, `virtio`)

veps/sig-compute/vm_video_configuration_field.md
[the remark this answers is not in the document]
dasionov: The preferred choice is adding API for VM Video configuration, I adjusted the design to clarify this better.

fabiand:
@dasionov howdy.

> since new discussion regarding the usage of virtio video device has been raised here: https://redhat-internal.slack.com/archives/C068X44C8VB/p1742881025301279

Seems to be an company/internal slack channel, please remove it from this pr, because it's not publicly accessible.

veps/sig-compute/vm_video_configuration_field.md
fabiand: KubeVirt is no VDI solution.

VDI includes graphics, but also multi monitor, tablets, usb passthrough, audio, video encoding. VNC does not provide all of this.

Thus let me echo Felix: How does setting the device to virtio help? Please provide some details.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,164 @@
...
+- **New API:** Introducing additional fields that need maintenance and testing.
+
+## Alternatives Considered:
+
+### Option 1: Default to Virtio Instead of Bochs for UEFI VMs
+#### Description
+Instead of relying on `Bochs` as the default video device for `AMD64` architectures, 
+KubeVirt would transition to `virtio` as the new default across all architectures.
+This change would provide better video performance, higher resolutions,
+and improved compatibility with modern operating systems,especially Windows VMs.
```
fabiand, line 80: This is not not always true.
The pre-requisite is the guest to have the relevant drivers installed.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
```
fabiand, line 83: Maybe virtio is leading to better in VM performance, but from an e2e perspective this is not necessarily given: VNC is not a great protocol for high resolutions, by default VNC traffic goes thorugh the API server, there high traffic is not desired.

fabiand (commented):
There is some use to specify the video device, but i want to highlight that VDI is not a good justification.
VDI is not only grpahics, but much more, KubeVirt, and VNC; are not the right elements to address VDI use-cases. In the worst case, we are setting the wrong expectations.

veps/sig-compute/vm_video_configuration_field.md
iholder101: > KubeVirt is no VDI solution.

Sure, let's get rid of this argument then.

> How does setting the device to virtio help? Please provide some details.

IIUC the main advantages of virtio are generally better performance (no need to emulate a "real" graphics device) and better support for higher resolutions and configurations.

Is there anything else you expect as a benefit?

veps/sig-compute/vm_video_configuration_field.md
dasionov: what about the firmware uuid? we are the ones that generates and persist it, it wont be change due to the video device switch

0xFelix: Yeah, let's move forward ruling out this alternative.

@dasionov We control the firmware UUID, but we don't control what Windows takes into account for computing a "hardware ID". That's why I was suggesting to simply test it.

veps/sig-compute/vm_video_configuration_field.md
0xFelix: We don't have customers, we have users. :)

I agree, let's remove the wording around VDI, just state that it allows to specify what type of video device is going to be used and why this is beneficial, i.e. better guest support or better support for higher resolutions and configurations.

veps/sig-compute/vm_video_configuration_field.md
dasionov: Thank you for clarifying, now I understand it better. :+1:

dasionov: adjusted the design to mark this alternative was ruled out due to compatibility issues

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed the problematic use case as our priority for now reside in addressing the resolution concern

veps/sig-compute/vm_video_configuration_field.md
dasionov: True, and also this alternative was ruled out due to compatibility reasons

veps/sig-compute/vm_video_configuration_field.md
dominikholler: I understood that this is about in-guest remote display mirroring, which would be reflected by a user story like
```suggestion
- As an AMD64 VM owner, I want my VM to provide high-resolution display, so that I can use high resolution when I access the VM display via an in-guest graphical remote access methods like an a in guest VNC server```

Comment (dasionov) [reply veps/sig-compute/vm_video_configuration_field.md]: Thanks, I adjusted the design.

File: veps/sig-compute/vm_video_configuration_field.md [hunk 36]
```diff
@@ -0,0 +1,164 @@
...
+
+**Pros:**
+- **Granular Control:** Enables precise configuration for each VM.
+
+**Cons:**
+- **New API:** Introducing additional fields that need maintenance and testing.
+
+## Alternatives Considered:
+
+### Option 1: Default to Virtio Instead of Bochs for UEFI VMs - [Ruled out due to compatibility issues]
```
dominikholler, line 75: We did not check if there are compatibility issues, the issues are just a theoretical possibility which we do not know if they could happen.
```suggestion
### Option 1: Default to Virtio Instead of Bochs for UEFI VMs - [Ruled out due to possible compatibility issues]
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,164 @@
...
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the `virtio` video device.
+
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+In these cases, the guest operating system defaults to a limited resolution during boot.
```
0xFelix, line 23:
```suggestion
- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
  In these cases, the guest operating system defaults to a limited resolution during boot.
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,164 @@
...
+
+## User Stories
+- Application requiring specific resolution in order to work
+
+- As an AMD64 VM owner, I want my VM to provide high-resolution display, so that I can use high resolution 
+when I access the VM display via an in-guest graphical remote display mirroring access methods like an in guest VNC server
```
0xFelix, line 29:
```suggestion
when I access the VM display via an in-guest graphical remote display mirroring access method like an in guest VNC server
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,164 @@
...
+Instead of relying on `Bochs` as the default video device for `AMD64` architectures, 
+KubeVirt would transition to `virtio` as the new default across all architectures.
+This change would provide higher resolutions, and improved compatibility with modern operating systems, especially Windows VMs.
+
+**Pros:**
+- **Reduced Configuration:** Users won't need to manually switch to virtio for better video performance.
+
+**Cons:**
+- **Backward Compatibility:** we need to adjust a fall-back logic to maintain full backward-computability
+- **pre-requisite driver** `virtio` driver must be installed for `virtio` video to work properly
```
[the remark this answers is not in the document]
dasionov: Thanks! adjusted the design to your suggestion

veps/sig-compute/vm_video_configuration_field.md
dasionov: Thanks, fixed.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,165 @@
...
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+In these cases, the guest operating system defaults to a limited resolution during boot.
+
+## User Stories
+- Application requiring specific resolution in order to work
```
0xFelix, line 26:
```suggestion
- As VM owner I want to run applications requiring a specific resolution in order to work
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,165 @@
...
+
+- As an AMD64 VM owner, I want my VM to provide high-resolution display, so that I can use high resolution
```
0xFelix, line 28:
```suggestion
- As an VM owner, I want my VM to provide high-resolution display, so that I can use high resolution 
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,165 @@
...
+### Preferred Option: Add `Video` Configuration Field to VM API
+
+#### Description
+Introduce a structured Video field under spec.template.spec.domain.devices with the following optional attributes:
+```yaml
+video:
+  type: virtio         # Optional: virtio, bochs, vga, etc.
+  resolution:          # Optional
+    x: 1920
+    y: 1080
```
0xFelix, line 52: I'd drop this, video type should be enough for a start. We don't want to bloat the API too much.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,165 @@
...
+2. **Phase 2 Documentation**
+    - Update KubeVirt documentation to include examples of configuring the video device type.
+
+## Feature Stage
+
+We are targeting this feature to be introduced in **Alpha**.
+
+- During Alpha:
+   - The `video` field will be **optional**.
+   - The feature will be **enabled by default**.
```
0xFelix, line 162: How so? If no video device was specified what will happen?

veps/sig-compute/vm_video_configuration_field.md
fabiand: +1

I also think that this can be controlled from within the guest. Maybe we want to set a modern resolution by default.

dasionov: done, adjusted the design and dropped this addition

veps/sig-compute/vm_video_configuration_field.md
dasionov: fixed

veps/sig-compute/vm_video_configuration_field.md
dasionov: fixed, thanks.

veps/sig-compute/vm_video_configuration_field.md
dasionov: done, fixed.

veps/sig-compute/vm_video_configuration_field.md
dasionov: done.

veps/sig-compute/vm_video_configuration_field.md
iholder101: Alpha features shouldn't be on by default, by rely on a feature gate.
Please remove this point.

dasionov: adjusted to ` - The feature will be **disabled by default**.`
Thanks for the update !

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+- We want to allow VM-owners to explicitly set the video device type when needed.
+- We want to allow VM-owners to explicitly set the desired video resolution to overcome the default resolution defined by libvirt/qemu
```
iholder101, line 11: leftover?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the `virtio` video device.
+
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
```
iholder101, line 18: leftover?

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+# Overview
+- We want to allow VM-owners to explicitly set the video device type when needed.
+- We want to allow VM-owners to explicitly set the desired video resolution to overcome the default resolution defined by libvirt/qemu
+
+## Motivation
+In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
+For Arm and s390x architectures, we have transitioned to using the `virtio` video device.
```
iholder101, line 15: I don't see the motivation here.
Can you please write something like:
```suggestion
In KubeVirt, the default video device type for AMD64 architectures is VGA for BIOS-based VMs and Bochs for EFI-based VMs.
For Arm and s390x architectures, we have transitioned to using the `virtio` video device.

In other words, the video device that is chosen for VMs is implicit and unclear.
Different video devices have pros and cons and fit to different use-cases.
For example, a guest with virtio drivers installed could enjoy from a virtio device which is generally faster and optimized for virtualization. Legacy guests, or guests with no virtio drivers, can use VGA for compatibility, and so on.

With this proposal, the VM creator could adapt the video deceive to their needs.
```

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime.
```
iholder101, line 22: leftover.
You can rephrase the motivation part into here.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+  In these cases, the guest operating system defaults to a limited resolution during boot.
+
+## User Stories
+- As VM owner I want to run applications requiring a specific resolution in order to work
```
iholder101, line 26: leftover

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Goals
+- Allowing VM-owners to have control over the resolution of the video device with their VMs
+- Allowing VM-owners to choose `virtio` or any other desired video device to override the existing default one.
+
+## Definition of Users
+- VM owners whose workloads require specific display resolutions but are unable to modify them at runtime. 
+  In these cases, the guest operating system defaults to a limited resolution during boot.
+
+## User Stories
```
iholder101, line 25: I'd add:
* as a VM creator with legacy guest or guest with no virtio drivers, I want to rely on VGA.
* as a VM creator running a guest with virtio drivers, I want to enjoy from an optimized virtualization-aware video device which is generally faster and lighter on resources.

These are the most simple user stories here AFAICT

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Repos
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+- [common-templates](https://github.com/kubevirt/common-templates)
+- [common-instancetypes](https://github.com/kubevirt/common-instancetypes)
+
+
+## Design
+
+### Preferred Option: Add `Video` Configuration Field to VM API
```
iholder101, line 43: Please mention that this approach was chosen to implement.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,156 @@
...
+Overhead of the virt-launcher/qemu will be measured and adjusted accordingly.
+
+## Update/Rollback Compatibility
+Currently, the virt-launcher arch-converter manages the default video type.
+We only provide a field for the user to specify explicitly if they want something else like virtio, so backward compatibility isn't affected.
+
+## Functional Testing Approach
+* Create a VM with video.Type set to `virtio` and expect launch successfully.
+
+# Implementation Phases
```
iholder101, line 139: This is section is for implementation phases that had already happened. IOW you'll add major PRs to here after they'll get merged.

So you can remove this section for now, or move the important stuff to the alpha requirements section.

iholder101 (commented):
some minor comments

iholder101 (commented):
some minor comments

veps/sig-compute/vm_video_configuration_field.md
dasionov: kinda, but i chose to leave it at that because it's an issue being solved by switching the video device type, and not by introducing new api for resolution anymore

dasionov: will drop it

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed this block.

veps/sig-compute/vm_video_configuration_field.md
dasionov: done

veps/sig-compute/vm_video_configuration_field.md
dasionov: done, thanks.

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed

veps/sig-compute/vm_video_configuration_field.md
dasionov: done.

veps/sig-compute/vm_video_configuration_field.md
dasionov: removed it

veps/sig-compute/vm_video_configuration_field.md
dasionov: done

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,151 @@
...
+- VM owners who need to select a specific video device for their virtual machines to match their workload requirements.
+
+## User Stories
+- As an VM owner, I want my VM to provide high-resolution display, so that I can use high resolution
+  when I access the VM display via an in-guest graphical remote display mirroring access method like an in guest VNC server
+
+- As a Windows 11 on ARM user, I like to install my VM using the ramfb display adapter,
+  and replace it by the `virtio` adapter after I installed the `virtio` driver
+
+- as a VM creator with legacy guest or guest with no `virtio` drivers, I want to rely on VGA.
```
dominikholler, line 36:
```suggestion
- As a VM creator with legacy guest or guest with no `virtio` drivers, I want to rely on VGA for my x64 VMs.
```
Please note if you request virtio from libvirt on x64, you will get virtio-vga which provides vga compatibility.
An non-x64 platforms there is no VGA at all.

veps/sig-compute/vm_video_configuration_field.md
```diff
@@ -0,0 +1,151 @@
...
+- as a VM creator running a guest with `virtio` drivers, I want to enjoy from an optimized virtualization-aware video device which is generally faster and lighter on resources.
```
dominikholler, line 37:
```suggestion
- As a VM creator running a guest with `virtio` drivers, I want to enjoy from an optimized virtualization-aware video device which is generally faster and lighter on resources.
```

dominikholler: I am not convicned if this is significant here, but it might not hurt.

iholder101:
Thank you very much for driving this @dasionov, for your patience and the lengthy discussion above.

I think the proposal is in a good state to be merged.
Since we've already missed the deadline and accepted an exception for up to today, I'd really prefer to merge this and create follow-up adjustments in the near future for nits and such, unless there are major objections.

@vladikr @0xFelix @xpivarc PTAL.

/lgtm

veps/sig-compute/vm_video_configuration_field.md
dasionov: even when drivers are already installed?

veps/sig-compute/vm_video_configuration_field.md
dasionov: I agree

veps/sig-compute/vm_video_configuration_field.md
dasionov: added pre-requisite as a con to the alternative and adjusted the design.

veps/sig-compute/vm_video_configuration_field.md
dasionov: Agreed.

veps/sig-compute/vm_video_configuration_field.md
dasionov: I will verify this on a windows machine myself when i get the chance :+1:

veps/sig-compute/vm_video_configuration_field.md
dominikholler: > even when drivers are already installed?

I do not understand the question. kubevirt creates xml which is send to libvirt which is generting the qemu command line from it. So if kubevirt generates xml to use virtio, libvirt generates a virtio-vga command line for qemu. This all happens before the vm is started and the driver is relevant.

dasionov: sorry, you are right, but virtio-vga would still work for the goal of this vep will it not?

vladikr:
Thank you @dasionov !

As per the discussion today on the sig-compute call, this VEP seems to be ready. 
@xpivarc has some additional comments and a request to validate a certian onfiguration
We should update this VEP if the implementation diverges for some reason.

/approve

xpivarc:
So I was wondering about the virtio and asked around. It seems plain `virtio` is interpreted differently on each arch. `virtio-vga` and `virtio-gpu-pci` for x86 and arm respectively. This might be confusing in the API. I also looked at the `virtio-vga` which promises a vga compatible mode and an enhanced features (such as multi-head) if the guest have a drivers. I found the vga compatibility quit interesting and possibly good default for KubeVirt.
@dasionov can you try this out, with uefi, bios and guests with and without drivers respectively?

@vladikr fyi

dasionov:
> So I was wondering about the virtio and asked around. It seems plain `virtio` is interpreted differently on each arch. `virtio-vga` and `virtio-gpu-pci` for x86 and arm respectively. This might be confusing in the API. I also looked at the `virtio-vga` which promises a vga compatible mode and an enhanced features (such as multi-head) if the guest have a drivers. I found the vga compatibility quit interesting and possibly good default for KubeVirt. @dasionov can you try this out, with uefi, bios and guests with and without drivers respectively?
> 
> @vladikr fyi

Just to clarify — KubeVirt configures video devices using the Libvirt API, and the only "virtio" accepted model value we currently pass is "virtio". Attempting to use "virtio-vga" directly results in an error from Libvirt:
```
{"component":"virt-launcher","kind":"","level":"error","msg":"Defining the VirtualMachineInstance failed.","name":"vm-cirros","namespace":"kubevirt","pos":"libvirt_helper.go:136","reason":"virError(Code=27, Domain=5, Message='XML error: Invalid value for attribute 'type' in element 'model': 'virtio-vga'.')","timestamp":"2025-05-09T13:22:02.044471Z","uid":"5920b8aa-97a5-4b69-b4d5-607ecbddceaa"}
```
This happens because Libvirt has a fixed list of supported video model names, and "virtio-vga" is not among them.
[ref](https://gitlab.com/libvirt/libvirt/-/blob/master/src/conf/domain_conf.h?ref_type=heads#L1826)

so how can we default to `virtio-vga` I still don't understand :disappointed:

xpivarc:
@dasionov for default I meant x86

Issue #41, opened 2025-04-10
VEP 41: Object Graph subresource


alromeros:
**Primary contact (assignee)**:

/assign @alromeros

**Current Feature Stage**:

New

**Feature Gate**:

ObjectGraph

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/42

**Timeline**:

* Alpha release target (x.y): 1.6
* Beta release target (x.y): 1.8
* Stable release target (x.y): 1.9

* [ ] Alpha
  * [x] VEP PRs:
      - https://github.com/kubevirt/enhancements/pull/42
      - https://github.com/kubevirt/enhancements/pull/78
  * [x] Code PRs:
      - https://github.com/kubevirt/kubevirt/pull/14807
      - https://github.com/kubevirt/kubevirt/pull/14935
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

alromeros:
/sig compute
Adding compute as secondary sig since the feature is also pertinent for them.

alicefr:
@alromeros thinking a bit more to the entire proposal. Why do we actually want this in kubevirt core? AFAIU, this is basically some discovery based on the VM specification and maybe some query to k8s API. Hence, why do we want this on the server side? Cannot we simply implement everything on the client side, maybe doing some library that backup tools can include?

vladikr:
@alromeros is there a PR that implements this feature?

alromeros:
> [@alromeros](https://github.com/alromeros) is there a PR that implements this feature?

The PR is in progress, I'll open it asap.

alromeros:
@fossedihelm sure, thanks for the suggesiton!

fossedihelm:
@alromeros Thank you! Let's add the IPAM part as a requirement for the beta?

orelmisan:
I would like to raise my concern that the feature, in its current state, introduces an awareness of third-party integrations into KubeVirt as a core project.

This awareness manifests in several ways:

1. The core project becomes reliant on and directly interacts with CRDs defined by external projects.
2. New RBAC rules are required within KubeVirt to accommodate the permissions needed for these third-party integrations.
3. Logic specific to these external integrations is being embedded directly within the KubeVirt codebase.

[...]

PR #42, opened 2025-04-10
Enhancement: Object Graph subresource

alromeros:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/41
**SIG label**: 
/sig storage
/sig compute

### What this PR does

This Pull Request introduces a enhancement proposal for an Object Graph API in KubeVirt, which aims to provide an authoritative way to represent VM and VMI dependencies and their relationships.

POC: https://github.com/kubevirt/kubevirt/pull/13875

### Release note

```release-note
NONE
```

### DIFF
--- a/veps/sig-storage/vm-object-graph.md
+++ b/veps/sig-storage/vm-object-graph.md
@@ -48,28 +48,29 @@ As new features continue to be added to KubeVirt, the graph of objects related t
 /apis/subresources.kubevirt.io/v1/namespaces/{namespace}/virtualmachineinstances/{name}/objectgraph
 ```
 
-### API Examples
+### Graph Representation
 
-#### **1. First option: Hierarchical Representation**
+We propose a **Hierarchical Object Graph** to represent dependencies.
+Each object can have child objects, capturing direct relationships between resources (e.g., VM → VMI → Pod → PVC).
 
-This format visualizes dependencies in a tree structure, showing parent-child relationships between resources.
+#### API Schema
 
 ```go
-// ObjectGraphNode represents an individual resource node in the graph.
+// ObjectGraphNode represents an individual node in the graph.
 type ObjectGraphNode struct {
-	ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
-	Labels          map[string]string          `json:"labels,omitempty"`
-	Optional        bool                       `json:"optional"`
-	Children        []ObjectGraphNode          `json:"children,omitempty"`
+    ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
+    Labels          map[string]string          `json:"labels,omitempty"`
+    Optional        bool                       `json:"optional"`
+    Children        []ObjectGraphNode          `json:"children,omitempty"`
 }
 
-// ObjectGraphNodeList represents a list of object graph nodes.
+// ObjectGraph represents the complete dependency graph.
 //
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-type ObjectGraphNodeList struct {
-	metav1.TypeMeta `json:",inline"`
-	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []ObjectGraphNode `json:"items"`
+type ObjectGraph struct {
+    metav1.TypeMeta `json:",inline"`
+    metav1.ListMeta `json:"metadata,omitempty"`
+    RootNodes       []ObjectGraphNode `json:"rootNodes"`
 }
 ```
 
@@ -131,84 +132,6 @@ type ObjectGraphNodeList struct {
 }
 ```
 
-#### **2. Second option: Flat Dependency List**
-
-This format provides a simple list of dependencies without indicating hierarchical relationships.
-
-```go
-// ObjectGraphNode represents a node in the object graph.
-type ObjectGraphNode struct {
-	ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
-	Labels          map[string]string          `json:"labels,omitempty"`
-}
-
-// ObjectGraphNodeList represents a list of object graph nodes.
-//
-// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-type ObjectGraphNodeList struct {
-	metav1.TypeMeta `json:",inline"`
-	metav1.ListMeta `json:"metadata,omitempty"`
-	Items           []ObjectGraphNode `json:"items"`
-}
-```
-
-##### Example Output
-
-```json
-{
-  "items": [
-    {
-      "objectReference": {
-        "apiGroup": "kubevirt.io",
-        "kind": "virtualmachineinstances",
-        "name": "vm-cirros-source-ocs",
-        "namespace": "default"
-      },
-      "labels": {
-        "type": "",
-        "optional": "false"
-      }
-    },
-    {
-      "objectReference": {
-        "apiGroup": "",
-        "kind": "pods",
-        "name": "virt-launcher-vm-cirros-source-ocs-frn9h",
-        "namespace": "default"
-      },
-      "labels": {
-        "type": "",
-        "optional": "false"
-      }
-    },
-    {
-      "objectReference": {
-        "apiGroup": "cdi.kubevirt.io",
-        "kind": "datavolumes",
-        "name": "cirros-dv-source-ocs",
-        "namespace": "default"
-      },
-      "labels": {
-        "type": "storage",
-        "optional": "false"
-      }
-    },
-    {
-      "objectReference": {
-        "apiGroup": "",
-        "kind": "persistentvolumeclaims",
-        "name": "cirros-dv-source-ocs",
-        "namespace": "default"
-      },
-      "labels": {
-        "type": "storage",
-        "optional": "false"
-      }
-    }
-  ]
-}
-```
-
 ### **User Flow**
 
 1. Access the ObjectGraph API through the subresource endpoint for a VM/VMI.
@@ -235,16 +158,72 @@ type ObjectGraphNodeList struct {
 **Backend Storage PVC**  
 Identified by the persistent state PVC label.
 
-**Other Resources:**  
-- Should we include `networkAttachmentDefinitions` and `networks` in the ObjectGraph?  
-- Should optional objects such as `VMExports` or `VMSnapshots` be considered?  
+**Other Resources:**
+- Should we include `networkAttachmentDefinitions` and `networks` in the ObjectGraph?
+- Should optional objects such as `VMExports` or `VMSnapshots` be considered?
 
 ## **Alternatives**
+
+## Alternatives Considered
+
+### Flat Dependency List
+
+An alternative was to return a **flat list** of dependent objects without hierarchical relationships.
+
+Example schema:
+
+```go
+type ObjectGraphNode struct {
+    ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
+    Labels          map[string]string          `json:"labels,omitempty"`
+}
+```
+
+Example output:
+
+```json
+{
+  "items": [
+    {
+      "objectReference": {
+        "apiGroup": "kubevirt.io",
+        "kind": "VirtualMachineInstance",
+        "name": "vm1",
+        "namespace": "default"
+      }
+    },
+    {
+      "objectReference": {
+        "apiGroup": "",
+        "kind": "Pod",
+        "name": "virt-launcher-vm1",
+        "namespace": "default"
+      }
+    },
+    {
+      "objectReference": {
+        "apiGroup": "cdi.kubevirt.io",
+        "kind": "DataVolume",
+        "name": "dv1",
+        "namespace": "default"
+      }
+    }
+  ]
+}
+```
+
+We chose the hierarchical representation because:
+
+- Dependency relationships between objects are meaningful and should be explicit.
+- Migration, backup, and restore processes often require processing objects in dependency order.
+- Hierarchical graphs allow easier extensibility (for example, marking optional nodes).
+- Flat lists can become complex as the number of dependencies grows.
+
+### Other Considerations
 
 1. **Naming:** Is `ObjectGraph` descriptive enough even if we are returning a flat list of objects? Would `DependencyList` be more accurate?
 2. **Extensibility:** How can we ensure the API is extensible for future enhancements? Should the API be made more intelligent (with fields such as `Optional`) or just rely on labels for extensibility?
 3. **Filtering:** Should the user handle filtering, or should we allow some kind of filtering in the ObjectGraph request?
-4. **API Design:** Should we use a hierarchical or flat list representation for the Object Graph API?
 
 ## **Scalability**
 
--- a/veps/sig-storage/vm-object-graph.md
+++ b/veps/sig-storage/vm-object-graph.md
@@ -1,0 +1,277 @@
+# VEP #41: Object Graph API for VM Dependencies
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] : https://github.com/kubevirt/enhancements/issues/41
+
+## **Overview**
+
+This is a proposal to include an `Object Graph` API in KubeVirt to represent VM and VMI dependencies and their relationships.
+
+## **Motivation**
+
+As new features continue to be added to KubeVirt, the graph of objects related to VMs (DataVolumes, PersistentVolumeClaims, InstanceTypes, Preferences, Secrets, ConfigMaps, etc.) continues to expand. Identifying all the objects that a VM depends on for tasks like backup, disaster recovery, or migration can be error-prone. We should simplify this process for users and partners by creating an authoritative way to retrieve a structured object graph.
+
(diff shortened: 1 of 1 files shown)


alromeros:
Migrating from https://github.com/kubevirt/community/pull/385 as suggested by @mhenriks

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,279 @@
...
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+- [KubeVirt Velero Plugin](https://github.com/kubevirt/kubevirt-velero-plugin)
+
+## **Design**
+
+### API Endpoints
+
+```
+/apis/subresources.kubevirt.io/v1/namespaces/{namespace}/virtualmachines/{name}/objectgraph
+/apis/subresources.kubevirt.io/v1/namespaces/{namespace}/virtualmachineinstances/{name}/objectgraph
```
alicefr, line 48: Do we need to define this also for VMIs? Can we just focus on VMs?

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+### API Examples
+
+#### **1. First option: Hierarchical Representation**
```
alicefr, line 53: Can you please add a section with your analysis and your preference for option 1 vs option 2? I don't think there are many parent-child relationship right? I mean, it is almost everything under volumes. But in general I tend to prefer a hierarchical representations.

alicefr:
@alromeros it already looks quite good. However, I would like to focus this on VMs only if possible

veps/sig-storage/vm-object-graph.md
alromeros: Are there any downsides to enabling the subresource for VMIs? I’d argue it’s equally important to have it enabled for both VMs and VMIs... for example, when backing up a VMI directly using Velero, we still need to get its object graph.

alicefr: Why are they backing up VMIs? VMIs are ephemeral objects.

mhenriks: I think it makes sense to support standalone VMIs. They are backed up with Velero for example

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,279 @@
...
+
+This format visualizes dependencies in a tree structure, showing parent-child relationships between resources.
+
+```go
+// ObjectGraphNode represents an individual resource node in the graph.
+type ObjectGraphNode struct {
+	ObjectReference k8sv1.TypedObjectReference `json:"objectReference"`
+	Labels          map[string]string          `json:"labels,omitempty"`
+	Optional        bool                       `json:"optional"`
+	Children        []ObjectGraphNode          `json:"children,omitempty"`
```
mhenriks, line 63: I think we should think about having Children being []k8sv1.TypedObjectReference instead. I know KubeVirt doesn't do this but it is theoretically possible for object references to be recursive. obj1 owns obj2 and obj2 own obj1 so I'm not sure we want to have graphs within graphs in the json

alromeros: Your suggestion seems more aligned with the second API example I proposed in the enhancement (flat dependency list). But in a hierarchical format, having full ObjectGraphNodes as children allows for stuff like applying custom labels or filtering subtrees based on those labels. It would feel a bit limiting if only the top-level node could carry those fields, but not its children, when precisely only makes sense for the children to have fields such as `optional`.

veps/sig-storage/vm-object-graph.md
alicefr: Still, it isn't a reason. Why Velero is doing this? I mean if backup VMs and VMIs, then is a VM backed up twice?

veps/sig-storage/vm-object-graph.md
alicefr: @mhenriks do you have an example in mind? Is recursion really possible? Since we are describing kubevirt object, we can refuse to put recursive elements in there. Seems a nightmare for deletion if there are those situation

veps/sig-storage/vm-object-graph.md
alromeros: >  I mean if backup VMs and VMIs, then is a VM backed up twice?

The velero plugin for kubevirt already handles this scenario and ensures only the required resources are backed up for a VM/VMI to run properly, but even if a VM were backed up twice, velero discards duplicates automatically.

> Why Velero is doing this? 

For example, if the backup is being used purely for metadata or debugging purposes (which I believe some customers are doing), it still makes sense to support backing up a standalone VMI imo.

veps/sig-storage/vm-object-graph.md
mhenriks: @alromeros why are children not included in your flat graph? How does a user know that the datavolume in your example is a child of the VM and not the VMI?

@alicefr Pretty sure there are no recursive references in KubeVirt, coming at this more from a theoretical level and the "flat" representation is more similar to an adjacency list

alicefr:
@alromeros thinking a bit more about this proposal, why are we implementing this on server side and in KuebVirt code, and not just a client library? AFAIU, this should only involve VM spec and maybe a couple of queries in the k8s API

alromeros:
> @alromeros thinking a bit more about this proposal, why are we implementing this on server side and in KubeVirt code, and not just a client library? AFAIU, this should only involve VM spec and maybe a couple of queries in the k8s API

I'm not against a client-side implementation in principle, but from an integration standpoint, having the graph exposed directly by the VM simplifies things a lot... For example, in Velero plugin or MTV, relying on a subresource would be much simpler and more authoritative than depending on a separate repository that needs to handle API differences and other compatibility concerns.

Also wanted to mention that we discussed this during the KubeVirt unconference 1.5, and the approach that got the most support was exposing the object graph as a VM subresource. Embedding it in the status was also considered, but the subresource option was seen as cleaner and more aligned with existing KubeVirt patterns.

alicefr:
> > @alromeros thinking a bit more about this proposal, why are we implementing this on server side and in KubeVirt code, and not just a client library? AFAIU, this should only involve VM spec and maybe a couple of queries in the k8s API
> 
> I'm not against a client-side implementation in principle, but from an integration standpoint, having the graph exposed directly by the VM simplifies things a lot... For example, in Velero plugin or MTV, relying on a subresource would be much simpler and more authoritative than depending on a separate repository that needs to handle API differences and other compatibility concerns.
> 

Not sure about this, I mean, the library will be hosted and released with KubeVirt so it isn't from some random user. Overlaying tools needs in any case be able to parse the object graph returned by the API endpoint, so, in any case, they need to include new API modifications.
Doing a separate library could in my opinion gives a clear ownership and separation, it can be owned by sig-storage only for example.
@mhenriks @vladikr what are your thoughts on this?
In any case, I'd like to see a summary and conclusion of this discussion in the alternative section.

> Also wanted to mention that we discussed this during the KubeVirt unconference 1.5, and the approach that got the most support was exposing the object graph as a VM subresource. Embedding it in the status was also considered, but the subresource option was seen as cleaner.

I have unfortunately missed the unconference. Do you have the notes where this was discussed.

veps/sig-storage/vm-object-graph.md
alromeros: > How does a user know that the datavolume in your example is a child of the VM and not the VMI?

The flat graph is meant to be more utilitarian, it gives you the full list of dependencies for a VM/VMI so users can know what it's exactly needed to recreate it but doesn't go into ownership or relationships. In that context, whether the DV is linked to the VM or VMI might not matter. But yeah that's a good point, I'm starting to think a hierarchical representation is more appropriate.

alromeros:
> I have unfortunately missed the unconference. Do you have the notes where this was discussed.

Seems the notes of the unconference 1.5 were replaced by the ones from 1.6... that said I don't think we are missing anything critical.

> Doing a separate library could in my opinion gives a clear ownership and separation, it can be owned by sig-storage only for example.

I might prefer the subresource route but I'm not against this. Let's see what @mhenriks and @vladikr think, also pinging @awels since he'll likely be one of the first to use this for cross-cluster migration

alromeros:
> > > @alromeros thinking a bit more about this proposal, why are we implementing this on server side and in KubeVirt code, and not just a client library? AFAIU, this should only involve VM spec and maybe a couple of queries in the k8s API
> > 
> > I'm not against a client-side implementation in principle, but from an integration standpoint, having the graph exposed directly by the VM simplifies things a lot... For example, in Velero plugin or MTV, relying on a subresource would be much simpler and more authoritative than depending on a separate repository that needs to handle API differences and other compatibility concerns.
> > 
> 
> Not sure about this, I mean, the library will be hosted and released with KubeVirt so it isn't from some random user. Overlaying tools needs in any case be able to parse the object graph returned by the API endpoint, so, in any case, they need to include new API modifications.
> Doing a separate library could, in my opinion, give a clear ownership and separation, it can be owned by sig-storage only for example.
> @mhenriks @vladikr what are your thoughts on this?
> In any case, I'd like to see a summary and conclusion of this discussion in the alternative section.
> 
> > Also wanted to mention that we discussed this during the KubeVirt unconference 1.5, and the approach that got the most support was exposing the object graph as a VM subresource. Embedding it in the status was also considered, but the subresource option was seen as cleaner.
> 
> I have unfortunately missed the unconference. Do you have the notes where this was discussed.

After thinking more about this, I’m definitely leaning towards the subresource approach. There’s already a library in the velero plugin that builds the object graph, but what we really want is to have this as part of the core project. Some reasons:

- We don’t want this to live in an external repo where most people forget about it and only a few maintainers update it. If a new dependency is introduced to VMs, the developer adding it should also update the object graph logic in the main project and consider it a single source of truth. This ensures consistency and prevents issues like the backend storage PVC being left out from the vm status and requiring complex filtering to fetch it.

- IMO It makes sense for VMs themselves to expose their dependencies directly, If we can’t include that info in the spec or status, then a subresource is the cleanest way to make that information discoverable without needing to rely on a external library.

-  I agree the current graph logic is mostly parsing the VM spec and doing basic client calls but this might change in the future and we cant discard needing to run libvirt api calls for some reason. Keeping this inside Kubevirt gives us some flexibility to extend it on the server side if needed.

- This is not too disruptive with existing subresources like fslist or userlist. It's true those require libvirt calls, but the way they work by exposing VM-specific information via a subresource is really similar to what we’re suggesting.

Let me know your thoughts.

mhenriks:
> > I have unfortunately missed the unconference. Do you have the notes where this was discussed.
> 
> Seems the notes of the unconference 1.5 were replaced by the ones from 1.6... that said I don't think we are missing anything critical.
> 
> > Doing a separate library could in my opinion gives a clear ownership and separation, it can be owned by sig-storage only for example.
> 
> I might prefer the subresource route but I'm not against this. Let's see what @mhenriks and @vladikr think, also pinging @awels since he'll likely be one of the first to use this for cross-cluster migration

I think the subresource API is the best way to serve up this data, that way client doesn't have to worry about having compatible client/server versions. Also provides programming language independence

mhenriks:
@alromeros can you make it clear what graph representation you are choosing in the proposal?

alicefr:
@mhenriks can you please take a look?

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,258 @@
...
+    }
+  ]
+}
+```
+
+We chose the hierarchical representation because:
+
+- Dependency relationships between objects are meaningful and should be explicit.
+- Migration, backup, and restore processes often require processing objects in dependency order.
+- Hierarchical graphs allow easier extensibility (for example, marking optional nodes).
```
mhenriks, line 221: What does this mean?

alromeros: There's an `optional` field in the hierarchical API that tells whether an object is optional for the VM to work or not. This would allow extend the graph with objects not yet included like VMSnapshots or VMExports.

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,258 @@
...
+
+### **User Flow**
+
+1. Access the ObjectGraph API through the subresource endpoint for a VM/VMI.
+2. Parse the response and filter unnecessary objects (e.g., in backup scenarios).
+3. Use the retrieved data as needed.
+
+### **Included Resources**
+
+- **InstanceType (`spec.instancetype.name`)**
```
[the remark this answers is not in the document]
alromeros: wondering if these are still worth including in the graph if we mark them as optional. Since they're part of the spec, would the graph be considered incomplete without them?

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,258 @@
...
+1. Implement ObjectGraph types in the API.
+2. Implement virt-api logic to construct and return graphs.
+3. Expose endpoints under VM and VMI subresources.
+4. Add virtctl integration for user access.
+
+## **Feature Lifecycle Phases**
+
+- **Alpha:** Initial implementation with basic functionality.
+- **Beta:** Bug fixes and improvements based on feedback.
+- **GA:** Adapt external repos (e.g., Velero plugin) to use the ObjectGraph API.
```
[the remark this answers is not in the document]
alromeros: > Apologies if this complicates things but could you make the behaviour of the plugin configurable to allow us to switch between the current implementation and this new subresource based kvgraph?

Why would we need that? I guess we could include a label or annotation in both backup and restore objects to choose either one but I don't see the point if the dependency list is the same.

> Are there any plans to also use the subresource in our in-tree snapshot and restore flows?

Don't think this is needed as vmsnapshot and vmrestore only require the volume list.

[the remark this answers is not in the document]
alromeros: > I don't think we will get the required feedback for GA until we adapt the plugin.
ACK

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,258 @@
...
+### **User Flow**
+
+1. Access the ObjectGraph API through the subresource endpoint for a VM/VMI.
+2. Parse the response and filter unnecessary objects (e.g., in backup scenarios).
+3. Use the retrieved data as needed.
+
+### **Included Resources**
+
+- **InstanceType (`spec.instancetype.name`)**
+  - **ControllerRevision (`spec.instancetype.revisionName`)**
```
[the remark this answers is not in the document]
alromeros: ack

veps/sig-storage/vm-object-graph.md
```diff
@@ -0,0 +1,258 @@
...
+- **Preference (`spec.preference.name`)**
+  - **ControllerRevision (`spec.preference.revisionName`)**
```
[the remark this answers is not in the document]
alromeros: ack

veps/sig-storage/vm-object-graph.md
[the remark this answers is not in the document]
lyarwood: > > Apologies if this complicates things but could you make the behaviour of the plugin configurable to allow us to switch between the current implementation and this new subresource based kvgraph?
> 
> Why would we need that? I guess we could include a label or annotation in both backup and restore objects to choose either one but I don't see the point if the dependency list is the same.

It was more to allow optional feedback of the API while it's still beta without forcing the plugin to use it. If you're comfortable just switching the plugin over then that's also cool but I'd recommend doing it while the API is beta so we can get feedback and still make changes if required.

> > Are there any plans to also use the subresource in our in-tree snapshot and restore flows?
> 
> Don't think this is needed as vmsnapshot and vmrestore only require the volume list.

That isn't true, we also backup and restore the instance type and preference controller revisions allowing the originals to be deleted. I would expect that would be the same for everything else hanging off the VM.

veps/sig-storage/vm-object-graph.md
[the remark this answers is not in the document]
alromeros: Removed them, we can always add them later during implementation if we are ok with adding optional dependencies.

veps/sig-storage/vm-object-graph.md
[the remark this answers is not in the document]
lyarwood: >> Are there any plans to also use the subresource in our in-tree snapshot and restore flows?
>
> Don't think this is needed as vmsnapshot and vmrestore only require the volume list.

We already capture and restore instance type and preference controller revisions as part of these flows so I assumed we'd want to do the same with other objects hanging off a VirtualMachine.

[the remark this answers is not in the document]
alromeros: > We already capture and restore instance type and preference controller revisions as part of these flows so I assumed we'd want to do the same with other objects hanging off a VirtualMachine.

You are right, I will look into it. Still, the snapshot flow requires a more targeted analysis of the VM and getting an unsorted list of graph nodes might require unnecessary filtering. It's still worth considering.

lyarwood (commented):
/lgtm

alromeros:
> @alromeros the proposal overall looks good, for this /approve However, you need to address Mike's comments about the choose type of the graph. You can also remove the example of the flat list. Maybe, just add a note to the alternatives, why we have choose the hierarchical representation. /hold

The /approve didn’t go through, looks like only @vladikr and @xpivarc have approval rights. Do you mind approving if everything looks good?

vladikr:
/approve

Thank you!

vladikr:
/hold
@mhenriks can you please release the hold if you are satisfied

Issue #14, opened 2025-03-05
VEP 14: Track the second phase of the firmware UUID implementation fix


dasionov:
**Primary contact (assignee)**:

/assign @dasionov 

**Current Feature Stage**:

Work In Progress

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/43

**Responsible SIGs**:

/sig compute

**Enhancement link**:
https://github.com/kubevirt/community/pull/347

**Timeline**:

* Alpha release target (x.y): 1.6
Implemented

* Stable release target (x.y): 1.9

* [ ] Requires removal of legacy handling

* [ ] Alpha
  * [x] VEP PRs:
    * https://github.com/kubevirt/enhancements/pull/43
    * https://github.com/kubevirt/community/pull/347
  * [x] Code PRs:
    * https://github.com/kubevirt/kubevirt/pull/14130
  * [ ] Docs PRs:

**Additional context**:
This issue is meant ot track the implantation of the second phase of the design  https://github.com/kubevirt/community/pull/347

vladikr:
/remove-lifecycle stale

@dasionov, where are we with this feature? anything to track in 1.8?

lyarwood:
@dasionov can you update the description to highlight the fact this GA'd in v1.9 and that there's essentially only clean up left in v1.10? You should be able to the tracking issue to tracked in the VEP project then.

dasionov:
Thanks @lyarwood. I've updated the description, the legacy name-based UUID removal was targeted for 1.9 but slipped, so it's now in 1.10. 
One thing on GA: I don't think we're there yet. The whole point of this VEP is to retire the legacy name-based UUID mechanism, and that code is still in the tree. Until it's actually removed in 1.10, I'd keep this at Beta. Let me know if you're thinking about it differently.

PR #43, opened 2025-04-14
enhancements: update persistence of Firmware UUID for Virtual Machines

dasionov:
### VEP Metadata

**Tracking issue**: #14
/sig compute

### What this PR does
This VEP continues the discussion from
https://github.com/kubevirt/community/pull/347

### Changes Introduced:
- Added a new alternative: implementing the UUID persistence logic in a dedicated controller within virt-controller, separate from the main VM controller.

- Removed the Validation Webhook logic that enforced non-empty firmware UUIDs during VM update operations.
This was removed due to potential race conditions, where existing VMs (created before the new UUID mechanism) may not yet have a UUID populated by the virt-controller.
If a user attempts to update such a VM before the controller patches in the UUID, the validation webhook would incorrectly reject the update due to the missing field.

### Release note
```release-note
none
```

### DIFF
--- a/veps/sig-compute/persist-vmi-firmware-uuid.md
+++ b/veps/sig-compute/persist-vmi-firmware-uuid.md
@@ -1,0 +1,221 @@
+# VEP 14: Persistence of Firmware UUID for Virtual Machines
+
+## Release Signoff Checklist
+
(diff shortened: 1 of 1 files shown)


dasionov:
/close
Closing since the implementation was already completed in kubevirt/kubevirt#14130.

Issue #45, opened 2025-04-21
VEP 45: Support in-place pod resize for memory/cpu hotplug


iholder101:
**Primary contact (assignee)**:
/assign @iholder101 

**Current Feature Stage**: Alpha

**Feature Gate**: InPlaceHotplug

**Responsible SIGs**: sig-compute

Primary SIG:
/sig compute
 
**Enhancement link**: https://github.com/kubevirt/enhancements/pull/46

**Timeline**:

* Alpha release target (x.y): v1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
    * [ ] https://github.com/kubevirt/enhancements/pull/46
  * [ ] Code PRs:
  * [ ] Docs PRs:

fabiand:
Maybe the proposal should also include to use in-place pod resize when limits/requests are changed as well.

We had cases where users had set limits.memory (not desired), and then wanted to fix this. This required them to restart VMs. With in-place resize, they could have done the resizing immediately.

iholder101:
> Maybe the proposal should also include to use in-place pod resize when limits/requests are changed as well.
> 
> We had cases where users had set limits.memory (not desired), and then wanted to fix this. This required them to restart VMs. With in-place resize, they could have done the resizing immediately.

Thank you @fabiand! That's a very good point!
I'll make sure to include this in the VEP PR.

BTW, another idea, maybe in the future we can do some automation in order to autoscale virt-launcher's virt overhead on launchers, since we know we have a problem to assess the overhead reliably.

vladikr:
@iholder101 let's defer this 1.8 when you will be available.

iholder101:
Deferring this. Moving to "removed from milestone".

PR #46, opened 2025-04-21
In-place VM Hotplug (without live-migration)

iholder101:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/45
**SIG label**: sig/compute

### What this PR does
Recently, memory and CPU hotplug was added to KubeVirt. This allows users to add memory and CPU to a running VM on the fly.

When a VM gets hotplugged, the underlying virt-launcher pod's resources need to be modified accordingly. Traditionally in Kubernetes, pods are immutable. Once a pod is created, its resource requests and limits cannot be changed. Therefore, the hotplug feature was implemented by live-migrating the VM which would result in a new virt-launcher pod with the updated resources.

On the hotplug's original [design proposal](https://github.com/kubevirt/community/blob/main/design-proposals/cpu-hotplug.md#goals) (that pre-dated the VEP process) it's written:
> Implementation should be achievable today, with Kubernetes APIs that are at least in beta.
> Unfortunately, at the time of writing, the Kubernetes vertical pod scaling API is only alpha.

Fortunately, the in-place pod resize feature was [graduated to GA](https://kubernetes.io/blog/2025/12/19/kubernetes-v1-35-in-place-pod-resize-ga/) in Kubernetes 1.35. Therefore, Kubevirt should aim to move away from live-migrating the VM on hotplug and instead use the in-place pod resize feature.

### Notes for reviewers
This VEP does not include new APIs.

### DIFF
--- a/veps/sig-compute/45-in-place-hotplug/vep.md
+++ b/veps/sig-compute/45-in-place-hotplug/vep.md
@@ -1,0 +1,427 @@
+# VEP #45: In-Place Pod Resize for CPU/Memory Hotplug
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [X] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Table of contents
+
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non Goals](#non-goals)
+- [Definition of Users](#definition-of-users)
+- [User Stories](#user-stories)
+- [Repos](#repos)
+- [Design](#design)
+  - [Feature Gate](#feature-gate)
+  - [Decision Tree](#decision-tree)
+  - [In-Place Resize Flow](#in-place-resize-flow)
+  - [Container Resize Policy](#container-resize-policy)
+  - [Component Responsibilities](#component-responsibilities)
+  - [Preventing Workload-Updater Races](#preventing-workload-updater-races)
+  - [VMI Conditions](#vmi-conditions)
+  - [Failure Handling](#failure-handling)
+  - [Infeasible Resize](#infeasible-resize)
+- [API Examples](#api-examples)
+- [Alternatives](#alternatives)
+- [Risk and Limitations](#risk-and-limitations)
+  - [Dedicated CPUs](#dedicated-cpus)
+  - [Hugepages (Memory Only)](#hugepages-memory-only)
+  - [Pre-existing Issue: CPU Hotplug for Non-Migratable VMs](#pre-existing-issue-cpu-hotplug-for-non-migratable-vms)
+- [Scalability](#scalability)
+- [Update/Rollback Compatibility](#updaterollback-compatibility)
+- [Functional Testing Approach](#functional-testing-approach)
+- [Implementation History](#implementation-history)
+- [Graduation Requirements](#graduation-requirements)
+  - [Alpha](#alpha)
+  - [Beta](#beta)
+  - [GA](#ga)
+
+## Overview
+
+KubeVirt supports CPU and memory hotplug for running VMs. Currently, every hotplug
+operation requires a live migration, because Kubernetes pods were traditionally
+immutable - pod resource requests and limits could not be changed after creation.
+The migration creates a new virt-launcher pod with the updated resources.
+
+Kubernetes in-place pod resize [graduated to GA](https://kubernetes.io/blog/2025/12/19/kubernetes-v1-35-in-place-pod-resize-ga/) in Kubernetes 1.35.
+This allows modifying a pod's CPU and memory requests/limits without recreating
+the pod. The kubelet updates the container's cgroup limits directly.
+
+This VEP proposes using in-place pod resize to eliminate the migration requirement
+for hotplug. The hotplug flow becomes two phases:
+
+1. **Pod-level resize**: The VM controller patches the virt-launcher pod's resources,
+   triggering Kubernetes to update cgroup limits.
+2. **Guest-level hotplug**: virt-handler applies the change inside the VM via libvirt
+   (onlining vCPUs for CPU, hotplugging virtio-mem devices for memory).
+
+When in-place resize is not possible (dedicated CPUs, hugepages for memory), the
+existing migration-based path remains as a fallback.
+
+## Motivation
+
+The current migration-based hotplug has several drawbacks:
+
+- **Resource overhead**: Live migration temporarily doubles resource consumption (source
+  and target pods run simultaneously).
+- **Downtime risk**: Migration involves a brief stun-time during cutover and can fail,
+  leaving the VM in an intermediate state.
+- **Non-migratable VMs excluded**: VMs with passthrough devices (GPUs, vFPGAs, etc.)
+  cannot migrate, so they cannot hotplug CPU or memory at all today.
+- **Quota pressure**: The temporary second pod may exceed namespace ResourceQuotas,
+  blocking the hotplug entirely.
+- **Complexity**: The migration path involves multiple controllers (VM controller,
+  workload-updater, migration controller, virt-handler on both source and target nodes),
+  making the hotplug flow harder to reason about and debug.
+
+In-place resize eliminates all of these issues for the common case.
+
+## Goals
+
+- Use Kubernetes in-place pod resize for CPU and memory hotplug under the `LiveUpdate`
+  rollout strategy.
+- Make in-place resize the preferred method; fall back to migration only when in-place
+  is not feasible.
+- Enable hotplug for non-migratable VMs (e.g., VMs with passthrough devices).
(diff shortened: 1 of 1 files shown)


iholder101:
FYI @vladikr @acardace @jean-edouard

vladikr:
@iholder101 can you please double check if this works with guaranteed resources? As far as I know, it doesn't.

iholder101:
> @iholder101 can you please double check if this works with guaranteed resources? As far as I know, it doesn't.

Very good observation, thank you!
So it turns out that every QoS class pod can be resized, however this is not the case for dedicated cpus/memory [[1](https://github.com/kubernetes/enhancements/tree/61abddca34caac56d22b7db48734b7040dc68b43/keps/sig-node/1287-in-place-update-pod-resources#static-cpu--memory-policy)]:
> Resizing pods with static CPU & memory policy configured is out-of-scope for the beta release of in-place resize. If a pod is a guaranteed QOS on a node with a static CPU or memory policy configured, then the resize will be marked as infeasible.
> 
> This will be reconsidered post-beta as a future enhancement.

I will definitely try to dig into understanding where this limitation is coming from and emphasizing its importance to us.

If it will be decided that indeed dedicated-cpus are out of scope, we can fallback to the migration strategy for these specific VMs. I will update the VEP accordingly.

esotsal:
/cc 

For resize question, please check PR

https://github.com/kubernetes/kubernetes/pull/129719 

( there is a screencast at  PRs description with current proposal status ) , please note solution considers lowest threshold to be the CPU pinned during startup. Looking forward for your comments/review in that PR, your use case is very interesting.

iholder101:
> I will definitely try to dig into understanding where this limitation is coming from and emphasizing its importance to us.
> 
> If it will be decided that indeed dedicated-cpus are out of scope, we can fallback to the migration strategy for these specific VMs. I will update the VEP accordingly.

Updated the KEP accordingly.

iholder101:
> /cc ( to follow up )
> 
> For resize question, please check PR
> 
> [kubernetes/kubernetes#129719](https://github.com/kubernetes/kubernetes/pull/129719)
> 
> ( there is a screencast at PRs description with current proposal status ) , please note solution considers lowest threshold to be the CPU pinned during startup. Looking forward for your comments/review in that PR, your use case is very interesting.

Thank you very much for the pointer! I'll review the PR soon!

veps/sig-compute/45-in-place-hotplug/vep.md
```diff
@@ -0,0 +1,125 @@
...
+Refer to https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#releases for more details
+-->
+
+### Alpha
+
+- [ ] Implement in-place pod resize for CPU and memory hotplugs.
+
+### Beta
+- [ ] Turn this feature on by default.
+
```
Barakmor1, line 123: I think we should add GA Kubernetes In-Place resize as a requirement

iholder101: My opinion is that beta features should (in most cases) be enabled by default, this is also Kubernetes' policy and I think it makes a lot of sense.

There's a related discussion about about this in https://github.com/kubevirt/kubevirt/pull/14427, which aims to enable a path forward for beta features being on by default.

Barakmor1: Yes, I understand that Beta features should be enabled by default. This is why I think it would be better to wait until this feature gate is GA in Kubernetes before making it Beta in KubeVirt. I'm not sure we should have default behavior that depends on Kubernetes Beta/Alpha feature gates. We have a similar requirement for the ImageVolume feature gate.

iholder101: The general rule is that our feature gate will be 1 graduation phase below Kubernetes.
ImageVolumes are Beta on Kubernetes, therefore we allow it to become Alpha in Kubevirt.
Same goes for Beta and Alpha with the pod resize FG.

veps/sig-compute/45-in-place-hotplug/vep.md
```diff
@@ -0,0 +1,125 @@
...
+* The user to explicitly decide whether to use in-place pod resize or live-migration (as migration doesn't really makes sense anymore).
+
+## Definition of Users
+
+* VM owners.
+* Admins / namespace owners.
+
+## User Stories
+
```
rthallisey, line 54: As a user, in order to hotplug additional devices, I also need to hotplug more memory to support the new pass through devices.

iholder101: Thanks @rthallisey! I've added that!

> @iholder101, what resources are currently supported, or is it entirely open?

@vladikr According to the feature's [non-goals](https://github.com/kubernetes/enhancements/tree/61abddca34caac56d22b7db48734b7040dc68b43/keps/sig-node/1287-in-place-update-pod-resources#non-goals):
> updating extended resources or any other resource types besides CPU, memory

So yeah, only memory and CPU resources are currently supported.

rthallisey: With PT devices, it's very difficult support to live migration - also kvm won't allow it.  So avoiding live migrate will make hot plug of host devices possible.  The tricky part to support this feature is you will often need to hot plug memory to be used only by the compute container, not given to the guest, as part of the device hotplug.

iholder101: > The tricky part to support this feature is you will often need to hot plug memory to be used only by the compute container, not given to the guest

That should be entirely possible!

esotsal: How will you secure when reducing memory that you will avoid out of memory situations, given that it is not possible to get the exact memory usage from the container runtime ?

iholder101: > How will you secure when reducing memory that you will avoid out of memory situations, given that it is not possible to get the exact memory usage from the container runtime ?

This topic is still under discussions, since with cgroups v2 lowering memory limits will be accepted and cause OOM kills, and if k8s will do the validation it's bound to an inherit [TOC/TOU race](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use). More info can be found here: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use.

In any case, hot-unplug is not yet supported in Kubevirt in general. We can think of relaxing this limitation if the k8s feature will come up with a decent solution to this problem.

veps/sig-compute/45-in-place-hotplug/vep.md
```diff
@@ -0,0 +1,427 @@
...
+This VEP proposes using in-place pod resize to eliminate the migration requirement
+for hotplug. The hotplug flow becomes two phases:
+
+1. **Pod-level resize**: The VM controller patches the virt-launcher pod's resources,
+   triggering Kubernetes to update cgroup limits.
+2. **Guest-level hotplug**: virt-handler applies the change inside the VM via libvirt
+   (onlining vCPUs for CPU, hotplugging virtio-mem devices for memory).
+
+When in-place resize is not possible (dedicated CPUs, hugepages for memory), the
+existing migration-based path remains as a fallback.
```
jean-edouard, line 75: That means maintaining 2 different hotplug implementations. That is perfectly fine as long as this is in Alpha, but I'd argue that support for these edge cases is a requirement for Beta, so that GA can retire migration-based hotplug.

iholder101: Thanks @jean-edouard!

The problem is that Kubernetes doesn't support that yet, see https://github.com/kubernetes/enhancements/issues/5294.

I'm not that involved in the process, but AFAICT this KEP is in very early stages and not even alpha.

For this reason, I do think it makes sense to consider supporting in-place whenever we can and pay the price of maintaining these two flows. However, if we think that supporting in-place is not worth paying this price, I think it's a blocker for this VEP and we shouldn't bother introducing it as alpha at this point.

WDYT?

jean-edouard: I stand by my comment, I don't think we should commit to maintaining 2 different (GA) implementations of the same thing for an indeterminate amount of time.
Just my opinion though!

iholder101: Yup, fair point :) Thanks for bringing it up.
I agree. This is a price to pay which is not negligible.

@vladikr what are your thoughts on this?

veps/sig-compute/45-in-place-hotplug/vep.md
```diff
@@ -0,0 +1,427 @@
...
+  cannot migrate, so they cannot hotplug CPU or memory at all today.
+- **Quota pressure**: The temporary second pod may exceed namespace ResourceQuotas,
+  blocking the hotplug entirely.
+- **Complexity**: The migration path involves multiple controllers (VM controller,
+  workload-updater, migration controller, virt-handler on both source and target nodes),
+  making the hotplug flow harder to reason about and debug.
+
+In-place resize eliminates all of these issues for the common case.
+
+## Goals
```
Barakmor1, line 95: As discussed in [here](https://github.com/kubernetes/kubernetes/issues/131835), In-place Pod Resource Updates allows modifying a pod’s resources without re-scheduling. This may bypass out-of-tree quota mechanisms such as AAQ, which fully supports KubeVirt.

I’m concerned that users depending on AAQ may not be able to safely adopt this enhancement, or benefit from follow-up work built on top of it.

@iholder101 @vladikr Would it make sense to introduce a more general control mechanism that allows external controllers to temporarily gate or coordinate pod resource updates?

theblazehen:
This likely wouldn't hold up quality wise for a proper PR, but I've got this working for my own use [in my fork](https://github.com/theblazehen/kubevirt/commit/30d675b8bdf54360cca0ab93305b27b6156c469a)

This is just slopped together, but I've tested and it works with Linux and Windows guests - might be useful as a starting point, or a cherry-pick if anyone wants to pick up the feature themselves

Issue #55, opened 2025-05-29
VEP NNNN: Extend EvictionStrategy to work with specific cluster initiated virt-launcher pods disruption


tiraboschi:
**Primary contact (assignee)**:
/assign tiraboschi

**Current Feature Stage**:
1. Enhancement proposal.

**Feature Gate**:
tbd

**Responsible SIGs**:
Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #56, opened 2025-05-29
VEP 56: Extend `EvictionStrategy` to Support Specific Cluster-Initiated `virt-launcher` Pod Disruptions

tiraboschi:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/pull/55
**SIG label**: /sig compute

### What this PR does
This proposal enhances KubeVirt's EvictionStrategy by enabling live migration of VMs in response to cluster-initiated pod deletions caused by scheduler preemption or NoExecute taints. It ensures consistent behavior across eviction types, improving reliability for VM owners and cluster administrators.

This does **not** include handling node pressure eviction, which is locally initiated on the node (it's not a pod deletion request). [A previous proposal](https://github.com/kubevirt/enhancements/pull/9) was covering it.

### Release note
```release-note
NONE
```

### DIFF
--- a/veps/sig-compute/55-evictionstrategy-extension.md
+++ b/veps/sig-compute/55-evictionstrategy-extension.md
@@ -1,0 +1,154 @@
+# VEP #55: Extend `EvictionStrategy` to Support Specific Cluster-Initiated `virt-launcher` Pod Disruptions
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [X] (R) Graduation criteria filled
+
+## Overview
+
+Under specific well-known circumstances (e.g., scheduler preemption, `NoExecute` taints), a `virt-launcher` pod can be terminated (read deleted).
+In these same cases, `PodDisruptionBudget` can also be bypassed. This can easily result in a VM being gracefully shut down, even if its `EvictionStrategy` is set to `LiveMigrate` or `LiveMigrateIfPossible`.
+
+This proposal aims to have our controllers identify well-known cases (initially scheduler preemption and `NoExecute` taints) and, if the `EvictionStrategy` requires `LiveMigrate` or `LiveMigrateIfPossible`, attempt a live migration.
+
+## Motivation
+
+If the `EvictionStrategy` is configured with `LiveMigrate` or `LiveMigrateIfPossible`, the VM owners can reasonably assume that cluster actions outside of their control will result in a live migration. This is currently true for some cases (e.g., node drains, descheduler-initiated evictions) but not for others (e.g., scheduler preemption, `NoExecute` taints), leading to confusing or surprising behavior.
+This proposal enhances the reliability and robustness of `LiveMigrate`-based `EvictionStrategies`.
+
+## Goals
+- Extend the current behavior of the `EvictionStrategy` (which only works in some scenarios) to ensure consistent live migration attempts regardless of the eviction cause.
+- Handle standard Kubernetes features like scheduler preemption and `NoExecute` taints (e.g. well-known `CriticalAddonsOnly=true:NoExecute` that is correctly tolerated by daemonset pods) consistently within KubeVirt relaying on live migrations if possible.
+
+## Non Goals
+
+- This does **not** include handling node pressure eviction, which is locally initiated on the node (it's not a pod deletion request). [A previous proposal](https://github.com/kubevirt/enhancements/pull/9) was covering it.
+- This does **not** mean triggering live migration for all `virt-launcher` pod deletions (that would implicitly make a pod deletion request a live migration request). Only specific disruptions identified via the `DisruptionTarget` condition and known reasons should trigger it.
+
+## Definition of Users
+
+- **VM owner**: the user who owns a VM in his namespace on a Kubernetes cluster with KubeVirt.
+- **Cluster admin**: the administrator of the cluster.
+
+## User Stories
+
+- As a cluster admin, I want consistent behavior where VMs are always live-migrated during evictions rather than sometimes being shut down.
+- As a VM owner, I want scheduler priority preemption to use live migration, not shutdowns, for rearranging workloads.
+- As a cluster admin, I don’t want an unprivileged user to cause shutdowns of other users’ VMs via resource-heavy dummy pods.
+- As a cluster admin, I expect the `EvictionStrategy` to be honored even when using a `NoExecute` taint for maintenance tasks.
+
+# Design
+This section outlines the theory behind how this design works and the code changes required to implement this theory into practice.
+
+## Design Theory
+### Actual behavior
+#### Scheduler Priority Eviction
+
+When a pod cannot be scheduled due to constraints, the scheduler may trigger preemption.
+If the incoming pod has a high priority, the scheduler will preempt lower-priority pods by setting a condition `Type: v1.DisruptionTarget`, `Reason: v1.PodReasonPreemptionByScheduler`, then delete the pod directly (not using the eviction API).
+Currently, this causes a graceful shutdown of the VM, not a live migration—even if `evictionStrategy: LiveMigrate` is set.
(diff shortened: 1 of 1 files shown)


tiraboschi:
@iholder101 FYI

veps/sig-compute/55-evictionstrategy-extension.md
```diff
@@ -0,0 +1,154 @@
...
+- `PreemptionEvictionLiveMigration`
+- `NoExecuteTaintEvictionLiveMigration`
+
+Once enabled, KubeVirt responds to these eviction causes based on the per-VM or cluster-wide `EvictionStrategy`.
+
+## Alternatives
+
+The ideal long-term solution is for the scheduler and taint manager to use the eviction API.
+However, these are sensitive components so such a change could require a long path. Thus, this proposal can act as a consistent workaround in the meantime.
+In the past, something around ideas in this area got discussed in https://github.com/kubernetes/kubernetes/issues/91492
```
tiraboschi, line 98: [KEP-3280](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/3280-guarantee-pdb-when-preemption-happens) is also mentioning:
```
if the priority of the preemptor is greater than or equal to the value of AllowDisruptionByPriorityGreaterThanOrEqual in victim pod, the implementation will remain consistent with the existing behavior, meaning that the scheduler will try to select victims whose PDBs are not violated by preemption, but if no such victims are found, preemption will still happen and lower priority pods will be preempted via the /evictions endpoint despite their PDBs being violated. If the /eviction endpoint returns a response 429 Too Many Requests and the scheduler will fallback to deletion as an alternative.
```

enp0s3:
@tiraboschi There is a [KEP](Maintenance) about declarative node maintenance, I think this could be a possible path forward because it also involves kubelet awareness

enp0s3 (commented):
@tiraboschi I also think that we may need a prioritization mechanism for migrations in these scenarios.

tiraboschi:
> Given the actual behavior of pod deletion, even if we have a finalizer on the pod its an API-controller level delay, I think that kubelet will initiate the pod deletion anyway by sending the signals to the container, and if eventually it will send SIGKILL then we have a problem.

The kubelet doesn’t care about finalizers but it's going to honor `terminationGracePeriodSeconds` and eventual `PreStop` hooks so we can somehow delay it up to a certain configurable (potentially also really long) value. But yes, after that the kubelet will send a SIGTERM first and then a SIGKILL.
So yes, we can potentially configure a long `terminationGracePeriodSeconds` to have more time but we aim to be 100% sure about preventing the kubelet from killing the pod. 

I'll add a paragraph to the open questions section.

enp0s3:
> > Given the actual behavior of pod deletion, even if we have a finalizer on the pod its an API-controller level delay, I think that kubelet will initiate the pod deletion anyway by sending the signals to the container, and if eventually it will send SIGKILL then we have a problem.
> 
> The kubelet doesn’t care about finalizers but it's going to honor `terminationGracePeriodSeconds` and eventual `PreStop` hooks so we can somehow delay it up to a certain configurable (potentially also really long) value. But yes, after that the kubelet will send a SIGTERM first and then a SIGKILL. So yes, we can potentially configure a long `terminationGracePeriodSeconds` to have more time but we aim to be 100% sure about preventing the kubelet from killing the pod.
> 
> I'll add a paragraph to the open questions section.

@tiraboschi I think its important to address possible solutions for this case. It is important to converge in time with live-migration before the pod will be force killed. I see that `preStop` is blocking which is good for this case.  We can discuss it in the design. Also the graceful termination period can be manually set by the user, so preStop would be more reliable in this case IMO.

veps/sig-compute/55-evictionstrategy-extension.md
```diff
@@ -0,0 +1,154 @@
...
+### Actual behavior
+#### Scheduler Priority Eviction
+
+When a pod cannot be scheduled due to constraints, the scheduler may trigger preemption.
+If the incoming pod has a high priority, the scheduler will preempt lower-priority pods by setting a condition `Type: v1.DisruptionTarget`, `Reason: v1.PodReasonPreemptionByScheduler`, then delete the pod directly (not using the eviction API).
+Currently, this causes a graceful shutdown of the VM, not a live migration—even if `evictionStrategy: LiveMigrate` is set.
+See: https://github.com/kubernetes/kubernetes/blob/3e5849972e619c58b3a8e8be2ca3ac35c4eb74da/pkg/scheduler/framework/preemption/preemption.go#L160-L202
+
+Also, the scheduler can bypass PDB protection when necessary.
+See: https://github.com/kubernetes/kubernetes/blob/3e5849972e619c58b3a8e8be2ca3ac35c4eb74da/pkg/scheduler/framework/preemption/preemption.go#L678-L679
```
vladikr, line 57: Can you please talk about how the virt-launcher will be kept alive during this time?
will the controller do with the `DisruptionTarget` condition? won't sigterm be already sent to the pod? 
Isn't there a race?

veps/sig-compute/55-evictionstrategy-extension.md
```diff
@@ -0,0 +1,154 @@
...
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [X] (R) Graduation criteria filled
+
+## Overview
+
+Under specific well-known circumstances (e.g., scheduler preemption, `NoExecute` taints), a `virt-launcher` pod can be terminated (read deleted).
+In these same cases, `PodDisruptionBudget` can also be bypassed. This can easily result in a VM being gracefully shut down, even if its `EvictionStrategy` is set to `LiveMigrate` or `LiveMigrateIfPossible`.
+
+This proposal aims to have our controllers identify well-known cases (initially scheduler preemption and `NoExecute` taints) and, if the `EvictionStrategy` requires `LiveMigrate` or `LiveMigrateIfPossible`, attempt a live migration.
```
vladikr, line 16: What happens if live migration fails?
What will be the fallback?

veps/sig-compute/55-evictionstrategy-extension.md
vladikr: This ties into another question of rate limiting. Can we block the deletion long enough for the VMIM to be picked up by the migration controller?

veps/sig-compute/55-evictionstrategy-extension.md
tiraboschi: Yes, this is the real question but in my opinion we already have this problem today.
Let's assume that a migration of a specific VM cannot complete for any possible reason.
Even when using the eviction API, today we do not have any fallback logic:
the VMI will remain marked for eviction and the eviction controller will try migrations forever.
On my opinion we already need a fallback logic now independently of this.

enp0s3:
@tiraboschi @vladikr I think that we also should suggest a mechanism that will monitor this process with some timeout and do a fallback to VM shutdown, because it may happen that a live-migration will stuck because of some scheduling constraints i.e. insufficient resources, node selector etc. In this case the target pod will be stuck in pending state.

vladikr (commented):
Thinking about this, we might have more success if we focus on prevention instead of reacting

vladikr:
@jean-edouard @xpivarc, what do you think?

veps/sig-compute/55-evictionstrategy-extension.md
tiraboschi: > I think the race is not well emphasised in the proposal. If the virt-controller is slow to react, the pod could start terminating before the VMIM is created and picked up by the migration controller.

It's even more problematic than this.
Let's say that a cluster admin will apply an untolerated `NoExecute` taint to a node: this is going to affect all the pods/VMs on that node at the same time but we also have a `parallelOutboundMigrationsPerNode` limit and so we cannot naively assume that all of then can complete in time.

> I wonder if we shouldn't pause these workload loads at the beginning of the migration - otherwise, aren't we just waiting for a disaster to happen?

Is pausing really enough if we know that sooner or later (exactly after `terminationGracePeriodSeconds` + 2 seconds) every pod is still going to receive a SIGKILL?
Only suspend to disk will allow us to resume it on another node (but it's completely another story, which disk...), otherwise I think that the current approach of attempting a clean shutdown will be nicer than pausing while waiting to be eventually killed.

enp0s3:
@tiraboschi Also how would we differ user initiated pod deletion from controller initiated one? should we differ them at all?

Issue #58, opened 2025-06-05
VEP NNNN: AMD SEV-SNP Enablement


ajcaldelas:
**Primary contact (assignee)**:

/assign @ajcaldelas

**Current Feature Stage**: Alpha

**Feature Gate**: WorkloadEncryptionSEV

**Responsible SIGs**: 

Primary SIG:
/sig compute

Additional SIGs (optional): N/A

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.6
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

[AMD SEV technology](https://github.com/kubevirt/kubevirt/issues/6991)

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
> I think this is safe to close since we have combined Intel TDX & AMD SEV-SNP

Thank you @alancaldelas

PR #59, opened 2025-06-09
VEP #58: AMD SEV-SNP Support Alpha feature

ajcaldelas:
### VEP Metadata

**Tracking issue**:  #58 
**SIG label**: /sig-compute 

### What this PR does

Adds the extension of the SEV feature to now support SEV-SNP.

### DIFF
--- a/veps/sig-compute/58-amd-sev-snp-enablement/vep.md
+++ b/veps/sig-compute/58-amd-sev-snp-enablement/vep.md
@@ -1,0 +1,174 @@
+# VEP #58: AMD SEV-SNP enablement for Confidential Compute
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+<!--
+Provide a brief overview of the topic)
+-->
+This enhancement is about adding support for the AMD Secure Encrypted Virtualization with Secure Nested Paging (SEV-SNP) in Kubevirt. SEV-SNP provides enhanced security for virtual machines by encrypting VM memory and protecting against memory integrity attacks, this builds upon the SEV/SEV-ES capabilities. SEV is currently supported in KubeVirt, and this VEP proposes to build on top of this existing feature to integrate SEV-SNP support.
+
+## Motivation
+
+<!--
+Why this enhancement is important
+-->
+KubeVirt supports SEV and SEV-ES for memory encryption, but lacks SEV-SNP enahnced memory integrity protection. Organizations are now asking for workloads to include the strongest confidential computing secruity guarantees.
+
+## Goals
+
+<!--
+The desired outcomeo
+-->
+- **Enable SEV-SNP VM Deployment**: Provide users the ability to deploy VMs that are utilizing AMD SEV-SNP technology. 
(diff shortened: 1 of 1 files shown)


veps/sig-compute/58-amd-sev-snp-enablement/vep.md
```diff
@@ -0,0 +1,161 @@
...
+List of repose this design impacts
+-->
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+<!--
+This should be brief and concise. We want just enough to get the point across
+-->
+The SEV-SNP feature extend the exisitn SEV features by adding a `secureNestedPaging` boolean flag to the current Launch Security API. When enabled the libvirt XML will contain `<launcSecurity type='sev-snp'>` using the default QEMU policy. the implementation will use existing SEV feature gate and extend the node labeling, by labeling nodes with `sev-snp.node.kubevirt.io`. The node labeller will detect SEV-SNP capabilities from libvirt domain capabilities then apply the label on the node, while the node selector renderer will be extended to include SEV-SNP scheudling. The feature will alo generate the domain XML to contain the required fields like `cbitpos`.
```
lyarwood, line 77: Can you provide an example of the generated libvirt domain XML below in API examples just so folks can understand how we end up configuring this?

veps/sig-compute/58-amd-sev-snp-enablement/vep.md
```diff
@@ -0,0 +1,161 @@
...
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: sev-snp-vm
+spec:
+  domain:
+    launchSecurity:
+      sev:
+        secureNestedPaging: true  # Enable SEV-SNP, uses QEMU default policy
```
lyarwood, line 95: Would a struct be better here and allow us to extend if required in the future?

ajcaldelas: Are you suggesting something like this
```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachineInstance
metadata:
  name: sev-snp-vm
spec:
  domain:
    launchSecurity:
      sev:
        snp:  # New struct for SEV-SNP specific configuration
          enabled: true
          # Future fields can be added here:
          # policy: "0x30000"
```
I think for the initial implementation we could go for this boolean instead to keep things simple but I think you're right this addresses future work like enabling specific policies.

veps/sig-compute/58-amd-sev-snp-enablement/vep.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+- Node labels added.
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  labels:
+    sev-snp.node.kubevirt.io: "true"  # New SEV-SNP label
```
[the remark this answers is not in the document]
ajcaldelas: While it does build on top of it I think explicitly listing what the feature is better suited and follows what the previous iterations have been using like sev-es and sev

veps/sig-compute/58-amd-sev-snp-enablement/vep.md
ajcaldelas: I think this would actually make more sense
```go
spec:
  domain:
    launchSecurity:
      amd:
        snp:
          enabled: true
          policy: "0x30000"
          authorKey: true
          vcek: false
```

this way we can just add to the launchSecurity structure for future work for Intel and others for launchsecurity.
Since this is translating to XML for LaunchSecurity I don' think we can have both types for launchSecurity with can be either `sev` or `sev-snp`.

Issue #64, opened 2025-06-25
VEP 64: vm Hibernation


Longchuanzheng:
**Primary contact (assignee)**:

/assign $ASSIGNEE

**Current Feature Stage**:

**Feature Gate**:New

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig api
/sig storage

**Enhancement link**:https://github.com/kubevirt/enhancements/pull/66

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

jean-edouard:
This is marked sig-compute but seems sig-storage related. Could you please add more details? Thank you

PR #66, opened 2025-06-25
VEP 64: vm Hibernation

Longchuanzheng:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/64
**SIG label**: /sig/compute

### What this PR does
This proposal introduces a VM hibernation mechanism for KubeVirt, enabling users to stop and start virtual machines by saving and restoring their running memory state.


veps/sig-compute/stop-vm-with-pmsuspend-to-disk/stop-vm-with-pmsuspend-to-disk.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Definition of Users
+
+<!--
+Who is this feature set intended for
+-->
+
+## User Stories
+
+A user exec `virtctl pmsuspendToDisk vm.name` commend, the vm stopped with memory state saved, exec `virtctl start vm.name` to start the stopped vm.
```
lyarwood, line 35: Why not `virtctl suspend $VM_NAME` with additional flags to suspend to disk etc?

Avoids leaking the fact that we are using libvirt underneath through our API/commands etc.

vladikr:
I wonder what guarantees we would provide around this functionality? 
How does the VM owner know that there is enough space left on the disk to save the memory state?

Also, have you considered a more gitops friendly approach?

My main concern here is that hibernation can potentially indefinitelly block system operations, drain for example. Since hibernation can take a very long time in some cases.
Perhaps we should think about a better  timeout handling?

Longchuanzheng:
Hi, @vladikr @mhenriks Thanks for the discussion.  Perhaps my question below is not very professional. When I mentioned this VEP, I only wanted to use the “dompmsuspend” interface. Is this interface not universal? From our previous discussion, it seems that we are discussing the use of the 'save' interface?

mhenriks:
> Hi, @vladikr @mhenriks Thanks for the discussion. Perhaps my question below is not very professional. When I mentioned this VEP, I only wanted to use the “dompmsuspend” interface. Is this interface not universal? From our previous discussion, it seems that we are discussing the use of the 'save' interface?

Good question! Yes, I was assuming "save" and not "dompmsuspend target disk". Nice thing about 'dompmsuspend' is that it doesn't require any additional storage. But it does require the guest agent. The 'save' function will work for any VM. I think I'd be open to supporting dompmsuspend with disk target. But we may want to do both eventually.

I think next step for this VEP is to design a declarative API (maybe just a new runstrategy) and map out the flow

veps/sig-compute/vm-Hibernation/vm-Hibernation.md
```diff
@@ -0,0 +1,333 @@
...
+
+3. Generate the corresponding `FileName`.
+
+4. Render `VirtualMachineHibernationStatuses` , FileName should related to the vmi, may FileName  hash(vmi.uuid):
+
+   ```yaml
+   HibernationStatuses:
+     Phase: Initial-->associating
+     Claim: PVCname
+     FileName: filename
```
iholder101: do we really need this?

veps/sig-compute/vm-Hibernation/vm-Hibernation.md
```diff
@@ -0,0 +1,333 @@
...
+
+---
+
+### 1. Hibernation
+
+#### Step 1: Trigger & Initial Check
+
+1. The user triggers hibernation by setting `spec.runStrategy` of the VM to `Hibernate`.
+
+2. If PVC is not specified, create PVC.
```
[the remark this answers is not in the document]
Longchuanzheng: PVC name will save in VirtualMachineResumeStatuses.Claim.

Longchuanzheng:
hi, @vladikr. I’ve developed some code recently https://github.com/kubevirt/kubevirt/pull/16775, which unfortunately conflicts with the runStrategy strategy discussed earlier https://github.com/kubevirt/enhancements/pull/66#discussion_r2699642131. I’d like to have a discussion to confirm whether I need to switch to the new strategy. Additionally, I’m happy to update the documentation to detail the current code implementation if you think it’s necessary. I’m determined to finish this feature, even though it’s a bit of a challenge for me. cc @mhenriks

mhenriks:
@Longchuanzheng are you interested in reopening this?

Longchuanzheng:
> @Longchuanzheng are you interested in reopening this?

@mhenriks HI, I’m happy to move forward with this. If anyone else in the community is already working on something similar, I’d appreciate a heads-up.
For now, I’d like to start directly from the save flow and skip dompmsuspend — the feature is already implemented, but it targets an older version of KubeVirt.
My next focus is to establish a well-agreed declarative API and workflow on save flow.
Communication in this PR has been a bit slow (which I’m okay with), so I wanted to ask: would it be more efficient if I just open a new PR with the working code changes directly?

PR #68, opened 2025-06-30
Enhancement: Add nad metric

avlitman:
### VEP Metadata

**SIG label**: 

### What this PR does

Add a metric that reports the Network Attachment Definitions and their properties.
The goal is to be able to identify what external network each VM is connected to.


EdDev:
@xpivarc , looks to be still a draft.

PR #79, opened 2025-07-30
VEP 80: Enable TDX and SEV-SNP

MatiasVara:
Add joint proposal document about the enablement of the confidential technologies named Intel TDX and AMD SEV-SNP.

### VEP Metadata

**Tracking issue**: #80 
**SIG label**: /sig computing

### What this PR does
This PR adds a proposal document for the Intel TDX and AMD SEV-SNP features. These features enable the deployment of confidential VMs in kubevirt.

### DIFF
--- a/veps/sig-compute/tdx-sev-snp-proposal.md
+++ b/veps/sig-compute/tdx-sev-snp-proposal.md
@@ -1,0 +1,243 @@
+# VEP #80: Enable TDX and SEV-SNP
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+This proposal deals with the enablement of the following technologies: Intel
+TDX & AMD Secure Encrypted Virtualization with Secure Nested Paging (SEV-SNP)
+in Kubevirt. This enablement includes the support for creating confidential VMs
+by using TDX & SEV-SNP by encrypting VM memory. For AMD, SEV is currently
+supported in KubeVirt, and this VEP proposes to build on top of this feature
+and integrates SEV-SNP support.
+
+## Motivation
+Organizations are now asking for workloads to include the strongest
+confidential computing security guarantees.
+
+For AMD, KubeVirt supports SEV and SEV-ES for memory encryption, but lacks
+SEV-SNP enhanced memory integrity protection.
+
+For Intel, Add support for TDX technology for in-use memory encryption of
+confidential guests.
+
+## Goals
+- Enable TDX VM Deployment: Provide users the ability to deploy VMs that are
+  utilizing Intel TDX technology.
+- Enable SEV-SNP VM Deployment: Provide users the ability to deploy VMs that
+  are utilizing AMD SEV-SNP technology.
+- Extend the existing SEV Feature: The SEV feature is currently in the alpha
+  phase. Extending it with SEV-SNP is the next logical step.
+- Extend Existing SEV Infrastructure: Use the existing SEV node labeling,
+  scheduling, and domain generation infrastructure.
+
+## Non Goals
+- TDX & SEV-SNP Live Migration: Live migrations are currently not supported by
+  either technologies. Goal: mark confidential-computing-enabled VMs as
+  non-migratable.
+- All TDX & SEV-SNP libvirt XML configurations those would be use cases situations.
+- Custom Attestation Services: Only focus on the enabling the capabilities
+  through the standard interfaces and not building a custom attestation
+  infrastructure.
+- SEV migration: No automatic upgrade path from SEV/SEV-ES to SEV-SNP, this
+  will be required to be done by explicit user configuration.
+- Confidential Devices Support.
+
+## Definition of Users
+- Cluster Admins: Responsible for enabling and managing SEV-SNP & TDX
+  capabilities in Kubernetes Clusters.
+- Developers: Deploys confidential workloads that require SEV-SNP & TDX.
+
+## User Stories
+- As a Cluster Admin, I want to enable TDX on cluster nodes so that Developers
+  can deploy Confidential Compute workloads with TDX enabled.
+- As a Cluster Admin, I need to ensure that my clusters have the correct
+  labeling that shows the correct TDX labels are showing up on my nodes.
+- As a Developer, I want to deploy VMs on a platform with TDX protection to
+  meet compliance requirements for memory integrity.
+- As a Cluster Admin, I want to enable SEV-SNP on cluster nodes so that
+  Developers can deploy Confidential Compute workloads with SEV-SNP enabled.
+- As a Cluster Admin, I need to ensure that my clusters have the correct
+  labeling that shows the correct SEV and SEV-SNP labels are showing up on my
+  nodes.
+- As a Developer, I want to deploy VMs on a platform with SEV-SNP protection to
+  meet compliance requirements for memory integrity.
+- As a Developer, I want to deploy VMs without needing to understand the
+  low-level confidential computing technologies.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Common Features
+Both designs rely on labelling those nodes capable of running CVMs
+(Confidential Virtual Machines). In addition to labeling, the node registers
+its available "key ID" resources, which determines how many CVMs can be created
+in the node. The scheduler consumes one key ID for each CVM. The number of key
+IDs is limited per node.  This is applicable to both TDX and SNP and the
+available resources are available under `/sys/fs/cgroup/misc.capacity`.
+
+### TDX Design
+A new TDX Feature Gate should be added first. Extend the node labeller to
+detect and label nodes capable of running TDX VMs, a corresponding node
+selector can be added to VM pods so they are scheduled correctly. TDX
+parameters need to be specified via the `launchSecurity` element of the Libvirt
+domain xml:
+```xml
+   <domain>
+     ...
+     <launchSecurity type='tdx'>
+       <policy>0x10000001</policy>
+       <mrConfigId>xxx</mrConfigId>
+       <mrOwner>xxx</mrOwner>
+       <mrOwnerConfig>xxx</mrOwnerConfig>
+       <quoteGenerationSocket path="/var/run/tdx-qgs/qgs.socket"/>
+     </launchSecurity>
+     ...
+   </domain>
+```
+All parameters are optional. Please refer to
+https://gitlab.com/libvirt/libvirt/-/blob/master/docs/formatdomain.rst?ref_type=heads#launch-security
+for the explanation of the parameters. From the implementation perspective,
+only extend the VMI spec to enable the basics in the proposal currently:
+
+```xml
+   spec:
+     domain:
+       launchSecurity:
+         tdx: {}
+```
+Let's hold off on adding more options to the VMI spec for now. We need further
+discussion on the use cases and how they would apply to VM creation.
+
+### AMD Design
+
+AMD SEV-SNP maps to a new type in the LibVirt XML of type “sev-snp” with its
+own attributes and elements that are similar to the type “sev”.
+
+```xml
+   <domain>
+     ...
+     <launchSecurity type='sev-snp'>
+       <cbitpos>47</cbitpos>
+       <policy>0x00030000</policy>
+       <reducedPhysBits>1</reducedPhysBits>
+	<guestVisibileWorkarounds>...</guestVisibleWorkaround>
+	<idBlock>...</idBlock>
+	<idAuth>...</idAuth>
+	<hostData>...</hostData>
+     </launchSecurity>
+     ...
+   </domain>
+```
+
+A new structure should be created for adding SEV-SNP on the `LaunchSecurity`
+field.
+```go
+type SEVSNP struct {
+  // 64-bit SEV-SNP Guest Policy
+  policy: "0x3000" # Default Policy
+}
+```
+Similar to TDX, all elements for type=’sev-snp’, this will be a basic enablement
+until further use cases arise to provide additional settings to configure a 
+Confidential VM. The default policy option will have a default value that QEMU specifies.
+The node labeller will detect SEV-SNP capabilities from the LibVirt domain capabilities then
+apply the label to the node, while the node selector renderer will be extended to include
+SEV-SNP scheduling.
+
+### Security Considerations
+- The infrastructure provider and the VMI author are untrusted. For example,
+  the infrastructure provider is always capable of injecting random cloud-init
+  stuff behind our back. So if an instance of a secret application runs inside
+  a CVM, via cloud-init, an infrastructure provider or a hacker could simply
+  execute random commands inside this VM.
+
+## API Examples
+The bellow yaml snippets provide examples of how to request TDX feature in the
+VMI spec.
+
+### API Examples for Intel TDX
+
+- TDX VM with default options
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  ...
+spec:
+  domain:
+    launchSecurity:
+      tdx: {}
+  ...
+```
+
+### API Examples for AMD
+
+- Basic SEV-SNP with default options
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: sev-snp-vm
+spec:
+  domain:
+    launchSecurity:
+      snp: {}
+  ...
+```
+
+## Alternatives
+
+## Scalability
+
+## Update/Rollback Compatibility
+- All new fields are optional and disabled by default
+- This should not impact existing VMs
+- AMD SEV Compatibility:
+  - The AMD SEV-SNP feature sits behind the existing SEV feature gate without
+    breaking changes
+  - Node labels are added, no labels are removed.
+
+## Functional Testing Approach
+- Unit testing to detect TDX & SEV-SNP from the libvirt capabilities.
+- Since TDX & SEV-SNP do not have support for nested virtualization this will
+  require bare metal hardware to conduct e2e testing.
+
+## Implementation Phases
+### Intel TDX Phases:
+The initial phase of implementation will focus on integrating basic
+functionality in kubevirt thus allowing the creation and deployment of
+confidential VMs using the TDX technology. The initial phase also includes the
+e2e tests. The subsequent phase will involve adding any missing use cases
+aligned with other confidential VM technologies.
+
+### AMD SEV-SNP Phases:
+The initial phase of implementation will focus on integrating basic
+functionality. The initial phase also includes the e2e tests. The subsequent
+phase will involve adding any missing use cases and necessary checks to prevent
+the creation of improperly configured VMs (e.g., preventing users from setting
+KernelHashes without configuring kernel booting).
+
+## Feature lifecycle Phases
+
+### Alpha
+The feature will be implemented in Alpha. We do not know if it will be possible
+to have e2e tests in Alpha due to lack of TDX hardware. We expect the feature
+to be merged without the e2e tests.
+
+### Beta
+The requirements for graduating to Beta include having hardware available, CI
+lanes created, and e2e tests that can run.
+
+### GA
+Remove feature gate


veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
+# VEP #77: Enable TDX and SEV-SNP
```
mythi, line 1:
```suggestion
# VEP #80: Enable TDX and SEV-SNP
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+  will be required to be done by explicit user configuration.
+- Confidential Devices Support.
+
+## Definition of Users
+- Cluster Admins: Responsible for enabling and managing SEV-SNP & TDX
+  capabilities in Kubernetes Clusters.
+- Developers: Deploys confidential workloads that require SEV-SNP & TDX.
+
+## User Stories
+- As a Cluster Admin, I want to enable TDX on cluster nodes so that tenants can
```
mythi, line 55:
```suggestion
- As a Cluster Admin, I want to enable TDX on cluster nodes so that Developers can
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### TDX Design
+A new TDX Feature Gate should be added first. Extend the node labeller to
+detect and label nodes capable of running TDX VMs, a corresponding node
```
mythi, line 81: In addition to labeling, we need to register the node "key ID" resources on how many CVMs per node can be created to the scheduler and automatically consume one resource per CVM. This is applicable to both TDX and SNP and the available resources are available under `/sys/fs/cgroup/misc.capacity`.

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+```
+Similar to TDX, all elements for type=’sev-snp’, users should be able to
+provide additional settings to configure their Confidential VM. Some options
+should have a default value such as the policy if the user does not specify the
+VM should be using the QEMU default policy.  The node labller will detect
+SEV-SNP capabilities from the LibVirt domain capabilities then apply the label
+to the node, while the node selector renderer will be extended to include
+SEV-SNP scheduling.
+
+## API Examples
```
mythi, line 181: The design proposal suggest that the API would be TEE specific. Would there be any desire to have more unified API so that the platform specifics could be hidden from the users? Just a very rough example of what I'm thinking:

```
 spec:
     domain:
       launchSecurity:
         tee: tdx
         debug: true
         hostConfig: <some host provided data that is measured into mrconfigid/hostdata>
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+```xml
+   spec:
+     domain:
+       launchSecurity:
+         tdx:
+           policy: 0x10000000
+           mrConfigId: xxx
+           mrOwner: xxx
+           mrOwnerConfig: xxx
+           qgs: true
```
mythi, line 113: I don't see much value for this `qgs` flag. what would be the case when this is set to `false`?

mythi (commented):
mostly moving my comments/questions from the original PR.

veps/sig-compute/tdx-sev-snp-proposal.md
MatiasVara: Is not tenants the right wording?

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: Qemu runs OK even when the socket is not available. The socket is only needed when the guest VM issues GetQuote request Qemu needs to service.

MatiasVara: maybe @zhencliu can comment on this.

vladikr:
@rmohr would you like to take a look as well?

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: +1 to @mythi that's not sumething a user has to care about. It may make sense to allow configuring in the main KubeVirt CR a QGS socket or not. In both cases you get quotes the same way inside the guest. Only difference, in one way you can get a quote inside the VM in the other case you can't get a quote. Also admins have to register the host so that the quotes are verifyable. Really not something which should be configured on the VM.

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+for the explanation of the parameters. From the implementation perspective,
+extend the VMI spec to map to the parameters:
+
+```xml
+   spec:
+     domain:
+       launchSecurity:
+         tdx:
+           policy: 0x10000000
+           mrConfigId: xxx
```
rmohr, line 110: I have been reading a little bit about these mr fields here: https://lore.kernel.org/qemu-devel/31d6dbc1-f453-4cef-ab08-4813f4e0ff92@intel.com/. 

We may want to hold off adding them now. They may be worth extra discussiosn on use-case and also if some of this data is more suited for secrets than direct VMI spec.

AFAIK cloud-providers don't support this atm. I would not expect high demeand for that, but maybe you have use-cases, but then would be great to see them in the doc.

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+```yaml
+   spec:
+     domain:
+       launchSecurity:
+         tdx:
+           policy: 0x10000000
+           mrConfigId: xxx
+```
+
+- TDX VM with qgs settings, note libvirt will use `/var/run/tdx-qgs/qgs.socket`
```
rmohr, line 212: That seems outdated when qgs path was planned on the VMI api. KubeVirt should wire this transparently and either run with just TDX support or TDX+quote support.

rmohr (commented):
I would double-check if we really want to pass through all the low-level condiguration options without discussing use-cases, sensitivy and impact of that data. LibVirt and QEMU operate on a different level from a management perspective than kubevirt. Consider trimming down to the must-have.

On the non-must fields I would suggest go through them by:

 1. analyzing their meaning again
 2. coming up with a user story
 3. see if this is needed now (I am not aware of any commercial provider at the moment who would allow more than enable/disable atm)
 4. If there are use-cases, lets' see if they are better fitted on the whole kubevirt install instead of the VM
  5. If on the VM, if they should be more on the VM or on e.g. secrets.

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: A minimal viable support which would offer TDX with or without QGS (where QGS would be configured KubeVirt wide) would imo be:

```
   spec:
     domain:
       launchSecurity:
         tdx: {}
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Definition of Users
+- Cluster Admins: Responsible for enabling and managing SEV-SNP & TDX
+  capabilities in Kubernetes Clusters.
+- Developers: Deploys confidential workloads that require SEV-SNP & TDX.
+
+## User Stories
+- As a Cluster Admin, I want to enable TDX on cluster nodes so that Developers
+  can deploy Confidential Compute workloads with TDX enabled.
+- As a Cluster Admin, I need to ensure that my clusters have the correct
```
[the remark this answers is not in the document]
rmohr: in order to get verifyable quote, one needs to register the machine with intel and the signed cert needs to be fetched form intel to be part of the quote, right? If a machine is registered or not is not easy to see on a machine. So if you are getting a verifyable quote or just a quote is not easy to say at scheduling time with the current proposal.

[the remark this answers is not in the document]
rmohr: What I try to express: It would be great to explore if we can somehow ensure scheduling to registered machines only, for the sake of providing consistent quote experience inside the VMs. But not sure we need this discussion at that point.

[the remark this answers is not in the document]
mythi: > ensure scheduling to registered machines only

it's a valid point, I can think about this on what'd be the ways to do it.

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: > any feedback on the suggestion unify the api between the TEEs? with the use-case (i.e., what KubeVirt can offer) based approach, one use case would be enabled the same way independent of the CVM arch.

I think it is tricky, due to a few factors:

 1. on the same arch (x86_64) there exist two implementations already (AMD, Intel)
 2. inside the guest it makes a difference if I land on Intel or AMD (quote lifecycle, ...)
 3. as a consequence, the VM creator has to know about image creation for which architecture to take measurements and potentially prepare in-guest flows to use the right tools to get and verify quotes
 
I guess that can be abstracted away by users and by smart usage of guest tooling (like probing if it is intel or amd and doing different stratup flows, by taking measurements for both, ...).

mythi: FWIW, For Linux guests there's a unified ABI (TSM reports) for generating quotes. It works the same way for TDX/SNP/CCA.

veps/sig-compute/tdx-sev-snp-proposal.md
alancaldelas: I don't think this is necessary, all we are doing is mapping to libvirt XML so wouldn't it make sense to make the type the top level API? That's why I initially created a separate entry for SEV-SNP because the type in the XML is either `sev` or `sev-snp`. So my question is how is it mapped for Intel?

zhencliu: It might be hard to use a unified API for all kinds of tees, e.g.  IBM SE even doesn't have a similar option as hostData, even for SNP and TDX, the libvirt launchSecurity elments are quite different. Please refer to https://gitlab.com/libvirt/libvirt/-/blob/master/docs/formatdomain.rst?ref_type=heads#launch-security

veps/sig-compute/tdx-sev-snp-proposal.md
zhencliu: Hi @mythi and @rmohr, just to make sure I'm following correctly: if QGS is configured in the main KubeVirt CR, would this be a global setting that enables it for all TDX VMs?

If the answer is yes, then my follow-up question: Will there be a way for a user to launch a TDX VM without QGS for use cases where it's not required(e.g. people don't care about attestation, just start a TDX VM)?"

veps/sig-compute/tdx-sev-snp-proposal.md
zhencliu: Hi @rmohr, in a previous trustee POC I worked on using both SNP and TDX, we used the SNP hostData and TDX mrConfigId fields to distinguish between different VMs in our resource policy, e.g.
```
    allow if {
      path[2] == "kbsres1"
      input["submods"]["cpu"]["ear.status"] != "contraindicated"
      input["submods"]["cpu"]["ear.veraison.annotated-evidence"]["init_data"] == init_data_vm1
    }
```
In which "init_data" equals to hostData for SNP or mrConfigId decoded by base64 for TDX.

Azure provides another way to inject init data to VM by Cloud-Init(https://learn.microsoft.com/en-us/azure/virtual-machines/custom-data).  

Of course, the mrOwner and mrOwnerConfig are also listed in the evidence claims, but currently I didn't know any use case for them:

```
      "quote": {
        "body": {
          "mr_config_id": "fe2756ef7ed92c6fafa9fd8a8f884c2394fd4b7c26c0638490a1a3aac4b1ffa0338ac13a6ef0005bd8802d1f03679cdd",
          "mr_owner": "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
          "mr_owner_config": "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
So IMO we need to consider mrConfigId and hostData at least.

mythi: `initdata` is a CoCo construct. What would be the KubeVirt equivalent use-case?

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: @zhencliu it's possible to start a TD VM with `quote-generation-socket` settings added even if the guest has no plans to use it.

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: If exposing libvirt XML API to the users is how KubeVirt works then it's clear. My suggestion was to see if we can make it use case driven and hide the TEE libvirt XML differences from users.

veps/sig-compute/tdx-sev-snp-proposal.md
MatiasVara: I am not sure if this is pertinent here but we have also to bear in mind that socket device plugin has a number of max instances, i.e., `c.maxDevices` in my poc, this limits the number of VMs that can access the qgs unix socket.

zhencliu: Thanks. I will look into this, currently it looks the file /sys/fs/cgroup/misc.capacity is not exposed yet.

veps/sig-compute/tdx-sev-snp-proposal.md
zhencliu: I will update this part once we decide whether QGS should be configured by KubeVirt

MatiasVara: I will remove this from the document.

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: > I am not sure if this is pertinent here but we have also to bear in mind that socket device plugin has a number of max instances, i.e., c.maxDevices in my poc, this limits the number of VMs that can access the qgs unix socket.

A socket is not a working target for a device plugin necessarily. It works on cri-o but containerd complains for instance. In general if you still end up with a socket device plugin, you can just give it a high number like 1000 (just more than a reasonable amount of VMs is).

MatiasVara: > A socket is not a working target for a device plugin necessarily. It works on cri-o but containerd complains for instance. In general if you still end up with a socket device plugin, you can just give it a high number like 1000 (just more than a reasonable amount of VMs is).

Are you suggesting to use a different method than a device plugin for the QGS socket?

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,341 @@
...
+       <mrOwnerConfig>xxx</mrOwnerConfig>
+       <quoteGenerationSocket path="/var/run/tdx-qgs/qgs.socket"/>
+     </launchSecurity>
+     ...
+   </domain>
+```
+All parameters are optional. Note the Libvirt support is not ready in upstream,
+refer to the
+https://github.com/intel/libvirt-tdx/commit/962325568e67014c2ff2e41a673374fafd44e385
+for the explanation of the parameters. From the implementation perspective,
```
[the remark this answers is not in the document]
zhencliu: hi @MatiasVara , would you please help to update it? libvirt supports TDX now. Thanks.

veps/sig-compute/tdx-sev-snp-proposal.md
zhencliu: Thanks, @mythi and @rmohr. Based on the discussion, I will remove the QGS flag from the API. I agree with the plan to proceed with the basic enablement for now and hold off on adding the mr configs

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: > I see, I think we won't rely on vsock. What do you mean with out-of-band?

virt-handle can bind-mount the socket into the container, but that would be kind  of the last resort.

Two ways to explore would come to my mind:
 1. A CSI plugin. We would basically mount a QGS volume where the CSI plugin can ensure the socket is mounted in to the volume >
  2. DRA maybe. Since it would be more flexible.
  
  There may be other ways as well.

mythi: > I'm looking for guidance on where to register this as a new resource.

In CoCo we register this as an extended resource using NFD and one key resource is consumed as part of `RuntimeClass`es podoverhead. I think we should also tie this to the same resource somehow so that both CoCo TEE pods and KubeVirt TEE VMs can run simultaneously.

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: @MatiasVara we should keep the SNP aligned with TDX: if no use-cases are identified for the XML APIs, we'd leave them out for now.

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: Sure, it's fairly simple: NFD let's you [register](https://kubernetes-sigs.github.io/node-feature-discovery/v0.17/usage/customization-guide.html?highlight=extended%20resource#extendedresources) kubernetes [extended resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/). It does not have to be NFD to do that but that's what we use since it knows about node TEE capabilities.

Each pod consumes one key resource. Instead of forcing users to specify the resources (and/or to avoid doing it in a webhook), we just piggyback Pod overhead spec that let's you add any resource type as the pod consumed node resource:

```
apiVersion: node.k8s.io/v1
kind: RuntimeClass
metadata:
  name: kata-cc
handler: kata-qemu-tdx
overhead:
  podFixed:
    memory: "350Mi"
    cpu: "250m"
    tdx.intel.com/keys: 1
scheduling:
  nodeSelector:
     intel.feature.node.kubernetes.io/tdx: "true"
```

It does not have to be done like this for KubeVirt if there already is a place where the resources consumed by the VM are added. For SNP, we register `sev-snp.amd.com/esids` and TDX `tdx.intel.com/key`.

veps/sig-compute/tdx-sev-snp-proposal.md
MatiasVara: > > There is potentially some cloud-init injection risk of unwanted things
> 
> I think this VEP could say a few words about (planned) mitigation options.

do you have something in mind?

veps/sig-compute/tdx-sev-snp-proposal.md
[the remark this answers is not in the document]
MatiasVara: > TDX also has a concept of registered and unregistered hosts, where quotes can be verified or not. I wonder if we want to express that with some scheduling constraint as well. It is not so easy to figure that out though. One possibility would be letting virt-handler ask PCCS (the cluster cache) for the host cert.
> 
> However that can probably de discussed in a follow-up proposal.

should I add something about this to a future section?

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: > Sorry, I do not understand something, if I only care about virt-handler, why can't I use a device plugin to end up mounting the unix socket in the virt-launcher?

Mostly because containerd will tell you: Hey this thing you try to mount is not a device node.

rmohr: > Sure, it's fairly simple: NFD let's you [register](https://kubernetes-sigs.github.io/node-feature-discovery/v0.17/usage/customization-guide.html?highlight=extended%20resource#extendedresources) kubernetes [extended resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/). It does not have to be NFD to do that but that's what we use since it knows about node TEE capabilities.

Nice. That looks like a nice solution for capacity. I wonder if we should still do that in virt-handler (with a kubevirt.io/ resource) and follow the example of NFD. In general self-sufficiency to avoid dependencies to other projects for small overlaps was quite useful for kubevirt, since other component remain more optional, making it easier to deploy in various envs.

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: Regarding `type` fields: API conventions in k8s are more leaning towards config subsections. Where fields with the same functions have the same name in their subsection. Otherwise you end up with a lot of "if that type then that field" which has to be propagated to documentation, validation, ...

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: > Mostly because containerd will tell you: Hey this thing you try to mount is not a device node.

Sounds like something is wrong with the KubeVirt implementation? Device plugin API allows you to separate `Mounts[]` and `Devices[]`. This would be about `Mounts[]`.

One more alternative to mounts could be a custom NRI plugin.

veps/sig-compute/tdx-sev-snp-proposal.md
zhencliu: hi @mythi 

> As I see it, the trust model here is such that whatever is specified in a `VirtualMachineInstance` is untrusted from the VM perspective and the (confidential) guest must somehow be aware of this. For this initial VEP (and implementation), I'd at least add a note about the trust model assumptions and the fact that some of the `VirtualMachineInstance` capabilities are insure by default and something the guest VM design needs to take into account.
> 
> `cloud-init` injection is only one example but even things like `emptyDisk` be encrypted if confidential data is written to it by the VM.

Could we add the note in the attestation proposal or other proposals which introduce new VMI spec options in future? Because currently we don't introduce any option mapping to the libvirt launchsecurity option.

mythi: @MatiasVara I should be able to help with this part. I'll wait until we get the `launchSecurity` API story agreed.

veps/sig-compute/tdx-sev-snp-proposal.md
rmohr: > Sounds like something is wrong with the KubeVirt implementation? Device plugin API allows you to separate Mounts[] and Devices[]. This would be about Mounts[].

That's probably it. Then disregard my comments regarding that before. I guess we could use this as well for tdx capacity then.

veps/sig-compute/tdx-sev-snp-proposal.md
alancaldelas: > As I see it, the trust model here is such that whatever is specified in a `VirtualMachineInstance` is untrusted from the VM perspective and the (confidential) guest must somehow be aware of this. For this initial VEP (and implementation), I'd at least add a note about the trust model assumptions and the fact that some of the `VirtualMachineInstance` capabilities are insure by default and something the guest VM design needs to take into account.
> 
> `cloud-init` injection is only one example but even things like `emptyDisk` be encrypted if confidential data is written to it by the VM.

I'm still not clear here, why would the VMI but untrusted from the VM perspective? Isn't that the source of truth?

mythi: > Let me try to rephrase: The VM/VMI spec is the source of truth, but one point of CC VMs is that we don't have to trust the infrastructure provider.

This is the main message the initial VEP (this PR) should highlight (I can draft something here). We can probably get into the specifics and how to secure various flows in later stages.

veps/sig-compute/tdx-sev-snp-proposal.md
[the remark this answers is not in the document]
MatiasVara: > However that can probably de discussed in a follow-up proposal.
Based on this comment, I will resolve the conversation and address it in a follow-up proposal.

veps/sig-compute/tdx-sev-snp-proposal.md
MatiasVara: I mark it as done since I addressed the first comment. Feel free to reopen otherwise.

alancaldelas:
> I was suggesting something similar over on the SEV-SNP PR [kubevirt/kubevirt#14845](https://github.com/kubevirt/kubevirt/pull/14845)
> 
> Can we design the API around SEV (possibly also TDX) like this?
> 
> ```
> Basic SEV:
> 
> launchSecurity:
>   sev:
>     policy:
>       encryptedState: false # This is the default, so we could omit it
>       snp: null             # This is the default, so we could omit it
> 
> SEV-ES:
> 
> launchSecurity:
>   sev:
>     policy:
>       encryptedState: true
>       snp: null             # This is the default, so we could omit it
> 
> SEV-SNP:
> 
> launchSecurity:
>   sev:
>     policy:
>       encryptedState: false # This is the default, so we could omit it
>       snp:
>         authorKey: true
>         vcek: false
>         [...]
> ```
> 
> Can we derive the policy bits internally from that without exposing them on KubeVirt's API?
> 
> For me all these things logically belong together, I can have SEV and possible also enabled ES or SNP below that. Same for TDX, we could add another field on the same level as SNP to the policy struct. Now we have all possible features neatly grouped and also have an easier time understanding the configuration from a human perspective and from validating it in an admitter?
> 
> I would say that grouping these configurations is the higher management perspective that KubeVirt works on.
> 
> > Consider trimming down to the must-have.
> 
> +1 Make the API easier to understand for the end user, we are not libvirt or even QEMU.

From the comments in the VEP, our focus will be on basic enablement for the initial release. Additional configuration options can be addressed in the next versions.

Regarding SEV-SNP, IMO using sev.policy.snp.policy or sev.policy.policy is not ideal. A clearer approach would be to separate these parameters into sev.policy, sev.encryptedState, and snp.policy. This structure aligns more closely with the Intel TDX proposal and provides a cleaner separation of concerns.

Policy definitions should be determined by the TEE type. For example, if TDX requires fewer configuration options, it should not need to extend the struct. If SEV-SNP requires additional parameters, we should be able to expand the struct to address those gaps. While I understand that from a KubeVirt API perspective the aim should be to abstract low-level Libvirt XML settings, this is a case where exposing such parameters is justified. If there are use cases where a user needs to provide a policy directly, the API should enable them to do so by enabling the user providing the needed info.

rmohr:
> From the comments in the VEP, our focus will be on basic enablement for the initial release. Additional configuration options can be addressed in the next versions.

+1

> If there are use cases where a user needs to provide a policy directly, the API should enable them to do so by enabling the user providing the needed info.

Totally reasonable.

veps/sig-compute/tdx-sev-snp-proposal.md
MatiasVara: @alancaldelas would not the sev design section look like the tdx section? I mean very simple without any example before we resolve it?

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+
+For AMD, KubeVirt supports SEV and SEV-ES for memory encryption, but lacks
+SEV-SNP enhanced memory integrity protection.
+
+For Intel, Add support for TDX technology for in-use memory encryption of
+confidential guests.
+
+## Goals
+- Enable TDX VM Deployment: Provide users the ability to deploy VMs that are
+  utilizing INTEL TDX technology.
```
jean-edouard, line 31:
```suggestion
  utilizing Intel TDX technology.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+- Enable SEV-SNP VM Deployment: Provide users the ability to deploy VMs that
+  are utilizing AMD SEV-SNP technology.
+- Extend the existing SEV Feature: Currently the SEV sits behind the
+  WorkloadEncryptionSEV adding SEV-SNP to feature is the next logical step.
```
jean-edouard, line 35:
```suggestion
- Extend the existing SEV Feature: The SEV feature is currently in the alpha phase.
  Extending it with SEV-SNP is the next logical step.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+- Extend Existing SEV Infrastructure: Use the existing SEV node labeling,
+  scheduling, and domain generation infrastructure.
+
+## Non Goals
+- TDX & SEV-SNP Live Migration: Live migrations are currently not supported by
+  either technologies.
```
jean-edouard, line 41:
```suggestion
  either technologies. Goal: mark confidential-computing-enabled VMs as non-migratable.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+
+## User Stories
+- As a Cluster Admin, I want to enable TDX on cluster nodes so that Developers
+  can deploy Confidential Compute workloads with TDX enabled.
+- As a Cluster Admin, I need to ensure that my clusters have the correct
+  labeling that shows the correct TDX labels are showing up on my nodes.
+- As a Developer, I want to deploy VMs on a platform with TDX protection to
+  meet compliance requirements for memory integrity.
+- As a Developer, I want to deploy VMs without needing to understand the low
+  level INTEL technology.
```
jean-edouard, line 63:
```suggestion
  level Intel technology.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+- As a Cluster Admin, I want to enable SEV-SNP on cluster nodes so that
+  Developers can deploy Confidential Compute workloads with SEV-SNP enabled.
+- As a Cluster Admin, I need to ensure that my clusters have the correct
+  labeling that shows the correct SEV and SEV-SNP labels are showing up on my
+  nodes.
+- As a Developer, I want to deploy VMs on a platform with SEV-SNP protection to
+  meet compliance requirements for memory integrity.
+- As a Developer, I want to deploy VMs without needing to understand the low
+  level AMD technology.
```
jean-edouard, line 72:
```suggestion
- As a Developer, I want to deploy VMs without needing to understand the low-level
  confidential computing technologies.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Common Features
+Both designs rely on labelling those nodes capable to running CVMs. In addition
```
jean-edouard, line 81: Please define CVM

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+to labeling, the node "key ID" resources registers how many CVMs per node can
```
jean-edouard, line 82: What are `"key ID" resources registers`?

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+      snp: {}
+  ...
+```
+
+## Alternatives
+
+## Scalability
+
+## Update/Rollback Compatibility
+- All new fields are fields should be optional
```
jean-edouard, line 204:
```suggestion
- All new fields are optional and disabled by default
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+- This should not impact existing VMs
+- AMD SEV Compatibility:
+  - The AMD SEV-SNP feature sits behind the existing SEV feature gate without
+    breaking changes
+  - Node labels are added, no labels are removed.
+
+## Functional Testing Approach
+- Unit testing to detect TDX & SEV-SNP from the libvirt capabilities.
+- Since TDX & SEV-SNP does not have support for nested virtualization this will
+  require bare metal hardware to conduct e2e testing.
```
jean-edouard, line 214: Do we have plans to acquire such hardware? Will it be available before the end of the implementation phase? Also, integrating the hardware to CI and creating test lanes should probably be an implementation phase.
```suggestion
- Since TDX & SEV-SNP do not have support for nested virtualization this will
  require bare metal hardware to conduct e2e testing.
```

veps/sig-compute/tdx-sev-snp-proposal.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Implementation Phases
```
jean-edouard, line 216: Please talk about functest here.

MatiasVara: @jean-edouard what are functest? Do you have an example? Thanks.

jean-edouard: Functional tests, e2e. Similar comment as above basically, sorry for the confusion (the command to run them is `make functest`)

veps/sig-compute/tdx-sev-snp-proposal.md
jean-edouard: Either way, it's kind of a goal about a non-goal :) Just though it was worth mentioning somewhere.

veps/sig-compute/tdx-sev-snp-proposal.md
jean-edouard: Ah yes, missed that, I think so.

jean-edouard:
Please consider consolidating commits. `Address review comments.` commits don't look great in the repo's git history... Fixing typos should also be done in the commit that introduced the typo. Thanks!

MatiasVara:
> Please consider consolidating commits. `Address review comments.` commits don't look great in the repo's git history... Fixing typos should also be done in the commit that introduced the typo. Thanks!

Right, I was thinking to squash everything once we are done with the review.

veps/sig-compute/tdx-sev-snp-proposal.md
jean-edouard: This is just to make sure the implementation prevents VMs from migrating, instead of letting them try and fail.

jean-edouard:
> > Please consider consolidating commits. `Address review comments.` commits don't look great in the repo's git history... Fixing typos should also be done in the commit that introduced the typo. Thanks!
> 
> Right, I was thinking to squash everything once we are done with the review.

Works for me, but that requires a hold.
/hold

MatiasVara:
> > > Please consider consolidating commits. `Address review comments.` commits don't look great in the repo's git history... Fixing typos should also be done in the commit that introduced the typo. Thanks!
> > 
> > 
> > Right, I was thinking to squash everything once we are done with the review.
> 
> Works for me, but that requires a hold. /hold

@jean-edouard I think now the PR is in good shape. Let me know otherwise.

veps/sig-compute/tdx-sev-snp-proposal.md
mythi: > Do we have plans to acquire such hardware? 

I'm working on this for TDX.

jean-edouard:
> @vladikr implementation phases have been populated since you approved. PTAL before I lgtm. Thanks

The latest changes were verbally approved by @vladikr
/lgtm

PR #81, opened 2025-08-11
VEP 143- Monitoring Stack Separation via Hybrid Architecture

sradco:
### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/143
**SIG label**: sig-observability

### What this PR does

This VEP proposes separating KubeVirt's monitoring and metrics collection from core components through a hybrid architecture that combines efficient informer-based collection with selective sidecar containers.

### DIFF
--- a/veps/sig-observability/monitoring-stack-separation.md
+++ b/veps/sig-observability/monitoring-stack-separation.md
@@ -1,0 +1,951 @@
+# VEP-81: Monitoring Stack Separation via Hybrid Architecture
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This VEP proposes separating KubeVirt's monitoring and metrics collection from
+core components through a **hybrid architecture** that combines efficient
+informer-based collection with selective sidecar containers.
+
+This approach uses Kubernetes API informers for API-derived metrics (virt-controller,
+virt-api, virt-operator) while only deploying sidecars where direct system
+access is required (virt-handler for libvirt statistics).
+
+Additionally, this proposal includes migrating Prometheus recording rules and alerts to a
+unified external repository for independent lifecycle management.
+
+This enables independent monitoring stack management, reduces coupling between monitoring and core
+functionality, and provides better observability modularity with minimal resource overhead.
+
+## Motivation
+
+### Current State
+
+KubeVirt components currently expose Prometheus metrics directly through
+their HTTP endpoints:
+- `virt-handler`: Exposes 60+ VMI runtime metrics via local libvirt domain
+  statistics
+- `virt-controller`: Exposes VM lifecycle and cluster-wide metrics
+- `virt-api`: Exposes API request and connection metrics
+- `virt-operator`: Exposes operator health and configuration metrics
+
+### Problems with Current Approach
+
+1. **Resource Overhead**: Metrics collection adds CPU/memory overhead to critical components
+2. **Coupling**: Metrics collection logic is tightly coupled with core component functionality
+3. **Limited Flexibility**: Difficult to customize metrics export formats, destinations, or processing
+4. **Scalability**: High-frequency metrics collection can impact component performance
+5. **Maintenance**: Metrics changes require core component updates and releases
+6. **Limited Backports**: New metrics, recording rules and alerts can't be backported to older versions
+8. **Limited Monitoring Configuration Options**: Recording rules and alerts are only available through KubeVirt releases,
+   limiting independent monitoring stack evolution
+
+### Goals
+
+- **Decouple** metrics collection from core KubeVirt component logic
+- **Migrate Out** metrics, recording rules, and alerts from core components to external repositories for independent lifecycle management
+- **Maximize Efficiency** by providing optimized collection alternatives through hybrid architecture
+- **Improve Performance** by minimizing resource overhead through architectural optimization
+- **Enable Flexibility** in metrics export formats (Prometheus, OpenTelemetry, custom)
+- **Support Independent Lifecycle Management** for monitoring stack updates separate from KubeVirt releases
+- **Maintain Backward Compatibility** by keeping embedded monitoring in older versions while migrating to external approach in new versions
+- **Enable Flexible Backports** of new monitoring features to older KubeVirt versions through external repositories
+
+### Non-Goals
+
+- Changing existing metrics schemas or breaking Prometheus compatibility
+- Implementing new metrics not currently exposed by KubeVirt
+- Replacing KubeVirt's internal monitoring for health checks
+- **Breaking compatibility** with existing monitoring setups during migration
+- **Forced immediate migration** - older versions maintain embedded approach
+
+## Definition of Users
+
+- **KubeVirt Operators**: System administrators managing KubeVirt clusters who need comprehensive monitoring and observability
+- **Platform Engineers**: Infrastructure teams responsible for monitoring stack configuration and maintenance
+- **Application Developers**: Users running VMs who need visibility into VM performance and resource usage
+- **Monitoring Tool Vendors**: Third-party monitoring solutions that integrate with KubeVirt metrics
+- **KubeVirt Contributors**: Developers working on KubeVirt who need decoupled monitoring for easier development and testing
+
+## User Stories
+
+- As a **KubeVirt operator**, I want to upgrade my monitoring stack independently from KubeVirt core components, so I can get new monitoring features without waiting for KubeVirt releases
+- As a **platform engineer**, I want to minimize resource overhead of monitoring, so I can run more VMs per node with the same hardware and I want to disable monitoring
+- As an **application developer**, I want consistent VM performance metrics regardless of KubeVirt version, so my monitoring dashboards work across upgrades
+- As a **KubeVirt operator**, I want to backport critical metrics to older KubeVirt versions, so I don't have to upgrade production clusters just for monitoring improvements
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt) - Core KubeVirt
+  modifications for hybrid architecture support
+- [kubevirt/monitoring](https://github.com/kubevirt/monitoring) - Existing repository to be extended with hybrid metrics collection implementation alongside current dashboards and monitoring guidelines
+
+## Design
+
+### Hybrid Architecture Overview
+
+This proposal introduces a **hybrid monitoring architecture** that optimizes resource usage by using the most appropriate collection method for each type of metric:
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│                    Hybrid Monitoring Architecture               │
+│                                                                 │
+│  ┌─────────────────┐    ┌─────────────────────────────────────┐ │
+│  │  Informer-Based │    │         Sidecar-Based               │ │
+│  │   Collection    │    │        Collection                   │ │
+│  │                 │    │                                     │ │
+│  │ ┌─────────────┐ │    │ ┌─────────────────────────────────┐ │ │
+│  │ │Central      │ │    │ │     virt-handler Pod            │ │ │
+│  │ │Metrics      │ │    │ │  ┌─────────────┐ ┌────────────┐ │ │ │
+│  │ │Collector    │ │    │ │  │virt-handler │ │libvirt-    │ │ │ │
+│  │ │             │ │    │ │  │             │ │metrics     │ │ │ │
+│  │ │:8080        │ │    │ │  └─────────────┘ │sidecar     │ │ │ │
+│  │ └─────────────┘ │    │ │                  │:9090       │ │ │ │
+│  │      │          │    │ │                  └────────────┘ │ │ │
+│  │      │ watches  │    │ └─────────────────────────────────┘ │ │
+│  │      ▼          │    │             │                       │ │
+│  │ ┌─────────────┐ │    │             │ Unix socket access    │ │
+│  │ │K8s API:     │ │    │             ▼                       │ │
+│  │ │- VMs        │ │    │ ┌─────────────────────────────────┐ │ │
+│  │ │- VMIs       │ │    │ │     /var/run/kubevirt-private   │ │ │
+│  │ │- Pods       │ │    │ │     (libvirt sockets)           │ │ │
+│  │ │- Services   │ │    │ └─────────────────────────────────┘ │ │
+│  │ └─────────────┘ │    │                                     │ │
+│  └─────────────────┘    └─────────────────────────────────────┘ │
+│                                                                 │
+│  Collects:                    Collects:                         │
+│  • virt-controller metrics    • VM runtime metrics              │
+│  • virt-api metrics          • libvirt domain statistics        │
+│  • virt-operator metrics     • Real-time performance data       │
+│  • Node capability metrics   • High-frequency VM metrics        │
+└─────────────────────────────────────────────────────────────────┘
+```
+
+### Key Benefits of Hybrid Architecture
+
+1. **Optional Resource Efficiency**:
+   - **Current embedded approach**: Zero additional memory (baseline)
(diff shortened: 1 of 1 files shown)


xpivarc:
Summarizing the discussion from VEP meeting:
1. Please open issue so we can add it to the tracking board
2. Please update diagram (to the latest version) and include old vs new comparison
3. Please include the discussion that happen before 11.November into issue or PR to include the context for other reviewers
4. Please talk to sig-network about how to scrap the metrics from the launcher sidecar, 
5. Please include how this enable adding metrics out of tree or eases work for forks

vladikr:
Hey, Thanks for putting together this VEP. We spoke about this in detail during the KubeVirt's unconference.
In general, separating the monitoring stack via a hybrid architecture is a great step toward improving maintainability and extensibility in KubeVirt, especially for custom metrics. 
However, as I voiced my concern during the call, and I still see this being part of the current design, direct access to libvirt is a no-go from my point of view. 

A few points:

 - libvirt is just an internal detail of the virt-launcher. It is not a stable KubeVirt API. There are no guarantees around it, and with KubeVirt exploring multiple VMMs (See #97 for example) this approach risks breaking monitoring in the future. Virt-launcher should self-report its stats to maintain encapsulation.
 - Security concern. virt-launcher is designed as an untrusted component, with strict isolation. Granting the sidecar direct libvirt access could expand the attack surface, even if read-only. virt-launcher already provides sufficient stats via GetDomainStats() in virt-launcher/virtwrap/manager.go -  why not build on that to avoid exposing internals? I'm not aware of any filtering virt-launcher is doing on the libvirt-provided data. I wonder if we can come up with an API that requests data from virt-launcher using its own API.
 - libvirt operations can be blocking and resource-intensive and since libvirtd runs in virt-launcher's compute container, external calls from the sidecar could interfere with core flows. This could lead to unexpected throttling or OOM issues. This consumption will be counted toward the virt-launcher limits, not the sidecar.
 - it doesn't look to me that virDomainListGetStats() provides meaningfully more data than GetDomainStats() - I wonder what is missing? There is no batching benefits in our setup since there is only one domain in the pod.

My main suggestion is that more stats are needed, or a more generic API, extend the DomainManager interface in virt-launcher rather than bypassing it. From my point of view, the communication should be with the virt-launcher API, not directly with libvirt.

iholder101:
ping @sradco.

The [VEP freeze](https://github.com/kubevirt/sig-release/blob/main/releases/v1.8/schedule.md) is less than a month away and we are approaching holiday seasons.

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+- `kubevirt_vm_created_total`
+- `kubevirt_vmi_phase_count`
+- `kubevirt_vmi_vcpu_seconds`
+- All VM/VMI label and annotation-derived metrics
+
+The controller exposes a `/metrics` endpoint scraped by Prometheus.
+
+**Runtime VM/VMI Metrics Generation:**
+
+The controller uses the `GetMonitoringData` RPC to collect the data from the guest agent and libvirt domain.
```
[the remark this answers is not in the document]
machadovilaca: virt-handler would no longer expose vmi runtime metrics on its /metrics endpoint

on the inital POC i did (https://github.com/kubevirt/kubevirt/pull/16580/changes#diff-228ab15f599acf22b58aaa097c96b3881ce7fac4ddb62359a4e855ddb0583017)

i kept that flow:

Prometheus -> observability controller (/metrics endpoint with VMI metrics) -> virt-api -> virt-handler -> virt-launcher (GetMonitoringData)

Eddy suggested bypassing all of the middle components and do
Prometheus -> observability controller (/metrics with VMI metrics) -> virt-launcher (GetMonitoringData)
but having the new metrics exporter connecting directly to the virt-launcher socket doesn't seem right

wdyt?

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
+# VEP: Monitoring Stack Separation via Hybrid Architecture
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
```
iholder101: Can you rebase so that the following is included?https://github.com/kubevirt/enhancements/blob/4a41a0954a441f8a2b71df4483110c9d395ba372/veps/NNNN-vep-template/vep.md?plain=1#L5-L19

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+  // ... existing RPCs stay for backward compat ...
+  rpc GetMonitoringData(MonitoringRequest) returns (MonitoringResponse) {}
+}
+
+message MonitoringRequest {
+  bool domainStats      = 1;
+  bool guestInfo        = 2;
+  bool guestFilesystems = 3;
+  bool guestAgent       = 4;
+  bool guestUsers       = 5;
```
iholder101: I think we can consider empty structs here instead of bools for this to be extendable in the future if needed (e.g. maybe in the future we'd want to collect one of those with specific parameters of partial information etc.).

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+2. **Inflexibility**: The current model limits what information is collected.
+   Adding a new libvirt statistic or guest agent field requires defining a new proto message,
+   writing a new RPC, adding a new scraper in `virt-handler`, and wiring the data through to
+   the metrics endpoint, all tightly coupled to the existing parsing pipeline. This rigidity
+   has led to useful data being left uncollected because the cost of plumbing it through is
+   disproportionate.
+
+#### Proposed Architecture
+
+A new `GetMonitoringData` RPC consolidates monitoring queries into fewer, more flexible calls.
```
iholder101: I wonder if we should generalize this a bit, maybe "GetVMStats" or something like that?
I wonder if other components would need to use it for different purposes in the future, e.g. autoscalers?

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+  transferred remains the same (or decreases slightly due to reduced framing overhead).
+- Selective serialization ensures the server only marshals requested fields.
+
+## Update/Rollback Compatibility
+
+The `kubevirt-observability-controller` is an **external** component, deployed and managed
+independently from `virt-operator`. It is not part of the core KubeVirt installation, has
+its own release lifecycle, and is installed/upgraded separately by the cluster administrator.
+This architectural decision provides clear separation of concerns and simplifies
+compatibility.
```
iholder101: So just to ensure I understand - the proposal is to develop and deploy this in a separate repo, is that correct?

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+onward, the only source for these metrics, recording rules, and alerts is the
+`kubevirt-observability-controller`. Users who depend on any of these **must**
+install the new controller before upgrading to this release.
+
+### Timeline summary
+
+| Release | virt-controller metrics | PrometheusRule reconciliation (virt-operator) | kubevirt-observability-controller | Action required |
+|---|---|---|---|---|
+| N | Emitted | Active | Emits metrics + reconciles rules (if installed) | None: install the new controller at your convenience |
+| N+1 | **Removed** | **Removed** | Sole source of metrics, rules, and alerts | Install kubevirt-observability-controller before upgrading |
```
iholder101, line 346: A bureaucratic question - how does this align with alpha/beta/GA stages?

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
+# VEP: Monitoring Stack Separation via Hybrid Architecture
```
iholder101: Can you also change the PR title?

```suggestion
# VEP 143: Monitoring Stack Separation via Hybrid Architecture
```

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,431 @@
...
+  string   domainStats      = 2;
+  string   guestInfo        = 3;
+  string   guestFilesystems = 4;
+  string   guestAgent       = 5;
+  string   guestUsers       = 6;
+  ...
+}
+```
+
+All response fields are JSON-encoded strings (consistent with the current approach). The
```
vladikr: Doesn't this defeat the purpose of protobuf's schema enforcement?
Shouldn't this be google.protobuf.Struct 


@0xFelix @xpivarc

veps/sig-observability/monitoring-stack-separation.md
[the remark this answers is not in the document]
iholder101: I think routing through virt-api adds a middleman that doesn't need to be in this data flow. The observability controller has to fan out to nodes regardless - putting virt-api in the middle just adds a hop and a failure domain with no added value. It also gives virt-api a data-forwarding responsibility that isn't its job.

I think the observability controller should talk directly to virt-handler pods, bypassing virt-api - the same way Prometheus already reaches virt-handler today via the headless `kubevirt-prometheus-metrics` Service. The controller would use VMI informers to know which nodes have VMIs, then call `GetMonitoringData` on each virt-handler. virt-handler would aggregate its local virt-launcher data over Unix sockets as usual.

For auth, I think we could have virt-operator issue a dedicated client cert for the controller. This would keep the controller external while using the existing infrastructure. The observability controller would need RBAC permissions to read secrets on the KubeVirt namespace, which downstream vendors may choose to give it.

What do you think? Does this make sense?

iholder101 (commented):
Thank you very much @machadovilaca for this VEP!

It's both very well written, and I think this is a very important topic. I'd be very happy to see this pushed forward.

veps/sig-observability/monitoring-stack-separation.md
iholder101: I think the JSON string pattern is intentional here - the existing RPCs (`GetDomainStats`, `GetGuestInfo`, etc.) all use `string` response fields the same way. The data originates from libvirt/guest agent as data that we don't control the schema for, so JSON strings seem like a reasonable choice. Would `google.protobuf.Struct` add meaningful type safety here, since it's still schemaless? Does this make sense? WDYT?

veps/sig-observability/monitoring-stack-separation.md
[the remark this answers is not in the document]
machadovilaca: Sounds good, maybe we can also somehow reduce the number of requests to virt-handler, by returning the data from all those node VMs together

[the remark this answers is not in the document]
machadovilaca: updated

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: done

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: changed

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: fine by me, changed to GetVMStats, also considered GetVMRuntimeData

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: yes

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: changed in the doc, can't change the PR title, cc @sradco

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: alpha/beta should be N, where metrics are being emitted by the new component, but the user would need to disable collection from core KubeVirt components and enable it for the new controller.

GA should be N+1, where we remove the exposition of migrated metrics from the core KubeVirt components, and only the new component exposes them

iholder101: Sounds good!

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,521 @@
...
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
iholder101: can you mark this as ticked?

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,521 @@
...
+   Adding a new libvirt statistic or guest agent field requires defining a new proto message,
+   writing a new RPC, adding a new scraper in `virt-handler`, and wiring the data through to
+   the metrics endpoint, all tightly coupled to the existing parsing pipeline. This rigidity
+   has led to useful data being left uncollected because the cost of plumbing it through is
+   disproportionate.
+
+#### Proposed Architecture
+
+A new `GetVMStats` RPC consolidates monitoring queries into fewer, more flexible calls.
+The caller specifies which data categories it needs via boolean flags; the server only
```
iholder101: nit: stale reference to boolean flags

veps/sig-observability/monitoring-stack-separation.md
```diff
@@ -0,0 +1,521 @@
...
+- **Rigid data model**: Each existing RPC has a fixed request/response shape with its own
+  parsing logic, making it difficult to collect additional libvirt or guest agent fields.
+
+- **Unclear ownership**: Recording rules and alert definitions live alongside operational
+  controller logic, complicating code review, testing, and on-call responsibility boundaries.
+
+Separating monitoring into a dedicated component with a unified gRPC interface addresses
+all of these concerns, simplifying the overall approach, giving more flexibility to evolve
+what data is collected and how it is parsed, and preserving backward compatibility with
+existing deployments.
```
iholder101: Can you just throw a word about it being in a dedicated repo so it will be clear from the start? Maybe something like


```suggestion
Separating monitoring into a dedicated component in a separete repository with a unified gRPC interface addresses
all of these concerns, simplifying the overall approach, giving more flexibility to evolve
what data is collected and how it is parsed, and preserving backward compatibility with
existing deployments.
```

iholder101 (commented):
Some tiny nits below, but generally looks great to me.

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: done

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: :raised_hands: updated

veps/sig-observability/monitoring-stack-separation.md
machadovilaca: added

veps/sig-observability/monitoring-stack-separation.md
0xFelix: Was the idea to forward "opaque" data from libvirt so that the intermediates wouldn't need to know anything about the content?

machadovilaca: yes, exactly

vladikr: okay ;)

vladikr:
@machadovilaca, overall, it looks like a significant improvement!

iholder101:
Thank you very much for this work @machadovilaca! Great stuff!

/lgtm

veps/sig-observability/monitoring-stack-separation.md
iholder101: ping @sradco

sradco: Done

Issue #85, opened 2025-08-17
VEP 85: Deploy NetworkPolicies


nunnatsa:
**Primary contact (assignee)**:
/assign @nunnatsa 

**Current Feature Stage**:
n/a

**Feature Gate**:
n/a

**Responsible SIGs**:
Primary SIG:
/sig compute

Additional SIGs (optional):
/sig network
/sig storage
/sig deployment

**Enhancement link**: 
https://github.com/kubevirt/enhancements/pull/86

**Timeline**:
v1.7

fossedihelm:
This can be closed since the NP implementation has been included in release 1.7 https://github.com/kubevirt/kubevirt/pull/16000
/close
Thank you

PR #86, opened 2025-08-17
VEP 85: First draft of the NettworkPolicy VEP

nunnatsa:
This VEP suggest to deploy NetworkPolicies as part of the KubeVirt deployment, to protect pods from unwanted traffic.

The VEP will suggest how to implement in KubeVirt, and all other components that are deployed with it, including the HCO, when it is used to deploy KubeVirt, but also CDI, CNAO, AAQ, and SSP.

### VEP Metadata

**Tracking issue**: 
https://github.com/kubevirt/enhancements/issues/85

**SIG label**: 
/sig compute

### What this PR does
Add the first draft of the Network Policy VEP, to start the discussion.


iholder101 (commented):
Hey @nunnatsa! Thanks for this proposal!

One alternative that was previously discussed is having a separated network-policy controller that would live in its own repo. I think that such an approach would be able to solve some potential issues with the approach currently suggested in this VEP:
1. It could support more installation methods (e.g. Helm) and not be OLM-specific.
2. It would be able to dynamically change network policies, i.e. due to configuration changes or cluster observed changes.
3. Kubevirt's context is already too large. From a maintenance perspective, I think it'd be much easier to keep the specialized code in a dedicated place, keeping the core as narrow as possible. In addition, changes would need to be done in one single place, instead of multiple repos (SSP, AAQ, etc).

Can you add some alternative approaches to this proposal so we can discuss their pros/cons?

nunnatsa:
Thanks for the comment, @iholder101!
 
> One alternative that was previously discussed is having a separated network-policy controller that would live in its own repo. I think that such an approach would be able to solve some potential issues with the approach currently suggested in this VEP:

I wouldn't go there. My main concern is that the NetworkPolicies will become "somebody else's problem". This is a key principle: component owners must be responsible for every aspect of their component, including the NetworkPolicies.

A future KubeVirt contributor will probably forget to open a PR for the new NP component, if they'll made a traffic related change in KV. We'll loose sync with the other components in the speed of light.

It is relatively challenging to test NetworkPolicies anyway. Implementing and testing it in some other place, will make it much harder to find networking issues, before releasing the component (e.g. KV).

> 1. It could support more installation methods (e.g. Helm) and not be OLM-specific.
I think there are many ways to solve it. For example, get the static NP file from the csv-generator (or a new dedicated application) from the operator image (that you must know anyway when you install the component). This can be a single line in the installation script.

> 2. It would be able to dynamically change network policies, i.e. due to configuration changes or cluster observed changes.

This is a good point. However, it also demonstrate that the component knows its networking need better than any other component. So, except for very few NPs (3, as mentioned in the VEP), the component should make sure the networking is working for it by fine tuning its network policies.

> 4. It would reduce the need of Kubevirt relying directly on HCO via labels.

This part is not mandatory anyway. I feel that it causes many objections and it can be removed, if no one want to use it. Personally, I think it's easier, but if it not - don't use it.

Acedus (commented):
I think it needs to be said clearly: NetworkPolicies are not operator-owned.

Operators must define RBAC to function, because without it they cannot even reconcile. RBAC is intrinsic. By contrast, NetworkPolicies are extrinsic: they are a security boundary that belongs to the cluster operator, not the operator maintainer. They enforce security intent that is highly cluster-specific and cannot be assumed by the operator.

If an operator ships operands with non-deterministic traffic patterns, such as egress to arbitrary endpoints and ports (which is common), the operator has only two choices:
- Allow all traffic (insecure by default), or
- Deny all traffic (breaking workflows until relaxed).

Neither reflects actual user intent. Only the cluster operator can decide how workloads in their environment should communicate. For example, CDI’s importer pod may fetch images from arbitrary endpoints, there is no "correct" policy the operator could enforce. This is not unique to CDI; any operator with user-driven or contextual communication faces the same dilemma.

The most sustainable practice that I saw across the ecosystem is therefore to provide well-documented example policies, not to reconcile them as part of operator lifecycle. Security-conscious admins can then adopt and adapt those examples to fit their environment.

With that out of the way, my opinions on implementation:
- We should not maintain NetworkPolicies as part of the operator's lifecycle. Doing so adds complexity for a feature that is optional by design, and conflicts with our goal of keeping projects lean.
- Any policies we do provide should be general and unsurprising (least astonishment principle). They should work in clusters without existing policies, while being easy for users to extend or tighten.
- Internal tooling to dynamically generate policies is, in my opinion, a non-starter. The simplest and most flexible path is to provide static template manifests that users can then manage with their preferred tooling (kustomize, Helm, etc.).

nunnatsa:
> Firstly, I'm not sure I agree with `component owners must be responsible for every aspect of their component`. Kubevirt already relies on CDI for storage management, AAQ for quota management, etc. I'd even advocate for the opposite: the core should be narrow and cover only what's crucial for every use-case, and it's a great practice to extract specialized logic to other components, like we already do.
> 
> I also feel that the "becoming someone else's problem" can be claimed towards CDI or AAQ. But further than that, I think this approach has many advantages. For example, it eliminates the need of common developers to understand network policies and how they work, which affects development speed and complexity. Delegating the problem to a specialized repo with their own maintainers seems like a good separation of concerns that would free the core of considering these concerns. Bear in mind that Kubevirt is already very complex and deals with many different considerations, I'm not sure it's desired to add another one (especially since you admin yourself that testing it is challenging as it is).

I feel that there is a mix here between functional and non-functional requirements. Dividing the functionality between different component makes a lot of sense. But as I see it, NPs are, in this context, similar to K8s services. They are basic K8s resources that you need in order to make your component working.

> "the component knows its networking needs better than any other component".
> 
> As said above, I'm not sure I agree with this statement. I think that most Kubevirt developers aren't familiar with network policies and don't have deep knowledge in this area. 

NetworkPolicies are K8s resources, and not a new ones. I'm too wasn't familiar with them before starting to implementing logic around them, but I think we can expect developers in the K8s ecosystem, to know the basic K8s resources, and more important, to know how to get information, find the relevant documentation and learn.

> I think a specialized group that actually understands this topic could take the lead, freeing the broad community to develop the core faster and easier.

Again: No one except the component owner can know what are the component needs. No one can decide for the component owner what is a legit egress or ingress activity, and if it should be allowed.

> @nunnatsa can you please add this (and perhaps more) approach to the VEP and describe the pros/cons?
 
> Even if you disagree, I think it's healthy to discuss and consider multiple approaches and present them to the community to have a thorough discussion.

:+1: Will do later. Will love if you can suggest the text.

Acedus:
> Question: is there a common template format that can satisfy Helm, kustomize and other tools? Is there some kind of a standard?

AFAIK there isn’t a single template format that works across all tools. The most common approaches in the wild are Helm, kustomize, Go templates, or any combination of the three. For KubeVirt's use case, we should weigh each option against our needs to find the best balance without adding too much complexity. Personally, I lean toward kustomize since it is relatively simple and k8s-native.

> Also, in terms of ownership, I think that if we're not using a separated repo (should we?), then sig-network should probably take over this area.

I don’t object to this suggestion, but I don’t fully agree either. In my opinion there’s an important distinction between KubeVirt network plumbing and k8s networking. The former falls under what the [sig-network charter](https://github.com/kubevirt/community/blob/0ea96fd79214bca4a2bddd1b860ae632535b63c5/sig-network/charter.md) defines, while the latter is an opt-in feature that should be reviewed by the maintainers of the components it affects. 

For example, virt-export-server is primarily under sig-storage's scope. It wouldn't make much sense for sig-network to be responsible for setting its policies, since they lack the context that sig-storage brings.

Issue #90, opened 2025-08-26
VEP 90: Utility Volumes for virt-launcher Pods


ShellyKa13:
**Primary contact (assignee)**:

/assign @ShellyKa13 

**Current Feature Stage**: New

**Feature Gate**: UtilityVolumes

**Responsible SIGs**: sig-storage

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig $SIG

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/91

**Timeline**:

* Alpha release target (1.8):
* Beta release target (1.9):
* Stable release target (1.10?):

* [x] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/91
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/15922
  * [ ] Docs PRs:

* [x] Beta
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/279
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

fossedihelm:
@ShellyKa13 This is VEP 90, not 86 :)

iholder101:
@ShellyKa13 would you mind updating the issue description?
Is it targeted for Beta or Alpha in 1.8?

vladikr:
@ShellyKa13 are we planning to target beta in 1.9? If yes, could you please post a pr to update the VEP?

akalenyu:
sig-storage summary; we think it makes sense to target 1.9 though we'd like to start using utilityvolumes for memorydump first
(this should be achievable within 1.9 cycle)

akalenyu:
/cc @ShellyKa13 
if we could update the VEP accordingly that'd be awesome 🙏

ShellyKa13:
@vladikr @akalenyu I have update the VEP with the updated beta target version. Review of the PR is appreciated

vladikr:
@akalenyu can you please take a look and see if we can retarget this one?

akalenyu:
/remove-lifecycle rotten

We are trying to pick this up in the 1.10 cycle, unless CBT plans have changed
/cc @Acedus

Acedus:
https://github.com/kubevirt/enhancements/pull/433

iholder101:
Hey!

[...]

Issue #25, opened 2025-03-16
VEP 22: Incremental backup with CBT


ShellyKa13:
**Primary contact (assignee)**:

/assign @ShellyKa13 

**Current Feature Stage**: New

**Feature Gate**: IncrementalBackup

**Responsible SIGs**: Storage

/sig storage

**Enhancement link**:

**Timeline**:

* [] Alpha (v1.6.0?)
* [] Beta (v1.8.0?)
  * [] Additional Beta criteria
* [] GA (v1.9.0?)
  * [] Additional GA criteria

**Additional context**:

The current backup options rely on CSI-storage and create a full snapshot of the storage with each backup. This leads to longer backup times and increased storage space usage, whether the backup is stored in the cluster or off-site. Additionally, when stored off-site, a large amount of data needs to be copied over which further extends the backup time.
This can be improved by leveraging QEMU's capability to create incremental backups, which use Changed Block Tracking (CBT) to save only the changes made since the last backup.

vladikr:
@mhenriks @ShellyKa13 What are we tracking here for beta? Are there any KubeVirt PRs?

Acedus:
Created a VEP extension to introduce pull-mode support: https://github.com/kubevirt/enhancements/pull/126

vladikr:
@Acedus are you going to work on this during 1.9 as well? 
Are you planning to graduate this to beta? 
If yes, it needs a VEP to update the target version.

Acedus:
> @Acedus are you going to work on this during 1.9 as well? 
> Are you planning to graduate this to beta? 
> If yes, it needs a VEP to update the target version. 

Hi, we're planning to work on this during 1.9 but we're not necessarily aiming to graduate to beta during that timeframe. Instead, we'll be looking for successful vendor integration and general feature stability as our indicator to decide whether to graduate this feature to beta in 1.10.

iholder101:
Moved to "removed from milestone".
Also, can you please keep the issue updated? cc @ShellyKa13

Acedus:
> Moved to "removed from milestone". Also, can you please keep the issue updated? cc [@ShellyKa13](https://github.com/ShellyKa13)

[...]

PR #91, opened 2025-08-26
VEP 90: Utility Volumes for virt-launcher Pods

ShellyKa13:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/90
**SIG label**: 
/sig storage

### What this PR does
Introduce a new VEP proposing utility volumes - hot-pluggable volumes attached directly to virt-launcher pods for out-of-band operations.

This VEP addresses the need for a generalized mechanism to dynamically attach storage for host-level operations like:
- Backup data collection (required for Incremental Backup VEP)
- Memory dumps and diagnostic data collection
- Logging/telemetry export
- and more

Key features:
- New spec.utilityVolumes API field in VMI/VM specs
- Hot-attach/detach capability using existing volume hotplug infrastructure
- Migration-aware design (utility volumes do not migrate)
- Deprecation path for existing memoryDump volume source
- Protected by UtilityVolumes feature gate with Alpha/Beta/GA lifecycle

This provides a foundation for backup solutions and other operational tools that need temporary, host-level storage access without requiring guest VM involvement.

### DIFF
--- a/veps/sig-storage/utility-volumes.md
+++ b/veps/sig-storage/utility-volumes.md
@@ -1,0 +1,280 @@
+# VEP \#86: Utility Volumes for virt-launcher Pods
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+  - [+] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+  - [ ] (R) Target version is explicitly mentioned and approved
+  - [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal introduces a new API mechanism within KubeVirt that enables the general capability of **hot-plugging volumes directly into the virt-launcher Pod**.
+These maintenance-focused volumes are attached to the Pod, bypassing the Virtual Machine guest. The primary goal is to establish this foundational mechanism to support various out-of-band operations that require direct Pod-level volume access, such as backup data collection, memory dumps, and diagnostic log retrieval.
+
+## Motivation
+
+KubeVirt currently lacks a **general mechanism for pod-level volume attachment** that enables operational workflows to dynamically attach storage directly to the `virt-launcher` Pod without exposing volumes to the VM guest. Example operational scenarios which require this capability:
+
+  * **Libvirt based Backup Operations:** Collect VM backup data from libvirt - required for [Incremental Backup VEP](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md). VEP 25 requires the ability to hot-plug a volume directly into the virt-launcher pod without exposing it to the virtual machine domain, enabling libvirt to collect backup output and write it directly to the attached volume for efficient backup operations.
+  * **Diagnostic Data Collection:** Collect memory dumps, libvirt/qemu logs or information, system diagnostics etc..
+
+While KubeVirt currently offers the `memorydump` volume source which performs such hotplug pod-level attachment, there is a broader need for a **generalized, extensible mechanism** that can support multiple operational workflows.
+
+This VEP establishes the foundational mechanism that will enable these operational workflows. Adding more individual volumeSource types would lead to API overload and semantic ambiguity - a new, clearly separated mechanism for `virt-launcher`-scoped operations is the optimal solution (more details in [Alternatives](#alternatives)).
+
+## Goals
+
+  * **Establish the foundational mechanism** for pod-level volume attachment in KubeVirt
+  * Introduce a new API field, `vmi.spec.utilityVolumes`, for hot-pluggable volumes mounted into the `virt-launcher` Pod without VM guest exposure
+  * Enable hot-attach and detach operations on running pods through a controller-managed approach
+  * Ensure extensible API design to support diverse operational workflows requiring pod-level volume access
+
+## Non Goals
+
+  * **Implementing specific operational workflows:** This VEP establishes the pod-level volume attachment mechanism; specific use case implementations (backup workflows, memory dump processes, etc.) are outside this scope
+  * **Detailing utility volume type implementations:** Focus is on the underlying attachment/detachment mechanism, not on how each utility volume type functions
+  * **Defining complete end-to-end workflows:** While this enables various operational capabilities, complete workflow implementations are covered in separate VEPs (e.g., VEP 25 for backup)
+
+## Definition of Users
+
+  * **KubeVirt Controllers:** Primary users with appropriate service account permissions to modify `utilityVolumes` for operational workflows
+  * **Kubernetes Administrators/Operators:** Initiate operations through higher-level APIs that trigger controller-managed utility volumes
+  * **Backup/DR Solution Vendors:** Develop controllers utilizing this API for VM backup data collection
+  * **KubeVirt Developers/Maintainers:** Debug KubeVirt components via controller-managed diagnostic collection
+
+## User Stories
+
+* As a cluster administrator, I want to initiate diagnostic operations through high-level APIs that automatically trigger controllers to manage utility volume attachment and data collection.
+* As a developer building KubeVirt extensions, I want a standardized mechanism for controllers to attach temporary volumes to virt-launcher pods for various operational needs.
+
+## Repos
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+The proposed design introduces a new field, `spec.utilityVolumes`, in the `VirtualMachineInstanceSpec`. This field defines **hot-pluggable volumes** mounted directly into the `virt-launcher` Pod, not exposed to the VM guest, ensuring clear separation from guest-facing volumes in `spec.volumes`.
+
+### Key Design Principles
+
+1. **VMI-Only Scope**: Utility volumes are exclusively supported in VMI spec. Attempt to edit VM spec with utility volumes would be rejected.
+2. **Controller-Managed Access**: Only KubeVirt controllers with appropriate service account permissions can modify utility volumes.
+3. **Foundational Mechanism**: This VEP focuses on establishing the underlying volume attachment/detachment mechanism rather than detailing specific utility volume type implementations.
+4. **Workflow Integration**: Controllers managing specific workflows handle both the utility volume management and the operational logic that uses those volumes.
+5. **Hot-Plug Consistency**: Monitoring and status reporting leverage existing hot-plug infrastructure and patterns.
+
+### Core Design Components
+
+**New API Field (`utilityVolumes`):**
+
+* `utilityVolumes []UtilityVolume` contains a `name` field, inlined `PersistentVolumeClaimVolumeSource`, and optional `type` field
+* `UtilityVolumeType` defines the type of utility operation (`MemoryDump`, `Backup`, etc.) for operational context and controller coordination
+* All utility volumes use PersistentVolumeClaims as the underlying storage mechanism for `virt-launcher` Pod operations
+* Name collision validation: utilityVolume names must be unique across both `spec.volumes` and `spec.utilityVolumes` lists to prevent naming conflicts
+* No corresponding entry in `domain.devices.disks` - explicitly not for the VM domain
+
+**Volume Hot-Attachment/Detachment Mechanism:**
+
+* **Hotplug Process:** Leverages KubeVirt's existing volume hotplug infrastructure, automatically recognized and mounted as directories in the `virt-launcher` Pod
+* **Mounting Behavior:** Unlike regular volumes containing `disk.img` files for the VM domain, utility volumes are mounted as filesystem directories directly accessible within the `virt-launcher` Pod environment
+* **Lifecycle Management:** Controllers managing specific workflows handle both attachment and detachment by updating the VMI specification. The `virt-controller` watches for changes and `virt-handler` manages Pod volume operations
+
+**Status Reporting:**
+
+Utility volumes use existing VMI `status.volumeStatus` field alongside regular volumes. The `Target` field remains empty since volumes aren't presented to the VM guest. Standard status fields report attachment/detachment lifecycle state only - process-specific monitoring is handled by respective workflow controllers.
+
+
+### Migration Compatibility
+
+**Design Principle:** Utility volumes should not indefinitely block VM migration. Migration waits with a configurable timeout until no utility volumes exist in the VMI spec before proceeding.
+
+**Migration Behavior:**
+
+VMs with utility volumes remain **migratable** - migration is not rejected outright. When migration is initiated, it will be kept in `Pending` state as long as utility volumes are present in the VMI spec.
+> Note: Pending migrations do not count against concurrent migration slots, hence it will not block other migrations from progressing.
+
+Each operation that attached the utility volume needs to be completed and the volumes detached by their respective controllers for the migration to continue. The migration priority mechanism can influence whether operations should be aborted or wait for completion based on migration urgency.
+
+A timeout mechanism will be introduced to prevent indefinite blocking caused by a utility volume that fails to detach. A timer begins at the migration request start, and the migration must transition to `Scheduling` state before timeout expires. If not, the migration will be marked as Failed, preventing system hang and allowing retries once the underlying issue is resolved.
+
+No utility volume attachment pods are created on the destination node during migration. Utility volumes have no point in migrating as they are temporary operational tools for pod-level operations, not part of the VM's persistent state. If desired, utility volumes can be reattached after migration completion.
+
+**Rationale:** This approach balances migration responsiveness with operational continuity by allowing utility operations to complete or abort gracefully while preventing indefinite migration blocking.
+
+## API Examples
+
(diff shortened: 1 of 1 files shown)


veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Motivation
+
+KubeVirt currently offers the `memorydump` volume source, which, upon request, hot-attaches a PersistentVolumeClaim (PVC) to the `virt-launcher` Pod to store a memory dump of the VM. While powerful for its specific use case, there is a broader need for similar **temporary, hot-pluggable, host-level data collection** that doesn't involve mounting inside the VM's guest OS. For example:
+
+  * **Backup Data Collection:** This proposal is a **required foundational component for the [Incremental Backup VEP](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md)**. The Incremental Backup VEP necessitates a mechanism to **dynamically attach a PVC** to the `virt-launcher` Pod to collect the backup output directly from libvirt. This allows for efficient data transfer and management of backup artifacts.
```
alromeros, line 19: I guess the changed block tracking would still be handled in the VM state PVC, right? Would it make sense to include it in this PVC?

alromeros: Makes sense, thanks!

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+## Design
+
+The proposed design introduces a new top-level field, `spec.utilityVolumes`, in the `VirtualMachineInstanceSpec` (and `spec.template.spec.utilityVolumes` in `VirtualMachine`). This field will define **hot-pluggable volumes**, mounted directly into the `virt-launcher` Pod, not exposed to the VM guest. This clean separation ensures clarity and avoids confusion with guest-facing volumes defined under `spec.volumes`.
+
+### Core Design Components
+
+1.  **New API Field (`utilityVolumes`):**
+
+      * Introduce `utilityVolumes []UtilityVolume` in `VirtualMachineInstanceSpec` (and `VirtualMachineSpec`).
+      * `UtilityVolume` will be a new KubeVirt type that mirrors the structure of KubeVirt's native `Volume` type, containing a `name` field for identification and a an inlined `UtilityVolumeSource`, maintaining consistency with KubeVirt's existing volume API patterns.
```
Acedus, line 58:
```suggestion
      * `UtilityVolume` will be a new KubeVirt type that mirrors the structure of KubeVirt's native `Volume` type, containing a `name` field for identification and an inlined `UtilityVolumeSource`, maintaining consistency with KubeVirt's existing volume API patterns.
```

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+      * **Target Field Handling:** For utility volumes, the `Target` field will be empty since these volumes are not presented to the VM guest.
+      * **Standard Status Fields:** Utility volumes will use the existing status fields (`Phase`, `Reason`, `Message`, etc.) to report their lifecycle state.
+
+
+### Migration Compatibility
+
+**Design Principle:** Utility volumes are temporary operational tools attached via pod-level attachment pods and are incompatible with libvirt's migration mechanisms. **Consequently, these volumes will not migrate with the VM.**
+
+**Migration Process:**
+1. **Pre-Migration:** In the unlikely event that a utility operation is active at migration time, each utility type must be designed to handle this possibility, either aborting the operation or ensuring its completion before migration proceeds.
```
Acedus, line 83: I lean toward the latter option, iirc we already delay migrations for certain operations.

Acedus (commented):
Thanks @ShellyKa13, the proposal looks very solid.
I noticed that all of the examples for utility volumes seem to follow the same pattern:
attach -> perform operation to populate the PVC -> detach.

This makes me wonder if we could do things differently and leverage volume poplators instead of adding utility volume types which could in theory help shift some load from the virt-launcher layer. This is just a thought of course, the currently suggested approach is great too IMO.

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+2. **During Migration:** Utility volumes remain attached to source VM. Migration controller filters utility volumes preventing the creation of attachment pods on destination node, namespace or cluster.
+3. **Migration Completion:** Just before migration transitions to succeeded state, utility volumes are automatically removed from both VMI and VM specifications to prevent post-migration re-attachment.
```
mhenriks, line 85: This would lead to the resource being out of sync with gitops. Ideally we don't touch any specs. Can we wait for migration to complete before starting hotplug pods for utility volumes?

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+        # remains unchanged
+        - name: system-disk
+          containerDisk:
+            image: kubevirt/fedora-cloud-container-disk-demo:latest
+      # NEW FIELD:
+      utilityVolumes:
+        - name: memorydump-target
+          memoryDump:
+            persistentVolumeClaim:
+              claimName: memorydump-pvc
```
mhenriks, line 136: `hotpluggable: true`? Or is that assumed?

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,256 @@
...
+      domain:
+        # remains unchanged
+      volumes:
+        # remains unchanged
+        - name: system-disk
+          containerDisk:
+            image: kubevirt/fedora-cloud-container-disk-demo:latest
+      # NEW FIELD:
+      utilityVolumes:
+        - name: memorydump-target
```
mhenriks, line 133: one edge case to consider, given the current hotplug implementation you have to make sure to not have name collisions with `spec.volumes` and `spec.utilityVolumes` or work around that somehow internally

mhenriks (commented):
I wonder about the use case for users specifying utility volumes on a VM. Am I correctly understanding that for things like memory dump and backup, a utility volume will only be temporarily added to the VMI then removed? If so, maybe we can hold off on giving the ability for users to specify them on the VM?

veps/sig-storage/utility-volumes.md
mhenriks: In general, we shouldn't update resource specs unless it's something created/managed internally (not in git). If a user defines a utility volume in git that is later removed by a controller, the resource will become "out of sync." Users can define policies in systems like Argo for what to do in this case. But we should avoid if possible

mhenriks:
> @mhenriks for memorydump the volume is first added to the vm and then the vmi..

Yeah but it is added and subsequently removed by the memorydump machinery. If I recall, it is technically possible for a user to add the volume manually but that can lead to unintended consequences (like the volume being removed when the dump is complete) and is not advised. I think we should avoid touching the VM spec with utility volumes if possible for now and add user defined (VM level) utility volumes later if possible

veps/sig-storage/utility-volumes.md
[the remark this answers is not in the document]
ShellyKa13: obviously it cant detect migration before it happens :)
Ill rephrase.. once migration is detected that is should happen it wont be able to start until either the utility volume action is completed or aborted.
In the case of memory dump, libvirt and qemu dsoesnt have a clean way to abort memory-dump, with memory dump it has to complete before migration can start.
With backup I will implement and abort option.
But again this is utilityvolume type specific detail that should be handled for each type independently irrelevant to the general utility volume type. its just a note. What is relevant is the volume being hotplugged to the virt-launcher.

veps/sig-storage/utility-volumes.md
ShellyKa13: So should we prevent the option of adding utilityvolumes? it should be added only by some supported command via our controllers like for current memorydump and future backup?

ShellyKa13:
@mhenriks so to sum up, you think for this VEP the utilityvolumes should be somewhat of internal controllers usability, allowed to be added only by kubevirt controllers and not by a user. Each utility-volume type will have its controller or mechanism that adds the volume to the vm spec. WDYT?

mhenriks:
> @mhenriks so to sum up, you think for this VEP the utilityvolumes should be somewhat of internal controllers usability, allowed to be added only by kubevirt controllers and not by a user. Each utility-volume type will have its controller or mechanism that adds the volume to the vm spec. WDYT?

I say for now any utility volumes added to the VM spec are rejected, only VMI volumes allowed, and only by the virt-controller SA (like all hotplug volumes)

veps/sig-storage/utility-volumes.md
mhenriks: yeah no utility volumes in VM specs for now is what I propose

vladikr:
> > @mhenriks so to sum up, you think for this VEP the utilityvolumes should be somewhat of internal controllers usability, allowed to be added only by kubevirt controllers and not by a user. Each utility-volume type will have its controller or mechanism that adds the volume to the vm spec. WDYT?
> 
> I say for now any utility volumes added to the VM spec are rejected, only VMI volumes allowed, and only by the virt-controller SA (like all hotplug volumes)

+1 I like @mhenriks suggestion. It keeps the vm api simple.
But then I don't understand how this mechanism will get triggered?
How exactly would users or external controllers trigger the backups or memory dumps without directly editing the spec?
Are we planning to introduce a new CRD or a subresource?

Regarding the status reporting: what is the plan for monitoring this operation progress/completion?

Since this VEP is speakingabout `backup` - I failed to find a comparison to the existing backup mechanism using snapshots+export.
Could you clarify the differences, advantages, and use cases. I mean why use utility volumes for incrementals instead of full snapshots/exports?
In scenarios where both approaches could work, when should users prefer utility volumes over snapshots/exports? 
For exmaple, #64 speaks about hybernation, would this scenario fit as a utility volume or snapshot?

I wonder if you thought about edge case: 
- What happens if the backup/memory dump operations (multiple concurrent or sequential operations); 
how would the controller handle them? Would it queue requests, allow multiple `utilityVolumes` entries?
how do we know that a backup operation is in progress? 

- during clean up, how the detachment will be triggered? 

@ShellyKa13 how important is this VEP for the implementation of #25 ? I see now that both a targeted for this cycle. 
The feature freeze is today, please ask for an execption if you belive we can converge this VEP in reasonable time.

ShellyKa13:
Hi @vladikr,
I want to make it clearer that this VEP's main goal is not to detail how each utility volume type functions. Instead, it focuses on the underlying mechanism for attaching and detaching volumes to the virt-launcher pod.
Based on our agreement that users should not be allowed to directly update the spec, all these modifications will be handled by a controller. Since the utilityvolumes field is just one component of a larger process for a specific functionality, the same controller that manages the full workflow (e.g., a backup or memory dump) will also handle the attachment and detachment of the volume by updating the VM/VMI spec.

The current implementation of memory dumps, adds memory dump volume source to the vm.spec.volumes and later also to vmi.spec.volumes. The plan was to keep the API as is which already does the same mechanisem of hot plugging to the virt-launcher pod and only replace the trigger of the attachment where instead of adding memorydumpvolume to the volumes list, add a memorydumputilityvolume to the utilityvolumes list.. and so if I want to keep it the same I need utilityvolumes also in the vm template spec. 

Regarding monitoring, since as mentioned the suggestion here is basically having a special struct that does hotplug to the pod the idea is that monitoring will be consistent with existing hot-plug methods. The only change is the introduction of a new struct to represent utility volumes; the progress and status of the operations will still be reported via the volumestatus field. Again the monitor here is only for the hotplug part not the process that uses the utilityvolumes struct.

This VEP is a foundational requirement for the backup functionality outlined in VEP 25. It is crucial for enabling a key backup feature: the ability to hot-plug a volume directly into the virt-launcher pod without also attaching it to the virtual machine itself. This is a requirement to allow libvirt to collect the backup output and write it directly to the attached volume.

While this VEP enables that functionality, the detailed justification and use cases for backup are explained in the separate backup VEP. Initially, this was planned as an implementation detail of VEP 25 which I already had an open PR for. @mhenriks reviewed the PR and [raised the suggestion ](https://github.com/kubevirt/kubevirt/pull/15235#pullrequestreview-3068906313) to have a seperate list for this operation which made me believe that a dedicated VEP was necessary to address this significant API change. This new API establishes a clear, separate mechanism for other future processes that may need to hot-plug volumes to the pod, justifying its separation from regular volumes. And so while required for backup its goal is not exclusive for backup but rather open this mechanism to be more generalized to be used for other functional types.

vladikr:
Hi @ShellyKa13, thank you for the clarification.  I understand that the foucs of this VEP is on the attachment/detachment mechanism itself, and that it's an important part of VEP #25 . It makes sense that the details of specific workflows (like backups) belong in their own VEPs/PRs. However, I'm still a bit confused on a few fronts, as the response doesn't fully address how this would work in practice without user-editable spec fields. I'll structure my follow-up questions/concerns below to keep it organized, grouping them by topic. My goal is to ensure we have a clear path forward, especially with the feature freeze today.

mhenriks:
> Hi @ShellyKa13, thank you for the clarification. I understand that the focus of this VEP is on the attachment/detachment mechanism itself, and that it's an important part of VEP #25 . It makes sense that the details of specific workflows (like backups) belong in their own VEPs/PRs. However, I'm still a bit confused on a few fronts, as the response doesn't fully address how this would work in practice without user-editable spec fields.

WRT [backup](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md), there is a `VirtualMachineBackup` CR with the PVC name

For memory dump there is a parameter passed to the [subresource API](https://kubevirt.io/api-reference/main/definitions.html#_v1_virtualmachinememorydumprequest)

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,252 @@
...
+          claimName: memorydump-pvc
+          hotpluggable: true
+```
+
+## Alternatives
+
+Two approaches involving placing utility volumes within the existing `spec.volumes` list rather than creating a separate `utilityVolumes` field were considered:
+
+**Option A:** Add individual volume source types (e.g., `backupVolumeSource`) alongside the existing `memoryDumpVolumeSource`
+**Option B:** Create a single `utilityVolumes` volumeSource type that contains multiple utility volume types, deprecating the standalone `memoryDumpVolumeSource` field
```
mhenriks, line 154: looking at the example yaml above, it looks like the utility volume contains a `memoryDumpVolumeSource` attribute. Maybe it would be useful to add the golang struct definitions for `spec.utilityVolumes`.

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,276 @@
...
+// Represents the type of a utility volume.
+// Only one of its members may be specified.
+type VolumeSource struct {
+	// MemoryDump represents a volume which will be used to get memory dump
+	// +optional
+	MemoryDump *MemoryDumpVolumeSource `json:"memoryDump,omitempty"`
+
+	// Backup represents a volume which will be used to get backup output
+	// +optional
+	Backup *BackupVolumeSource `json:"backup,omitempty"`
```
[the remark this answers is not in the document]
ShellyKa13: Ill keep the type optional, for future flexibility for possible user specific actions.

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,276 @@
...
+No utility volume attachment pods are created on the destination node during migration. Utility volumes have no point in migrating as they are temporary operational tools for pod-level operations, not part of the VM's persistent state. If desired, utility volumes can be reattached after migration completion.
+
+**Rationale:** This approach balances migration responsiveness with operational continuity by allowing utility operations to complete or abort gracefully while preventing indefinite migration blocking.
+
+## API Examples
+
+```
+type UtilityVolume struct {
+	// UtilityVolume's name.
+	// Must be unique within the vmi, including regular volumes.
```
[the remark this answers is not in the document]
mhenriks: then wouldn't be able to hotplug the same PVC for two different volumes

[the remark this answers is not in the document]
ShellyKa13: well then each controller that adds the utility volumes will add it at its discretion, having a special prefix for the volume with a hash of the pvc name sounds good, but again not something that I think is part of the VEP here only the fact that the volume name is unique is the important part.

vladikr:
Thank you for making the changes!
/approve

veps/sig-storage/utility-volumes.md
ShellyKa13: @aglitke @alromeros do you know what was the original design rationale for having the VMState PVC only in the status and not in the spec? this is one of the major difficulties in the design of this feature.. definitely if it can be changed it would be for the best.
I do stand behind what I said where the VMState PVC does not belong as a utilityVolume. Those are inherently hotplug volumes - not persistent, and does not interact with the domain at all.

veps/sig-storage/utility-volumes.md
```diff
@@ -0,0 +1,280 @@
+# VEP \#86: Utility Volumes for virt-launcher Pods
```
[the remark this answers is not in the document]
ShellyKa13: damn I missed changing that. Sure

PR #102, opened 2025-10-20
sig-storage: Declarative Volume Hotplug to Beta for KubeVirt 1.9

mhenriks:
### VEP Metadata

**Tracking issue**: 
**SIG label**: 
/sig storage

### DIFF
--- a/veps/sig-storage/declarative-volume-hotplug.md
+++ b/veps/sig-storage/declarative-volume-hotplug.md
@@ -5,8 +5,8 @@
 ### Target releases
 
 - This VEP targets alpha for version: v1.6
-- This VEP targets beta for version:
-- This VEP targets GA for version:
+- This VEP targets beta for version: v1.9
+- This VEP targets GA for version: v1.10
 
 ### Release Signoff Checklist
 
@@ -14,7 +14,7 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 - [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
 - [x] (R) Alpha target version is explicitly mentioned and approved
-- [ ] (R) Beta target version is explicitly mentioned and approved
+- [x] (R) Beta target version is explicitly mentioned and approved
 - [ ] (R) GA target version is explicitly mentioned and approved
 
 ## Overview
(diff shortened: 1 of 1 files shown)


vladikr:
/approve
There are no code changes required, and we didn't encounter any issues. 
I'm +1 for beta
/cc @xpivarc

veps/sig-storage/declarative-volume-hotplug.md
```diff
@@ -328,16 +328,16 @@ A comprenhensive test suite that checks the guest state will be important for th
 
 ## Feature lifecycle Phases
 
-### Alpha
+### Alpha (KubeVirt 1.6)
 
 There will be two relevent featuregates:
 
-1. `HotplugVolumes` - This existing featuregate will enable declarative adding/removing volumes and their corresponding disks in pairs
-2. `InjectEjectCDROM` - This new feature gate will allow for VM definitions that contain a CD-ROM and no volume
```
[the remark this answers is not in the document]
mhenriks: @wheatdog yes that is correct

veps/sig-storage/declarative-volume-hotplug.md
```diff
@@ -328,16 +328,16 @@ A comprenhensive test suite that checks the guest state will be important for th
...
+1. `DeclarativeHotplugVolumes` - New featuregate that enables declarative volume hotplug and inject/eject CD-ROM
+2. `HotplugVolumes` - This featuregate will be deprecated, but if enabled will take precedence over `DeclarativeHotplugVolumes` to preserve legacy behavior
 
-### Beta
+### Beta (KubeVirt 1.7)
 
-Perhaps after one or two releases, when we are confident that the feature is working as expected, move to beta.
+Declarative volume hotplug goes beta. No code changes required.
```
EdDev: By this stage, can we formally deprecate `HotplugVolumes` (but not remove yet).
I.e. Kubevirt is recommending to avoid using it from 1.7 and plans to remove it in the next release (1.8?).

Having two competing features at the beta stage is confusing.
I would also block enabling both FG in parallel, it is confusing and depends on some internal logic. I find it beneficial to be explicit, especially when one is planned to be dropped.

EdDev: It is not GA, we can change it to whatever we want. The validation can be on the VM creation.

mhenriks: At VM creation time, the only invalid configuration would be to have an "empty" cdrom and that is already checked. There is no ambiguous config.

I don't think adding validation at VM update makes much sense either unless it's a warning but virt-controller will also get that warning when it updates the VM spec for non declarative hotplug as well.

My point is that we can't get rid of the "internal logic" that gives HotplugVolumes precedence since there is no way to guarantee that both FGs will not be set in existing or newly created configs

mhenriks:
cc @Dsanatar

Issue #104, opened 2025-10-22
VEP 104: Allow disabling feature gates


iholder101:
**Primary contact (assignee)**: @iholder101.

/assign @iholder101 

**Current Feature Stage**:

New

**Feature Gate**:

No feature gate is added.

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

VEP: https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/104_allow_disabling_feature_gates/vep.md

**Timeline**:

**Exception** to go straight to GA. See VEP for more info,

* Alpha release target (x.y): v1.8
* Beta release target (x.y): v1.8
* Stable release target (x.y): v1.8

* [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/105.
* [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/16370.
* [x] Docs PRs: https://github.com/kubevirt/user-guide/pull/953.

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
This done and can be closed.

/remove-lifecycle rotten
/close

PR #105, opened 2025-10-23
VEP 104: Allow disabling feature gates

iholder101:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/104
**SIG label**: /sig compute

### What this PR does
Allows to explicitly disable feature gates.

VEP in a human readable format: https://github.com/iholder101/kubevirt-enhancements/blob/vep104/initial-vep-pr/veps/sig-compute/104_allow_disabling_feature_gates/vep.md.

### DIFF
--- a/veps/sig-compute/104_allow_disabling_feature_gates/vep.md
+++ b/veps/sig-compute/104_allow_disabling_feature_gates/vep.md
@@ -1,0 +1,275 @@
+# VEP 104: Allow disabling feature gates
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+Today, Kubevirt's API for feature gate management is a `kv.spec.configuration.developerConfiguration.featureGates`
+which is a [list of strings](https://github.com/kubevirt/kubevirt/blob/18c78c0f4d5f4c155ff2b425d0a213b563ac720e/staging/src/kubevirt.io/api/core/v1/types.go#L2813-L2815),
+each string is an enabled feature gate.
+
+The API is defined as follows:
+```go
+ type DeveloperConfiguration struct { 
+ 	// FeatureGates is the list of experimental features to enable. Defaults to none 
+ 	FeatureGates []string `json:"featureGates,omitempty"
+```
+
+With this approach it is impossible to explicitly disable a feature gate.
+This VEP aims to make it possible.
+
+## Motivation
+
+Disabling a feature gate is very important for different reasons, the main one arguably being that it opens the door to
+enable beta feature gates by default.
+This capability is very important, because it allows widely testing a feature upstream
+(which will at least be tested by CI and developers, alongside small users) while possibly disabling it downstream.
+
+This approach enables to gain wider feedback and confidence before the feature becomes GA,
+which has the potential to ensure features are much more stable and get a much wider feedback before graduating.
+
+This VEP however is only scoped to allow disabling feature gates.
+Enabling Beta feature gates is out-of-scope for this VEP and should be addressed in a follow-up.
+
+## Goals
+
+* Allow to explicitly disable feature gates.
+
+## Non Goals
+
+* Enable feature gates by default.
+
+## Definition of Users
+
+* Feature developers.
+* Cluster admins.
+
+## User Stories
+
+* As a feature developer, I want my feature to be widely tested and used before graduating,
+so I can gain wide feedback and confidence in it.
+* As a cluster administrator, I want to be able to decide on whether to use Alpha/Beta features or not.
+
+## Repos
+
+kubevirt/kubevirt.
+
+## Design
+
+Kubevirt CR's API needs to change in order to allow disablement of feature gates.
+
+See the different alternatives below.
+
+### Chosen approach - Use a complementary slice for disabled feature gates
+
+The chosen approach is approach number 3 from below.
+
+The subject was discussed in an upstream-facing meetings.
+The AI-generated summary can be found in:
+https://docs.google.com/document/d/10aqnfV4EWYHGilsWvVCqaX3sNdtUqKb1sTKOFrm-W38/edit?usp=sharing
+
+Briefly, despite there was a complete consensus that approach 1 is ideally better,
+there was also an agreement that since features gates are highly unlikely to be extended in terms of configuration options
+(other than being either enabled or disabled) there is almost no real value in creating another field and exposing
+two ways to achieve the same goal.
+If in the future we will find the need to extend feature gate configuration, we can always circle back to deprecating
+the current fields and fallback to approach number 1.
+
+## API Examples
+
+See below.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,264 @@
...
+As part of this approach, a feature enablement is determined by its configuration in the new `kv.spec.configuration.featureGates` field.
+This field is a slice of `FeatureGateConfiguration` objects which currently contain two fields, `name` and `enabled`:
+```go
+type FeatureGateConfiguration struct {
+	Name string `json:"name"`
+	// Enabled indicates whether the feature gate is enabled or not. Defaults to true.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+}
+```
```
iholder101: I vote in favor of this approach BTW.

nunnatsa: Me too

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,264 @@
...
+### Approach #1 - Use a `FeatureGateConfiguration` struct slice with an `Enabled` field
+
+As part of this approach, a feature enablement is determined by its configuration in the new `kv.spec.configuration.featureGates` field.
+This field is a slice of `FeatureGateConfiguration` objects which currently contain two fields, `name` and `enabled`:
+```go
+type FeatureGateConfiguration struct {
+	Name string `json:"name"`
+	// Enabled indicates whether the feature gate is enabled or not. Defaults to true.
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
Barakmor1: I wonder if we should make this string follow best practice conventions, even though the enabled field is only expected to have true or false values. What do you think?

Barakmor1 (commented):
Thank you for taking the initiative to move this forward! Since we currently don’t have a way to disable feature gates, we cannot enable them by default. This affects the feature gate lifecycle and makes it harder to test beta feature gates at scale. That is why I believe this proposal is very important and should be a high priority.

Regarding the proposed solutions, I am fine with approach 1, and I am also okay with approach 4, even though it adds some parsing overhead that needs careful handling. Maintaining an extra API field for something as central as feature gates has a high cost, so this trade-off is worth considering.

lyarwood (commented):
Thanks for pushing this forward as a VEP @iholder101, hopefully something we can agree on and land for v1.8.0.

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,253 @@
...
+The api conventions specifically mention using maps as an anti-pattern:
+"Lists of named subobjects preferred over maps ...  There are no maps of subobjects in any API objects".
+See [Kubernetes api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps)
+for more info.
+* Maps are, by definition, unordered. This means that users/clients always have to sort it in a consistent manner
+to avoid entities mistakenly assuming that the state had changed because a change of ordering.
+* Non-extensible: If we'll need to add more configurations to feature gates in the future, we'll be in the same problem
+of having to extend the API.
+
+### Approach #3 - Use a complementary slice for disabled feature gates
```
EdDev: From my point of view, this is the only solution that preserves "comparability" as elaborated in the 3rd comment at https://github.com/kubevirt/enhancements/pull/105/files#r2596246960 .

It may look less attractive, but it seems the safest and most compatible solution, without adding logic to sync two inputs.

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,253 @@
...
+        - DownwardMetrics
+        # - ImageVolume  -> invalid, would result in an error since already provided above
+```
+
+Pros:
+* No need to deprecate the old `kv.spec.configuration.developerConfiguration.featureGates` field.
+* Easy adoption: the API is very familiar to what we have to day.
+
+Cons:
+* Non-extensible: If we'll need to add more configurations to feature gates in the future, we'll be in the same problem
```
EdDev: FG should have only on-off specifications, anything beyond that are not expected to be added.
Extensibility could be added through yet-another-list.

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
iholder101: I've switched to using `type FeatureGateEnablement string` now :+1:

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+Pros:
+* Extensible: a struct opens the door for further configuration moving forward.
+* Aligns with Kubernetes best practices: "the convention is to use a list of subobjects containing name fields ... 
+This rule maintains the invariant that all JSON/YAML keys are fields in API objects".
+See [Kubernetes api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps)
+for more info.
+
+Cons:
```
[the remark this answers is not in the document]
iholder101: Added it as a con

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,253 @@
...
+
+Pros:
+* Avoid adding / changing API fields.
+* Syntax is fairly simple and elegant.
+
+Cons:
+* Special parsing would be needed by any user/client. This means:
+  * Kubevirt's code would need to include many parsing logic. 
+  * Standard tooling would have trouble working with that.
+* We keep the API field under `kv.spec.configuration.developerConfiguration` instead of finding a new home under `kv.spec.configuration`.
```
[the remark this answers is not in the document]
iholder101: What I meant is that if we were designing this API today from scratch we would probably place it under `kv.spec.configuration`.

veps/sig-compute/104_allow_disabling_feature_gates/vep.md
```diff
@@ -0,0 +1,260 @@
+# VEP 104: Your short, descriptive title
```
dollierp: Title needs to be updated.

Barakmor1:
Sorry if this was already discussed, but I’m wondering whether we’ll have a way to see which feature gates are enabled without needing to look through the codebase, especially now that the default won’t necessarily be false.

iholder101:
> Sorry if this was already discussed, but I’m wondering whether we’ll have a way to see which feature gates are enabled without needing to look through the codebase, especially now that the default won’t necessarily be false.

Yes this was raised by others. I think this is out-of-scope for this VEP as we don't turn on beta features by default (yet).

But generally, my take on this is that feature gates are intended for advanced users / developers, hence I feel it's ok to go through a manual process for things to happen (verify which other FGs are enabled / having to kill virt-handler after enabling an FG / etc). It is possible however to expose this in KubeVirt CR's status which we can debate elsewhere.

BTW, @alaypatel07 suggested to add a section to the VEP template about advanced suggestions of this sort. I think it's a good idea.

Issue #106, opened 2025-10-27
VEP 106: Decouple net-attach-def from KubeVirt


nirdothan:
**Primary contact (assignee)**:

/assign @nirdothan 

**Current Feature Stage**:

**Feature Gate:** DisableNADResourceInjection

**Responsible SIGs**:

Primary SIG:
/sig network

**Enhancement link**: 
https://github.com/kubevirt/enhancements/blob/main/veps/sig-network/decouple-nad.md 

**Timeline**:

* FG release target (1.8):
* Stable [no FG] release target (1.10):

* [ ] FG
  * [x] VEP PRs:  https://github.com/kubevirt/enhancements/pull/107
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@nirdothan @orelmisan is there a PR that implements this VEP?

orelmisan:
@vladikr, yes:
- https://github.com/kubevirt/kubevirtci/pull/1607
- https://github.com/kubevirt/kubevirtci/pull/1609
- https://github.com/kubevirt/kubevirtci/pull/1610
- https://github.com/kubevirt/kubevirt/pull/16592
- https://github.com/kubevirt/kubevirt/pull/16734

iholder101:
@nirdothan Please keep the issue updated.

Unfortunately, it's not the first time you're poked for this. Both @vladikr and I already pinged you specifically to keep this updated.

It's your responsibility to keep it updated - please ensure to do so.

iholder101:
> Thanks [@orelmisan](https://github.com/orelmisan) . I didn't find it because these didn't have an approved-vep label and were not referenced here. I added it a reference to the open pr, now it has it.

Also please link the tracking issue in future implementation PRs to avoid this moving forward.

nirdothan:
> [@nirdothan](https://github.com/nirdothan) [@orelmisan](https://github.com/orelmisan) is there a PR that implements this VEP?

Sorry missed this. Added now.

orelmisan:
@nirdothan could you please adjust the FG name to `ExternalNetResourceInjection`? Thanks :)

orelmisan:
@nirdothan, could you please attach the following in the GA section? Thanks.
- https://github.com/kubevirt/kubevirt/pull/18497
- https://github.com/kubevirt/user-guide/pull/1027

PR #107, opened 2025-10-27
VEP 106: Decouple net-attach-def from KubeVirt 

nirdothan:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/106
**SIG label**: /sig network

### What this PR does

Remove dependency on net-attach-def

### DIFF
--- a/veps/sig-network/decouple-nad.md
+++ b/veps/sig-network/decouple-nad.md
@@ -1,0 +1,237 @@
+# VEP 106: Decouple net-attach-def from KubeVirt
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Target version is explicitly mentioned and approved
+- [x] (R) Graduation criteria filled
+
+## Overview
+
+KubeVirt utilizes Network Attachment Definition (NAD) annotations to enforce node-level network resource requirements. When a NAD is annotated with a resource name, KubeVirt automatically adds its value as a resource request to the virtual machine's virt-launcher pod. This mechanism ensures the Kubernetes scheduler only places the VM on nodes that can provide the required network.
+Common use cases include SR-IOV, bridge network when [bridge-marker](https://github.com/kubevirt/bridge-marker) is used and [macvtap](https://github.com/kubevirt/macvtap-cni).
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM. It does so by querying the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.
+`NetworkAttachmentDefinition` is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to eventually remove that query, in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.  
+In addition, the virt-controller and virt-operator ClusterRole that currently allow it to access the NAD will eventually be removed.
+
+## Motivation
+
+- This replacement provides the architectural advantage of decoupling KubeVirt from an external dependency and simplifying code maintenance.
+- The query is currently a performance bottleneck that has been [observed](https://github.com/kubevirt/kubevirt/issues/14615) in mass VM activations. It can be alleviated by caching NAD data, thus reducing direct API calls.
+- Security will be improved by removing access permissions to the external CRD from KubeVirt's control plane components.
+
+
+## Goals
+
+- KubeVirt will no longer be aware of NetworkAttachmentDefinition objects.
+- Decrease the number of API calls required to spin-up and migrate a VM
+
+## Non Goals
+
+Removal of other multus related dependencies
+
+## Definition of Users
+
+This enhancement is intended for:
+
+- VM owners
+- Cluster Admins
+
+## User Stories
+
+- As a VM owner, I want to make sure that my VM keeps running following the upgrade. There is no change of VM spec.
+
+- As a Cluster Admin I want to reduce the scope of ClusterRoles as much as possible, so that my system would use the principle of least privilege.
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+- https://github.com/kubevirt/kubevirtci
+
+## Design
+
+### Current Architecture
+
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements (`k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limits.
+
+### Proposed Architecture
+
+Instead of [querying the NAD directly](https://github.com/kubevirt/kubevirt/blob/v1.7.0-rc.0/pkg/network/multus/nad.go#L52) and [processing it in the VMI controller](https://github.com/kubevirt/kubevirt/blob/v1.7.0-rc.0/pkg/virt-controller/services/template.go#L370), an existing external component, [network-resources-injector](https://github.com/k8snetworkplumbingwg/network-resources-injector), can perform the same operation.
+This component includes a mutating webhook that automatically mutates Pods, upon creation, to inject network resource requirements from NetworkAttachmentDefinition annotations into pod resource requests and limits.
+network-resources-injector, although a standalone project, is also part of the [sriov network operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator/tree/v1.6.0/bindata/manifests/webhook). It is known to run successfully in KubeVirt clusters that already include the `sriov network operator`, 
+even though it does not mutate virt-launcher pods, since the existing KubeVirt code pre-populates the pod template with the resource requests before the admission webhook operates.
+This architecture decouples KubeVirt from NAD resources, creating a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while the admission controller handles network resource injection.
+A new Feature Gate will be introduced named `DisableNADResourceInjection`, in preparation for NAD query code removal.
+If the FG is enabled:
+
+- The VMI/Migration controller code that queries NADs and populates custom resources in the virt-launcher pod template, will be skipped.
+- The RBAC rules allowing NAD objects fetch, deployed by virt-operator, will be disabled.
+
+After 2-3 releases, the code that executes NAD querying will be removed along with the code that deploys associated RBAC rules (applies to both options below).
+
+### Deployment
+
+#### Deprecate existing injection code and require manual deployment of network-resources-injector
+
+Documentation will indicate that network-resources-injector is a required dependency for KubeVirt installations that use secondary networks, and will refer users to the github repo for installation instructions.
+Documentation will provide configuration instructions for MutatingWebhookConfiguration to optimize performance, by limiting pod interception to only pods that have the `k8s.v1.cni.cncf.io/networks`.
+In addition, a deprecation notice has been [published](https://groups.google.com/g/kubevirt-dev/c/5AvvhNYAtqU) in the kubevirt-dev mailing list to let users prepare and respond.
+Relevant CI lanes will be enhanced to perform the deployment, using [official image](https://ghcr.io/k8snetworkplumbingwg/network-resources-injector:v1.8.0) clones. This approach is equivalent to KubeVirt's dependency on Multus and how it is currently deployed.
+Other deployment systems in the KubeVirt ecosystem can be considered to assist users in deployment, such as [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) or [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator), however this proposal's scope is limited to KubeVirt/KubeVirt only.
+
+
+### Deployment Alternatives: 
+
+##### Use virt-operator to deploy network-resources-injector
+
+The network-resources-injector repo includes all the source code, build infrastructure and deployment yamls required for successful deployment of the operator.
+Built server images are available in ghcr.io, e.g. ghcr.io/k8snetworkplumbingwg/network-resources-injector:v1.8.0
+The image will be cloned and published to the official kubevirt public registry.
+Virt-operator will be enhanced to deploy and reconcile the network-resources-injector along with its deployment, service, MutatingWebhookConfiguration and RBAC.
+The manifests will be copied from the original repo and converted to go code.
+
+### Pros and Cons
+
+#### Selected deployment option: Deprecation
+
+##### Pros
+
+- Easiest option to implement and maintain
+- Does not add any code or build steps to current repo
+
+##### Cons
+
+- Breaks backward compatibility 
+
+#### Alternative deployment option: virt-operator
+
+##### Pros
+
+- No backward compatibility breach
+- Synchronized upgrade, where removed code is replaced with added code in a controlled manner 
+- Allows for opinionated configuration, e.g. optimize to process only virt-launcher pods. 
+
+###### Cons
+
+- Virt-operator currently only deploys internal components and does not reconcile 3rd party tools. Letting it deal with external components may architecturally be a slippery slope that may furthermore contribute to blurring responsibilities between KubeVirt, [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) and [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator).
+It will be hard to reason why virt-operator should deploy network-resources-injector, but not other network dependencies such as multus, sriov-operator/device-plugin, various CNIs etc.
+- `network-resources-injector` requires multus to be deployed as it is essentially an extension of the project. It implements an informer on the NAD CRD, which is deployed by multus. Multus is not a strict dependency of KubeVirt and is not even deployed in all CI test lanes. Missing the NAD schema is expected to cause errors in virt-operator. 
+- Adds a significant maintenance burden to KubeVirt project.
+- NAD related RBAC rules remain in virt-operator though bound to network-resources-injector.
+- The objective of decoupling is not achieved.
+- Since sriov-operator deploys network-resources-injector, clusters with both sriov-operator and KubeVirt deployed, will result in duplicate deployment/reconciliation attempts of network-resources-injector.
+
+> [!NOTE] The rest of this document assumes Option #1
+
+## API Examples
+
+### NetworkAttachmentDefinitions
+
+**SR-IOV Network:**
+
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  name: sriov-net-attach-def
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: intel.com/sriov_net
+spec:
+  config: |
+    {
+      "cniVersion": "0.3.1",
+      "type": "sriov",
+      "vlan": 100
+    }
+```
+
+
+### Expected virt-launcher Pod (After network-resources-injector)
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-testvm-xxxxx
+  annotations:
+    # KubeVirt applied annotation
+    k8s.v1.cni.cncf.io/networks: '[{"name":"sriov-net-attach-def","namespace":"default"}]'
+spec:
+  containers:
+  - name: compute
+    resources:
+      requests:
+        # network-resources-injector applied request
+        intel.com/sriov_net: "1"
+      limits:
+        # network-resources-injector applied limit
+        intel.com/sriov_net: "1"
+```
+
+## Scalability
+
+The network-resources-injector approach provides several scalability advantages over the current implementation:
+
+- Caching of NADs avoids redundant API calls.
+- By default runs in 2 replicas, but can be scaled to run in multiple replicas, thus highly available.
+
+## Update/Rollback Compatibility
+
+Existing VMs in clusters opting in with the feature gate must deploy network-resources-injector before upgrade, so that new VMs and migrated VMs will be able to consume custom network resources.
+
+Note that since the mechanism is replaced. There will be minor changes in the manner in which errors are handled. For example, if a referenced NAD is missing,
+The emitted error event will indicate:   
+```
+Message: failed to create virtual machine pod: admission webhook "network-resources-injector-mutating-config.k8s.cni.cncf.io" denied the request
+```
+Whereas it currently indicates:
+```
+Message: failed to render launch manifest: failed to locate network attachment definition default/ptp-conf
+```
+
+### Missing Injector Impact
+
+After code is removed (or if FG is enabled), in case system admins fail to install `network-resources-injector`, VMs requiring network devices will fail to run. VMIs remain in scheduling state and virt-launcher pods fail to initialize, as Kubelet does not request a device from the device plugin.
+
+Examples:
+
+- In case of SR-IOV, sriov-cni fails as it does not receive an allocated PCI device from kubelet. The following event is emitted:
+
+```
+Warning  FailedCreatePodSandBox  11s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "6be33fd689d91f280d6f82b9c6e3f018968fed39cd7762eb03f046bae10b8773": plugin type="multus" name="multus-cni-network" failed (add): [default/virt-launcher-testvmi-txhzt-p4hxv:sriov]: error adding container to network "sriov": SRIOV-CNI failed to load netconf: LoadConf(): VF pci addr is required
+```
+
+- In case of macvtap, macvtap cni fails for a similar reason: It cannot find a device because it does not receive an allocated device from kubelet. The following event is emitted:
+
+```
+Warning  FailedCreatePodSandBox  0s (x6 over 72s)  kubelet  : Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_virt-launcher-testvmi-vnmz7-rdm9r_kubevirt-test-default1_d3ea53de-aa74-4128-b91a-a94d9c126e89_0(1f24220d8d124bd8552c287f16e06a35fd5e510f1ed9a08890927aa93094fc4e): error adding pod kubevirt-test-default1_virt-launcher-testvmi-vnmz7-rdm9r to CNI network "multus-cni-network": plugin type="multus-shim" name="multus-cni-network" failed (add): CmdAdd (shim): CNI request failed with status 400: 'ContainerID:"1f24220d8d124bd8552c287f16e06a35fd5e510f1ed9a08890927aa93094fc4e" Netns:"/var/run/netns/b5eaba98-9d38-42bc-b1d2-c2a101b65a90" IfName:"eth0" Args:"IgnoreUnknown=1;K8S_POD_NAMESPACE=kubevirt-test-default1;K8S_POD_NAME=virt-launcher-testvmi-vnmz7-rdm9r;K8S_POD_INFRA_CONTAINER_ID=1f24220d8d124bd8552c287f16e06a35fd5e510f1ed9a08890927aa93094fc4e;K8S_POD_UID=d3ea53de-aa74-4128-b91a-a94d9c126e89" Path:"" ERRORED: error configuring pod [kubevirt-test-default1/virt-launcher-testvmi-vnmz7-rdm9r] networking: [kubevirt-test-default1/virt-launcher-testvmi-vnmz7-rdm9r/d3ea53de-aa74-4128-b91a-a94d9c126e89:net1]: error adding container to network "net1": failed to lookup device "": Link not found
+```
+
+## Functional Testing Approach
+
+Existing e2e tests already rely on mapping of custom resources and thus will validate regression.
+
+## Feature Lifecycle
+
+### Beta (Skipping Alpha)
+
+  Since no new code addition is planned (other than validation warnings), there's not much to protect with a multi-phased FG.
+  The FG is mainly used in order to provide users with sufficient preparation time, and a means to roll back if they failed to deploy network-resources-injector.
+  As such:
+  - In release 1.8, the `DisableNADResourceInjection` FG will be introduced in **disabled by default** mode.
+    While the FG is enabled, KubeVirt will **not** map custom-resources, and will **not** deploy associated RBAC rules.
+    The NAD query and RBAC code will be marked as deprecated.
+    This provides users with 1 release period to learn about the new required dependency.
+  - In release 1.9 or 1.10 (user feedback dependent), the FG will be redesignated as **enabled by default**, essentially functioning as if the code was removed. However, users can still roll back this behavior by explicitly disabling the FG. 
+
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+  A warning will be issued for the VM API in case secondary networks exist and FG is not enabled. A validation webhook will be implemented to issue the warning at VM creation. 
+
+
+### GA
+
+  In release 1.10 or 1.11, if there's no significant negative feedback from users, the NAD query and RBAC code will be removed, and the FG discontinued.


veps/sig-network/decopuld-nad.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Overview
+
+**Target Version**: v1.8
+
+VMs that use secondary networks inject custom resource requests into the virt-launcher pod template whenever custom resources are required to exist on the node in order to fulfill the secondary network.
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM that is launched.
```
orelmisan: The virt-controller queries the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+## Overview
+
+**Target Version**: v1.8
+
+VMs that use secondary networks inject custom resource requests into the virt-launcher pod template whenever custom resources are required to exist on the node in order to fulfill the secondary network.
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM that is launched.
+Network-attachment-definition is an external project maintained by the k8s network plumbing WG. The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
```
orelmisan:
```suggestion
NetworkAttachmentDefinition is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
```

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+## Motivation
+
+The replacement has the architectural advantage of decoupling KubeVirt from an external dependency by simplifying code maintenance and supporting future replacement technologies.
+Security will be improved by removing access permissions to the external CRD.
+In addition, a performance improvement can be gained from using a project that caches NAD data. 
+    
+
+## Goals
+
+NAD query code will be deleted from virt-controller and a replacement injector will be introduced.
```
orelmisan: Please consider the following alternative:
```suggestion
KubeVirt will no longer be aware of NetworkAttachmentDefinition objects.
```

The current wording describes the "how" instead of the "what".

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+## Definition of Users
+
+This enhancement is intended for:
+- VM owners
+- Namespace owners
+
+## User Stories
+
+- As a VM owner, I want to accelerate VM load time
+- As a VM owner and a namespace owner, I want my system to use the principle of least privilege, such that service accounts will only have access to resources that they absolutely need.
```
orelmisan: I think it matters more to the cluster admin, as virt-controller and virt-operator use a cluster role for reading the NAD objects.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements ( `k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limit.
+
+### Proposed Architecture
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
+
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+Current KubeVirt functionality will be deprecated and removed within 3 KubeVirt releases.
```
orelmisan: How will KubeVirt decide whether to perform the query and add the resources to the virt-launcher pod?

orelmisan (changes requested):
Thank you for the detailed proposal @nirdothan!
Please see the inline comments.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+ 
+
+##### Pros
+- No BWC breach
+
+##### Cons
+- Virt-operator currently only deploys internal components and does not reconcile 3rd party tools. Letting it deal with external components may architecturally be a slippery slope that may furthermore contribute to blurring responsibilities between KubeVirt, [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) and [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator). 
+It will be hard to explain why it's important to have virt-operator deploy network-resources-injector, but other network dependencies such as multus, sriov-operator/device-plugin, various CNIs etc.
+- Considerably bloat the code base and/or build system, just to replace a few lines of code.
+- "Swallowing" an external project somehow defeats the objective of decoupling dependencies.
```
oshoval: and prone for doing it again, which is bad practice.
it also miss the point of this effort.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Overview
+
+**Target Version**: v1.8
+
+VMs that use secondary networks inject custom resource requests into the virt-launcher pod template whenever custom resources are required to exist on the node in order to fulfill the secondary network.
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM that is launched.
```
nirdothan: Updated.

veps/sig-network/decouple-nad.md
nirdothan: Added, though I don't think that cluster admins care more about the security of their VMs than the owners of the VMs.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+Current KubeVirt functionality will be deprecated and removed within 3 KubeVirt releases.
+Documentation will indicate that network-resource-injector is a required dependency for KubeVirt installations and will refer users to the github repo for installation instructions.
+Relevant CI lanes will be enhanced to perform the deployment. This approach is equivalent to KubeVirt's dependency on Multus and how it is currently deployed. 
+Optionally, log warnings can be added to virt-operator in case network-resources-injector's MutatingWebhookConfiguration is missing, in order to assist in identifying BWC breach associated issues.
+
+Other deployment systems in the KubeVirt group can be considered to assist users in deployment, such as [CNAO](https://github.com/kubevirt/cluster-network-addons-operator), however this proposal's scope is limited to KubeVirt/KubeVirt only.
```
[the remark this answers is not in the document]
nirdothan: > ghcr.io/k8snetworkplumbingwg/network-resources-injector:v1.8.0

I didn't know that there were official images. In CI we can surely clone those like we do for the Istio installer. Added above.

Regarding building internally. That's part of Option 2. Under option 1 we will not go into this effort.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Proposed Architecture
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
+
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+Current KubeVirt functionality will be deprecated and removed within 3 KubeVirt releases.
+Documentation will indicate that network-resource-injector is a required dependency for KubeVirt installations and will refer users to the github repo for installation instructions.
```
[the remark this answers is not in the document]
nirdothan: Added.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+- Does not add any code or build steps to current repo
+
+##### Cons
+- Breaks BWC
+
+#### Option 2: Use virt-operator to deploy network-resources-injector
+
+The https://github.com/k8snetworkplumbingwg/network-resources-injector repo includes all the source code, build infrastructure and deployment yamls required for successful deployment of the operator.
+However, it does not have an official published image for the server. 
+1. The server code will be imported into the KubeVirt repository as vendored code [optional - not recommended]
```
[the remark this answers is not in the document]
nirdothan: changed the design of this option following your suggestion. Please review.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
+
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+Current KubeVirt functionality will be deprecated and removed within 3 KubeVirt releases.
+Documentation will indicate that network-resource-injector is a required dependency for KubeVirt installations and will refer users to the github repo for installation instructions.
+Relevant CI lanes will be enhanced to perform the deployment. This approach is equivalent to KubeVirt's dependency on Multus and how it is currently deployed. 
+Optionally, log warnings can be added to virt-operator in case network-resources-injector's MutatingWebhookConfiguration is missing, in order to assist in identifying BWC breach associated issues.
```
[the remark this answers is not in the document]
nirdothan: > wouldnt the vm simply fail with an error, visible on its describe ?

Visible: yes. Comprehensible? I doubt that users would be able to understand the reason why a CNI failed to consume a resource.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+Removal of other multus related dependencies 
+
+## Definition of Users
+
+This enhancement is intended for:
+- VM owners
+- Namespace owners
+- Cluster Admins
+
+## User Stories
```
oshoval: naive question, is there also a "dev story", i.e about maintaining / code complexity and such ?

it is written all over, just wondering if it also fits here as a bullet, acting as highlight

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+Built server images are available in ghcr.io e.g. ghcr.io/k8snetworkplumbingwg/network-resources-injector:v1.8.0
+The image will be cloned and published to the official kubevirt public repo.
+Virt-operator will be enhanced to deploy and reconcile the network-resources-injector along with its deployment, service, MutatingWebhookConfiguration and RBAC.
+The manifests will be copied from the original repo and converted to go code.
+
+
+##### Pros
+- No backward compatibility breach
+
+##### Cons
```
oshoval: what about adding to the discussion the issue of what if it is already presented ?
so it will need to detect and uninstall it if so, back and forth if it is removed / installed etc
I think it is a big cons

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [x] (R) Graduation criteria filled
+
+## Overview
+
+**Target Version**: v1.8
+
+VMs that use secondary networks inject custom resource requests into the virt-launcher pod template whenever custom resources are required to exist on the node in order to fulfill the secondary network.
```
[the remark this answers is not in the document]
nirdothan: Done.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM. It does so by querying the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.
+`NetworkAttachmentDefinition` is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to remove that query and code dependency in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.
```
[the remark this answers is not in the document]
nirdothan: I removed it because it was an incorrect statement. The project if full of nad imports in many places.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements ( `k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limit.
+
+### Proposed Architecture
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
+
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+A new Feature Gate will be introduced named `decoupleNAD`, in preparation for NAD query code removal.
```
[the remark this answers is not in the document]
nirdothan: Yes. Much better.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+If the FG is enabled, the code that queries NADs (for secondary networks only) and populates custom resources in the virt-launcher pod template, will be skipped.   
+After graduation, within 3 releases, the code that executes that will be removed.
```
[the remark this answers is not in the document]
nirdothan: Done.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+Virt-operator will be enhanced to deploy and reconcile the network-resources-injector along with its deployment, service, MutatingWebhookConfiguration and RBAC.
+The manifests will be copied from the original repo and converted to go code.
+
+
+##### Pros
+- No backward compatibility breach
+
+##### Cons
+- Virt-operator currently only deploys internal components and does not reconcile 3rd party tools. Letting it deal with external components may architecturally be a slippery slope that may furthermore contribute to blurring responsibilities between KubeVirt, [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) and [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator). 
+It will be hard to explain why it's important to have virt-operator deploy network-resources-injector, but other network dependencies such as multus, sriov-operator/device-plugin, various CNIs etc.
```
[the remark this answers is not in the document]
nirdothan: Unfortunately there is no single common method/tool. 
In CI we use [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) to install CNIs, but Multus may be on it's was to become a standalone deployment as Istio already is.
Not all KuebVirt users use CNAO, and they are free to install the dependencies that they require in any way that they see fit.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,182 @@
...
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM. It does so by querying the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.
+`NetworkAttachmentDefinition` is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to remove that query and code dependency in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.  
+In addition, the virt-controller role binding that currently allows it to access the NAD can be reduced in scope.
```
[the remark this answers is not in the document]
orelmisan: Since virt-operator deploys virt-controller's RBAC rules, it needs to have the same (or greater) permissions.
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-creation-or-update

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Target version is explicitly mentioned and approved
+- [x] (R) Graduation criteria filled
+
+## Overview
+
+**Target Version**: v1.8
+
+KubeVirt utilizes Network Attachment Definition (NAD) annotations to enforce node-level network resource requirements. When a NAD is annotated with a resource name, KubeVirt automatically adds its value as a resource request to the virtual machine's virt-launcher pod. This mechanism ensures the Kubernetes scheduler only places the VM on nodes that can provide the required network.
+
```
orelmisan: Please consider giving SR-IOV, bridge network when bridge-marker [1] is used and macvtap [2] as examples of known use-cases.

[1] https://github.com/kubevirt/bridge-marker
[2] https://github.com/kubevirt/macvtap-cni

veps/sig-network/decouple-nad.md
oshoval: i meant that the fact it is tricky + that it might do problems (or not, but we didn't actually test it yet with two) + the fact there might be 2 at the same time, is a cons by itself that can result in selecting other alternative
hence might worth to add

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,178 @@
...
+One such example is SR-IOV free VFs that must be available on a node for the VM to make use of.
+The names of those custom resources are listed as an annotation in the network-attachment-definition (NAD) CR that corresponds to the secondary network requested by the VM.
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM that is launched.
+Network-attachment-definition is an external project maintained by the k8s network plumbing WG. The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to remove that query and code dependency in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.  
+In addition, the virt-controller role binding that currently allows it to access the NAD can be reduced in scope.
+
+
+## Motivation
```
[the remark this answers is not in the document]
orelmisan: Please resolve this thread.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+### Current Architecture
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements ( `k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limits.
+
+### Proposed Architecture
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
```
orelmisan: Since `network-resources-injector` is first introduced here, consider adding a link to its GitHub repository and providing a brief explanation of what it does.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Scalability
+
+The network-resources-injector approach provides several scalability advantages over the current implementation:
+- Caching of NADs avoids redundant API calls.
+- By default runs in 2 replicas, but can be scaled to run in multiple replicas, thus highly available.
+
+## Update/Rollback Compatibility
+
+Existing VMs in clusters opting in with the feature gate will have to deploy network-resources-injector before upgrade so that new VMIs and migrated VMs will be able to consume custom network resources.
```
orelmisan: nit:
```suggestion
Existing VMs in clusters opting in with the feature gate will have to deploy network-resources-injector before upgrade so that new VMs and migrated VMs will be able to consume custom network resources.
```

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Implementation History
+
+<!--
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Graduation Requirements
```
orelmisan: In this case I suggest changing the title to "Deprecation Requirements".

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Alpha
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
```
orelmisan: How will this change be communicated?

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+### Proposed Architecture
+The new architecture decouples KubeVirt from NAD resources by leveraging the network-resources-injector as an external admission controller. This creates a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while network-resources-injector handles network resource injection.
+
+### Deployment alternatives 
+
+#### Option 1: Deprecate existing injection code and require manual deployment of network-resources-injector
+
+A new Feature Gate will be introduced named `DisableNADResourceInjection`, in preparation for NAD query code removal.
+If the FG is enabled, the VMI controller code that queries NADs (for secondary networks only) and populates custom resources in the virt-launcher pod template, will be skipped.   
+After 2-3 releases, the code that executes that will be removed.
```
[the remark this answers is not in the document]
orelmisan: Will the RBAC also be disabled?

veps/sig-network/decouple-nad.md
nirdothan: Done.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,183 @@
...
+## Graduation Requirements
+
+### Alpha
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources. 
+
+### Beta
+  Within one release, assuming `network-resources-injector` works as expected without failures in e2e testing, FG will be advanced to Beta. 
+
+### GA
+  Within one more release, if there's no or only little negative feedback from users, the feature will be graduated and the NAD code removed.
```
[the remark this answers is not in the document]
nirdothan: I wrote that although I don't what it means. Do we delete them?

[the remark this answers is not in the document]
orelmisan: Please see:
https://github.com/kubevirt/kubevirt/blob/7531f28db3a94679b881bb537a20794e46e847e6/pkg/virt-config/featuregate/inactive.go#L133

nirdothan:
[Removed](https://github.com/kubevirt/enhancements/compare/65dfa6d837ce832ca203f3000b6cbe4a12ca826a..eadc7577037667fb215ecd14774049fc977641b4) the log warning proposal, because it's not possible to reliably determine if the webhook is deployed or not.

veps/sig-network/decouple-nad.md
orelmisan: Communicating this deprecation via the mailing list gives the community time to adjust.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+Common use cases include SR-IOV, bridge network when [bridge-marker](https://github.com/kubevirt/bridge-marker) is used and [macvtap](https://github.com/kubevirt/macvtap-cni).
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM. It does so by querying the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.
+`NetworkAttachmentDefinition` is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to remove that query, in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.  
+In addition, the virt-controller and virt-operator role bindings that currently allow it to access the NAD can be reduced in scope.
```
orelmisan:
```suggestion
In addition, the virt-controller and virt-operator ClusterRoles that currently allow it to access the NAD will be reduced.
```

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Motivation
+
+The replacement has the architectural advantage of decoupling KubeVirt from an external dependency by simplifying code maintenance and supporting future replacement technologies.
+Security will be improved by removing access permissions to the external CRD.
```
orelmisan: ...from KubeVirt's control plane components.

veps/sig-network/decouple-nad.md
orelmisan: Please resolve this thread.

veps/sig-network/decouple-nad.md
nirdothan: This means that the service accounts that are used in virt-controller/operator will have fewer permissions.
There is no such thing as "virt-controller and virt-operator ClusterRoles" those are RoleBindings.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+Common use cases include SR-IOV, bridge network when [bridge-marker](https://github.com/kubevirt/bridge-marker) is used and [macvtap](https://github.com/kubevirt/macvtap-cni).
+
+Currently, virt-controller queries the NAD CR for every secondary network in every VM. It does so by querying the API server each time it templates a new virt-launcher pod, which occurs during initial VM startup and every migration.
+`NetworkAttachmentDefinition` is an external CRD maintained by the [k8s network plumbing WG](https://github.com/k8snetworkplumbingwg). The mentioned query is the only direct dependency that KubeVirt production code has on this external project.
+This proposal aims to remove that query, in favor of an external replacement that will map the custom resources to the virt-launcher pod instead.  
+In addition, the virt-controller and virt-operator role bindings that currently allow it to access the NAD can be reduced in scope.
+
+## Motivation
+
+The replacement has the architectural advantage of decoupling KubeVirt from an external dependency by simplifying code maintenance and supporting future replacement technologies.
```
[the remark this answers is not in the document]
nirdothan: Exactly. Therefore it is the motivation.

veps/sig-network/decouple-nad.md
nirdothan: Done.

nirdothan:
[Pushed changes](https://github.com/kubevirt/enhancements/compare/c68df4c2829b35a80ab0d5074540e6efd9f9d4a3..4120197edcc7aba0f991e08919176e94d15ab0ff) requested by @orelmisan.

[Added code references](https://github.com/kubevirt/enhancements/compare/4120197edcc7aba0f991e08919176e94d15ab0ff..1fa954fd0b707d66e646b71b41f35b6fbf597b1b) requested by @xpivarc.

nirdothan:
[Changes](https://github.com/kubevirt/enhancements/compare/1fa954fd0b707d66e646b71b41f35b6fbf597b1b..9eb193d2160c78f3795770391e508156d8168887) from offline review with @orelmisan.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Expected virt-launcher Pod (After network-resources-injector)
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: virt-launcher-testvm-xxxxx
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[{"name":"sriov-net-attach-def","namespace":"default"}]'
```
xpivarc: Annotate that this is what KubeVirt specifies and that resources below are from injector

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+    resources:
+      requests:
+        intel.com/sriov_net: "1"
+      limits:
+        intel.com/sriov_net: "1"
+```
+
+## Scalability
+
+The network-resources-injector approach provides several scalability advantages over the current implementation:
```
xpivarc: Do we have any numbers to back this up?

xpivarc:
@alromeros Does this conflict with https://github.com/kubevirt/enhancements/issues/41 in any way?

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Motivation
+
+This replacement provides the architectural advantage of decoupling KubeVirt from an external dependency and simplifying code maintenance.
+Security will be improved by removing access permissions to the external CRD from KubeVirt's control plane components.
+In addition, a performance improvement can be gained from using a project that caches NAD data, thus reducing overall number of API calls.
+
+## Goals
+
+- KubeVirt will no longer be aware of NetworkAttachmentDefinition objects.
```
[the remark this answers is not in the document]
nirdothan: Yes. There's no current plan to reduce dependency on multus (although IMHO ideally there should be).

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Design
+
+### Current Architecture
+
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements (`k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limits.
+
+### Proposed Architecture
+
+Instead of [querying the NAD directly](https://github.com/kubevirt/kubevirt/blob/release-1.7/pkg/network/multus/nad.go#L52) and [processing it in the VMI controller](https://github.com/kubevirt/kubevirt/blob/release-1.7/pkg/virt-controller/services/template.go#L370), an existing external component, [network-resources-injector](https://github.com/k8snetworkplumbingwg/network-resources-injector), can perform the same operation.
```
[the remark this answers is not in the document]
nirdothan: Done.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+This component includes a mutating webhook that automatically mutates Pods, upon creation, to inject network resource requirements from NetworkAttachmentDefinition annotations into pod resource requests and limits.
+network-resources-injector, although a standalone project, is also part of the [sriov network operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator/tree/v1.6.0/bindata/manifests/webhook). It is known to run successfully in KubeVirt clusters that already include the `sriov network operator`, 
+even though it does not mutate virt-launcher pods, since the existing KubeVirt code pre-populates the pod template with the resource requests before the admission webhook operates.
+This architecture decouples KubeVirt from NAD resources, creating a cleaner separation of concerns where KubeVirt focuses on VM lifecycle management while the admission controller handles network resource injection.
+
```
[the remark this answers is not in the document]
nirdothan: Good point. Added a label to virt-launcher pod template for VMs with secondary networks. We will provide instructions in documentation how to manually configure MutatingWebhookConfiguration to only query relevant pods,  and implement it that way in HCO (outside the scope of this VEP).

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+After 2-3 releases, the code that executes NAD querying will be removed along with the code that deploys associated RBAC rules (applies to both options below).
+
```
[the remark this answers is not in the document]
nirdothan: Yes. CNI/DevicePlugin will fail to allocate the required resource thus failing pod creation.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+##### Cons
+
+- Breaks backward compatibility 
+
+#### Option 2: Use virt-operator to deploy network-resources-injector
+
+The network-resources-injector repo includes all the source code, build infrastructure and deployment yamls required for successful deployment of the operator.
+Built server images are available in ghcr.io, e.g. ghcr.io/k8snetworkplumbingwg/network-resources-injector:v1.8.0
+The image will be cloned and published to the official kubevirt public repo.
```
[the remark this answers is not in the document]
nirdothan: Done.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+-->
+
+## Deprecation Requirements
+
+### Alpha
+
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+
+### Beta
```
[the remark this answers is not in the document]
nirdothan: Good idea. Added.

veps/sig-network/decouple-nad.md
nirdothan: Sorry no.

veps/sig-network/decouple-nad.md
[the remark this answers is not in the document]
orelmisan: @xpivarc, which warning did you have in mind? Following which operations?

veps/sig-network/decouple-nad.md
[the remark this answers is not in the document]
orelmisan: @nirdothan, please note that adding this label would require the implementation of additional logic to support hotplug scenarios.
I am not in favor of adding it.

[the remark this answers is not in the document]
nirdothan: That's a great idea! will try that out tomorrow,

vladikr (commented):
My main trouble with this proposal is the new component's optional deployment.

Critically, while optional deployment sounds principled (modular and flexible), it optimizes for developers at users' expense.
Senodery networks are a core KubeVirt feature, why are we ok with breaking it by not forcing a hard dependency on the new component? 

From the user-first perspective, forcing virt-operator to deploy the network-resources-injector is the better path. It ensures "it just works" without breaking modularity.  This is prioritizing reliability and UX, reducing the burden of manual deployment. The proposal's rejection of this feels overly cautious.

Other than this, I think it's a great solution :)

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,202 @@
...
+- https://github.com/kubevirt/kubevirt
+- https://github.com/kubevirt/kubevirtci
+
+## Design
+
+### Current Architecture
+
+KubeVirt currently has a tight coupling with NetworkAttachmentDefinition (NAD) resources. When a VM with secondary networks is launched, the virt-controller directly queries NAD CRs to extract custom resource requirements (`k8s.v1.cni.cncf.io/resourceName` annotation) and injects them into the virt-launcher pod specification, as resource requests and limits.
+
+### Proposed Architecture
```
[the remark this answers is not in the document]
orelmisan: As you've demonstrated in an offline discussion, there is a minor difference in the user-facing error message:
Before:
```yaml
Status:
  Conditions:
    Last Probe Time:       2025-11-27T11:53:10Z
    Last Transition Time:  2025-11-27T11:53:10Z
    Message:               virt-launcher pod has not yet been scheduled
    Reason:                PodNotExists
    Status:                False
    Type:                  Ready
    Last Probe Time:       <nil>
    Last Transition Time:  2025-11-27T11:53:10Z
    Message:               failed to render launch manifest: failed to locate network attachment definition default/ptp-conf
    Reason:                FailedCreate
```

After:
```yaml
Status:
  Conditions:
    Last Probe Time:       2025-11-27T12:32:51Z
    Last Transition Time:  2025-11-27T12:32:51Z
    Message:               virt-launcher pod has not yet been scheduled
    Reason:                PodNotExists
    Status:                False
    Type:                  Ready
    Last Probe Time:       <nil>
    Last Transition Time:  2025-11-27T12:32:51Z
    Message:               failed to create virtual machine pod: admission webhook "network-resources-injector-mutating-config.k8s.cni.cncf.io" denied the request: could not find network attachment definition 'default/ptp-conf': could not get Network Attachment Definition default/ptp-conf: the server could not find the requested resource
```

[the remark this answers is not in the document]
nirdothan: Added to the VEP.

nirdothan:
[Added](https://github.com/kubevirt/enhancements/compare/a048a8ad1dcf8a351f262acb24c8ce720add4d96..1d187b3f50bcc25f552a19f17b8a7b1371132e55) reference to https://github.com/kubevirt/kubevirt/issues/14615.
Indicated difference in error messages.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,231 @@
...
+- Allows for opinionated configuration, e.g. optimize to process only virt-launcher pods. 
+
+##### Cons
+
+- Virt-operator currently only deploys internal components and does not reconcile 3rd party tools. Letting it deal with external components may architecturally be a slippery slope that may furthermore contribute to blurring responsibilities between KubeVirt, [CNAO](https://github.com/kubevirt/cluster-network-addons-operator) and [HCO](https://github.com/kubevirt/hyperconverged-cluster-operator).
+It will be hard to reason why virt-operator should deploy network-resources-injector, but not other network dependencies such as multus, sriov-operator/device-plugin, various CNIs etc.
+- Adds a significant maintenance burden to KubeVirt project.
+- NAD related RBAC rules remain in virt-operator though bound to network-resources-injector.
+- The objective of decoupling is not achieved.
+
```
orelmisan: NRI may fail in case NAD CRDs are not deployed as it uses a NAD informer.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,231 @@
...
+## Scalability
+
+The network-resources-injector approach provides several scalability advantages over the current implementation:
+
+- Caching of NADs avoids redundant API calls.
+- By default runs in 2 replicas, but can be scaled to run in multiple replicas, thus highly available.
+
+## Update/Rollback Compatibility
+
+Existing VMs in clusters opting in with the feature gate will have to deploy network-resources-injector before upgrade, so that new VMs and migrated VMs will be able to consume custom network resources.
```
orelmisan:
```suggestion
Existing VMs in clusters opting in with the feature gate will must deploy network-resources-injector before upgrade, so that new VMs and migrated VMs will be able to consume custom network resources.
```

veps/sig-network/decouple-nad.md
nirdothan: Yes!

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,231 @@
...
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Deprecation Requirements
+
+### Alpha
+
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+  A warning will be issued by the VM/VMI API in case secondary networks exist and FG is not enabled.
```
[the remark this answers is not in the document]
nirdothan: Validation webhook added,

nirdothan:
> IMHO there are two disparate goals being tackled here - captured in the goal section.
> 
> However, the end user only benefits from one of them - reducing the number of API calls.
> 
> Thus, I wonder why a simpler option - just using a NAD informer in the kubevirt code - was not evaluated.

The option was evaluated but was less favored by the SIG since:
- It achieves the performance goal at the price of deepening our coupling with NAD.
- We're not in favor of developing something that has already been developed, battle tested  and maintained for a long time. Let alone when the end of life of multus and NAD is already in sight.

maiqueb:
> > IMHO there are two disparate goals being tackled here - captured in the goal section.
> > However, the end user only benefits from one of them - reducing the number of API calls.
> > Thus, I wonder why a simpler option - just using a NAD informer in the kubevirt code - was not evaluated.
> 
> The option was evaluated but was less favored by the SIG since:
> 
>     * It achieves the performance goal at the price of deepening our coupling with NAD.
> 
>     * We're not in favor of developing something that has already been developed, battle tested  and maintained for a long time. Let alone when the end of life of multus and NAD is already in sight.

IMHO all of this should be captured in the document, linking the original information sources.

nirdothan:
> > > In [your reply](https://github.com/kubevirt/enhancements/pull/107#issuecomment-3650547739) you are saying (and I quote): "Let alone when the end of life of multus and NAD is already in sight."
> > > I'm asking you to link the source of that information. I.e. where was sun-setting multus decided, voted, and in which date will that take place.
> > 
> > 
> > The stated goal of this proposal is to decouple KubeVirt from NAD. Adding new NAD-dependent logic—particularly logic that must account for non-Multus clusters—reintroduces the very coupling this change is intended to remove, and therefore defeats the objective.
> 
> I don't see the relationship between the request I made (links about multus sun-set) and this reply.
> 
> FWIW, out of the 2 goals in this proposal, I only "buy" one.

Meaning you are against decoupling? It's important for me to understand.

xpivarc:
@EdDev Please remove hold when you are happy with this.

veps/sig-network/decouple-nad.md
EdDev: Can this be resolved?

veps/sig-network/decouple-nad.md
EdDev: Please resolve.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,229 @@
...
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Deprecation Requirements
+
+### Alpha
+
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+  A warning will be issued for the VM API in case secondary networks exist and FG is not enabled. A validation webhook will be implemented to issue the warning at VM creation.
```
EdDev: I suggest the warning to be introduced in the Beta stage. Alpha is too early.
If this jumbs to Beta directly (with proper reasoning), it makes sense.

EdDev (commented):
Thank you for this decoupling suggestion, we waited a long time for this one.

veps/sig-network/decouple-nad.md
nirdothan: I'm not sure what to do with this comment.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Implementation History
+
+<!--
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Deprecation Requirements
```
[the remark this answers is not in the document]
nirdothan: It was already there: https://github.com/nirdothan/enhancements/blob/979c54173af873249c91c343f979a02c08ece4ee/veps/sig-network/decouple-nad.md?plain=1#L81

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,229 @@
...
+
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+  A warning will be issued for the VM API in case secondary networks exist and FG is not enabled. A validation webhook will be implemented to issue the warning at VM creation. 
+
+### Beta
+
+  Assuming option 1 is selected, Beta stage will be skipped as it entails no associated code changes.
+
+### Final Deprecation
```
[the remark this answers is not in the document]
nirdothan: Done. Please see if it's clear enough now.

nirdothan:
[Changes](https://github.com/kubevirt/enhancements/compare/39a0842ccb98968eaf29aad0c7ad122c5bd56f73..15f9eed8b75bf7e5f3157c5b6fe44573e69e0165) in lifecycle phases following offline conversation with @EdDev re https://github.com/kubevirt/enhancements/pull/107#discussion_r2642202544.

veps/sig-network/decouple-nad.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Alpha
+
+  Will be introduced in KubeVirt v1.8, protected by the `DisableNADResourceInjection` FG. If the FG is set, KubeVirt will **not** map custom-resources.
+  Documentation will introduce the FG, highlight the deprecation and reference the network-resources-injector installation instructions.
+  A warning will be issued for the VM API in case secondary networks exist and FG is not enabled. A validation webhook will be implemented to issue the warning at VM creation. 
+
+### Beta
+
+  Assuming option 1 is selected, Beta stage will be skipped as it entails no associated code changes.
```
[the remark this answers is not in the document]
EdDev: > @iholder101 Re:
> 
> > we plan to enable beta feature by default soon
> 
> IIUC, It's not that all beta FGs will be enabled by default, but rather that we have the ability to enable them by default.

This is my expectation. Hopefully it will be a FG specific thing.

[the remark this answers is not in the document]
iholder101: This hasn't been determined yet and demands more discussion, but IMO it should be on a per-FG basis like @EdDev suggested but the policy should be that one requires an exception in order to not enable a beta FG by default and it would block GA in the sense that a GA requirement would be to enable the beta feature by default for at least one cycle. All of this should not apply retroactively for old feature gates.

Generally speaking, I think there's almost no reason to not enable an FG by default on beta, and that in these cases it should stay in alpha.

veps/sig-network/decouple-nad.md
nirdothan: Addressed in [this change](https://github.com/kubevirt/enhancements/compare/15f9eed8b75bf7e5f3157c5b6fe44573e69e0165..60692ff9fdbd7a24df3b987ee48fca32b1582279).

veps/sig-network/decouple-nad.md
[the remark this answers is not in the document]
EdDev: I think the original intention was to have it disabled by default (like today) and then go directly to GA.
As the functionality here has already field experience, I think it is safe for Alpha skipping. Although, if the idea was to disable it first and then enable it by default later, Alpha could be left here.

I think the only reason to leave the Beta with default enabled is for a quick user revert in case anything is found problematic. It mainly helps with gradual progressing here.

For now, I am fine with the last state described here.
If we will see problems with this, we can change it to Alpha in final implementation.

@iholder101 , thank you for the feedback!

EdDev (approved):
Looks great, thank you!

Issue #109, opened 2025-10-27
VEP 108: Implement vGPU Enabled Live Migration


csomani1:
**Primary contact (assignee)**:

/assign @csomani1

**Current Feature Stage**:

Alpha

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

xpivarc:
Reviewing this VEP against 1.9 as it missed 1.8:

## Alpha

* Implement basic functionality and testing.
* Limitations
    * Users must ensure all worker nodes have identical GPU driver versions since KubeVirt will not take this into account when scheduling the migration
    * KubeVirt is unable to estimate the maximum period for the migration. Use a hard limit that is equal to the existing   calculated values (which ignore gpu info)
* Figure out how to handle any data loss during the migration.

From these deliverables I am missing the testing part, @Aseeef is this correct? And as well have we progressed with the data loss?

Further for Beta (1.10?) we will need Docs and update the VEP.

Aseeef:
I'm not too familiar with what is meant by "Figure out how to handle any data loss during the migration" to be honest. The PR implementation did have basic unit tests. This VEP does need to be updated to include:
- e2e testing is a beta requirement
- clarify requirements on "KubeVirt will take driver version into account when scheduling the migration" as per https://docs.nvidia.com/vgpu/latest/grid-vgpu-release-notes-red-hat-el-kvm/index.html#vgpu-migration-support.
- update alpha to be 1.9

I'll try get to this next week.

Barakmor1:
> I'm not too familiar with what is meant by "Figure out how to handle any data loss during the migration" to be honest. The PR implementation did have basic unit tests. This VEP does need to be updated to include:
> 
> * e2e testing is a beta requirement
> * clarify requirements on "KubeVirt will take driver version into account when scheduling the migration" as per https://docs.nvidia.com/vgpu/latest/grid-vgpu-release-notes-red-hat-el-kvm/index.html#vgpu-migration-support.
> * update alpha to be 1.9
> 
> I'll try get to this next week.

Hey @Aseeef any updates regarding this VEP?

Aseeef:
[...]

PR #110, opened 2025-10-27
VEP 109: Implement vGPU Enabled Live Migration

csomani1:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/109
**SIG label**: /sig compute

### What this PR does
Allow VMs with an mdev vGPU to live migrate

### DIFF
--- a/veps/sig-compute/vgpu_live-migration.md
+++ b/veps/sig-compute/vgpu_live-migration.md
@@ -1,0 +1,122 @@
+# VEP #109: Implement vGPU Enabled Live Migration
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [] (R) Target version is explicitly mentioned and approved
+- [X] (R) Graduation criteria filled
+
+## Overview
+
+This is a proposal to allow live migrations in KubeVirt to work for VMs with a single NVIDIA vGPU, exposed by mdev, between two nodes in the same cluster with identical GPUs and GPU drivers.
+
+## Motivation
+
+GPU usage is increasing with more and more companies running AI workloads, so companies are now requesting live migration to support GPU enabled VMs.
+
+## Goals
+
+* Address a common live migration problem where the target needs to update the destination Libvirt XML. In the case of mdevs, it needs to update the mdev UUID in the XML.
+* Support single vGPU enabled live migrations for both nodes that are using the Nvidia GPU Operator and clusters that are using KubeVirt’s generic device plugin for mdev.
+* Support single vGPU enabled live migrations with minimal data lost due to high dirty rates.
+
+## Non Goals
+
+* Do not want to change the live migration workflow for non vGPU enabled VMs.
+* Do not support live migration for passthrough or SRIOV vGPU
+* Do not support cross-cluster live migrations
+* Do not support live migrations for VMs with multiple vGPUs
+
+## Definition of Users
+
+* **KubeVirt Administrators:** Users who have cluster wide privileges to trigger APIs to manage a cluster.
+* **KubeVirt Owner:** VM workload owners who want high availability for their VMs.
+
+## User Stories
+
+As a KubeVirt admin/owner, I want to be able to live migrate my VMs that have an NVIDIA vGPU.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+For Alpha, GPU driver versions on all worker nodes must be identical. The migration will not be successful if there is a version mismatch, so users must ensure this. This will be addressed and updated during Beta.
+
+[VEP 141](https://github.com/kubevirt/enhancements/issues/141) introduces a feature gate in KubeVirt, TargetSideMigrationHooks, to register and write QEMU hooks for the target `virt-launcher`. We will use this new infrastructure to mutate the domain XML with the updated mdev UUID, which will be the one assigned to the target `virt-launcher` by `gpu.CreateHostDevices()` in `manager.go`. VGPU live migration will only be available with the TargetSideMigrationHooks feature gate enabled. 
+
+Once the destination XML contains the correct fields, the live migration can begin. Libvirt/QEMU already support vGPU live migration for mdev (since Libvirt 8.6.0 and QEMU 8.1.0) and will do the actual migration, so no further work is needed by KubeVirt to migrate the vGPU. Some migration configs at the Libvirt/QEMU level, such as the migration method or downtime limit, may be necessary however.
+
+### Example 
+XML snippet before hook:
+```
+<hostdev mode='subsystem' type='mdev' managed='no' model='vfio-pci' display='on' ramfb='on'>
+      <source>
+        <address uuid='bb4a98d8-60c1-40c6-b39b-866b1e82bd8c'/>
+      </source>
+      <alias name='ua-gpu-gpu1'/>
+      <address type='pci' domain='0x0000' bus='0x09' slot='0x00' function='0x0'/>
+    </hostdev>
+```
+
+XML snippet after hook (address uuid updated):
+```
+<hostdev mode='subsystem' type='mdev' managed='no' model='vfio-pci' display='on' ramfb='on'>
+      <source>
+        <address uuid='05b59010-d19c-47d2-9477-33b4579edc90'/>
+      </source>
+      <alias name='ua-gpu-gpu1'/>
+      <address type='pci' domain='0x0000' bus='0x09' slot='0x00' function='0x0'/>
+    </hostdev>
+```
+
+**Failed migrations:** Cleanup will be performed by existing code and by code introduced in [16212](https://github.com/kubevirt/kubevirt/pull/16212).
+
+## API Examples
+
+N/A
+
+## Alternatives
+
+Instead of relying on a QEMU hook, a Libvirt API could be introduced to allow KubeVirt to update the destination XML at the start of migration via callbacks. However, previous discussions asking for this API haven’t made progress.
+
+## Scalability
+
+The unix socket used will be `/var/run/kubevirt/migration-hook-socket` introduced in PR [16212](https://github.com/kubevirt/kubevirt/pull/16212). A target `virt-launcher` pod will have at most one of this socket open at a time, so it should be possible to live migrate a large number of VMs concurrently without significant performance issues. KubeVirt also imposes its own limitations on the number of live migrations on a node and cluster-wide level.
+
+## Update/Rollback Compatibility
+
+* Needs TargetSideMigrationHooks feature gate from PR [16212](https://github.com/kubevirt/kubevirt/pull/16212) to be enabled
+* Will be safe during upgrades as long as the newer node's mdev uuids don't change unexpectedly.
+
+## Functional Testing Approach
+
+* Unit tests: Verify that the VM is able to live migrate with the vGPU given the proper conditions.
+* [Optional] Also verify that this works with the NVIDIA GPU Operator.
+
+## Implementation History
+
+N/A
+
+## Graduation Requirements
+
+### Alpha
+
+* Implement basic functionality and testing.
+* Limitations
+    * Users must ensure all worker nodes have identical GPU driver versions since KubeVirt will not take this into account when scheduling the migration
+    * KubeVirt is unable to estimate the maximum period for the migration. Use a hard limit that is equal to the existing   calculated values (which ignore gpu info)
+* Figure out how to handle any data loss during the migration.
+
+### Beta
+
+* No longer require users to ensure all worker nodes have identical GPU driver versions. KubeVirt will take driver version into account when scheduling the migration
+* Find a way to estimate the maximum period for the migration 
+* Needs [VEP 141](https://github.com/kubevirt/enhancements/issues/141) to be in Beta.
+
+### GA
+
+* Needs [VEP 141](https://github.com/kubevirt/enhancements/issues/141) to be in GA.


veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+In order to live migrate a vGPU, the destination XML needs to be updated with the mdev UUID that Kubevirt allocates on the target virt-launcher. To do this, we introduce a QEMU hook that interacts with the target `virt-launcher` pod.
+
+This will be the flow:
```
vladikr: I wonder if you could describe a bit more about how this mechanism will work. 
If I understand correctly, the idea is to rely https://libvirt.org/hooks.html#qemu-guest-migration

Is the idea to completely bypass the existing [migratableDomXML](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/live-migration-source.go#L198) ?
If not , then how will the approaches not collide?

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,91 @@
...
+
+1.  **`virt-launcher` (Target):** Target `virt-launcher` creates a vGPU `api.HostDevice`, containing some fields including the mdev UUID.
+2.  **`virt-launcher` (Target):** Target `virt-launcher` listens on the `"/tmp/kube-migration-hook-socket-{unique_id}"` unix socket. At this point, the socket hasn’t been opened yet, so use a Go func so that the rest of the migration code can execute.
+3.  **QEMU hook:** During the "migrate begin" phase on the target, the QEMU hook will open the `"/tmp/kube-migration-hook-socket-{unique_id}"` unix socket and pass the destination XML.
+4.  **`virt-launcher` (Target):** Target `virt-launcher` accepts the connection and receives the destination XML. It then updates the XML with the target’s mdev UUID and writes the XML to the socket.
+5.  **QEMU hook:** The hook receives the updated XML and outputs to stdout to complete the hook.
```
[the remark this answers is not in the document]
csomani1: added

Barakmor1:
Hey @csomani1 ,
I created [this PR](https://github.com/kubevirt/kubevirt/pull/16212) as a first version / PoC to move source-based XML modifications to the target using libvirt hooks.

The goal is to make target-side modifications extendable and easier to add in the future.
This can also support your VEP by providing a simple way to handle XML changes and avoid having to support both source- and target-side modifications.

I’d love your feedback and happy to discuss how this could work with your VEPs.

rthallisey (changes requested):

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,94 @@
...
+## Design
+
+In order to live migrate a vGPU, the destination XML needs to be updated with the mdev UUID that Kubevirt allocates on the target virt-launcher. To do this, we introduce a QEMU hook that interacts with the target `virt-launcher` pod. Live migrations have multiple phases, and the QEMU hook is called at each phase. The hook will have conditional statements to only execute at
+the "prepare begin" phase. Hooks receive the phase and XML from stdin, and the updated XML is sent to stdout.
+
+Once the destination XML contains the correct fields, the live migration can begin. Libvirt/QEMU already support vGPU live migration for mdev and will do the actual migration, so no further work is needed by KubeVirt to migrate the vGPU. Some migration configs at the Libvirt/QEMU level, such as the migration method or downtime limit, may be necessary however.
+
+This will be the flow:
+
+1.  **`virt-launcher` (Target):** Target `virt-launcher` creates a vGPU `api.HostDevice`, containing some fields including the mdev UUID.
```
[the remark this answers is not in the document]
rthallisey: Also, nvidia's driver will require the same physical GPU model on both the source and target.  As well as the vGPU provide must match.  You can get this profile info in `/sys/.../mdev_supported_types/`.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,94 @@
...
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [] (R) Target version is explicitly mentioned and approved
+- [] (R) Graduation criteria filled
+
+## Overview
+
+This is a proposal to allow live migrations in Kubevirt to work for VMs with a single NVIDIA vGPU. The live migration will be supported between two nodes with identical, Kubevirt compatible GPUs that use mdev.
```
iholder101: Do we want to scope this VEP only to NVIDIA vGPUs?

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,94 @@
...
+* Support single vGPU enabled live migrations with minimal data lost due to high dirty rates.
+
+## Non Goals
+
+Do not want to change the live migration workflow for non vGPU enabled VMs.
+
+## Definition of Users
+
+**Kubernetes Administrators:** Users who have cluster wide privileges to trigger APIs to manage a cluster.
+
```
xpivarc: I would add workload owner / user of KubeVirt  as this persona might want HA for their VM.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,94 @@
...
+As a cluster admin, I want to be able to live migrate my VMs that have a vGPU.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+In order to live migrate a vGPU, the destination XML needs to be updated with the mdev UUID that Kubevirt allocates on the target virt-launcher. To do this, we introduce a QEMU hook that interacts with the target `virt-launcher` pod. Live migrations have multiple phases, and the QEMU hook is called at each phase. The hook will have conditional statements to only execute at
+the "prepare begin" phase. Hooks receive the phase and XML from stdin, and the updated XML is sent to stdout.
```
xpivarc: You can reference https://github.com/kubevirt/enhancements/issues/141

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,94 @@
...
+2.  **`virt-launcher` (Target):** Target `virt-launcher` listens on the `"/tmp/kube-migration-hook-socket-{unique_id}"` unix socket. At this point, the socket hasn’t been opened yet, so use a Go func so that the rest of the migration code can execute.
+3.  **QEMU hook:** During the "migrate begin" phase on the target, the QEMU hook will open the `"/tmp/kube-migration-hook-socket-{unique_id}"` unix socket and pass the destination XML.
+4.  **`virt-launcher` (Target):** Target `virt-launcher` accepts the connection and receives the destination XML. It then updates the XML with the target’s mdev UUID and writes the XML to the socket.
+5.  **QEMU hook:** The hook receives the updated XML and outputs to stdout to complete the hook.
+
+**Security:** Communication between the `virt-launcher` pod and QEMU hook should be limited to the unix socket, and the unix socket should be exclusively read and written to by the QEMU hook and `virt-launcher` pod.
+
+**Failed migrations:** If the migration fails, the unix socket should close, and cleanup will be performed by existing code.
```
xpivarc: Can't we simply close it once we reply with modified xml?

veps/sig-compute/vgpu_live-migration.md
csomani1: added

csomani1:
> > > I’d love your feedback and happy to discuss how this could work with your VEPs.
> > 
> > 
> > @Barakmor1 Thanks, I've added your PR as a reference in the design section. Registering the vGPU hook should be similar to the CPU hook that you alr added.
> 
> Thank you! Can we explicitly specify in the design that this feature is dependent on @Barakmor1's feature gate?

Added

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Graduation Requirements
+
+### Alpha
+
+Implement all functionality and testing.
+
+### Beta
+
+### GA
```
xpivarc: Please add explicit dependency on https://github.com/kubevirt/enhancements/issues/141

xpivarc (commented):
I think we are good to go for Alpha after these last adjustments.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+In order to live migrate a vGPU, we must first check that the host-driver versions for the GPU are the same on each node for compatability. Then, the destination XML needs to be updated with the mdev UUID that KubeVirt allocates on the target virt-launcher. To do this, we use a QEMU hook that interacts with the target `virt-launcher` pod. Live migrations have multiple phases, and the QEMU hook is called at each phase. The hook will have conditional statements to only execute at the "prepare begin" phase. Hooks receive the phase and XML from stdin, and the updated XML is sent to stdout.
+
+Once the destination XML contains the correct fields, the live migration can begin. Libvirt/QEMU already support vGPU live migration for mdev (since Libvirt 8.6.0 and QEMU 8.1.0) and will do the actual migration, so no further work is needed by KubeVirt to migrate the vGPU. Some migration configs at the Libvirt/QEMU level, such as the migration method or downtime limit, may be necessary however.
```
orelmisan: Can you please add an xml snippet of the information that would need to change in the domain?
Similar to what we have in https://github.com/kubevirt/enhancements/blob/main/veps/sig-network/ordinal-naming-scheme-upgrade-path.md?plain=1#L100 .

orelmisan (commented):
Thank you for the VEP @csomani1!

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,97 @@
...
+
+## User Stories
+
+As a KubeVirt admin/owner, I want to be able to live migrate my VMs that have a vGPU.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
```
EdDev: I am missing information on what is actually going to be changed in Kubevirt and how it interacts with the Nvidia GPU operator.
The expected information and protocol is important as Kubevirt should define the requirement for any GPU operator in order to allow integration. We should be open here for extensions and assure there is no vendor locking.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,97 @@
...
+* Address a common live migration problem where the target needs to update the destination Libvirt XML. In the case of mdevs, it needs to update the mdev UUID in the XML.
+* Support single vGPU enabled live migrations for both nodes that are using the Nvidia GPU Operator and clusters that are using KubeVirt’s generic device plugin for mdev.
+* Support single vGPU enabled live migrations with minimal data lost due to high dirty rates.
+
+## Non Goals
+
+* Do not want to change the live migration workflow for non vGPU enabled VMs.
+* Do not support live migration for passthrough or SRIOV vGPU
+* Do not support cross-cluster live migrations
+* Do not support live migrations for VMs with multiple vGPUs
```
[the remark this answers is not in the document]
csomani1: Not hard because I believe qemu supports multi vgpu migration

veps/sig-compute/vgpu_live-migration.md
csomani1: GPU operator providing mdev is the only req (for now until SRIOV is supported in kubevirt). 

Kubevirt's interaction with the NVIDIA GPU op and any other op will remain untouched by this VEP because the vgpu code only cares that an mdev is available to assign to a VM, doesn't matter how the mdev gets created whether by the generic dev plugin or a GPU op.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,125 @@
...
+    GPUDriverVersion string `json:"gpuDriverVersion,omitempty"`
+}
+```
+## Alternatives
+
+Instead of relying on a QEMU hook, a Libvirt API could be introduced to allow KubeVirt to update the destination XML at the start of migration via callbacks. However, previous discussions asking for this API haven’t made progress.
+
+## Scalability
+
+Each unix socket will be unique to allow concurrent live migrations. Linux can support a very large number of open sockets at the same time, so it should be possible to live migrate a large number of VMs concurrently without significant performance issues. KubeVirt also imposes its own limitations on the number of live migrations on a node and cluster-wide level.
```
orelmisan: Could you clarify which specific Unix socket is referred to here?

veps/sig-compute/vgpu_live-migration.md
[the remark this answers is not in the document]
EdDev: Thank you for the clarification.

IMO it will be more elegant not to have this as a non-goal then.
Maybe just as the current limitation.

But this is not something that worth too much attention.

veps/sig-compute/vgpu_live-migration.md
```diff
@@ -0,0 +1,125 @@
...
+
+As a KubeVirt admin/owner, I want to be able to live migrate my VMs that have an NVIDIA vGPU.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+One requirement for the live migration is the gpu driver versions on the source and target nodes must be identical. To check this, the source `virt-handler` will update the new GPUDriverVersion field in VirtualMachineInstanceMigrationState in `updateVMIStatus()` and the target `virt-handler` will compare that version with the version on the target node in `migration-target.go`. The gpu driver version can be extracted from `/proc/driver/nvidia/version` on the `virt-handler` pods.
```
EdDev: The introduction of a new `GPUDriverVersion` under `VirtualMachineInstanceMigrationState` is a bit controversial. This communication media is sub-optimal in general, but in this case, it seems not sufficient.

Per my discussion with @vladikr , we are missing here the scheduling part. On migration, we need to assure the scheduler will place the target pod on a node that has the correct version.
To do so, please consider using the mechanism we have with CPUs.

With CPUs, we have the `node-labeler` who adds labels to the node resource of available CPU types.
Then, the migration controller will examine the existing node the VM is running on, query the relevant labels from the node resource, and restrict the target pod using a selector, so the pod will be scheduled on a node that can serve the VM.

Can we use a similar mechanism with one tweak?
The GPU operator should be responsible to add the labels on the node resource, and that label needs to be agreed on with Kubevirt so the migration controller will know to read it and use it as a selector on the target pod.
If the label will be missing while there is a vGPU defined on the VM, the migration will not be initiated.

iholder101:
FYI @xpivarc @EdDev @orelmisan that this is the last day to merge this VEP for v1.8.

EdDev (changes requested):
@csomani1 , this VEP has been discussed in the sig-compute meeting today.

It has been agreed to accept this VEP with the following changes:
- Driver version requirement: Drop the existing proposal to pass this information from the source to the target node and add a limitation during Alpha stage that the driver versions must be identical on all worker nodes.
  - Beta stage should require this to be resolved. This VEP will need to be updated accordingly by then.
- Migration duration: Kubevirt is unable to estimate the maximum period for the migration. During Alpha stage, we accept this as a limitation and place a hard limit that is equal to the existing calculated values (which ignore gpu info).
  - Beta stage should require this topic to be addressed.

If you manage to update the VEP with the above, we will take in this VEP.
Please update us to the earliest, as the freeze date is today. We can extend for this specific one, by a day to let you update it, considering all the other details are accepted by the reviewers.

csomani1:
> Please update us to the earliest, as the freeze date is today. 

Added these changes

EdDev:
Thank you!

/unhold
/lgtm

Issue #108, opened 2025-10-27
VEP 108: Enable attestation for Confidential VMs (TDX)


MatiasVara:
**Primary contact (assignee)**:

/assign @MatiasVara 

**Current Feature Stage**:

Alfa

**Feature Gate**:

`EnableAttestation`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
    * [x] VEP PRs: 
      - https://github.com/kubevirt/enhancements/pull/113
      - https://github.com/kubevirt/enhancements/pull/243
    * [x] Code PRs:
      - https://github.com/kubevirt/kubevirt/pull/15958 (Draft)

> [!IMPORTANT]
> Please keep this description up to date.

Aseeef:
With regards to https://github.com/kubevirt/kubevirt/pull/15958 at least the following open questions remain answered:

1. How should we modify the VMI API (if at all) to indicate that a VM should require attestation? Alternatively, we can also simply have VM indicate that it requires the devices.kubevirt.io/qgs device.
  - Pros/Cons: Modifying VMI spec directly increases user-friendliness, allows building a more platform-agnostic API-spec for other future technologies.

2. Should we always try to mount the QGS socket unconditionally given we are running a TDX VM?
  - Pros/Cons: I don't see a harm in unconditionally attempting to mount the QGS Socket if it's there either from host -> virt-handler or from virt-launcher -> QEMU. Perhaps leave a log or error to indicate to the user if you want attestation, then you must explicitly enable it? I don't know, any one have any thoughts?

mythi:
> Personally, I think the best path forward here is to modify `socket_plugin.go` to not modify permissions (like it does now) and let this be the responsibility of whatever sets up the QGS socket.

+1. Theoretically there can also be non-kubevirt uses on the same node that uses the same socket. I think we just need to figure out how to configure the GID/Selinux labels correctly in places.

Two additional opens:
1. can we already set libvirt XML to enable `quote-generation-socket":{"type":"unix","path":"/var/run/tdx-qgx/qgs.sock"}}` for QEMU
2. do we need to worry about making `path` configurable

MatiasVara:
[...]

PR #113, opened 2025-10-28
VEP 108: Add proposal document about the enablement of attestation in Intel TDX

MatiasVara:
### VEP Metadata

**Tracking issue**:  #108 
**SIG label**: /sig computing

### What this PR does
This PR adds the proposal document for enabling the attestation of confidential guests that rely on Intel TDX.

### DIFF
--- a/veps/sig-compute/attestation-proposal.md
+++ b/veps/sig-compute/attestation-proposal.md
@@ -1,0 +1,168 @@
+# VEP #107: Enable attestation for Confidential VMs in Intel TDX
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+This proposal introduces attestation for Confidential VMs in Intel TDX.
+Attestation is the mechanism used by a Confidential VM to attest that it is
+running on confidential hardware (HW). For Intel TDX, attestation requires to
+deploy the Quote Generation Service (QGS) on the host that communicates with
+QEMU through a UNIX socket when a confidential VM wants to generate a blob. The
+VEP does not include managing or installing QGS components on the host but
+instead only proposes adding the required mechanisms to KubeVirt to utilize an
+existing QGS device socket. Bear in mind that SEV-SNP does not require a
+host-side service for the blob generation.
+
+## Motivation
+Organizations are now asking for workloads to include the strongest
+confidential computing security guarantees. Together with confidential VMs, we
+need to provide a way for confidential VMs to be attested. By doing so,
+confidential VMs can certify that they are running on confidential HW.
+
+## Goals
+- Enable attestation for confidential VMs on Intel TDX
+
+## Non Goals
+- Deployment of the Quote Generation Service (QGS)
+
+## Definition of Users
+- VM users: Deploy confidential workloads that require attestation.
+
+## User Stories
+- As a VM user, I want to deploy VMs on a platform with TDX protection and be
+  able to attest these VMs.
+- As a developer, I want to continue to be able to launch TDX VMs with out
+  setting up attestation for testing purposes.
+- As a cluster admin, I want to be to fail-fast by adding an addition check to
+  verify that QGS is setup correctly (i.e. the socket exists) before deploying
+  before deploying my TDX VMs.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+From the implementation perspective, we extend the KubeVirt CR spec to enable
+attestation:
+
+```xml
+configuration:
+  confidentialCompute:
+    tdx:
+      attestation:
+        enforced: true
+        qgsSocketPath: "/var/run/tdx-qgs/qgs.socket"
+```
+
+We add a new top-level section in configuration for all TEE(Trusted Execution
+Environment)-related configuration items named `confidentialCompute`. We add a
+new section for tdx. The attestation section contains two attributes:
+`enforced` and `qgsSocketPath`. When `enforced` is true, KubeVirt shall enforce
+that the QGS socket exists before scheduling any TDX VMs. The enforcement
+mechanism is discussed later. Next, `qgsSocketPath` is an optional attribute
+that indicates the path to the UNIX socket on the host.
+
+We introduce the TDX device plugin, using the resource `devices.kubevirt.io/tdx`,
+which runs on each TDX capable node and handles both TDX hardware capability
+advertisement and QGS socket mounting. This plugin replaces the current
+node-labeller for TDX detection. It relies on Kubernetes native extended
+resource scheduling for placement. The plugin checks for TDX hardware support
+by checking for the existance of a `tdx` key in `/sys/fs/cgroup/misc.capacity`
+(which is added by the kernel given TDX support). Moreover, the value associated
+with the `tdx` key inside `misc.capacity` also reports the maximum number of
+concurrent TDs supported by the hardware. Therefore, this file is also used to
+derive the device capacity of the device plugin.
+
+It should be noted that all mounts are unconditional. Mount points are created
+even if the QGS socket does not exist which allows attestation to function once
+the QGS socket is created even as the VM is running.
+
+The plugin's resource health is based on the cluster-wide
+confidentialCompute.tdx.attestation.enforced configuration:
+
+* If attestation is enforced: The resource is healthy only if the QGS socket is
+  present. If the socket is missing, it becomes unhealthy and enters a retry
+  loop. This acts as a scheduling gate by ensuring that: (a) TDX VMs requesting
+  the resource won't allocate until QGS is deployed. (b) If there are multiple
+  nodes in the cluster supporting TDX hardware, but only some have attestation
+  support via the QGS socket then the scheduler will only deploy TDX VMs to
+  nodes that have the QGS socket configured.
+* If attestation is not enforced: The resource is always healthy if TDX
+  hardware is available. QGS mounting is opportunistic - if the socket exists,
+  it's mounted; otherwise, the VM starts without attestation.
+
+TDX VMs automatically request the tdx resource in their VM spec to make sure
+that these are scheduled on the TDX capable nodes.
+
+## API Examples
+The below YAML snippets provide an example of how to enable attestation in the
+KubeVirt CR. In the default configuration, enforced is false and the
+qgsSocketPath is "/var/run/tdx-qgs/qgs.socket".
+
+```xml
+configuration:
+  confidentialCompute:
+    tdx:
+      attestation:
+        enforced: false
+```
+
+The below YAML snippets provide an example of how to enforce the enabling of
+attestation.
+
+```xml
+configuration:
+  confidentialCompute:
+    tdx:
+      attestation:
+        enforced: true
+        qgsSocketPath: "/var/run/qgs.socket"
+```
+
+## Alternatives
+- Request the `devices.kubevirt.io/qgs` resource in the VMI spec. We chose a
+  simpler approach where attestation is set up cluster-wide.
+- Use DRA instead of a device plugin.
+- Use a device plugin or an NRI plugin served by the QGS pod itself
+- Use a hostPath mount patching handled by a mutating webhook
+
+## Scalability
+
+## Update/Rollback Compatibility
+- All new fields are optional and disabled by default
+- This should not impact existing VMs
+
+## Functional Testing Approach
+- e2e tests to check that the blob from the guest point of view is correctly
+  generated.
+- Since TDX & SEV-SNP do not have support for nested virtualization this will
+  require bare metal hardware to conduct e2e testing.
+
+## Implementation Phases
+The initial phase of implementation will focus on integrating basic
+functionality. For example, a socket device plugin would be required to mount
+the UNIX socket into the virt-launcher.
+
+## Feature lifecycle Phases
+
+### Alpha
+The feature will be implemented in Alpha. We do not know if it will be possible
+to have e2e tests in Alpha due to lack of TDX hardware. We expect the feature
+to be merged without the e2e tests.
+
+### Beta
+We expect e2e tests in Beta. We expect the API to be stable. We need to decide
+if we keep the path to the UNIX socket as an attribute or we remove it in favor
+of a default location. Finally we need to re-evaluate whether there is any use
+case for enforcing the existance of QGS on the KubeVirt side (i.e use of the
+`attestation.enforced` field) and if so whether it makes sense to extend these
+enforcements to support per-VMI enforcements (as opposed to only cluster-wide).
+
+### GA
+Remove feature gate.


veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,104 @@
...
+## User Stories
+- As a Developer, I want to deploy VMs on a platform with TDX protection and be
+  able to attestate these VMs.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+From the implementation perspective, we extend the Kubevirt CR spec to enable
```
jean-edouard: Nit:
```suggestion
From the implementation perspective, we extend the KubeVirt CR spec to enable
```
More occurrences below.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Design
+From the implementation perspective, we extend the KubeVirt CR spec to enable
+attestation:
+
+```xml
+configuration:
+    attestation:
+      enabled: true
+```
```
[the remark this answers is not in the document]
vladikr: Yes, apologies. 

My thinking was that a TDX device plugin could advertise `devices.kubevirt.io/tdx` solely based on hardware support (replacing the node-labeller), and during allocation, it checks and mounts the QGS socket if it exists.

This way, we get opportunistic attestation without a separate plugin or cluster config. VMs are scheduled to any TDX-capable node, but only get QGS if deployed there. If we need to require QGS, we could add a config to make the allocation fail if the socket is missing.

I thought that the enforcement would come from the QGS service deployment. 

I see now that this isn't a good idea.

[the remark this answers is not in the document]
MatiasVara: > should use `attestation.enforced` instead of `enabled`

Sounds good. What is the difference between `enforced` and `enabled`?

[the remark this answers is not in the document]
Aseeef: > > should use `attestation.enforced` instead of `enabled`
> 
> Sounds good. What is the difference between `enforced` and `enabled`?

This is related to @vladikr's previous comment on https://github.com/kubevirt/enhancements/pull/113#discussion_r2581950966 where he proposed opportunistically always mounting QGS when it's there and only using the KubeVirt CR to "enforce" a requirement that the socket must be there before we can schedule. (So when enforced=false, if QGS is not there the VMI would still schedule, and if QGS is there, it will schedule AND opportunistically mount the socket if its there). `enabled=false` implies that QGS attestation functionality is disabled.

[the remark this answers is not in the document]
vladikr: > spec:
>   tdx:
>     attestation:
>       enforced: true
>       qgsSocketPath: "/var/run/qgs/qgs.sock"

do you mean `spec.configuration.confidentialCompute` and then the above ^^^ right?

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,125 @@
...
+The below YAML snippets provide an example of how to enable attestation in the
+KubeVirt CR. In the default configuration, enforced is False and the qgsSocket
+is "/var/run/tdx-qgs/qgs.socket".
+
+```xml
+configuration:
+    confidentialCompute:
+      tdx:
+        attestation:
+          enforced: {}
```
vladikr: Will it be a boolean or a struct?

vladikr:
@MatiasVara, overall, looks good, one question inline. 
But I think what's missing is the discussion we had around the device plugin's role (in [#discussion_r2581950966](https://github.com/kubevirt/enhancements/pull/113#discussion_r2581950966)). Specifically, details on how the TDX device plugin handles QGS detection, health checks for enforcement, and opportunistic mounting seem missing.

veps/sig-compute/attestation-proposal.md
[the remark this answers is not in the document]
MatiasVara: I did not get the question but if you refer to how qgs is deployed, we do not deal with that here. qgs may be deployed as a part of an operator.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,150 @@
...
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+This proposal introduces attestation for Confidential VMs. Attestation is the
```
Aseeef:
```suggestion
This proposal introduces attestation for Confidential Intel TDX VMs. Attestation is the
```

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,150 @@
...
+mechanism used by a Confidential VM to attest that it is running on
+confidential hardware (HW). The VEP does not include managing or installing QGS
+(Quote Generation Service) components on the host. It only references their
+presence. This service is required for the blob generation on confidential
+guests using Intel TDX. Bear in mind that SEV-SNP does not require a host-side
+service for the blob generation.
```
Aseeef: The part about "it only references their presence" can be clarified.
How about:
```
mechanism used by a Confidential VM to attest that it is running on
confidential hardware (HW). Attestation in TDX requires access to a 
user space application that lives on the host known as the "Quote 
Generation Service" (QGS). This service is required for the blob 
generation on confidential guests using Intel TDX. Bear in mind 
that SEV-SNP does not require a host-side service for the blob 
generation. The VEP does not include managing or installing QGS
components on the host but instead only proposes adding the neccesary
mechanisms to KubeVirt to utilize an existing QGS device socket.
```

Aseeef (changes requested):

Aseeef:
Besides the feedback mentioned above, looks good to me!

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,150 @@
...
+When enforced is true, KubeVirt shall require the socket to exist or it fails.
+When enforced is false, KubeVirt shall try to mount the UNIX socket only if it
+exists. If the socket does not exist, KubeVirt shall not fail. qgsSocketPath is
+an optional attribute that indicates the path to the UNIX socket. We implement
+a device plugin that mounts the UNIX socket depending on the enforced
+attribute.
+
+We introduce the TDX device plugin, i.e., `devices.kubevirt.io/tdx`, which runs
+on nodes and handles both TDX hardware capability advertisement and QGS socket
+management. This plugin replaces the current node-labeller for TDX detection.
```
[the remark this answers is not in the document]
Aseeef: Ok. The PR looks fine from my PoV, I think this is something we can continue to discuss  the fine details off as we implement this.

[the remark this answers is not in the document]
zhencliu: It might be problematic to use a single resource (e.g., devices.kubevirt.io/tdx) to represent both capacity and the QGS socket device.Consider this scenario: 
Even if attestation.enforced is set to true, a user can still start a TDX VM without requiring attestation (Is this assumption correct?). If the QGS socket then becomes unavailable for some reason, the single resource representing both capacity and the socket would also become unhealthy and disappear.
Consequently, a TDX VM that does not require attestation would also be unable to start.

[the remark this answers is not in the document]
zhencliu: > What is the danger of relying only on `/sys/fs/cgroup/misc.capacity` to see whether or not it has a field called "tdx" in there to determine whether on not TDX is configured? If I understand, the kernel only sets this field if TDX is properly enabled right?

It's safe for TDX, but it may not work for SNP, the content is something like the following on a SNP enabled host:
sev: 509
sev-es: 99
No 'snp' appears in the misc.capacity, because snp and sev-es share the capacity 99, so the device plugin cannot tell if snp is enabled or not on the host, it relies on the approach that the node labeller does


> * Only create the devices.kubevirt.io/tdx device plugin if TDX is enabled
> * When `attestation.enforced = true` the device plugin is configured to fail health checks when QGS socket is not present (effecting VM scheduling). When `attestation.enforced = false`, the device plugin will never fail health checks so you can always acquire the device plugin. In this case, device plugin will still attempt to mount the QGS socket if its there (opportunistic).
> * Number of devices.kubevirt.io/tdx devices = grep tdx /sys/fs/cgroup/misc.capacity.

Thanks, I see, a socket device plugin is used for ' devices.kubevirt.io/tdx ' and the max number of devices is exactly the capacity. Then for SNP it's quite different, a generic device plugin can work for SNP capacity instead of the socket device plugin, so we may need  to handle TDX/SNP in different way.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,152 @@
...
+It relies on Kubernetes native extended resource scheduling for placement. The
+plugin checks for TDX hardware support in the same way node-labeller used to
+do. If it is supported, it advertises it as a resource:
+`devices.kubevirt.io/tdx`. During resource allocation the plugin probes for the
```
Aseeef: One more point mentioned in the confidential computing meeting which doesn't seem to be mentioned here yet but we should converge on soon:

If we use device plugins, one natural consideration is what should the count of this device plugin be? Similar to https://github.com/kubevirt/kubevirt/pull/16262, the answer seems to be to read `/sys/fs/cgroup/misc.capacity`.

Another point brought up with regard to this strategy is: What if something other than KubeVirt also creates a TDX VM? In that case, some of the VM capacity would be in use but unaccounted for in our device plugin.

One general way to resolve this to have your device plugin create a watcher in `/sys/fs/cgroup/misc.current`. The device plugin can track how much of that in-use capacity is from KubeVirt. Then for each in-use capacity not coming from KubeVirt's device plugin we mark one of the device plugins as "unhealthy".

Example:
* There are 32 TDX VMs you can have in total.
* KubeVirt "ownes" 10 of those VMs.
* But according to `/sys/fs/cgroup/misc.current` there are 11 TDX VMs currently in use.
* This means 1 of those VMs was NOT created by KubeVirt.
* Thus, KubeVirt can mark 1 of its 32 TDX VMs inside its device plugin as "unhealthy".

While this approach is mostly fool-proof, there is a small window where you can have a race condition of both KubeVirt and this external service both try to allocate a confidential VM. However, in practice, we don't expect this to be a problem.

Finally, since we would like to have first class compatibility with Kata containers, we also discussed potentially converging with Kata on the approach to take for tracking capacity. For example, maybe we can share a device plugin with Kata?

vladikr: @Aseeef Thank you. Yes, this is true that we need to consider the co-existence with kata in terms of device counting.
However, the situation in this design is not worse than it is today with simple node labels.
From a design perspective, we know that we need a device plugin, but how it will be implemented and how we will avoid collisions belongs to the implementation, from my POV. 
This doesn't have to be perfect as we are going to alpha and will learn and improve as during the development. 
I think your idea https://github.com/kubevirt/kubevirt/pull/16262/files#r2623295882 is a great start.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+        attestation:
+          enforced: true
+          qgsSocketPath: "/var/run/tdx-qgs/qgs.socket"
+```
+
+We add a new top-level section in configuration for all TEE-related
+configuration items named confidentialCompute. We add a new section for tdx.
+The attestation section contains two attributes: enforced and qgsSocketPath.
+When enforced is true, KubeVirt shall require the socket to exist or it fails.
+When enforced is false, KubeVirt shall try to mount the UNIX socket only if it
```
mythi: Can you clarify what is the KubeVirt component that "fails"?  If the device plugin does not register any resources, the virt-launcher pods stay in Pending state since nodes have no resources. 

I think the only logic here should be to skip adding the device plugin resource request.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+virt-launcher pod as a volume.
+
+The plugin's health is based on the cluster-wide
+confidentialCompute.tdx.attestation.enforced configuration:
+
+* If attestation is enforced: The plugin marks itself as healthy only if both
+  TDX hardware and the QGS socket are present. If the socket is missing, it
+  becomes unhealthy and enters a retry loop. This acts as a scheduling gate:
+  TDX VMs requesting the resource won't allocate until QGS is deployed.
+* If attestation is not enforced: The plugin is always healthy if TDX hardware
```
mythi: "the plugin is healthy" is a bit overloaded. The plugin pod can run OK (status=Ready) but the resource it manages can be healthy/unhealthy.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+```xml
+configuration:
+    confidentialCompute:
+      tdx:
+        attestation:
+          enforced: true
+          qgsSocketPath: "/var/run/qgs.socket"
+```
+
+## Alternatives
```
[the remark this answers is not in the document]
MatiasVara: Can you elaborate a bit on those items so I can add it to the document, Or, Do you think it is clear enough?

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+do. If it is supported, it advertises it as a resource:
+`devices.kubevirt.io/tdx`. During resource allocation the plugin probes for the
+QGS socket at the configured path. If the socket exists, it mounts it into the
+virt-launcher pod as a volume.
+
+The plugin's health is based on the cluster-wide
+confidentialCompute.tdx.attestation.enforced configuration:
+
+* If attestation is enforced: The plugin marks itself as healthy only if both
+  TDX hardware and the QGS socket are present. If the socket is missing, it
```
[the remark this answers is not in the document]
MatiasVara: I am bit confuse now because in the document we propose:
```We introduce the TDX device plugin, i.e., `devices.kubevirt.io/tdx`, which runs
on nodes and handles both TDX hardware capability advertisement and QGS socket
management. This plugin replaces the current node-labeller for TDX detection.```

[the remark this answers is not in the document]
Comment (Aseeef) [reply veps/sig-compute/attestation-proposal.md]: If I understand correctly, @mythi is pointing out that when attestation=enforced, then DP health is based *only* on whether QGS socket is there. As for TDX detection, then the DP would have never been registered in the first place if TDX is not there.

File: veps/sig-compute/attestation-proposal.md (hunk 9 above)
Comment (MatiasVara) [reply veps/sig-compute/attestation-proposal.md]: I'll use the word resource instead of plugin.

File: veps/sig-compute/attestation-proposal.md (hunk 8 above)
Comment (Aseeef) [reply veps/sig-compute/attestation-proposal.md]: > Essentially my question was just that what this "fails" means: is one of thevirt-* components crashing? If not, perhaps clarify what users will see.

Good point. What "fails" means in the sentence `When enforced is true, KubeVirt shall require the socket to exist or it fails.` is unclear.

> I still believe the socket availability is not relevant because the absence of it does not prevent TD VM from starting. 

Adding to the Matias' explanation: when enforced=true, the mechanism that prevents the pod from being scheduled is device plugin health being "unhealthy" and in this case the TD fails to start. And when enforced=false, we still try to mount the socket, but if it doesn't exist we don't prevent scheduling and the TD still starts (because the device is always marked "healthy").

I'm not sure if I understood what you were saying or if I answered the question but let me know if something was unclear.

Comment (MatiasVara) [reply veps/sig-compute/attestation-proposal.md]: > Maybe we need to take one step back. Not sure where this `enforced` came from but as I see it, the goal should be that nothing prevents the TD from starting. (QGS is not needed for TD to run.)
> 

The `enforced` parameter was introduced following the discussion in
https://github.com/kubevirt/enhancements/pull/113#discussion_r2577692690 more than a month ago. The idea we discussed (with @vladikr) was the following:

`enforced` is a cluster-wide setting: based on the administrator’s decision, TDs either have access to the QGS socket or none of them do. We explicitly chose not to make this a per-VM setting.

The goal is not to make QGS a technical requirement for TDs to run. Rather, when `enforced=true`, the administrator is expressing a policy that TDs should not be launched in a configuration where attestation is expected but cannot be provided. In that mode, TDs that would rely on attestation are
intentionally blocked if QGS is not deployed.

When `enforced=false`, attestation is opportunistic: the TD is always launched, and it may or may not get access to attestation depending on whether QGS is present. This is the default value, and it reflects the default TDX behavior where TDs are allowed to run even if QGS is not available.

File: veps/sig-compute/attestation-proposal.md [hunk 12]
```diff
@@ -0,0 +1,154 @@
...
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+From the implementation perspective, we extend the KubeVirt CR spec to enable
+attestation:
+
+```xml
+configuration:
+    confidentialCompute:
```
jean-edouard: Nit: extra indentation
```suggestion
  confidentialCompute:
```

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+      tdx:
+        attestation:
+          enforced: true
+          qgsSocketPath: "/var/run/tdx-qgs/qgs.socket"
+```
+
+We add a new top-level section in configuration for all TEE-related
```
jean-edouard: First mention of `TEE`, please define.

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,154 @@
...
+- e2e tests to check that the blob from the guest point of view is correctly
+  generated.
+- Since TDX & SEV-SNP do not have support for nested virtualization this will
+  require bare metal hardware to conduct e2e testing.
+
+## Implementation Phases
+The initial phase of implementation will focus on integrating basic
+functionality. For example, a socket device plugin would be required to mount
+the UNIX socket into the virt-launcher. The initial phase also includes the e2e
+tests.
```
jean-edouard: Isn't the initial phase `Alpha`? Below it says that Alpha may or may not come with e2e...

jean-edouard (commented):
Looks great! Few nits in-line but in my opinion this is pretty much ready to go.
Not LGTMing right away to give time to resolve comments (mines and others).

veps/sig-compute/attestation-proposal.md
```diff
@@ -0,0 +1,155 @@
...
+
+We introduce the TDX device plugin, i.e., `devices.kubevirt.io/tdx`, which runs
+on nodes and handles both TDX hardware capability advertisement and QGS socket
+management. This plugin replaces the current node-labeller for TDX detection.
+It relies on Kubernetes native extended resource scheduling for placement. The
+plugin checks for TDX hardware support in the same way node-labeller used to
+do. If it is supported, it advertises it as a resource:
+`devices.kubevirt.io/tdx`. During resource allocation the plugin probes for the
+QGS socket at the configured path. If the socket exists, it mounts it into the
+virt-launcher pod as a volume.
```
Aseeef: > If the socket exists, it mounts it into the
virt-launcher pod as a volume.

Similar to the previous comment, this statement is also a bit misleading because the mount point is created unconditionally. If the directory corresponding to the mount point doesn't exist, most container runtimes will auto create them. Consequently (if I understand correctly), if the QGS socket is created while a pod is running since the mount point already exist, then even if it didn't exist when the pod started, attestation can still work.

Issue #115, opened 2025-10-29
VEP 115: PCIe NUMA Topology Awareness


mresvanis:
**Primary contact (assignee)**: @mresvanis @fanzhangio

/assign @mresvanis
/assign @fanzhangio

**Current Feature Stage**: New

**Feature Gate**: PCINUMAAwareTopology

**Responsible SIGs**: compute

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**: https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md

**Timeline**:

* Alpha release target (1.8):
* Beta release target (1.9): 
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
      * https://github.com/kubevirt/enhancements/pull/116
  * [ ] Code PRs:
      * https://github.com/kubevirt/kubevirt/pull/16632
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

lyarwood:
@mresvanis @fanzhangio Apologies for not reaching out before posting draft PRs extending the design of this VEP, the GB200 issue it describes came up during testing and might be something you want to address yourself during v1.9.0 in this VEP or the separate https://github.com/kubevirt/enhancements/issues/199 VEP. Happy to discuss more during the next wg-aie call or in upstream slack anytime.

johnahull:
**Extending VEP-115 for v1.10: DRA metadata topology and PCIe root grouping**

I've opened #424 to amend this VEP with a v1.10-scoped extension. Requesting it be tracked against the v1.10 milestone.

**Scope of the v1.10 extension:**

- Read device NUMA topology from DRA KEP-5304 metadata (`resource.kubernetes.io/numaNode`, `resource.kubernetes.io/pcieRoot`), with fallback to the existing sysfs path.
- Group devices per PCIe root complex rather than per NUMA node. This addresses the "Mirror PCIe topology" open question in the current VEP, since a single NUMA node can expose multiple PCIe root complexes (for example AMD NPS1, Intel SNC-off).
- Build guest NUMA cells from DRA device metadata when the kubelet CPU Manager cpuset is unavailable (`cpuManagerPolicy=none`), enabling composition with DRA CPU allocation (VEP-152).
- Inject host ACPI SLIT inter-NUMA distances into the guest.
[...]

PR #116, opened 2025-10-29
VEP 115: Add PCIe NUMA topology awareness VEP

mresvanis:
### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/115
**SIG label**:  /sig compute

### What this PR does
When `guestMappingPassthrough` is set any PCIe GPU and host devices with NUMA affinity information will be placed to a PCI controller hierarchy matching their NUMA node affinity (i.e. not directly on the default PCI root). The devices that don't have any NUMA alignment will be assigned to the default PCI root. Any error during this process will result to all devices falling back to their assignment under the default PCI root.

### DIFF
--- a/veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
+++ b/veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
@@ -1,0 +1,653 @@
+# VEP 115: PCIe NUMA Topology Awareness
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Target version is explicitly mentioned and approved
+- [x] (R) Graduation criteria filled
+
+## Overview
+
+This VEP enables AI workloads in VMs to achieve near-native performance by preserving
+host NUMA topology for PCIe devices. KubeVirt will automatically mirror device placement
+so AI frameworks can optimize communication paths and avoid cross-socket traffic. This is
+achieved by using QEMU's [pcie-expander-bus](https://github.com/qemu/qemu/blob/711a1ddf899bef577907a10db77475c8834da52f/include/hw/pci/pci_bridge.h#L92-L100)
+controllers to create NUMA-aware PCIe hierarchies in the guest.
+
+## Motivation
+
+AI workloads in VMs experience significant performance degradation when GPU and network devices
+(InfiniBand, RoCE) are presented without proper NUMA topology awareness. AI frameworks such as
+NVIDIA's Collective Communication Library (NCCL) and Unified Communication X (UCX) rely on
+accurate hardware topology information to optimize communication paths for distributed mode
+training. On physical hosts, these frameworks use NUMA affinity information to enable GPUDirect
+peer-to-peer communication between co-located devices and select optimal GPU-NIC pairs that
+minimize communication latency.
+
+KubeVirt currently presents all PCIe devices as uniformly accessible under a single host bridge,
+obscuring their actual NUMA placement. This causes AI frameworks to make suboptimal routing
+decisions that result in cross-NUMA memory traffic, even when guest vNUMA topology is properly
+configured. The performance impact is substantial: cross-socket communication incurs significantly
+higher latency and bandwidth costs, and frameworks cannot leverage direct device-to-device
+communication paths.
+
+Current limitations:
+- Guest VMs lack visibility into GPU and NIC NUMA placement and device locality.
+- AI frameworks cannot optimize communication topology for multi-GPU distributed model training.
+- Cross-NUMA memory traffic degrades performance despite correct NUMA-aware device allocation.
+- KubeVirt depends on Kubernetes' capabilities for NUMA-aligned resource allocation, e.g.:
+    - [CPU Manager](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/)
+    - [Topology Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/)
+    - [Dynamic Resource Allocation](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
+
+## Goals
+
+- Support VM creation with NUMA-aware PCIe device topologies that map directly
+  to the underlying host topology.
+
+## Non Goals
+
+- This VEP does not propose changes to core Kubernetes device allocation
+  mechanisms. Instead, it works with existing solutions (Topology Manager,
+  CPU Manager, NUMA-aware device plugins) and future enhancements (Dynamic Resource
+  Allocation) that can provide NUMA-aligned device placement.
+- Hot-plug support, as `pcie-expander-bus` controllers are [not hot-pluggable](https://github.com/qemu/qemu/blob/711a1ddf899bef577907a10db77475c8834da52f/docs/pcie.txt#L222-L228).
+  Since each pcie-pxb is associated with a specific NUMA node, adding a device at
+  runtime might require creating a new pcie-pxb if the device needs to be on a
+  different NUMA node than the existing controllers. Because the pcie-pxb can't
+  be hot-plugged, this would require a VM reboot. This feature only applies
+  during VM creation.
+- [Live migration](https://kubevirt.io/user-guide/compute/live_migration/) suppport.
+  This VEP does not address live migration of VMs with NUMA-aware PCIe
+  topologies to another compute node.
+- [Live migrate](https://kubevirt.io/user-guide/cluster_admin/updating_and_deletion/#updating-kubevirt-workloads)
+  when updating KubeVirt workloads. VMs with NUMA-aware PCIe topologies will
+  need to be shut down before a cluster upgrade.
+- Support for non-PCIe devices (e.g. USB host devices).
+
+## Assumptions, Constraints, and Dependencies
+
+- **Machine Type**: Requires q35 architecture.
+- **Device Support**: PCIe devices only, configured with VFIO passthrough.
+- **NUMA Inheritance**: VFs and mdevs inherit NUMA affinity from parent devices.
+- **Platform**: Requires QEMU 2.6 version that includes support for `pcie-expander-bus` controllers.
+
+## Definition of Users
+
+- Infrastructure administrators managing multi-socket NUMA systems with hardware accelerator workloads.
+- Application developers running AI/ML workloads requiring optimal GPU and network performance.
+- Cloud providers offering high-performance computing instances.
+
+## User Stories
+
+- As an infrastructure administrator, I want AI workloads in VMs to achieve near-native performance by
+  preserving NUMA topology awareness.
+- As a developer running distributed ML training, I want my chosen AI framework to automatically detect
+  optimal communication paths between GPUs and network devices.
+- As a cloud provider, I want to offer GPU instances with predictable, high-performance characteristics.
+
+## Repos
+
+- [kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+When creating a VM with NUMA CPU passthrough enabled (i.e.
+`spec.domain.cpu.numa.guestMappingPassthrough: {}`), KubeVirt will automatically
+mirror the host's PCIe device NUMA topology in the guest. This preserves device
+locality information that AI frameworks need for optimal performance.
+
+This feature will be introduced behind a feature gate named
+`PCINUMAAwareTopology`, as this will allow the community to test, validate and
+provide feedback before enabling it by default.
+
+When this feature is enabled [live migration](https://kubevirt.io/user-guide/compute/live_migration/)
+of VMs with NUMA-aware PCIe topologies will be blocked. Those VMs will be marked
+as non-live-migratable and need to be shut down before a [cluster upgrade](https://kubevirt.io/user-guide/cluster_admin/updating_and_deletion/#updating-kubevirt-workloads)
+or before being moved to another node.
+
+### Core Approach
+
+KubeVirt creates a NUMA-aware PCIe topology using:
+- one `pcie-expander-bus` (pxb-pcie) controller per NUMA node.
+- one `pcie-root-port` controller per device.
+- device placement based on host NUMA affinity.
+
+This approach ensures each passthrough device appears on a guest PCIe bus that corresponds to
+its actual host NUMA node, enabling AI frameworks to make optimal communication and memory access
+decisions.
+
+A non NUMA-aware PCIe topology of a VM looks like this:
+
+```
+   pcie.0
+   --------------------------------------------------------------------------------------
+         |                   |                     |                 |
+   ---------------    ------------------    ---------------    ---------------
+   | PCIe Device |    | pcie-root-port |    | PCIe Device |    | PCIe Device |
+   |  (GPU/NIC)  |    ------------------    |  (GPU/NIC)  |    |  (GPU/NIC)  |
+   ---------------                          ---------------    ---------------
+```
+
+While a NUMA-aware PCIe topology will look like this:
+
+```
+   pcie.0
+   --------------------------------------------------------------------------------------
+         |                   |                     |                         |
+   --------------     ------------------    ---------------           ---------------
+   | PCIe Device |    | pcie-root-port |    |  pxb-pcie   |           |  pxb-pcie   |
+   --------------     ------------------    | (NUMA node) |           | (NUMA node) |
+                                            ---------------           ---------------
+                                                   |                         |
+                                          --------------------      --------------------
+                                          |  pcie-root-port  |      |  pcie-root-port  |
+                                          --------------------      --------------------
+                                                   |                         |
+                                            ---------------           ---------------
+                                            | PCIe Device |           | PCIe Device |
+                                            |  (GPU/NIC)  |           |  (GPU/NIC)  |
+                                            ---------------           ---------------
+```
+
+### Scope
+
+- This feature is enabled only when the virtual machine's `spec.domain.cpu.numa.guestMappingPassthrough` is set.
+- It applies only to passthrough PCIe host devices (GPUs, IB/RoCE NICs, SR-IOV PF/VFs, DRA-backed host devices).
+- Mediated devices (mdevs) follow the parent physical PCI device's NUMA node when available. Mediated devices
+  should be backed by Virtual Functions (VFs) using VFIO passthrough.
+- Devices without NUMA affinity are placed on the default `pci.0` bus and report NUMA node `-1` (absent).
+
+### Device NUMA Affinity Discovery
+
+KubeVirt's `virt-launcher` component [converts](https://github.com/kubevirt/kubevirt/blob/0358eea1e1b9faf2717700b7b0599268e8488706/pkg/virt-launcher/virtwrap/converter/converter.go#L1159)
+the VirtualMachineInstance Spec to the equivalent libvirt domain XML. During
+this conversion, it determines device NUMA affinity by reading
+`/sys/bus/pci/devices/<BDF>/numa_node` from the host filesystem and comparing
+the device's NUMA node against the guest's configured NUMA nodes. When the
+device's NUMA node matches one of the guest's NUMA nodes, the device is aligned
+to that node; otherwise, it defaults to NUMA node `-1`.
+
+Mediated devices: while PCIe devices already include this information, mediated
+devices require additional resolution. We resolve the parent PCIe device address
+via `/sys/bus/mdev/devices/<mdev-uuid>`. The mdev UUID is propagated to the
+`virt-launcher` pod with the respective environment variable, i.e.
+`MDEV_PCI_RESOURCE_<resource-name>`, by either the device plugin or the DRA
+driver.
+
+DRA devices: the status controller must include the PCIe address in
+`DeviceResourceClaimStatus.Attributes.PCIAddress`. If DRA drivers cannot provide
+this information, those devices will default to NUMA node `-1`.
+
+### Domain Generation
+
+We propose a new PCI NUMA-aware assigner (in addition to the existing PCI root
+slot assigner). This assigner will be executed before any host devices are added
+to the domain. It will perform the following steps to create the NUMA-aware PCIe
+topology for the devices with NUMA affinity information:
+
+1. Discovers the NUMA node for each passthrough PCIe device and if it matches any of the guest's NUMA nodes.
+2. Filters out devices without NUMA alignment to be placed on the default `pci.0` bus using the existing root slot assigner.
+3. Groups devices by NUMA node: `numaNode -> []api.HostDevice`.
+4. Creates `pcie-expander-bus` controllers with `<target busNr="X"><node>Y</node></target>` elements for each NUMA node.
+5. Creates one `pcie-root-port` controller per device under the respective `pcie-expander-bus` controller.
+6. Assigns device addresses to the `pcie-root-port` aligned with its NUMA node.
+7. Inserts the controllers created into the domain before adding the host devices.
+
+### Schema Changes
+
+Since the `virt-launcher/virtwrap/api/schema.go` currently lacks a `target` field in the `Controller`
+struct, we need to add a new `ControllerTarget` struct to represent the `<target>` element:
+
+```go
+// pkg/virt-launcher/virtwrap/api/schema.go
+type Controller struct {
+	Type    string            `xml:"type,attr"`
+	Index   string            `xml:"index,attr"`
+	Model   string            `xml:"model,attr,omitempty"`
+	Driver  *ControllerDriver `xml:"driver,omitempty"`
+	Alias   *Alias            `xml:"alias,omitempty"`
+	Address *Address          `xml:"address,omitempty"`
++   Target  *ControllerTarget `xml:"target,omitempty"`
+}
+
++ type ControllerTarget struct {
++   BusNr     *uint32 `xml:"busNr,attr,omitempty"`
++   NUMANode  *uint32 `xml:"numaNode,omitempty"`
++ }
+```
+
+These schema additions follow the existing [libvirt go XML module](https://gitlab.com/libvirt/libvirt-go-xml-module/-/blob/master/domain.go?ref_type=heads#L46-55)
+for PCIe controllers.
+
+BusNr values for `pcie-expander-bus` controllers can be assigned starting from `254` and decrementing
+to avoid conflicts with existing controllers.
+
+According to the libvirt [documentation](https://libvirt.org/formatdomain.html#controllers):
+> pci-expander-bus and pcie-expander-bus controllers can have an optional busNr attribute (1-254).
+  This will be the bus number of the new bus; All bus numbers between that specified and 255 will be
+  available only for assignment to PCI/PCIe controllers plugged into the hierarchy starting with this
+  expander bus, and bus numbers less than the specified value will be available to the next lower
+  expander-bus (or the root-bus if there are no lower expander buses). If you do not specify a busNumber,
+  libvirt will find the lowest existing busNumber in all other expander buses (or use 256 if there are
+  no others) and auto-assign the busNr of that found bus - 2, which provides one bus number for the
+  pci-expander-bus and one for the pci-bridge that is automatically attached to it (if you plan on adding
+  more pci-bridges to the hierarchy of the bus, you should manually set busNr to a lower value).
+
+### Hardware Utils Extension
+
+We will extend the [hardware](https://github.com/kubevirt/kubevirt/tree/7d7c72f1a0ec134175a750d1d89d261fd6138823/pkg/util/hardware)
+package with a function to retrieve a PCIe device's NUMA node from
+`/sys/bus/pci/devices/<BDF>/numa_node` and verify it matches the guest's vCPU
+NUMA affinity. This enables NUMA-aware device placement during domain generation.
+
+```go
+// pkg/util/hardware/hardware.go
+
+func PCIAddressToString(pciBusID *api.Address) string {
+       prefix := "0x"
+       return fmt.Sprintf("%s:%s:%s.%s",
+               strings.TrimPrefix(pciBusID.Domain, prefix),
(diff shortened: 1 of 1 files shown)


veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,544 @@
...
+
+Current limitations:
+- Guest VMs lack visibility into GPU and NIC NUMA placement and device locality.
+- AI frameworks cannot optimize communication topology for multi-GPU distributed model training.
+- Cross-NUMA memory traffic degrades performance despite correct NUMA-aware device allocation.
+
+## Goals
+
+- Add a new feature gate `PCINUMAAwareTopology`.
+- When the afore-mentioned feature gate and `guestMappingPassthrough` are enabled, mirror PCIe
```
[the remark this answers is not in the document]
fanzhangio: Thanks @lyarwood for your feedback!
Our current design is based on the understanding that PCIe NUMA passthrough is inherently coupled with CPU NUMA passthrough. In practice, enabling PCIe NUMA passthrough without aligning the guest’s CPU topology is not feasible. Therefore, the workflow we designed assumes that:

- If users want to enable PCIe NUMA awareness, they must explicitly enable the feature gate and set `guestMappingPassthrough: true`
- If users only want CPU NUMA passthrough, the current behavior continues to work as expected, as long as the feature gate remains disabled.
 
That said, you bring up an excellent point: when the feature gate graduates, enabling it could inadvertently alter behavior for users already relying on `guestMappingPassthrough`. Introducing an explicit API field early on is a more robust and future-proof solution, as it avoids implicit behavior shifts.

I agree with your suggestion. We can introduce a dedicated policy field to clearly express PCIe NUMA passthrough intent. One possible design might look like this:
```yaml
spec:
  domain:
    devices:
      pciNumaAwareTopology: true
```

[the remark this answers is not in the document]
vladikr: @lyarwood, I'm wondering if you could explain why we would want another tunable?
Are there users who don't want to expose the NUMA affinity of the passed-through devices?
So far, we have been avoiding constructing PCI controllers, but with this VEP, we have a path forward. 
I think that constructing these controller hierarchies for NUMA passthrough is a first step, but later we will need to find a way to advertise the passed devices' NUMA affinity with dedicated CPUs as well  - hopefully with DRA integration, this will be easier

if we must have a new api field, then I would vote for
 ```
spec:
  domain:
    cpu:
      numa:
        guestMappingPassthrough: {}
        pciNumaAwareTopology: {}
```

However, I don't know if it's needed.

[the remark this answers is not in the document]
Comment (lyarwood) [reply veps/sig-compute/pcie-numa-topology-awareness.md]: @vladikr guest ABI, ultimately to implement this we are going to rewire PCI device topology right? So without an additional configurable all users of `guestMappingPassthrough` will see this enabled in their guests once the FG graduates or is enabled. That's why I'm suggesting another configurable.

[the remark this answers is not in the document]
Comment (vladikr) [reply veps/sig-compute/pcie-numa-topology-awareness.md]: @lyarwood I still fail to see a reason for a VMI level api to enable to disable this. For me, this will be just another tunable that the user will forget to set.
From a live migration point of view, these additional controllers will come into play only if we have a passthrough device - for most, we don't allow live migration. So there is no upgrade path for these VMs. The only exception is network SRIOV, which in some cases we unplug and re-plug the device on the target - I'm not sure here @EdDev @orelmisan fyi
I think there shouldn't be any effect since we are reusing the same domain xml from the running domain and not creating a new one.

In the static case, when a stopped VM gets new controllers, we will need to evaluate this. I don't believe there is an issue with Linux, but I don't know about Windows guests. 
However, I don't believe we need a VM-level tunable; instead, we could have a cluster-wide setting.

File: veps/sig-compute/pcie-numa-topology-awareness.md [hunk 2]
```diff
@@ -0,0 +1,561 @@
...
+  host device NUMA topology in the guest.
+
+## Non Goals
+
+- This VEP focuses solely on ensuring optimal guest PCIe topology generation
+  when devices are locally available. It does not address changes to Kubernetes
+  device/resource allocation for guaranteeing NUMA-aligned device allocation.
+  Dynamic Resource Allocation (DRA) will help address that in the near future.
+- Hot-plug support, as `pcie-expander-bus` controllers are [not hot-pluggable](https://github.com/qemu/qemu/blob/711a1ddf899bef577907a10db77475c8834da52f/docs/pcie.txt#L222-L228).
```
Aseeef: pcie-expander-buses are not hot pluggable as stated. But what about devices connected through the pcie-root-port (which if I understand do implement hot plugging support)?

Basically when we say "no hotplug support" is a non-goal does that include the downstream devices that would be connecting into the pcie-root-port?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+          - deviceName: rdma/ib_device
+            name: ib1
+        resources:
+          requests:
+            memory: 8Gi
+            nvidia.com/gpu: 4
+            rdma/ib_device: 2
+```
+
+## Implementation Roadmap
```
Aseeef: I noticed graduation requirements put mediated devices in a separate stage and looking at that I think we should explicitly mention at what stage here we intend to do the implementation for VFs and mdevs.

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+</devices>
+```
+
+## Scalability
+
+The solution should scale to:
+
+- Large clusters with hundreds of NUMA-enabled nodes.
+- Multiple PCIe devices per NUMA node (up to 14+ devices as shown in examples).
+- Complex multi-socket systems with 4+ NUMA nodes.
```
Aseeef: About the "14+ devices" and "4+ NUMA nodes" as mentioned: is there a fundamental limitation for why we can't or shouldn't support systems with more devices per NUMA node or systems with more NUMA nodes besides the 255 devices hard-limit on the root-complex (pcie.0)?

Meaning for example, would a system with 8 NUMA nodes with 2 devices each (16 total) still not be valid?

mresvanis: I apologize, I will rephrase these 2 sentences to better reflect their intention.  The `14+ devices` and `4+ NUMA nodes`  are there to show that this VEP's proposed solution should cover systems with more that 4 NUMA nodes and more than 14 devices per NUMA node. 

> Meaning for example, would a system with 8 NUMA nodes with 2 devices each (16 total) still not be valid?

That system should be properly supported by this VEP's proposed solution.

veps/sig-compute/pcie-numa-topology-awareness.md
Aseeef: OK, I see. So this implementation road map is specifically referring to the "Alpha" part of this later section, right?
```
## Graduation Requirements

### Alpha
- Basic NUMA topology awareness implemented
- Unit tests and basic integration tests
- Documentation for feature usage

### Beta
- Support for mediated devices and VFs
- Performance validation with real AI workloads
- Comprehensive testing coverage including edge cases
- Integration with DRA (Dynamic Resource Allocation)

### GA
- Production-ready performance and reliability
- Full documentation and operational guides
```

I was getting confused how this section corresponded to this later section.

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,560 @@
...
+
+- Add a new feature gate `PCINUMAAwareTopology`.
+- When the afore-mentioned feature gate and `guestMappingPassthrough` are enabled, mirror PCIe
+  host device NUMA topology in the guest.
+
+## Non Goals
+
+- Changes in Kubernetes device/resource allocation to guarantee NUMA-aligned device assignment.
+  This VEP focuses solely on guest PCIe topology generation.
+- Hot-plug support, as `pcie-expander-bus` controllers are [not hot-pluggable](https://github.com/qemu/qemu/blob/711a1ddf899bef577907a10db77475c8834da52f/docs/pcie.txt#L222-L228).
```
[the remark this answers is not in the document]
xpivarc: @orelmisan @EdDev cc

veps/sig-compute/pcie-numa-topology-awareness.md
mresvanis: That's a very good point, I now realized that we can improve this and be clear about the milestones for each version/implementation. I'll update the doc and ping you back.

veps/sig-compute/pcie-numa-topology-awareness.md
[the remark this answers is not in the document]
EdDev: I would prefer the non-goals to be generic as well, not specific and without too much details.
At this level, we are interested to know if hotplug of hostdevices (or devices in general?) is supported or not.
The implications regarding SR-IOV is dependent on the ability to migrate + hotplug. If we cannot migrate and/or hotplug the device, SR-IOV is not hotpluggable, like many other hotplug functionalities.
Therefore, I would mention that any dependent migration functionality, like hotplug, is not supported.

> **SR-IOV Virtual Functions (VFs)**: VFs inherit NUMA affinity from their parent Physical Function (PF). KubeVirt resolves the parent PF by reading the `physfn` symlink at `/sys/bus/pci/devices/<VF_BDF>/physfn`, which points to the parent PF's sysfs directory. The VF is then assigned to the same NUMA node and `pcie-expander-bus` controller as its parent PF. This ensures that all VFs from the same PF maintain locality with their shared hardware resources, enabling optimal performance for SR-IOV workloads.

I do not think the virt-launcher has access to the PF. Only the VF is passed to the pod and the PF is touched by the CNI which has access to the node namespace.

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+- Add a new feature gate `PCINUMAAwareTopology`.
+- When the afore-mentioned feature gate and `guestMappingPassthrough` are enabled, mirror PCIe
+  host device NUMA topology in the guest.
+
+## Non Goals
+
+- This VEP focuses solely on ensuring optimal guest PCIe topology generation
+  when devices are locally available. It does not address changes to Kubernetes
+  device/resource allocation for guaranteeing NUMA-aligned device allocation.
+  Dynamic Resource Allocation (DRA) will help address that in the near future.
```
EdDev: I do not understand this part. Without it, how can we provide a solution that works?
AFAIU the NUMA topology awareness is dependent on the availability of other resource managers to assign correct devices into the pod, such that Kubevirt will be able to continue and assign those to the domain.

If that initial step is not something we can assure, the feature is not complete.
I would expect that Kubevirt is able to request this in a way from the resource manager and be assured it is aligned.

To elaborate a bit more, if this is dependent on an external solution that does this correct assignment, I would like to understand how it interacts with Kubevirt. As a community project, we should at the minimum give some guidelines and possibly references to existing solutions which can make this happen.

We'll also need to test this, therefore it is also a practical scenario which needs attention.

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+                                            ---------------           ---------------
+                                            | PCIe Device |           | PCIe Device |
+                                            |  (GPU/NIC)  |           |  (GPU/NIC)  |
+                                            ---------------           ---------------
+```
+
+### Scope
+
+- This feature is enabled only when:
+  - the feature gate `PCINUMAAwareTopology` is enabled and
```
EdDev: This is irrelevant for the feature itself. The FG protects the feature lifecycle and is not used to control it as a operational knob. The design should focus on graduation and keep the lifecycle in the relevant section.
If you need to control this feature operation, a formal API is needed.

veps/sig-compute/pcie-numa-topology-awareness.md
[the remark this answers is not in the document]
orelmisan: The current suggestion is written in the present tense, but I haven't found any NUMA-aware logic for defining SR-IOV VFs in our current implementation.
I assume adding this logic is one of the intended goals of the PR?

Does the virt-launcher pod have enough privileges to read the relevant information from the operating system?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+
+1. Discovers the NUMA node for each passthrough PCIe device and if it matches any of the guest's NUMA nodes.
+2. Filters out devices without NUMA alignment to be placed on the default `pci.0` bus using the existing root slot assigner.
+3. Groups devices by NUMA node: `numaNode -> []api.HostDevice`.
+4. Creates `pcie-expander-bus` controllers with `<target busNr="X"><node>Y</node></target>` elements for each NUMA node.
+5. Creates one `pcie-root-port` controller per device under the respective `pcie-expander-bus` controller.
+6. Assigns device addresses to the `pcie-root-port` aligned with its NUMA node.
+7. Inserts the controllers created into the domain before adding the host devices.
+
+### Schema Changes
```
EdDev: Is this in sync with https://gitlab.com/libvirt/libvirt-go-xml-module ?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+  expander bus, and bus numbers less than the specified value will be available to the next lower
+  expander-bus (or the root-bus if there are no lower expander buses). If you do not specify a busNumber,
+  libvirt will find the lowest existing busNumber in all other expander buses (or use 256 if there are
+  no others) and auto-assign the busNr of that found bus - 2, which provides one bus number for the
+  pci-expander-bus and one for the pci-bridge that is automatically attached to it (if you plan on adding
+  more pci-bridges to the hierarchy of the bus, you should manually set busNr to a lower value).
+
+### Hardware Utils Extension
+
+We will extend the `hardware` package with a function to retrieve a PCIe device's NUMA node
```
EdDev: What package is this referring to? Can you provide a ref and/or overview of what is extended?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+The solution should scale to:
+
+- Large clusters with hundreds of NUMA-enabled nodes.
+- Multiple PCIe devices per NUMA node (up to 14+ devices as shown in examples).
+- Complex multi-socket systems with 4+ NUMA nodes.
+- Efficient controller allocation avoiding resource conflicts.
+
+## Update/Rollback Compatibility
+
+- Feature is opt-in via existing `spec.domain.cpu.numa.guestMappingPassthrough` flag.
```
EdDev: What about compatibility with this opt-in setting to this new addition? It was discussed before that it will change the topology. Can you add that information here?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+- Existing VMs without NUMA specifications continue to work unchanged.
+- Devices reporting NUMA node `-1` fallback to the default `pci.0` layout.
+- Rollback gracefully disables NUMA-aware features without affecting running workloads.
```
EdDev: What does "rollback" means in this context? Kubevirt is not supporting downgrading, therefore this may effect a single VM to disable the mapping. Was that the case referred here?

How will a guest OS behave if you change the scheme?

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+
+## Functional Testing Approach
+
+- Unit tests: add cases under `virt-launcher` `virtwrapper` codebase verifying NUMA grouping, controller
+  generation, and fallback when NUMA is `-1`.
+- Integration tests: craft NUMA VMI functional tests (similar to the ones found in `tests/numa/numa.go`)
+  launching with multiple passthrough GPUs on different NUMA nodes; assert domain XML includes `pxb-pcie` `node='0/1'`
+  and devices placed under the correct root ports.
+- Provide a golden XML fixture akin to `numa-topo-vm-domain.xml` for regression checks.
```
EdDev: Is this referring to unit tests?

EdDev (changes requested):
Thank you for this interesting proposal.

The overall feature proposal is clear.
I am missing details on how this is going to play in the different kubevirt components and the API seems not to reach agreement yet.

I may not have all the context the other reviewers of this VEP had, therefore I apologize in advance.

veps/sig-compute/pcie-numa-topology-awareness.md
mresvanis: I updated the text to make this point clear. PTAL

veps/sig-compute/pcie-numa-topology-awareness.md
mresvanis: Great point, I removed the feature gate from the scope. PTAL

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+- Feature is opt-in via existing `spec.domain.cpu.numa.guestMappingPassthrough` flag.
+- Existing VMs without NUMA specifications continue to work unchanged.
+- Devices reporting NUMA node `-1` fallback to the default `pci.0` layout.
+- Rollback gracefully disables NUMA-aware features without affecting running workloads.
+
+## Functional Testing Approach
+
+- Unit tests: add cases under `virt-launcher` `virtwrapper` codebase verifying NUMA grouping, controller
+  generation, and fallback when NUMA is `-1`.
+- Integration tests: craft NUMA VMI functional tests (similar to the ones found in `tests/numa/numa.go`)
```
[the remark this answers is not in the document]
mresvanis: Good point, I have updated all text referring to "integration tests" to use "end-to-end test" instead. PTAL

veps/sig-compute/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+  launching with multiple passthrough GPUs on different NUMA nodes; assert domain XML includes `pxb-pcie` `node='0/1'`
+  and devices placed under the correct root ports.
+- Provide a golden XML fixture akin to `numa-topo-vm-domain.xml` for regression checks.
+- Performance benchmarks comparing NUMA-aware vs NUMA-unaware configurations.
```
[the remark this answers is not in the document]
mresvanis: Since we are not ready to answer these questions I opted to remove this point from that section. I am also not confident that these benchmarks would be useful. Does this make sense?

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+  expander bus, and bus numbers less than the specified value will be available to the next lower
+  expander-bus (or the root-bus if there are no lower expander buses). If you do not specify a busNumber,
+  libvirt will find the lowest existing busNumber in all other expander buses (or use 256 if there are
+  no others) and auto-assign the busNr of that found bus - 2, which provides one bus number for the
+  pci-expander-bus and one for the pci-bridge that is automatically attached to it (if you plan on adding
+  more pci-bridges to the hierarchy of the bus, you should manually set busNr to a lower value).
+
+### Hardware Utils Extension
+
+We will extend the `hardware` package with a function to retrieve a PCIe device's NUMA node
```
mresvanis: Great catch, I added a link to the [respective](https://github.com/kubevirt/kubevirt/tree/7d7c72f1a0ec134175a750d1d89d261fd6138823/pkg/util/hardware) package and a sample implementation. PTAL

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+  - the feature gate `PCINUMAAwareTopology` is enabled and
+  - the virtual machine's `spec.domain.cpu.numa.guestMappingPassthrough` is set.
+- It applies only to passthrough PCIe host devices (GPUs, IB/RoCE NICs, SR-IOV PF/VFs, DRA-backed host devices).
+- Mediated devices (mdevs) follow the parent physical PCI device's NUMA node when available. Mediated devices
+  should be backed by Virtual Functions (VFs) using VFIO passthrough.
+- Devices without NUMA affinity are placed on the default `pci.0` bus and report NUMA node `-1` (absent).
+
+### Device NUMA Affinity Discovery
+
+KubeVirt discovers device NUMA affinity by reading `/sys/bus/pci/devices/<BDF>/numa_node`
```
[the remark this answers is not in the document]
mresvanis: > Which component is expected to read it and at what stage?
 
Very good question, I completely missed adding the relevant details. What we have already tested (PoC-ed) is gathering the NUMA node of the host devices via the `virt-launcher`'s VMI Spec to libvirt domain XML [conversion](https://github.com/kubevirt/kubevirt/blob/0358eea1e1b9faf2717700b7b0599268e8488706/pkg/virt-launcher/virtwrap/converter/converter.go#L1159). In the same flow we can also setup the respective PCIe controllers to reflect the devices' NUMA affinity when that is available. I added these details to the VEP. PTAL

> In addition, resolving parent devices and attempting to query them may also hit the permission and visibility challenge. E.g. with SR-IOV, while VFs are assigned to the pod, their parent device (PF) is not.

> There may be a need to run a PoC to check all these points.
> It will be valuable for this information to be explicitly described in the VEP.

That's is another great point. We do indeed need to run a PoC for VFs and NUMA alignment. I am currently trying to set it up and provide you with updates asap.

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+          - deviceName: rdma/ib_device
+            name: ib1
+        resources:
+          requests:
+            memory: 8Gi
+            nvidia.com/gpu: 4
+            rdma/ib_device: 2
+```
+
+## Implementation Roadmap
```
enp0s3: @mresvanis Hi,
I would recommend implementing all the use-cases in Alpha, leaving Beta to increase stability by adding testing coverage and addressing changes that may come from user feedback, then having one release cycle without changes, positive feedback and stable CI I would move to GA

mresvanis: Great point, this makes a lot of sense. Introducing new use-cases in Beta seems confusing and complicates the transition from Alpha to Beta. Thank you for pointing this out, I updated the text to reflect this.

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+configured. The performance impact is substantial: cross-socket communication incurs significantly
+higher latency and bandwidth costs, and frameworks cannot leverage direct device-to-device
+communication paths.
+
+Current limitations:
+- Guest VMs lack visibility into GPU and NIC NUMA placement and device locality.
+- AI frameworks cannot optimize communication topology for multi-GPU distributed model training.
+- Cross-NUMA memory traffic degrades performance despite correct NUMA-aware device allocation.
+
+## Goals
```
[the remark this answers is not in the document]
EdDev: Thank you.
Please resolve this thread.

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+- Add a new feature gate `PCINUMAAwareTopology`.
+- When the afore-mentioned feature gate and `guestMappingPassthrough` are enabled, mirror PCIe
+  host device NUMA topology in the guest.
+
+## Non Goals
+
+- This VEP focuses solely on ensuring optimal guest PCIe topology generation
+  when devices are locally available. It does not address changes to Kubernetes
+  device/resource allocation for guaranteeing NUMA-aligned device allocation.
+  Dynamic Resource Allocation (DRA) will help address that in the near future.
```
EdDev: > KubeVirt can request devices (and other resources) for a VM/Pod

I would expect that ether Kubevirt abstracts this for the user (i.e. it asks for this through the Kubevirt API) or it needs to configure/integrate something else to make this happen.
I separate the component responsible and the actual implementation from the request method/API.

Most of it should be document, I agree. But some initial explanation here will help clarify Kubevirt part better IMO.

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+- Multiple PCIe devices per NUMA node (up to 14+ devices as shown in examples).
+- Complex multi-socket systems with 4+ NUMA nodes.
+- Efficient controller allocation avoiding resource conflicts.
+
+## Update/Rollback Compatibility
+
+- Feature is opt-in via existing `spec.domain.cpu.numa.guestMappingPassthrough` flag.
+- Existing VMs without NUMA specifications continue to work unchanged.
+- Devices reporting NUMA node `-1` fallback to the default `pci.0` layout.
+- Rollback gracefully disables NUMA-aware features without affecting running workloads.
```
EdDev: Yes, thank you.
If you can, lets call it "Fallback" to clarify this.

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,625 @@
...
+- As a cloud provider, I want to offer GPU instances with predictable, high-performance characteristics.
+
+## Repos
+
+- [kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+When creating a VM with NUMA CPU passthrough enabled, KubeVirt will automatically mirror
+the host's PCIe device NUMA topology in the guest. This preserves device locality information
```
EdDev: Starting the discussion here on the "default" operation.
An active discussion is on the root discussion page: https://github.com/kubevirt/enhancements/pull/116#issuecomment-3666959580

The decision is to have this enabled by default + reflect the existing topology.
And to allow opt-out in order to support the previous legacy behavior?
Is that the bottom line?

@vladikr , @xpivarc

veps/sig-compute/115-pcie-numa-topology-awareness/pcie-numa-topology-awareness.md
```diff
@@ -0,0 +1,561 @@
...
+
+1. Discovers the NUMA node for each passthrough PCIe device and if it matches any of the guest's NUMA nodes.
+2. Filters out devices without NUMA alignment to be placed on the default `pci.0` bus using the existing root slot assigner.
+3. Groups devices by NUMA node: `numaNode -> []api.HostDevice`.
+4. Creates `pcie-expander-bus` controllers with `<target busNr="X"><node>Y</node></target>` elements for each NUMA node.
+5. Creates one `pcie-root-port` controller per device under the respective `pcie-expander-bus` controller.
+6. Assigns device addresses to the `pcie-root-port` aligned with its NUMA node.
+7. Inserts the controllers created into the domain before adding the host devices.
+
+### Schema Changes
```
mresvanis: I updated the text to reflect the fact that those schema changes are in sync with the libvirt Go XML module.

mresvanis:
> > > I like this proposal; to me, it looks excellent and thoughtful. However, I would like to avoid any VMI spec API changes for this.
> > > I believe there is only one open issue/risk related to guest ABI breakage, particularly for existing Windows VMs.
> > > What I would like to suggest is to persist the initial startup behavior in the VMI status to mitigate this without disrupting existing VMs.
> > > Specifically, On start we could add `PCITopologyMode: Legacy` for existing VMs and `PCITopologyMode: NUMAAware` for new ones, or just add a simple boolean `numaAwarePCITopologyApplied: true` for newly created ones. And then honor the stored value: rebuild the domain XML with the same topology mode.
> > > Then, for users who want to opt-out of this persistence we can introduce an annotation, something like `kubevirt.io/force-pci-numa-topology-upgrade: "true"`
> > > What do you think?
> > 
> > 
> > Chatting with @iholder101, Itamar suggested avoiding the more static vmi status api. Storing temporary or internal state in .status can feel like overkill, and the idea is that it may not be long-term. It also has only limited value to the users. At some point, most VMs will migrate to the new topology.
> > With that said we could use annotations to store this internal/temporary state: `internal.kubevirt.io/pci-topology-mode: "legacy"` And add a validating webhook to reject user and only allow controller writes (we already have a similar mechanism).
> > For users who wish to opt out ( to force the upgrade), we can still use a separate user facing annotation: `kubevirt.io/upgrade-pci-topology: "true"` Then if it's present, the controller clears or overrides the internal annotation and applies the new topology.
> 
> @mresvanis I can't find a reference to this suggestion in the VEP, maybe worth to describe it in the "Upgrade & Rollback" section.

@enp0s3 thank you for pointing this out and your suggestion! I updated the section `Update/Rollback Compatibility` with all this information. PTAL

EdDev:
> > > Chatting with @iholder101, Itamar suggested avoiding the more static vmi status api. Storing temporary or internal state in .status can feel like overkill, and the idea is that it may not be long-term. It also has only limited value to the users. At some point, most VMs will migrate to the new topology.
> > > With that said we could use annotations to store this internal/temporary state: `internal.kubevirt.io/pci-topology-mode: "legacy"` And add a validating webhook to reject user and only allow controller writes (we already have a similar mechanism).
> > > For users who wish to opt out ( to force the upgrade), we can still use a separate user facing annotation: `kubevirt.io/upgrade-pci-topology: "true"` Then if it's present, the controller clears or overrides the internal annotation and applies the new topology.
> > 
> > 
> > @mresvanis I can't find a reference to this suggestion in the VEP, maybe worth to describe it in the "Upgrade & Rollback" section.
> 
> @enp0s3 thank you for pointing this out and your suggestion! I updated the section `Update/Rollback Compatibility` with all this information. PTAL

I think this is controversial and we do not have time for this discussion. Can we keep it in the open-questions and work on it through the Alpha stage?

orelmisan (approved):
Thank you @mresvanis !
Please check the boxes in the checklist :)

vladikr:
/approve
Thank you!

PR #117, opened 2025-10-30
VEP118: image volume with digest proposal

Barakmor1:
### VEP Metadata

**Tracking issue**: 
https://github.com/kubevirt/enhancements/issues/118
**SIG label**: 
/sig-compute

### DIFF
--- a/veps/sig-compute/image-volume-digest-proposal.md
+++ b/veps/sig-compute/image-volume-digest-proposal.md
@@ -1,0 +1,119 @@
+# VEP #NNNN: Your short, descriptive title
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


Barakmor1:
> > > /cc @Barakmor1 is this ready for review?
> > 
> > 
> > Hey Adi, Unfortunately, [kubernetes/enhancements#5365](https://github.com/kubernetes/enhancements/issues/5365) didn’t make it into alpha for k8s 1.35, so there isn’t much we can do on our side until k8s 1.36 becomes available.
> 
> @Barakmor1, what about this cycle? :)

Hey, this enhancement can only be tested on Kubernetes versions 1.36 and above, since it relies on a feature gate that is Alpha in 1.36.

At the moment, our CI does not support 1.36, and we also cannot GA this feature until it is GA in the three latest Kubernetes versions. Based on that, we will likely only be able to GA this around Kubernetes 1.40, which is still quite far out.

Given these constraints, this work is currently low priority. I plan to start working on it in the next VEP cycle, once kubevirtci supports 1.36. Working on it now is not expected to accelerate the GA timeline due to these limitations.

iholder101:
> At the moment, our CI does not support 1.36

We do support it now :)
Any plans for v1.10?

Barakmor1:
> > At the moment, our CI does not support 1.36
> 
> We do support it now :) Any plans for v1.10?

@iholder101 Since this one is relatively low priority, I'd like to focus more on reviews during this release cycle. If that works for you, I think it makes sense to defer it to the next release.

By the way, since you're the KEP owner in Kubernetes, do you have a rough estimate of when you expect this to reach beta?

iholder101:
> By the way, since you're the KEP owner in Kubernetes, do you have a rough estimate of when you expect this to reach beta?

The window for 1.37 is already missed. I'll try to graduate this for v1.38.

PR #126, opened 2025-11-12
VEP 25 Incremetal Backup extension: Introduce pull mode design

Acedus:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issue/25
**SIG label**: /sig storage 
VEP Tracker: https://github.com/kubevirt/enhancements/issues/25

### What this PR does

This extension to VEP 25 proposes adding support to the pull backup mode in addition to the previously established push mode.

The pull mode is designed for backup clients that require direct interaction with the backup endpoint, rather than receiving sparse qcow2 images, as is the case with push mode.

The extension specifies a preferred M.O, evaluates several alternative approaches, and examines the respective pros and cons of each.

### DIFF
--- a/veps/sig-storage/incremental-backup.md
+++ b/veps/sig-storage/incremental-backup.md
@@ -93,18 +93,20 @@ As outlined in Libvirt, there are two general modes for backup:
 ```
 A `push` mode (where the hypervisor writes out the data to the destination file, which may be local or remote), and a `pull` mode (where the hypervisor creates an NBD server that a third-party client can then read as needed, and which requires the use of temporary storage, typically local, until the backup is complete).
 ```
-* In the initial phase, we will implement **push** mode. In this mode, the user provides a filesystem-based PersistentVolumeClaim (PVC) to store the backup data for all relevant VM disks. Before the process begins, an estimation of the required backup size will be performed. If the provided PVC size is insufficient, an error will be returned.
+We will be implementing both modes in the following manner:
+* **Push mode**: In this mode, the user provides a filesystem-based PersistentVolumeClaim (PVC) to store the backup data for all relevant VM disks. Before the process begins, an estimation of the required backup size will be performed. If the provided PVC size is insufficient, an error will be returned.
 The PVC will be hot-plugged into the virt-launcher pod as a directory (rather than as a disk attached to the VM, as is typical for hot-plugged volumes). The backup process will then proceed, writing data directly to the mounted PVC. Once the backup is complete, the PVC will be detached. It is then the user’s responsibility to manage the backup data—for example, by transferring it to a remote storage location, if needed.
-* In the second phase, we will implement **pull** mode. This will involve defining a complete network API that exposes an endpoint the user can connect to in order to retrieve the backup.
->> Note: More details on the pull mode will be described in a subsequent VEP.
+* **Pull mode**: In this mode, the user also provides a filesystem-based PersistentVolumeClaim (PVC) to serve as temporary storage for scratch files that will be used to facilitate the backup process by storing writes during it for every disk that's included in the backup. Once the backup is initialized, libvirt creates a Unix domain socket that exposes the NBD export internally. A user-facing endpoint is then exposed to allow external components to interact with the backup, e.g., to retrieve the backup data or query the dirty bitmap of an export.
 
 Once the backup is initialized, the controller will pass a backup command to the virt-launcher via the virt-handler using a subresource, containing all the relevant information. Before the backup begins, an FSFreeze command will be issued to ensure file system consistency during the backup. This will be the default behavior, with an option to skip filesystem quiescing if desired.
 Then, [`virDomainBackupBegin`](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainBackupBegin) will be invoked, which, as documented in libvirt, starts a point-in-time backup job for the desired disks of the running domain. This job captures the domain's disks state at the time of initiation, allowing to then call FSThaw (if needed). This minimizes guest downtime and enables the backup to be fetched while the guest continues its workload.
 
-In `push` mode, the backup job in libvirt automatically terminates once all data has been successfully backed up. The controller will be notified when the job completes and will update the `VirtualMachineBackup` phase to `Done`. At this stage, the PVC will be detached from the VM and made available for user operations. Since the `VirtualMachineBackup` resource is no longer needed, it can be safely deleted.
+In `push` mode, the backup job in libvirt automatically terminates once all data has been successfully backed up. The controller will be notified when the job completes and will update the `VirtualMachineBackup` phase to `Done`. At this stage, the PVC will be detached from the VM and made available for user operations. Since the `VirtualMachineBackup` resource is no longer needed, it can be safely deleted
+
+In `pull` mode, the backup job does not terminate on its own, because there is no inherent signal that the transfer has completed. Instead, the `VirtualMachineBackupStatus` gains a `Ready` condition to indicate that the backup data is prepared and available for retrieval. At that point, the client should begin interacting with the exposed endpoints to pull the backup. When the client has finished, it must explicitly delete the corresponding `VirtualMachineBackup` resource to signal that the process is complete. This deletion initiates the cleanup of all associated artifacts, including aborting the active libvirt backup job and deleting the Unix socket inside the virt-launcher.
 
 **Backup-PVC hotplugging**
-As mentioned, the PVC that will store the backup output must be hot-plugged into the virt-launcher pod. We will leverage an existing approach already used in the memory dump process, where a volume is hot-plugged as a directory directly into the pod—without being attached to the VM as a disk.
+As mentioned, the PVC that will store the backup output in the case of `push` mode and scratch space file in the case of `pull` mode must be hot-plugged into the virt-launcher pod but without being attached to the guest. To achieve this, we will leverage [Utility Volumes](https://github.com/kubevirt/enhancements/pull/91).
 This is considered an implementation detail rather than an exposed API, as it is not user-configurable and will be handled automatically as part of the backup workflow.
 
 ### Incremental backup
@@ -286,21 +288,38 @@ Optional. A reference to a `VirtualMachineBackupTracker` CR. Required to enable 
 - `mode`<br>
 Optional. Should be either `push` or `pull`. Initially, as mentioned, only `push` will be allowed. If not specified, `push` will be the default behavior.
 - `pvcName` <br>
-Optional. Required in `push` mode. Specifies the name of the PVC where the backup output will be stored. A PVC with that name needs exist otherwise the backup will wait for it to exist.
+Required in `push` and `pull` mode. Specifies the name of the PVC where the backup output will be stored for `push` mode and where the scratch files will be created for `push` mode. A PVC with that name needs exist otherwise the backup will wait for it to exist.
 - `skipQuiesce` <br>
 Optional. If set to true, the VM's filesystem will not be quiesced before the backup. By default, the system attempts to quiesce the filesystem to ensure data consistency.
 - `forceFullBackup` <br>
 Optional. If a backupTracker is provided and this field is set to true, force a full backup (instead of incremental using `latestCheckpoint`). The `latestCheckpoint` will be updated with the new full backup checkpoint.
+- `tokenSecretRef`<br>
+Optional. Required in `pull` mode. Specifies the name of the Secret containing the authentication token for the backup export. 
+- `ttl`<br>
+Optional. Only relevant for `pull` mode. Specifies the time to live for the backup, defaults to 2 hours if not specified.
 
 **Status:**
 - `checkpointName`<br>
 The name of the checkpoint created for the current backup. This field is updated by the backup controller.
 - `conditions`<br>
-Represents the current state of the backup process, such as `Initializing`, `Done`, `Failed` or `Deleting`
+Represents the current state of the backup process, such as `Initializing`, `Progressing`, `Ready`, `Done`, `Failed`, or `Deleting`.
 - `backupType` <br>
 Either `Full` or `Incremental`. Indicates the backup's scope, allowing users to confirm it aligns with their expectations or understand any deviation.
+- `cert`<br>
+CA certificate used for connecting to the backup endpoints in pull mode.
+- `includedVolumes` <br>
+A list of the names of the volumes that were included for the backup, each entry containing:
+  - `name` <br>
+    The name of the volume that was included.
+  - `mapEndpoint` <br>
+    The URL for querying the bitmap of the volume.
+  - `dataEndpoint` <br>
+    The URL for querying and pulling the data of the volume.
+- `pullTunnelSecret` <br>
+The name of the secret that contains the token created by the backup controller that is used for authenticating the internal tunnel connection between the virt-launcher and the virt-exportserver.
 
-Example:
+
+Examples:
 ```yaml
 apiVersion: backup.kubevirt.io/v1alpha1
 kind: VirtualMachineBackup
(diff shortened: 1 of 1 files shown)


veps/sig-storage/incremental-backup.md
```diff
@@ -316,12 +327,34 @@ status:
...
+PVC name containing the backup is provided in the `VirtualMachineBackup` CR status. The backup output will be stored as sparsed qcow2 images, one per disk. In the `Restore` section we document an example of how to stitch these files together to construct a restorable image. To move the images, a data-mover pod can be spawned to attach to the PVC and copy the data over to a remote storage. After the data is moved, the PVC can be deleted.
+
+**Pull mode:**
+Clients connect via NBD to an endpoint exposed in the `VirtualMachineBackup` CR status. The connection is secured with TLS, and the backup proxy validates the client certificate against the trusted CA and expected CN. The client can choose which data to pull (e.g., individual disk exports). The NBD endpoint is exposed via a Kubernetes Service; external access requires the Service to be exposed outside the cluster (e.g., LoadBalancer or NodePort). The backup endpoint remains active until the `VirtualMachineBackup` resource is deleted.
```
ShellyKa13: do we need to explain here how to collect the backup? supposedly its still part of the backup process since thats how you eventually receive the backup output..
Also not sure if we want to say that the client can choose which data to pull cause Im not sure if we want them to use the backup any other way other then straight out the data of the backup.. I know that technically they can also only pull the CBT but they have no actual value if not taken exactly with the snapshot so I dont think we really want them to do it. We cant prevent them but probably dont want to advertise it?

ShellyKa13 (commented):
Looks good, I think the need to expend more about the Alternatives to be able to present all the options in a clear way to understand that the chosen M.O is indeed the best one

veps/sig-storage/incremental-backup.md
```diff
@@ -93,18 +93,19 @@ As outlined in Libvirt, there are two general modes for backup:
...
 
 In `push` mode, the backup job in libvirt automatically terminates once all data has been successfully backed up. The controller will be notified when the job completes and will update the `VirtualMachineBackup` phase to `Done`. At this stage, the PVC will be detached from the VM and made available for user operations. Since the `VirtualMachineBackup` resource is no longer needed, it can be safely deleted.
+In `pull` mode, the backup job does not terminate automatically upon completion. It is the client's responsibility to explicitly delete the corresponding `VirtualMachineBackup` resource to indicate that the backup process has finished. Doing so triggers the cleanup of all related backup artifacts, including aborting the active libvirt backup job, deleting the scratch PVC, and removing the Unix socket within the virt-launcher.
```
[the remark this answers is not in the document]
Acedus: Ack, I'll be adjusting to controller-provisioned.

veps/sig-storage/incremental-backup.md
Acedus: > do we need to explain here how to collect the backup? supposedly its still part of the backup process since thats how you eventually receive the backup output..

I think it's necessary to explain the way of interacting with the endpoint, maybe section isn't the right place for it? I couldn't find a better one (could also just create a new one), WDYT?

> Also not sure if we want to say that the client can choose which data to pull cause Im not sure if we want them to use the backup any other way other then straight out the data of the backup.. I know that technically they can also only pull the CBT but they have no actual value if not taken exactly with the snapshot so I dont think we really want them to do it. We cant prevent them but probably dont want to advertise it?

Sure, let's do that.

ShellyKa13: > I think it's necessary to explain the way of interacting with the endpoint, maybe section isn't the right place for it? I couldn't find a better one (could also just create a new one), WDYT?

Maybe it wasnt clear that I think it wasnt explained enough lol not that its not the right place. like go more into details of how to interact with NBD endpoint. not sure

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +400,86 @@ Disadvantages:
...
+
+**Cons**
+- Requires an additional Pod per backup to establish the connection.
+- Requires additional setup from backup providers and cluster administrators to expose (if needed) and secure the backup endpoint.
+
+### Alternatives
+
+#### Network Push
+
+The backup process begins by initializing the Unix socket. The virt-launcher opens a bidirectional channel to the socket and connects to a remote endpoint provided in the `VirtualMachineBackup` CR. Once connected, it orchestrates a similar process to push mode by transferring all disk backups to the remote endpoint over the established channel.
```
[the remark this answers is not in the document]
Acedus: > When you say similar process to push mode, we have established you cant have the libvirt backup process run in push mode. So you basically call pull in the pod and then stream the output to the channel right?

Pretty much.

> which protocol will you use to to this process?

AFAIK, we should be able to have the virt-launcher use an NBD client to stream the disk data directly to the remote endpoint over TCP. The key requirement is to ensure that all available exports are streamed in a way that keeps them distinguishable. To achieve this, we could wrap the streams using a tar writer or a similar mechanism (definitely open to more suggestions, though).

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +400,103 @@ Disadvantages:
...
+
+    %% 2: virt-launcher connects to backup proxy via Service
+    CH -- "2. Connects to internal port" --> SVC -- Internal Port <--> BP
+
+    %% 3: Client accesses using external port
+    Client -- "3. Client accesses using external port" --> SVC -- External Port <--> BP
+```
+
+**Security considerations and approaches**
+1. TLS with client authentication is used to secure the transport layer and verify client identity. The backup proxy exposes a TLS-enabled endpoint using its own certificate and key. The client provides a Secret containing the CA certificate and the expected Common Name (CN) of the client certificate. The client implementation remains minimal and does not perform server certificate verification, relying solely on TLS encryption and server-side client authentication to establish a trusted channel.
```
mhenriks: Why should the client not validate the server CA?

Acedus: It's up to the backup provider to implement that validation, not necessarily discouraged just optional right now.

veps/sig-storage/incremental-backup.md
```diff
@@ -316,12 +327,34 @@ status:
...
 
+```yaml
+apiVersion: backup.kubevirt.io/v1alpha1
+kind: VirtualMachineBackup
+metadata:
+    name: backup1
+    namespace: ns1
+spec:
+    backupTracker: my-backup-tracker
+    mode: pull
```
hao-zhou-rubrik: Hi, I'm looking at this from a backup-vendor perspective, integrating with pull-mode backups via the NBD endpoint. A few clarifications would be helpful:

1. My understanding is that in `pull` mode the `NBD` export always presents the full block device view of the disk (offset 0 → end), and that: Full backups read the entire export, Incremental backups still see the full disk, but the client uses the dirty-bitmap to decide which ranges to read.
   - Can you confirm that this is correct, and `pull` mode does not expose a "delta-only" export format?
   - How will the dirty bitmap be exposed to the client side? NBD metadata context?

2. The example shows that a single `status.endpoint` is exposed for the backup. 
   - Can you confirm that this also applies to VMs with multiple disks? I.e., a single NBD endpoint per VirtualMachineBackup, and each VM disk is exposed as a separate NBD export on that same endpoint?
   - Will there be one dirty bitmap per disk on its export, or is the changed-block information represented in some other way?

3. I see the filesystem-consistent snapshot was discussed in the doc. Any options for application consistency?

4. It was mentioned that `it is the client's responsibility to explicitly delete the associated VirtualMachineBackup resource` and also `only one backup allowed per VM at a time`
   - In the `pull` mode, if a backup vendor takes longer to ingest a backup, or didn't clean up the backup CR due to some issues, will this block all new backups on the VM?

veps/sig-storage/incremental-backup.md
```diff
@@ -316,12 +327,34 @@ status:
...
+PVC name containing the backup is provided in the `VirtualMachineBackup` CR status. The backup output will be stored as sparsed qcow2 images, one per disk. In the `Restore` section we document an example of how to stitch these files together to construct a restorable image. To move the images, a data-mover pod can be spawned to attach to the PVC and copy the data over to a remote storage. After the data is moved, the PVC can be deleted.
+
+**Pull mode:**
+Clients connect via NBD to an endpoint exposed in the `VirtualMachineBackup` CR status. The connection is secured with TLS, and the backup proxy validates the client certificate against the trusted Certificate Authority (CA) and expected Common Name (CN). The NBD endpoint is exposed via a Kubernetes Service; external access requires the Service to be exposed outside the cluster (e.g., LoadBalancer or NodePort). The backup endpoint remains active until the `VirtualMachineBackup` resource is deleted. Clients should interact with the exposed endpoint via an NBD client as defined in [The NBD Protocol](https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md) documentation.
```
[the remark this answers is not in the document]
carlbraganza: Thanks for confirming. It would be nice if you added such details to the proposal.

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +400,103 @@ Disadvantages:
...
+    %% 2: virt-launcher connects to backup proxy via Service
+    CH -- "2. Connects to internal port" --> SVC -- Internal Port <--> BP
+
+    %% 3: Client accesses using external port
+    Client -- "3. Client accesses using external port" --> SVC -- External Port <--> BP
+```
+
+**Security considerations and approaches**
+1. TLS with client authentication is used to secure the transport layer and verify client identity. The backup proxy exposes a TLS-enabled endpoint using its own certificate and key. The client provides a Secret containing the CA certificate and the expected Common Name (CN) of the client certificate. The client implementation remains minimal and does not perform server certificate verification, relying solely on TLS encryption and server-side client authentication to establish a trusted channel.
+2. Securing the backup proxy endpoint may also be delegated to the cluster operator, using an L4 ingress that provides traffic encryption and routes to the exposed Service.
```
carlbraganza: What about cross-namespace access to this endpoint? Will the backup controller create the necessary **NetworkPolicy** object for a backup application running in another namespace to connect to the NBD service? It may be necessary to provide the backup application's desired namespace in the input arguments in order to set the NamespaceSelector field of an ingress NetworkPolicyPeer.

Ideally the NetworkPolicy object should be owned by your associated runtime pod so that it (and the Service) can be cleaned up properly.

veps/sig-storage/incremental-backup.md
```diff
@@ -93,18 +93,19 @@ As outlined in Libvirt, there are two general modes for backup:
...
+In `push` mode, the backup job in libvirt automatically terminates once all data has been successfully backed up. The controller will be notified when the job completes and will update the `VirtualMachineBackup` phase to `Done`. At this stage, the PVC will be detached from the VM and made available for user operations. Since the `VirtualMachineBackup` resource is no longer needed, it can be safely deletedIn `pull` mode, the backup job does not terminate automatically since there's nothing to indicate completion. Instead, once the backup can be retrieved by a client, the `VirtualMachineBackupStatus` will be changed to `Ready` field and a `Ready` condition. At that point, it is the client's responsibility to explicitly delete the associated `VirtualMachineBackup` resource to signal that the backup process has concluded. This deletion triggers the cleanup of all related artifacts, including aborting the active libvirt backup job, removing the controller-provisioned scratch PVC, and deleting the Unix socket inside the virt-launcher.
```
ShellyKa13: accidental double past of `In pull mode... ` which is also in the next line?

veps/sig-storage/incremental-backup.md
```diff
@@ -291,16 +292,22 @@ Optional. Required in `push` mode. Specifies the name of the PVC where the backu
...
 **Status:**
 - `checkpointName`<br>
 The name of the checkpoint created for the current backup. This field is updated by the backup controller.
 - `conditions`<br>
-Represents the current state of the backup process, such as `Initializing`, `Done`, `Failed` or `Deleting`
+Represents the current state of the backup process, such as `Initializing`, `Progressing`, `Ready`, `Done`, `Failed`, or `Deleting`.
 - `backupType` <br>
 Either `Full` or `Incremental`. Indicates the backup's scope, allowing users to confirm it aligns with their expectations or understand any deviation.
+- `includedVolumes` <br>
+A list of the names of the volumes that were included for the backup.
```
ShellyKa13: I believe its better to also include the endpoints from which the users can pull the disks over. I see in the explanations below that you intended to have the user get this info from vmexport. I think we would prefer to have all the relevant info for backup in the same crd of the virtualmachinebackup instead of refering the user to get another CR of the export and get the data from there.. I know its basically copy the info that is in the vmexport status to the vmbackup status but I think the vmexport is somewhat of and implementation detail on how we expose the endpoints and dont want to bother the user with extra info it doesnt need to know exists. WDYT?

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +395,229 @@ Disadvantages:
...
+  - Internal Listener: The server is configured to listen on a protected internal port (9090) for the backup tunnel.
+  - External Listener: The server listens on the standard external port (8443) for user HTTP traffic.
+  - Configuration: The controller injects the included volume list (from the `VirtualMachineBackupStatus`) into the pod's environment variables.
+5. Once the export service is ready, the backup controller retrieves the service's name and the internal CA certificate. It then issues yet another `Backup` RPC, this time with the `Connect` cmd to the virt-launcher.
+6. The virt-launcher acts as the TLS client. It establishes a persistent, self-healing TLS tunnel to the virt-exportserver's internal port. This reverse-tunnel architecture ensures no network changes are required for the virt-launcher pod.
+7. Data Serving:
+  - The virt-exportserver registers dynamic HTTP handlers for every volume included in the backup under the new `Backups` status field..
+  - When a client requests a specific endpoint, the server translates this into a JSON-RPC command (`read`, `map`, or `info`) and sends it over the TLS tunnel.
```
ShellyKa13: If you agree with my previous comment probably need to add here update of the backup status with the endpoints. while also updating the condition `Ready` to true

ShellyKa13 (commented):
Over all for me the design sounds good!
Just a couple of suggestions mostly with regards of how to report the endpoints to the user

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +404,232 @@ Disadvantages:
...
+5. Once the export service is ready, the backup controller retrieves the service's name and the internal CA certificate. It then issues yet another `Backup` RPC, this time with the `Connect` cmd to the virt-launcher.
+6. The virt-launcher acts as the TLS client. It establishes a persistent, self-healing TLS tunnel to the virt-exportserver's internal port. This reverse-tunnel architecture ensures no network changes are required for the virt-launcher pod.
+7. Data Serving:
+  - The virt-exportserver registers dynamic HTTP handlers for every volume included in the backup under the new `Backups` status field.
+  - The backup controller recognizes that the `VirtualMachineExport` is serving and updates the `Ready` condition of the `VirtualMachineBackupStatus` to true.
+  - The `VirtualMachineBackupStatus` includes the created URLs for every volume in its `includedVolumes`'s list by populating the `mapEndpoint` and `dataEndpoint` fields respectively.
+  - When a client requests a specific endpoint, the server translates this into a JSON-RPC command (`read`, `map`, or `info`) and sends it over the TLS tunnel.
```
mhenriks: Why JSON RPC and not REST? I'm not necessarily opposed to JSON RPC. May make things a slightly more complicated for clients using curl. I don't think we use it anywhere else in kubevirt.

Maybe if we have virtctl wrappers for client API it is less of a concern. Have you thought about that (regardless of API implementation)?

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +404,232 @@ Disadvantages:
...
+3. Once the backup job is progressing, the backup controller automatically creates an owned `VirtualMachineExport` CR, setting its source to the `VirtualMachineBackup`.
+4. The export-controller reconciles the new `VirtualMachineExport`. It identifies the source type and launches the virt-exportserver pod.
+  - Internal Listener: The server is configured to listen on a protected internal port (9090) for the backup tunnel.
+  - External Listener: The server listens on the standard external port (8443) for user HTTP traffic.
+  - Configuration: The controller injects the included volume list (from the `VirtualMachineBackupStatus`) into the pod's environment variables.
+5. Once the export service is ready, the backup controller retrieves the service's name and the internal CA certificate. It then issues yet another `Backup` RPC, this time with the `Connect` cmd to the virt-launcher.
+6. The virt-launcher acts as the TLS client. It establishes a persistent, self-healing TLS tunnel to the virt-exportserver's internal port. This reverse-tunnel architecture ensures no network changes are required for the virt-launcher pod.
```
mhenriks: I assume the virt-launcher uses the the same token as external clients to connect to the exportserver pod? That's probably fine but I wonder if it would be better to somehow have a separate token for this connection (one that is not created by the user)

veps/sig-storage/incremental-backup.md
```diff
@@ -93,18 +93,20 @@ As outlined in Libvirt, there are two general modes for backup:
...
-* In the second phase, we will implement **pull** mode. This will involve defining a complete network API that exposes an endpoint the user can connect to in order to retrieve the backup.
->> Note: More details on the pull mode will be described in a subsequent VEP.
+* **Pull mode**: In this mode, an estimation of the backup size is performed beforehand and a filesystem-based PVC is provisioned by the backup controller. The provisioned PVC serves as temporary storage for a scratch file used to facilitate the backup process by storing further writes during it. Once the backup is initialized, libvirt creates a Unix domain socket that exposes the NBD export internally. A user-facing endpoint is then exposed to allow external components to interact with the NBD server, e.g., to retrieve the backup data or query the dirty bitmap of an export.
```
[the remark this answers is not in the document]
Acedus: > How would the controller estimate backup size?

The currently proposed heuristic is: "A sum of 10% of each included volume will be used to calculate the size.", we could decide to defer sizing to the user, but then we'd need to decide how we want the sizing logic to work since there are a few options as I see it:
1. The user sets the exact capacity (e.g., 10Gi).
2. The user sets a % of the sum of the total capacity.
3. The user explicitly selects which volumes will be included in the backup and _can_ configure their scratch capacity.

WDYT?

veps/sig-storage/incremental-backup.md
Acedus: > I assume the virt-launcher uses the the same token as external clients to connect to the exportserver pod?

No, we can't provide the virt-launcher with the user provided token since that would require us to read the secret which is something we can't do due to lack of sufficient permissions.

> That's probably fine but I wonder if it would be better to somehow have a separate token for this connection (one that is not created by the user)

I did think about that too, the main issue with a runtime generated token is that the export controller would need to be aware of it and add it as a mounted secret and in order for configuration to persist we'd need to add an additional field to the VMExport API to specifically support that. I don't see it as too big of an issue, but at the same time I'm not sure it's worth it to further secure a connection that's already namespace-bound and TLS encrypted.

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +404,232 @@ Disadvantages:
...
+    Launcher->>ES: TLS Dial (Port 9090)
+    activate ES
+    ES-->>Launcher: Handshake OK
+    Note right of Launcher: Persistent Bidirectional Tunnel Established
+
+    %% Phase 3: The Data Pull
+    Note over User, Launcher: Phase 3: Data Streaming
+    User->>ES: HTTP GET /exports/disk1/map
+    
+    ES->>Launcher: Tunnel JSON {op: "map"...}
```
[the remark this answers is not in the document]
Acedus: The general idea I had looks something like this:
```go
type TunnelCommand struct {
	Op     string `json:"op"`
	Disk   string `json:"disk"`
	Offset uint   `json:"offset"`
	Length uint   `json:"length"`
}
```

Where valid operations are `map` and `read` for obtaining the bitmap and reading the data respectively.
And then we'd have this for the virt-exportserver:
```go
func executeRPC(client *NbdClient, op, disk string, offset int64, length int) ([]byte, error) {
	client.mu.Lock()
	defer client.mu.Unlock()

	cmd := TunnelCommand{Op: op, Disk: disk, Offset: offset, Length: length}

	if err := json.NewEncoder(client.conn).Encode(cmd); err != nil {
		return nil, fmt.Errorf("tunnel write error: %v", err)
	}

	header := make([]byte, 9)
	if _, err := io.ReadFull(client.conn, header); err != nil {
		return nil, fmt.Errorf("tunnel header read error: %v", err)
	}

	status := header[0]
	payloadLen := int64(binary.BigEndian.Uint64(header[1:]))

	payload := make([]byte, payloadLen)
	if _, err := io.ReadFull(client.conn, payload); err != nil {
		return nil, fmt.Errorf("tunnel payload read error: %v", err)
	}

	if status != StatusSuccess {
		return nil, fmt.Errorf("remote error: %s", string(payload))
	}

	return payload, nil
}
```

We could also just decide to use gRPC bidirectional streaming to have an easier time, WDYT?

veps/sig-storage/incremental-backup.md
mhenriks: We can't let just anyone connect to exportserver and serve up backup data. Also have to make sure that virt-launcher has the exportserver CA.

I was assuming that the `BackupConnect()` RPC was passing the exportserver CA and the token as params. But I just realized that it won't have permission to read the secret from a token. So we need a way to have exportserver accept connections from valid virt-launchers

To secure virt-launcher->exportserver connection on client side, virt-launcher should get the exportserver CA cert. Can pass this to `BackupConnect()` correct? May have to somehow deal with exportserver cert rotation though. Exportserver exits before server cert expires and export server will recreate the pod with a new cert

To secure virt-launcher->exportserver connection on server side, I think we have to either give virt-launcher a client cert that is validated by exportserver or have a token. If we have to extend the export api, so be it. But maybe it would be enough to have export server only accept connections for this endpoint from virt-launcher IP (could be an env var set by backup controller)

Acedus: Good points.
So as you correctly surmised, we pass the exportserver CA via the `Backup` RPC using the `Connect` CMD.
I find relying on pod IP a bit too fragile, supposedly it shouldn't change for the lifetime of the export (if it changes it necessarily means that the backup concluded since the original virt-launcher pod has terminated) but it relies on CNI behavior which varies. 

The main issue I have with a runtime generated token for the internal connection is idempotency. We'd need to initiate the connect request and have it succeed before the creation of the VMExport, otherwise we risk a scenario where we've created the VMExport with the generated token, proceeded to issuing the connection RPC but failed, and that will render the token unusable since it is stateless as far as the backup controller is concerned. I don't necessarily see that as too much of an issue, the virt-launcher backup tunnel is designed to self-heal and reattempt connections every set duration, so at worst it'll just spam connection attempts before the export-server is ready. WDYT?

veps/sig-storage/incremental-backup.md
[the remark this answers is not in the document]
Acedus: So in that case I suppose we can just rely on the existing `pvcName` field and just leverage it differently depending on the backup mode.

veps/sig-storage/incremental-backup.md
Acedus: As always, you're one step ahead :smile:, I thought about using HMAC, just wasn't sure about what the key would be. I suppose that we could create a key secret in the kubevirt namespace and use that, let's do that!

veps/sig-storage/incremental-backup.md
```diff
@@ -367,6 +456,231 @@ Disadvantages:
...
+    Launcher-->>ES: Tunnel JSON response
+    
+    ES-->>User: HTTP 200
+    deactivate ES
+```
+
+**Security considerations and approaches**
+- Both internal traffic and external traffic are both secured using the existing export certificate API.
+- Authorization for external traffic is done via a user-provided token.
+- Authorization for the internal tunnel is established using a deterministic, runtime-generated token. This token is derived via HMAC using the Backup UID as the message and a lifecycle-managed key within the kubevirt namespace (whose lifecycle is bound to the `IncrementalBackup` feature gate status). To facilitate verification, the token is projected into a secret within the target backup namespace, allowing the virt-exportserver mount it and validate the incoming connection request.
```
mhenriks: Do we have to extend export API to support this token?

mhenriks (commented):
Looking good!

veps/sig-storage/incremental-backup.md
Acedus: Yes, we will, otherwise we won't have a way of telling the export controller which secret to mount. With that said, we can have a utility function that's responsible for deterministically deriving the secret name, that way we can avoid extending the export API and only alter the backup source handling logic. WDYT?

Acedus: Not sure I like that, adding the internal secret name as a status field exposes an implementation detail.

Acedus: I guess it should be fine, we have some other components in KV doing similar things.

Issue #135, opened 2025-11-20
VEP 135: Automatic NUMA passthrough enablement for VMs based on node topology


ksimon1:
**Primary contact (assignee)**:
/assign @ksimon1

**Current Feature Stage**:
New

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:
TODO

**Timeline**:
* Alpha release target (x.y): 1.8
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #136, opened 2025-11-20
VEP #134: Remove dependency between NUMA passthrough and Huge pages

ksimon1:
### VEP Metadata

**Tracking issue**: #134 
**SIG label**: /sig compute

### What this PR does
This VEP proposes to remove a specific validation check within KubeVirt's admission logic that currently rejects a Virtual Machine Instance (VMI) when a NUMA passthrough is enabled, but the VMI has not explicitly requested HugePages.

Issue #134, opened 2025-11-20
VEP 134: Remove dependency between NUMA passthrough and Huge pages


ksimon1:
**Primary contact (assignee)**:
/assign @ksimon1

**Current Feature Stage**:
New

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/136

**Timeline**:
* Alpha release target (x.y): 1.8
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #137, opened 2025-11-20
VEP #135: Automatic VM NUMA passthrough Enablement Based on Node Topology

ksimon1:
### VEP Metadata

**Tracking issue**: #135 
**SIG label**: /sig compute

### What this PR does
This proposal proposes an automated mechanism to enable vNUMA passthrough within a VirtualMachineInstance (VMI) when the VMI's requested resources (vCPUs or memory) have exceeded the capacity of a single physical NUMA node on the host.

### DIFF
--- a/veps/sig-compute/135-automatic-NUMA/automatic-numa.md
+++ b/veps/sig-compute/135-automatic-NUMA/automatic-numa.md
@@ -1,0 +1,119 @@
+# VEP #135: Automatic VM NUMA passthrough Enablement Based on Node Topology
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal proposes an automated mechanism to enable NUMA passthrough within a VirtualMachineInstance (VMI) when the VMI's requested resources (vCPUs or memory) have exceeded the capacity of a single physical NUMA node on the host. This feature will improve the performance of large VMs by reducing cross-NUMA latency and making the underlying topology transparent to the guest VM, without requiring manual user configuration. The behavior will be controllable via a KubeVirt configuration flag.
+
+## Motivation
+
+Currently, users running large VMs on multi-NUMA host machines must manually configure the NUMA passthrough in the VMI spec. Failing to do so, it may result in poor performance due to the operating system inside the VM not being aware of the underlying physical NUMA boundaries, which leads to frequent cross-NUMA access.
+
+This manual process faces the following drawbacks:
+
+1. **Error-prone**: Users might forget the necessary NUMA passthrough configuration.  
+2. **Sub-optimal**: KubeVirt should strive to automatically provide the best performance configuration for the resources requested.
+
+Automating this configuration will provide performance benefits and a better out-of-the-box experience for users running large workloads.
+
+## Goals
+
+- Introduce an KubeVirt configuration field to globally enable or disable this automatic NUMA enablement feature.  
+- Implement a mechanism in node-labeller to discover the host's physical NUMA topology (specifically, the maximum memory and CPU count of a single NUMA node).  
+- Report the discovered NUMA topology information on the Kubernetes Node object (via annotations or labels).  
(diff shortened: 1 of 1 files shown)


veps/sig-compute/135-automatic-NUMA/automatic-numa.md
```diff
@@ -0,0 +1,119 @@
...
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal proposes an automated mechanism to enable NUMA passthrough within a VirtualMachineInstance (VMI) when the VMI's requested resources (vCPUs or memory) have exceeded the capacity of a single physical NUMA node on the host. This feature will improve the performance of large VMs by reducing cross-NUMA latency and making the underlying topology transparent to the guest VM, without requiring manual user configuration. The behavior will be controllable via a KubeVirt configuration flag.
```
dominikholler: Would the feature be simpler if the automatic enablement would be not node dependent?

veps/sig-compute/135-automatic-NUMA/automatic-numa.md
```diff
@@ -0,0 +1,119 @@
...
+- As a Developer deploying a large VM (e.g. 64 vCPUs, 256Gi memory), I want the NUMA passthrough feature to be automatically enabled for my VM, so I don't have to manually tune the spec for best performance.
+- As a Cluster Admin, I want `node-labeller` to reliably report the max number of cpu and size of memory in single NUMA, so that the admission controller can make correct decisions about NUMA enablement.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+A new configuration field will be introduced in the `KubeVirtConfiguration` to act as a switch for this feature.
```
dominikholler: What might be the reasons for not enabling the feature?

veps/sig-compute/135-automatic-NUMA/automatic-numa.md
```diff
@@ -0,0 +1,119 @@
...
+
+```
+spec:
+  configuration:
+    developerConfiguration:
```
lyarwood: As discussed previously this shouldn't sit under `developerConfiguration`

mtessun:
> * Relying on Kubernetes Topology Manager for auto-alignment is solid (and it can work without a device anchor like GPUs or SR-IOV), but in practice, CPU allocations could still result in fragmented topologies, especially in loaded heterogeneous clusters. For large VMs, this might lead to uneven guest NUMA nodes, complicating application optimizations and increasing latency. Perhaps this needs to be enabled only if the VM requests more than 50% of the node resources to target those "special" large cases?

Well the question is: If the guest kernel knows about the physical NUMA even if it is "unevenly" allocated, doesn't the kernel still do a better job with scheduling?

> * The VMI spec already allows setting a fine-grained CPU topology (sockets/cores/threads) but lacks a generic vCPU field, the resulting topology could feel inconsistent. Users might end up with unexpected passthrough enabled, overriding their intent.

Ack. Ideally the CPU topology should be automated as well - mimicking the physical one as close as possible but still giving the possibility to run NUMA, e.g. first fill up the sockets similar to physical sockets and then the cores. Finally the hyperthreads.

> * Another issue is that there is an ongoing effort to integrate with DRA (especially with the DRA CPU driver) - this could provide more granular NUMA-aware claims and scheduling. I'm worried that it might create overlap.
> 
> What really increases my concern is the combination with the follow-up in #136 (VEP #134: Remove dependency between NUMA passthrough and HugePages). Decoupling NUMA passthrough from HugePages makes sense for flexibility, but auto-enabling it without requiring HugePages could increase the performance risks. HugePages often go together with NUMA for efficiency, and without them, VMs might see even more cross-NUMA penalties or allocation failures. HugePages are NUMA-namespaced, so it was easy from a scheduling perspective in Kubernetes. In special large-VM cases, users typically pair them anyway...

Well NUMA doesn't necessarily require Hugepages and the host kernel should do proper scheduling anyways, so the question is rather does the guest kernel do the same? Likely not.
What I would be curious about is the performance impact if the guest kernel is NUMA aware or if the config is the same but as UMA.
If there is no difference, then this configuration should be requested on a per VM basis (and likely including Hugepages) but still simplify the configuration to "host-copy" or something.

Issue #125, opened 2025-11-03
VEP 125: Clarify/Improve usage of shared disks in VMs


gaohoward:
**hgao@redhat.com**:

/assign hgao

**Current Feature Stage**:
The VM spec supports shared disks among VMs. There is few information regarding how to properly use it. It is important to have something in the document and code to make sure people use it properly.
1. we need to make clear to users how to properly uses the shared disks to avoid issues like syncing and data loss as describe in the following.
2. When doing snapshot of shared disks what to expect regarding the sync issue, i.e. snapshot of one VM while other vms are reading/writing the shared disks, also pehaps for vm exporting
3. When shared disks gets deleted by the owner vm, it affects other vms that use this shared disks. How to prevent that from happening.
4. we have already some issues that related to this

https://issues.redhat.com/browse/CNV-53145
https://issues.redhat.com/browse/CNV-56067

We need to have some thoughts on those topics around the shared disks feature.

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig $SIG

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #139, opened 2025-11-24
VEP 125: Clarify/Improve usage of shared disks in VMs (#125)

gaohoward:
The VEP is aimed at improving the handling of shared disks in KubeVirt, including but not limited to:

  1. Better documentation on the usages, limitations and warnings. Currenly there is little.
  2. Add necessary validations and improvements to avoid wrong usages
  3. Take measures to protect shared disks from being accidentally deleted

### VEP Metadata

**Tracking issue**: 
**SIG label**:

### DIFF
--- a/veps/sig-compute/125-shared-disks/shared-disks.md
+++ b/veps/sig-compute/125-shared-disks/shared-disks.md
@@ -1,0 +1,189 @@
+# VEP #125: Clarify/Improve usage of shared disks in VMs
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+<!--
+Provide a brief overview of the topic)
+-->
+[The KubeVirt VM spec](https://github.com/kubevirt/kubevirt/blob/ada6d32348b68cf82956c7f711549c8f1aea8054/staging/src/kubevirt.io/api/core/v1/schema.go#L724) allows a user to specify whether a disk device is shareable.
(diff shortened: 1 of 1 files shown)


mhenriks:
Do we really need a VEP for this? Sounds like bug fixes

veps/sig-compute/125-shared-disks/shared-disks.md
```diff
@@ -0,0 +1,189 @@
...
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+<!--
+Provide a brief overview of the topic)
+-->
+[The KubeVirt VM spec](https://github.com/kubevirt/kubevirt/blob/ada6d32348b68cf82956c7f711549c8f1aea8054/staging/src/kubevirt.io/api/core/v1/schema.go#L724) allows a user to specify whether a disk device is shareable.
+
+Shareable disks can used by applications (like clustered database servers) running in VMs in a cluster. However not very use-case
```
jean-edouard:
```suggestion
Shareable disks can used by applications (like clustered database servers) running in VMs in a cluster. However not every use-case
```

jean-edouard (commented):
As-is, I agree with @mhenriks that this VEP isn't necessary.
The only point that could be discussed is preventing the deletion of VM disks.
But even for that point, I suggest you bring it up to a sig-storage meeting first before reworking this VEP around it.

@vladikr I suggest we close this VEP for now.

xpivarc:
As of now me @vladikr doesn't seem a need for VEP as there are no API or behavior changes planed as far as we understand. We appreciate if you want to have a VEP merged for this feature as dev documentation...

Issue #10, opened 2025-02-25
VEP NNNN: Support DRA devices in KubeVirt VMIs


alaypatel07:
**Primary contact (assignee)**:

/assign @alaypatel07 

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

/sig compute

**Enhancement link**:

**Timeline**:

* [] Alpha (v1.6.0/TBD)
* [] Beta (TBD)
  * [] Additional Beta criteria
* [] GA (TBD)
  * [] Additional GA criteria

**Additional context**:

alaypatel07:
cc @vladikr @xpivarc @jean-edouard @orelmisan 

As promised, created this VEP, PTAL and guide if I got anything wrong.

alaypatel07:
also cc @rthallisey @varunrsekar

aglitke:
@akalenyu mentioned that this VEP will have some storage integrations so adding sig-storage here.

/sig sig-storage

alaypatel07:
@aglitke eventually, for the alpha version of this feature, storage stuff like nvme is explicit non-goal, but we would like to eventually support it in one of the beta+ versions. So not a lot of interesting stuff for sig-storage specifically in v1.6.

iholder101:
Hey @alaypatel07!

We've been discussing this in the sig-compute weekly meeting.
Would you mind sharing a status of this VEP? Is there a PR? Are you blocked by anything? Can we help in any way?

Thank you!

alaypatel07:
@iholder101 https://github.com/kubevirt/kubevirt/pull/14365

I have a draft PR open with an up-to- date checklist of this implemented and pending. I believe it will be a big PR to review so I am organizing the work in reviewable commits.

Not blocked on anything. Will market it ready when the checklist is complete.

alaypatel07:
There was a discussion on Jan 12 here is a summary with regards to this VEP:
1. Should this VEP be considered for Beta in release 1.8?
    @EdDev @vladikr @iholder101 @enp0s3 and I discussed the following:
   - The VEP 10 beta design here: https://github.com/kubevirt/enhancements/pull/156 proposes that an existing status block be removed and the drivers uses a standard CDI mount path to populate the relevant data, hence it should not qualify for beta.
[...]

PR #158, opened 2025-12-17
VEP-10 (extension): Add SR-IOV Network DRA support

oshoval:
### VEP Metadata

First 2 commits are cherry-pick of other PRs.

**Tracking issue**: 
https://github.com/kubevirt/enhancements/issues/10

**SIG label**: 
sig-network

### What this PR does

Extends the DRA devices VEP to document support for DRA-provisioned network
devices, specifically SR-IOV NICs.

Thanks @SchSeba for the co-op, and to all who contributed
See https://github.com/kubevirt/kubevirt/issues/15995

Key additions:
- ResourceClaimNetworkSource API for specifying DRA networks in spec.networks
- SR-IOV integration details: device allocation via DRA, configuration via
  existing KubeVirt networks API.
- Custom MAC address support through `kubevirt.io/dra-network-macs` annotation.
- DRANetworkDevices feature gate (Alpha).
- Example VMI YAML with DRA SR-IOV network configuration.

Network DRA maintains mutual exclusivity with traditional Multus-based SR-IOV
per VM. The existing Multus SR-IOV API remains fully supported and unchanged.

Migration is not included in this stage yet.
It will has its own VEP, more info in the vep.

Didn't include yet about how we detect there is no mix of types (i.e requested SR-IOV and got GPU).
It is still an open issue, mentioned it on the CDI design.

### DIFF
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -25,6 +25,9 @@ control of their devices using Virtual Machines and Containers.
 ## Non Goals
 
 - Replace device-plugin support in KubeVirt
+- Replace existing SR-IOV network integration (Multus-based approach remains fully supported)
+- Deploy DRA SR-IOV driver (will be done as part of sriov-network-operator)
+- Support co existance of both DRA SR-IOV driver and Device plugin
 - Align on what drivers KubeVirt will support in tree
 - Address any complexity issues when DRA APIs in KubeVirt
 
@@ -50,7 +53,8 @@ control of their devices using Virtual Machines and Containers.
 1. Devices where the DRA driver sets agreed upon attributes for resources in ResourceSlice.
     1. PCIe Bus address for pGPU.
     2. Mediated device uuid for vGPU.
-1. Devices have to either be of the type GPU or HostDevices.
+    3. PCIe Bus address for SR-IOV network devices.
+2. Devices have to be of the type GPU, HostDevices, or Network devices (SR-IOV or Scalable Funtions in the future).
 
 ### Future Usecases
 1. NVMe devices https://github.com/kubevirt/community/issues/254
@@ -62,14 +66,10 @@ control of their devices using Virtual Machines and Containers.
    be supported.
     1. In the future a standardization could be envisioned for a PCI device where the attributes are set
        automatically through the DRA framework. When this is achieved, any DRA device should be available in KubeVirt VMs
-2. NetworkDevices or Storage Devices which need a lifecycle of its own will not be supported through HostDevices in alpha
-   release.
+2. Storage Devices which need a lifecycle of their own will not be supported through HostDevices in alpha release.
    - For storage, NVMe devices may need to persist beyond pod deletion (e.g., during VM pause). However, current K8s 
      DRA implementation deletes resource claims when the virt-launcher pod is deleted. Managing claims beyond pod 
-     lifecycle (for example VM pause case) is out of scope for this design. 
-   - For network devices, there is already existing mechanism of configuring the network, any DRA network device will have
-     to work existing functionality. There is a need to explore ideas on how to best achieve it potential in sig-network.
-     Hence, it is out of scope, will have to be picked up as a separate VEP
+     lifecycle (for example VM pause case) is out of scope for this design.
 
 ## Repos
 
@@ -82,13 +82,14 @@ For allowing users to consume DRA devices, there are two main changes needed:
 1. API changes and the plumbing required in KubeVirt to generate the domain xml with the devices.
 2. Driver Implementation to set the required attributes for KubeVirt to use
 
-This design document focuses on part 1 of the problem. This design is introducing two new feature gates:
+This design document focuses on part 1 of the problem. This design is introducing three new feature gates:
 1. GPUsWithDRA
 2. HostDevicesWithDRA
-All the API changes will be gated behind either one of these feature gates so as not to break existing functionality.
+3. DRANetworkDevices
 
-Both the GPUs as well as HostDevices have separate lifecycle but essentially have the same API, hence two separate
-feature gates are required.
+All the API changes will be gated behind one of these feature gates so as not to break existing functionality.
+
+GPUs, HostDevices, and NetworkDevices have separate lifecycles but share similar DRA integration patterns. The three separate feature gates allow independent control of each device type.
 
 ### API Changes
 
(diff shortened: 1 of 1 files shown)


veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -116,50 +115,52 @@ type VirtualMachineInstanceSpec struct {
 type GPU struct {
 	// Name of the GPU device as exposed by a device plugin
 	Name string `json:"name"`
-	// DeviceSource is the name of the device provisioned either by device plugins
-	// or by DRA enabled device
-	// DeviceName string `json:"deviceName"`    <-- inlined into DeviceSource
-	DeviceSource      DeviceSource `json:",inline"`
+	// DeviceName is the name of the device provisioned by device-plugins
```
SchSeba: are you sure it's `device-plugins` here? show it be dra driver?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -183,67 +185,140 @@ type DeviceStatusInfo struct {
...
+(using `resourceClaim` network source) are **mutually exclusive per VM**. A single VMI should not mix both approaches.
+The existing Multus-based SR-IOV API remains fully supported and unchanged.
+
+##### Custom MAC Address Support
+
+To support custom MAC addresses for DRA-based SR-IOV networks, KubeVirt will annotate the virt-launcher pod with the 
+requested MAC addresses. The MAC address is taken from the existing `spec.domain.devices.interfaces[].macAddress` field:
+
+```
+kubevirt.io/dra-network-macs: '{"<claimName>/<requestName>":"de:ad:00:00:be:ef"}'
```
SchSeba: I prefert to have something more generic and not kubevirt specific.
maybe continue to use the already supported and battle tested CNI annotation


```
  annotations:
    k8s.v1.cni.cncf.io/networks: '[
      {
       "name": "<claimName>/<requestName>",
       "mac": "de:ad:00:00:be:ef",
      }
    ]'
```


but we can have multiple interfaces for the same requestName so maybe the mac can be a list (future implementation)

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -354,24 +433,33 @@ Note: The feature gate verification when VMI requesting DRA device will either b
...
-    1. The resourceslice corresponding to the node running the VMI has information about the allocated device.
+    2. The resourceClaim has information about the individual requests in the claim and their allocated device names.
+    3. The resourceslice corresponding to the node running the VMI has information about the allocated device.
 
 ### Virt launcher changes
 
 1. For devices generated using DRA, virt-launcher needs to use the vmi.status.deviceStatus to generate the domain xml
    instead of environment variables as in the case of device-plugins
-1. The standard env variables `PCI_RESOURCE_<deviceName>` and `MDEV_PCI_RESOURCE_<deviceName>` may continue to be set
+2. The standard env variables `PCI_RESOURCE_<deviceName>` and `MDEV_PCI_RESOURCE_<deviceName>` may continue to be set
```
SchSeba: I think in case of DRA we should stop using this one and start adding CDI information

SchSeba (commented):
nice!
I left some small comments
thanks!

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -171,6 +172,7 @@ type DeviceStatus struct {
 	// +optional
 	GPUStatuses []DeviceStatusInfo `json:"gpuStatuses,omitempty"`
```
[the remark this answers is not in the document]
oshoval: not sure tbh, but this is not in the scope of this VEP amendment please, the VEP is foucsed on network devices
lets have @alaypatel07 answer about it

veps/sig-compute/10-dra-devices/vep.md
oshoval: maybe reusing it will make co-existence with SR-IOV / multus harder
it already does problems https://github.com/kubevirt/kubevirt/issues/15480, since there are couple entities that update it.
moreover not sure we want to use `k8s.v1.cni.cncf.io` prefix for something kubevirt specific, 
lets discuss it deeper please

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -637,17 +807,40 @@ type PodResourceClaimStatus struct {
...
+### Network DRA Migration (Planned Approach)
+
+Live migration for network DRA devices will be supported in a follow-up implementation. The planned approach addresses
+the limitation that `vmi.status.deviceStatus` is not the appropriate place for pod-specific context during migration
+(when both source and target pods exist).
+
+Instead, CDI (Container Device Interface) will inject device information as a file into each pod. The file structure 
+will contain mappings of request/claim to PCI addresses:
+```
+request/claim: <list of pci addresses>
```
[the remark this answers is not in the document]
oshoval: ok lets open a draft about it and reference it
alternative is to write initial draft inline, and reference once we have an external doc

veps/sig-compute/10-dra-devices/vep.md
oshoval: as discussed offline, we will have different annotation but with the same structure, 
since otherwise it triggers multus (the annotation is the "opt-in" method for multus)

oshoval: Please see the updated one and the reasoning
https://github.com/kubevirt/enhancements/pull/158/files#diff-ca56d7cf7c1e00e1a9f18d95a66ad56535d734121d47fcddb0831dca511cc506R306

`kubevirt.io/dra-networks: '[{"claimName":"sriov","requestName":"vf","mac":"de:ad:00:00:be:ef"}]'`

> the interface name is yet another input of the network selection element. No need to add a list imho.

we want to keep the same structure, and there might be 2 tuples of (claimName,requestName) that require MAC address, hence we do need a list for those two reasons, if i understood you correctly

> what are you planning to do if in the future we want to pass more parameters along ?

in this VEP only MAC, this is the only missing feature comparing DRA to DP (ran all the current SR-IOV e2e tests beside the migration ones to map the missing features)
in the future if any parameter is needed, it can be added as parameter / cni-args to the annotation

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -183,67 +187,145 @@ type DeviceStatusInfo struct {
...
+```
+kubevirt.io/dra-networks: '[{"claimName":"sriov","requestName":"vf","mac":"de:ad:00:00:be:ef"}]'
+```
+It will preserve the structure of `k8s.v1.cni.cncf.io/networks`, but for claimName,requestName instead of NAD.
+
+The SR-IOV DRA driver reads this annotation and passes the claim/request identifier along with the MAC address to the SR-IOV CNI, ensuring the network interface is configured with the specified MAC address.
+
+Note: it is planned that Kubevirt will support "auto" mode, where the ResourceClaim is created by Kubevirt,
+and a ResourceClaim includes a config that supports MAC address field.
+Since Kubevirt also support cases where the ResourceClaim / ResourceClaimTemplate are created by the admin,
```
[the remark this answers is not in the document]
oshoval: Kubevirt does support now cases where the admin is creating the ResourceClaim
It is required because this is one of the major features of DRA, to allow having special constrains, for example a Claim can allocate two different devices on the same NUMA.

The "auto" is the unsupported yet case where no special constrains are needed, and on this case the ResourceClaim will be created by Kubevirt, it is still a plan, not implemented yet.
But the current manual method is here to stay, therefore we need a solution for that, which was also discussed to be the general solution since, instead for example having a webhook that alters admin's claims etc.

maiqueb:
/uncc

I'm afraid I don't have the mental energy to follow this.

oshoval:
note, will need to rebase on top of https://github.com/kubevirt/enhancements/pull/155 once that one is merged
since both change the same file

iholder101:
Sorry, I didn't see @enp0s3's [message](https://github.com/kubevirt/enhancements/pull/158#pullrequestreview-3649126443) above.

TBH, I was wondering if it's time to move DRA to beta or if we should wait in Alpha another cycle.
If we go this route and add SRIOV support to the same VEP, I think we definitely need to stay in Alpha for another cycle.

CC @alaypatel07 WDYT?

oshoval:
> Since we're using a separate feature gate (which makes total sense) I think it would be better to separate this to a different VEP.
> 
> I think it would be more cohesive and clear this way, and you can document that the FG depends on the more general dra FG.

Hi @enp0s3 @iholder101 @alaypatel07 @EdDev 
will need a final decision about it please, 
and if it means new VEP, should it be in the same folder of VEP10, as and additional file
or totally new, with all the skeleton etc (i guess the latter)

iholder101:
> > Since we're using a separate feature gate (which makes total sense) I think it would be better to separate this to a different VEP.
> > I think it would be more cohesive and clear this way, and you can document that the FG depends on the more general dra FG.
> 
> Hi @enp0s3 @iholder101 @alaypatel07 @EdDev will need a final decision about it please, and if it means new VEP, should it be in the same folder of VEP10, as and additional file or totally new, with all the skeleton etc (i guess the latter)

I vote for staying in Alpha in v1.8 (too many changes + dependency on a KEP that is not even alpha yet).

I don't think I mind merging the VEPs, especially if the VEP owner (@alaypatel07) prefers it this way.
I do think we better have one FG per VEP though.

oshoval:
> I vote for staying in Alpha in v1.8 (too many changes + dependency on a KEP that is not even alpha yet).
> 
> I don't think I mind merging the VEPs, especially if the VEP owner (@alaypatel07) prefers it this way. 


Thanks
sgtm, so will keep as it is now, and just rebase once #155 is merged,
unless someone insist otherwise

> I do think we better have one FG per VEP though.

There will be a new FG for network devices, undependent on the others.
We shouldn't mix GPU / hostDevices with network devices.
VEP 10 already has 2 FGs.

oshoval:
> > There will be a new FG for network devices, undependent on the others.
> > We shouldn't mix GPU / hostDevices with network devices.
> 
> If that's the case, why not add a separate VEP?

I think that since this what Alay preferred and we already started, and since they have lot in common, 
it is better to keep it, but if it makes noise we can consider decouple.
Note that we already depend on #155.
P.S in the future each one will support migration on different stage (so it might make more sense to decouple at least in the future)
I do think that amendments cause burden, merge conflicts and duplication tbh.

> 
> > VEP 10 already has 2 FGs.
> 
> Are you sure? I see that the [VEP](https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/10-dra-devices/vep.md#unsupported-usecases) lists network devices under "Unsupported Usecases".

It has 2 FGs for sure
GPU and HostDevices
It doesn't support yet Network devices, this one will have a new 3rd FG, this is what this extension is about

Issue #169, opened 2026-01-04
VEP 169: File-Level Restore


arnongilboa:
**Primary contact (assignee)**:
/assign @arnongilboa 

**Current Feature Stage**:
New

**Feature Gate**:
`VirtualMachineFileRestore`

**Responsible SIGs**:
Primary SIG:
/sig storage

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:
* Alpha release target (1.9):
* Beta release target (1.10):
* Stable release target (1.11):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

arnongilboa:
> Hey [@arnongilboa](https://github.com/arnongilboa)!
> 
> Since the VEP freeze is tomorrow, targeting this for v1.8 is unrealistic IMO. Can you target alpha to v1.9 instead?

Sure, my mistake.

akalenyu:
sig-storage meeting summary; are we still looking for this to go into 1.9?
I think we are still not quite converged on the VEP, with some considerations like subresources/crd not being ideal for guest agent commands.

arnongilboa:
> sig-storage meeting summary; are we still looking for this to go into 1.9? I think we are still not quite converged on the VEP, with some considerations like subresources/crd not being ideal for guest agent commands.

@akalenyu we target this VEP to 1.9, and there are already several backup vendors interested ([1](https://github.com/kubevirt/enhancements/pull/170#issuecomment-4029039354), [2](https://github.com/kubevirt/enhancements/pull/170#issuecomment-4031251609)). I agree with the reviewers, and will update the VEP dropping the guest command CRD and using an internal RPC instead. The ssh reference will be generalized, mentioning we can start initially with ssh over the network, and add vsock support if there is demand.

iholder101:
Moved this to "Tracked". FYI @vladikr

arnongilboa:
@iholder101 as decided in the merged [VEP](https://github.com/kubevirt/enhancements/pull/170), [kubevirt/vm-file-restore-operator](https://github.com/kubevirt/vm-file-restore-operator) is outside of `kubevirt/kubvirt`, and it's progressing toward DP aligned with 1.9 as expected, so should we keep this tracker or close it?

iholder101:
[...]

PR #170, opened 2026-01-04
VEP #169: File-Level Restore

arnongilboa:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/169
**SIG label**: /sig storage

### What this PR does
This VEP proposes file-level restore capabilities for KubeVirt VMs, allowing users to restore specific files or directories instead of entire volumes.

### DIFF
--- a/veps/sig-storage/file-restore.md
+++ b/veps/sig-storage/file-restore.md
@@ -1,0 +1,240 @@
+# VEP #169: File-Level Restore
+
+## Release Signoff Checklist
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+Proposal to support virtual machine file-level restore, applying to specific files or directories. This will give users and data protection partners the ability to restore files or a subset of user data into a virtual machine.
+
+## Motivation
+Currently, you need to restore an entire VM or a specific volume separately, and retrieve the data from it. This enhancement will allow the user to restore the small set of files needed.
+
+## Goals
+- Provide a declarative API allowing users to partially restore VM data.
+- A guest-OS-agnostic solution, supporting Windows (NTFS) in addition to Linux.
+- A backup-vendor-agnostic solution.
+
+## Non Goals
+- Creating backup PVCs or volume snapshots, indexing their content, browsing them offline to locate specific items, and making them accessible in the appropriate namespace, are all the responsibility of the backup vendor.
+- Backup vendors are encouraged to provide customized guest-side file restore helpers. Reference working helpers will be provided.
+
+## Definition of Users
+* VM owners
+* Cluster Admins
+* Backup vendors
+
+## User Stories
+* As a KubeVirt user, I would like to partially restore my VM data with specific files or directories from a block-level backup of the VM persistent storage created by a backup vendor.
+* As a backup vendor, I would like to offer file-level restore capability similar to vSphere.
+* As a backup vendor, I would like to offer a declarative KubeVirt API for file-level restore.
+* As a user of a Linux VM, I would like to restore a file or directory on a LUKS-encrypted volume in my running VM.
+* As a user of a Windows VM, I would like to restore a file or directory on a BitLocker-encrypted filesystem in my running VM.
+
+## References
+
+### Existing File-Level Restore Solutions
+Several backup vendors already offer their own file-level restore solutions; however, when integrating with KubeVirt they encounter some gaps and pain points, which our KubeVirt-integrated, guest-cooperative solution addresses:
+* Simplifying guest OS credentials handling, built for confidential guests.
+* Supporting encrypted volumes (LUKS, BitLocker).
+* Managing filesystem-specific metadata (ACLs, xattrs, etc.).
+Furthermore, our approach provides a simple declarative k8s API, supports both Linux and Windows guests, and enables both automated and manual FLR. KubeVirt will offer an integrated, open FLR solution for users who need FLR from a specific backup PVC or VolumeSnapshot, without depending on a backup vendor solution.
+
+### VMware File-Level Restore
+For VMware File-Level Restore, many backup vendors (Veeam, Commvault, Veritas, Dell PowerProtect, etc.) use a guest agent (a persistent one, or an ephemeral one pushed via `VMware Tools`). VMware [vSphere Guest Operations API](https://techdocs.broadcom.com/us/en/vmware-cis/vsphere/vsphere-sdks-tools/7-0/web-services-sdk-programming-guide/virtual-machine-guest-operations/running-guest-os-operations.html) allows interaction with files inside the guest through the hypervisor without requiring a guest network connection. Its capabilities include `InitiateFileTransferToGuest`, `MakeDirectoryGuest`, `ListGuestFiles`, etc. It is based on the `VMware Tools` control channel. For authentication it requires valid guest OS credentials passed via the API for `root` or a user with specific `sudo` permissions for file operations.
+
+### OADP VM File Restore
+The [OADP VM File Restore](https://github.com/migtools/oadp-vm-file-restore?tab=readme-ov-file#oadp-vm-file-restore) project is Velero-oriented, while we suggest a vendor-agnostic approach. Their external pod file-serving architecture is not designed to support encrypted backup volumes; for those, we suggest hotplugging the backup PVCs to the VM so we can mount and access their filesystems. However, for non-encrypted volumes, we can use OADP file-serving to browse backups in a temporary namespace before applying our file restore.
+
+## Repos
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+### Guest-Level Cooperation
+The solution is designed to support file-level restore for both Linux and Windows guests. On Windows, NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including access control lists (ACLs), alternate data streams (ADS), hard links and junction points, and file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are properly restored from encrypted backup PVC or snapshot.
+
+### Implementation Approach
+* We introduce the `VirtualMachineFileRestore` CRD, which is namespace-scoped and created in the VirtualMachine namespace. Access control is enforced through the relevant RBAC rules.
+* A hotplug volume will be used to attach a backup PVC or a VolumeSnapshot-restored PVC to the VM. The `DeclarativeHotplugVolumes` feature gate must be enabled.
+* SSH over the network will be used for specific command execution on the guest.
+* The guest OS-specific operations will be wrapped in a restricted file-restore guest helper script or binary. We will provide reference scripts, but backup vendors are encouraged to implement their own helpers.
+* Encrypted volume key management is in the responsibility of the guest file restore helper, looking for them in a specific location (e.g. next to the helper, or in the VM encrypted volume that we restore to from its encrypted backup PVC/VolumeSnapshot).
+
+### File-Level Restore
+virt-controller:
+* Watch `VirtualMachineFileRestore`
+* If the source is a VolumeSnapshot, restore it to a PVC in the VM namespace
+* If the source is a PVC in another namespace, clone it to a PVC in the VM namespace
+* Hotplug the volume (read-only)
+* Label `VirtualMachineFileRestore` with the target node (e.g. `kubevirt.io/target-node=node01`).
+* When the restore completes or the CR is deleted:
+  * Unplug the volume
+  * Delete the PVC if temporarily created
+
+virt-handler:
+* Watch `VirtualMachineFileRestore` using a filtered informer to process only restores to local VMIs.
+* SSH into the guest and execute the restore helper with the relevant arguments.
+* Update status upon completion (whether succeeded or failed) or timeout.
+
+Guest file restore helper:
+* Unlock the volume if encrypted (key file in target volume)
+* Mount the backup filesystem (read-only)
+* If automated restore:
+  * Restore the specified files or directories
+  * Unmount the backup filesystem
+
+## API Examples
+
+### Directory Restore from PVC
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileRestore
+metadata:
+  name: filerestore1
+  namespace: ns1
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: "fedora"
+  source:
+    pvc:
+      name: filebackup1
+      #namespace: ns2
+  sourcePath: /home/donald # optional
+  #sourcePartition: 1      # optional
+  #targetPath: /home/duck  # optional
+status:
+  conditions:
+  - type: Completed
+    status: "True"
+    reason: RestoreCompleted
+    message: "All 123 files restored successfully"
+    lastTransitionTime: "2026-01-01T01:23:45Z"
+  - type: Progressing
+    status: "False"
+    reason: Finished
+    lastTransitionTime: "2026-01-01T01:23:45Z"
+```
+
+### Directory Restore from Volume Snapshot
+Similar to PVC restore, except the source is a VolumeSnapshot.
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileRestore
+metadata:
+  name: filerestore1
+  namespace: ns1
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: "fedora"
+  source:
+    snapshot:
+      name: snap1
+  sourcePath: /home/donald
+```
+
+### Manual Restore
+When `sourcePath` is not provided, we allow manual restore from the mounted filesystem of the hotplugged backup volume using any preferred guest tool. When done, the user deletes the CR so we unmount the filesystem and unplug the volume.
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileRestore
+metadata:
+  name: filerestore1
+  namespace: ns1
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: "fedora"
+  source:
+    snapshot:
+      name: snap1
+```
+
+### Direct restore from external storage (TBD)
+For some backup vendors, backup data is not stored on the same cluster and is often moved to external storage like S3, requiring a two-step copy process for restoration, as we expect the backup data on an accessible PVC or volume snapshot within the cluster. To solve the double-copy process, we should support direct file transfer:
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileRestore
+metadata:
+  name: filerestore1
+  namespace: ns1
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: "fedora"
+  source:
+    remote:
+      name: s3_backup
+      bucket: buck1
+  sourcePath: /home/donald # optional
+  #targetPath: /home/duck  # optional
+```
+This reference example assumes the guest helper uses `rclone` and S3 remote was already configured via `rclone config`.
+
+## Guest command execution
+`virt-handler` connects to the guest `sshd` as restore-user, a user restricted to `sudo` execution of only the specifically-named trusted file-restore helper (via `.ssh/authorized_keys` and `/etc/sudoers.d`). `virt-operator` generates an SSH key pair and stores it in a Secret. The public key can be added to the guest either manually or propagated by `virt-controller` patching the VM `accessCredentials.sshPublicKey` with `propagationMethod: qemuGuestAgent` so it is injected via qemu-guest-agent. `virt-handler` mounts the private key Secret at `/etc/kubevirt/ssh/id_rsa` and uses it for file-restore SSH connections. If we find that private key compromise is an issue, we can create an ephemeral key pair per SSH connection.
+
+### Guest file restore helper CLI
+For covering the mentioned file restore flows, a guest helper CLI should support:
+* Restore (automatic):\
+  `filerestore.sh restore --serial <SERIAL> --mount-path <MNT_PATH> --source-path <SRC_PATH>`
+* Restore (manual - mount only):\
+  `filerestore.sh restore --serial <SERIAL> --mount-path <MNT_PATH>`
+* Cleanup (unmount and remove mount point):\
+  `filerestore.sh cleanup --mount-path <MNT_PATH>`
+
+`SERIAL` is the string that identifies the hotplugged disk inside the guest.\
+`MNT_PATH` is the mount path, named specifically for identifying the backup PVC or snapshot.\
+`SRC_PATH` is the file or directory path to be restored.
+
+## Alternatives
+### SSH over VSOCK
+We considered using SSH over VSOCK, which allows guest command execution for VMs without networking. Since this is a relatively rare case, we decided to start with SSH over the network and support SSH over VSOCK later if there is demand. To allow it, the guest SSH daemon needs to be configured to listen for incoming connections on a VSOCK. On Linux this can be done with `systemd-ssh-generator`, `socat`, etc. Windows supports VSOCK via `virtio-win`, so we need to install the viosock driver and viosock-tcp bridge service.
+
+### Hotplug utility volume
+A simpler solution that has been suggested is to hotplug a utility volume in `virt-launcher` with the backed-up files on a filesystem PVC, instead of inside a disk image on a PVC. By running an `rsync` client in `virt-handler`, we can copy files from the utility volume to the guest over `ssh+vsock` (using `ProxyCommand`). One problem with this direction is that backup PVCs are usually created and maintained by backup vendors, so we have no control over them; they can use block volume mode, and they are expected to be encrypted when the VM volumes are encrypted, so we can access their filesystem outside the guest only if we have the encryption passphrase, which is unlikely. We also need to support Windows guest NTFS, which is not fully supported by Linux utilities such as `rsync` and `scp`.
+
+### qemu-agent-command
+We initially considered using qemu-agent-command for the file-level operations. However, direct use of the qemu-agent-command APIs is strongly discouraged by libvirt. QGA commands such as guest-file-*, guest-exec-*, etc., are considered host-admin backdoors and will be blocked for confidential guests due to security risks. RHEL builds already disable these commands. QGA is also not designed for efficient large-scale data transfer and lacks support for file permissions, xattrs, symlinks, and hard links, making it unsuitable for robust restore.
+
+## Scalability
+We use rsync or similar in the guest for file-level restore. File transfer is memory- and I/O-intensive. It is also CPU-intensive due to file comparison and checksumming. However, we assume file-level restores are relatively rare operations that transfer only deltas and are not performed on many VMs at once.
+
+## Update/Rollback Compatibility
+- The backup volume is temporarily hotplugged to the VMI only for as long as needed
+- No changes to existing APIs or objects
+- No changes to existing VM/VMI specs
+
+## Functional Testing Approach
+A comprehensive test suite that checks the guest filesystem state is important for this feature. The following cases should be covered:
+* Automated directory restore from a PVC source to Linux guest — verify files and metadata match the source.
+* Automated directory restore from a VolumeSnapshot source — verify the snapshot is restored to a temporary PVC, files are restored, and the temporary PVC is cleaned up.
+* Restore from a PVC in a different namespace — verify the PVC is cloned to the VM namespace and cleaned up after restore.
+* Restore of a LUKS-encrypted volume — verify the volume is unlocked and the files are restored.
+* Manual restore — verify the backup filesystem is mounted read-only in the guest and remains available until the CR is deleted, then is unmounted and the volume is unplugged.
+* Windows guest restore — verify files and NTFS-specific metadata are restored correctly.
+* Missing guest helper — verify the VirtualMachineFileRestore reports a Failed phase with a clear error condition.
+
+## Implementation History
+...
+
+## Graduation Requirements
+
+### Alpha
+The `VirtualMachineFileRestore` feature gate will enable file-level restore.
+
+### Beta
+- Adoption by 2 backup and recovery applications or vendors.
+- After one or two releases, when we are confident that the feature is working as expected, move to beta.
+
+### GA
+Move to GA once the feature has been running in production without issues. Remove feature gates.
--- a/veps/sig-storage/file-restore.md
+++ b/veps/sig-storage/file-restore.md
@@ -40,7 +40,7 @@ Several backup vendors already offer their own file-level restore solutions; how
 * Simplifying guest OS credentials handling, built for confidential guests.
 * Supporting encrypted volumes (LUKS, BitLocker).
 * Managing filesystem-specific metadata (ACLs, xattrs, etc.).
-Furthermore, our approach provides a simple declarative k8s API, supports both Linux and Windows guests, and enables both automated and manual FLR. KubeVirt will offer an integrated, open FLR solution for users who need FLR from a specific backup PVC or VolumeSnapshot, without depending on a backup vendor solution.
+Furthermore, our approach provides a simple declarative k8s API, supports both Linux and Windows guests, and enables both automated and manual FLR. A dedicated file-restore operator aims to offer an integrated, open FLR solution for users who need FLR from a specific backup PVC or VolumeSnapshot, without depending on a backup vendor solution, while keeping KubeVirt at the infrastructure layer.
 
 ### VMware File-Level Restore
 For VMware File-Level Restore, many backup vendors (Veeam, Commvault, Veritas, Dell PowerProtect, etc.) use a guest agent (a persistent one, or an ephemeral one pushed via `VMware Tools`). VMware [vSphere Guest Operations API](https://techdocs.broadcom.com/us/en/vmware-cis/vsphere/vsphere-sdks-tools/7-0/web-services-sdk-programming-guide/virtual-machine-guest-operations/running-guest-os-operations.html) allows interaction with files inside the guest through the hypervisor without requiring a guest network connection. Its capabilities include `InitiateFileTransferToGuest`, `MakeDirectoryGuest`, `ListGuestFiles`, etc. It is based on the `VMware Tools` control channel. For authentication it requires valid guest OS credentials passed via the API for `root` or a user with specific `sudo` permissions for file operations.
@@ -49,7 +49,7 @@ For VMware File-Level Restore, many backup vendors (Veeam, Commvault, Veritas, D
 The [OADP VM File Restore](https://github.com/migtools/oadp-vm-file-restore?tab=readme-ov-file#oadp-vm-file-restore) project is Velero-oriented, while we suggest a vendor-agnostic approach. Their external pod file-serving architecture is not designed to support encrypted backup volumes; for those, we suggest hotplugging the backup PVCs to the VM so we can mount and access their filesystems. However, for non-encrypted volumes, we can use OADP file-serving to browse backups in a temporary namespace before applying our file restore.
 
 ## Repos
-[KubeVirt](https://github.com/kubevirt/kubevirt)
+[VM File Restore Operator](https://github.com/kubevirt/vm-file-restore-operator) a standalone operator in the kubevirt organization. [KubeVirt](https://github.com/kubevirt/kubevirt) is used only via the public API.
 
 ## Design
 ### Guest-Level Cooperation
@@ -58,27 +58,21 @@ The solution is designed to support file-level restore for both Linux and Window
 Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are properly restored from encrypted backup PVC or snapshot.
 
 ### Implementation Approach
-* We introduce the `VirtualMachineFileRestore` CRD, which is namespace-scoped and created in the VirtualMachine namespace. Access control is enforced through the relevant RBAC rules.
+* The file-restore operator introduces the `VirtualMachineFileRestore` CRD, which is namespace-scoped and created in the VirtualMachine namespace. Access control is enforced through the relevant RBAC rules.
 * A hotplug volume will be used to attach a backup PVC or a VolumeSnapshot-restored PVC to the VM. The `DeclarativeHotplugVolumes` feature gate must be enabled.
 * SSH over the network will be used for specific command execution on the guest.
 * The guest OS-specific operations will be wrapped in a restricted file-restore guest helper script or binary. We will provide reference scripts, but backup vendors are encouraged to implement their own helpers.
 * Encrypted volume key management is in the responsibility of the guest file restore helper, looking for them in a specific location (e.g. next to the helper, or in the VM encrypted volume that we restore to from its encrypted backup PVC/VolumeSnapshot).
 
 ### File-Level Restore
-virt-controller:
-* Watch `VirtualMachineFileRestore`
+VM File Restore Operator:
+* Watches `VirtualMachineFileRestore` CRs
 * If the source is a VolumeSnapshot, restore it to a PVC in the VM namespace
 * If the source is a PVC in another namespace, clone it to a PVC in the VM namespace
-* Hotplug the volume (read-only)
-* Label `VirtualMachineFileRestore` with the target node (e.g. `kubevirt.io/target-node=node01`).
-* When the restore completes or the CR is deleted:
-  * Unplug the volume
-  * Delete the PVC if temporarily created
-
-virt-handler:
-* Watch `VirtualMachineFileRestore` using a filtered informer to process only restores to local VMIs.
-* SSH into the guest and execute the restore helper with the relevant arguments.
-* Update status upon completion (whether succeeded or failed) or timeout.
+* Declaratively Hotplug the volume (read-only) via the KubeVirt API
+* SSH into the guest over the network and run the restore helper
+* Update CR status on completion, failure, or timeout
+* Unplug the volume and delete the PVC if temporarily created
 
 Guest file restore helper:
 * Unlock the volume if encrypted (key file in target volume)
@@ -181,7 +175,15 @@ spec:
 This reference example assumes the guest helper uses `rclone` and S3 remote was already configured via `rclone config`.
 
 ## Guest command execution
-`virt-handler` connects to the guest `sshd` as restore-user, a user restricted to `sudo` execution of only the specifically-named trusted file-restore helper (via `.ssh/authorized_keys` and `/etc/sudoers.d`). `virt-operator` generates an SSH key pair and stores it in a Secret. The public key can be added to the guest either manually or propagated by `virt-controller` patching the VM `accessCredentials.sshPublicKey` with `propagationMethod: qemuGuestAgent` so it is injected via qemu-guest-agent. `virt-handler` mounts the private key Secret at `/etc/kubevirt/ssh/id_rsa` and uses it for file-restore SSH connections. If we find that private key compromise is an issue, we can create an ephemeral key pair per SSH connection.
+The operator connects to the guest `sshd` as restore-user, a user restricted to using `sudo` to execute only the specifically-named, trusted file-restore helper binary or script (configured via `.ssh/authorized_keys` and `/etc/sudoers.d`).
+
+## Key management
+The SSH key lifecycle is managed by the operator, which creates a single SSH key pair and shares the public key via a ConfigMap.
+
+To activate the helper in the guest adding the restore-user, or to update its public key, one will need to:\
+`$ sudo filerestore.sh activate --public-key-file <PUBLIC_KEY_FILE>`
+
+The operator will also patch specifically labeled VMs `accessCredentials.sshPublicKey` with `propagationMethod: qemuGuestAgent` so the public key is automatically propagated by the `qemu-guest-agent`. This method allows us to support key rotation; however, it is currently supported only for Linux guests.
 
 ### Guest file restore helper CLI
 For covering the mentioned file restore flows, a guest helper CLI should support:
@@ -211,7 +213,7 @@ We use rsync or similar in the guest for file-level restore. File transfer is me
 
 ## Update/Rollback Compatibility
 - The backup volume is temporarily hotplugged to the VMI only for as long as needed
-- No changes to existing APIs or objects
+- No changes to existing KubeVirt APIs or objects
 - No changes to existing VM/VMI specs
 
 ## Functional Testing Approach
@@ -230,7 +232,7 @@ A comprehensive test suite that checks the guest filesystem state is important f
 ## Graduation Requirements
 
 ### Alpha
-The `VirtualMachineFileRestore` feature gate will enable file-level restore.
+The file-restore operator enables file-level restore.
 
 ### Beta
 - Adoption by 2 backup and recovery applications or vendors.


veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,205 @@
...
+#### Directory Restore from Remote Host
+Similar to PVC except the source
+```yaml
+  source:
+    host:
+      address: backup-01.us-east.internal # ip or name
+```
+
+### Guest Command
+To perform the actual guest operations the controller will create a VirtualMachineGuestCommand, which will execute the required commands inside the guest via SSH over, for example:
```
noamasu: i wonder if `VirtualMachineGuestCommand` accepts raw shell commands, there must be an admission controller or Validating Webhook that strictly limits allowed binaries (for exampl only allow rsync, mkdir, mount). Without this, a user could craft a command to execute `dd if=/dev/zero of=/dev/vda`, destroying the VM

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,205 @@
...
+### Restore
+#### Directory Restore from PVC
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileRestore
+metadata:
+  name: filerestore1
+spec:
+  vmiName: fedora
+  source:
```
noamasu: is adding an optional backup reference field (pointing to a VirtualMachineFileBackup CR) worth it? i mean If provided, the controller auto-fills the pvc and sourcePath.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,205 @@
...
+    pvc:
+      name: filebackup1
+  sourcePath: /home/donald
```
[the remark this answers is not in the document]
arnongilboa: It is recreated by rsync if possible, and if it fails the `VirtualMachineGuestCommand` will report the failure and the `VirtualMachineFileRestore` will fail as well.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,205 @@
...
+kind: VirtualMachineFileBackup
+metadata:
+  name: filebackup1
+spec:
+  vmiName: fedora
+  sourcePath: /home/donald
+  target:
+    pvc:
+      name: filebackup1 # optional
+      size: 1Gi # optional
```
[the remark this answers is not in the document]
arnongilboa: The original idea was to also support volume expansion of existing backup PVC using the `size` parameter. I don't think we should provision the backup PVC based of what `rsync --dryrun` says, as we want to allow limiting its size. In addition, it may need much more space when the following backup CR is applied. I guess that failing with a clear error specifying the required size is a reasonable approach.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,205 @@
...
+The solution is designed to support file-level backup and restore for both Linux and Windows guests. Windows NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including: access control lists (ACL), alternate data streams (ADS), hard links and junction points, file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are backed up in their encrypted state and properly restored.
+
+### Implementation Approach
+* Backup and restore are independent operations, and restore can be supported even without the backup support
+* Backup and restore are idempotent operations, so the same CRs can be applied whenever one wants to backup or restore
+* Hotplug volume will be used to attach the VM a PVC for backup and restore, or VolumeSnapshot restored PVC for restore
+* Backup and restore with guest ssh-able target will also be supported
+* SSH over VSOCK will be used to execute guest operations (mount backup volume filesystem, rsync etc.) using a helper  VirtualMachineGuestCommand CRD
```
dominikholler: While VSOCK provides some benefits over the virtual networking, is there a benefit from limiting the design to VSOCK, or would it be possible to have the design more generic, so that TCP could be utilized, too?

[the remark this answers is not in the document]
arnongilboa: We use only a minimal set of guest commands (e.g. mkdir, mount, rsync), so I don't think it creates dependency on its internal configuration. As mentioned, since we don't involve a guest agent, the controller creates guest OS specific commands (Linux/Windows).

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+### Guest-Level Cooperation
+The solution is designed to support file-level backup and restore for both Linux and Windows guests. Windows NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including: access control lists (ACL), alternate data streams (ADS), hard links and junction points, file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are backed up in their encrypted state and properly restored.
+
+### Implementation Approach
+* Backup and restore are independent operations, and restore can be supported even without the backup support
+* Backup and restore are idempotent operations, so the same CRs can be applied whenever one wants to backup or restore
+* Hotplug volume will be used to attach the VM a PVC for backup and restore, or VolumeSnapshot restored PVC for restore
+* Backup and restore with guest ssh-able target will also be supported
```
awels: This assumes that ssh support is enabled in the guest, I assuming this won't work if ssh is not enabled in the guest?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+
+### File-Level Backup
+When VirtualMachineFileBackup is applied:
+* If the target is PVC
+  * If the PVC does not exist, create it
+  * Hotplug the volume
+  * Using VirtualMachineGuestCommand
+    * Format the volume if needed
+    * Mount the backup filesystem
+    * Backup the specified files or directories
```
awels: How is the backup achieved? something like rsync? Does the guest have to have something installed to do the backup?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+    * Restore the specified files or directories
+    * Unmount the backup filesystem
+  * Unplug the volume
+* If remote source
+  * Using VirtualMachineGuestCommand
+    * Restore the specified files
+
+## API Examples
+
+For using this feature the DeclarativeHotplugVolumes and VSOCK feature gates must be enabled.
```
awels:
```suggestion
In order to use this feature both the DeclarativeHotplugVolumes and VSOCK feature gates must be enabled.
```

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+metadata:
+  name: filebackup1
+spec:
+  vmiName: fedora
+  sourcePath: /home/donald # optional, otherwise manual copy
+  target:
+    pvc:
+      name: filebackup1    # optional, otherwise use CR name
+      size: 1Gi            # optional
+  #targetPath: /home/duck  # optional
```
awels: I am assuming if omitted, that the targetPath will match the source path but from the base of the volume?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+  #targetPartition: 1      # optional
```
awels: I don't understand this, are we planning on more than one partition per PVC?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+  source:
+    pvc:
+      name: filebackup1
+  sourcePath: /home/donald # optional, otherwise manual copy
+  #sourcePartition: 1      # optional
+  #targetPath: /home/duck  # optional
+status:
+  phase: Succeeded
+```
+#### Directory Restore from Volume Snapshot
```
awels: I don't see an explanation of how we arrived at this snapshot. Was it an entire VMSnapshot or something else? How was this snapshot created?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+  name: cmd1
+spec:
+  vmiName: fedora
+  transport:
+    vsock:
+      name: fedora
+  command:
+  - /bin/sh
+  - -c
+  - rsync -avR /home/donald /backup/
```
awels: This assumes that rsync is installed in the guest OS if I understand correctly. This might not always be the case especially with windows guests.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+apiVersion: guestcommand.kubevirt.io/v1alpha1
+kind: VirtualMachineGuestCommand
+metadata:
+  name: cmd1
+spec:
+  vmiName: fedora
+  transport:
+    vsock:
+      name: fedora
+  command:
```
awels: I am super uncomfortable with being able to execute arbitrary commands from a k8s resource, even if the RBAC provides some protection. Would it make more sense to make a list of possible acceptable commands, and the controller takes the name of the command and arguments to that command (like the paths and names of the files) and executes only the acceptable commands. We would need to make sure we sanitize the arguments as well.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+  - /bin/sh
+  - -c
+  - rsync -avR /home/donald /backup/
+status:
+  phase: Succeeded
+  reason: SuccessfulExecution
```
awels: Some conditions here would be good as well. Might not be able to execute the command due to missing binaries etc. Some conditions to indicate issues and help trouble shoot will be helpful.

awels (commented):
One thought I had that is not covered here at all is how to handle missing binaries in the guest. In particular the rsync binary. There is no guarantee the guest has it installed. Would it be an interesting exercise to explore if we could put the binary in the volume we are writing to. We would have to have some mechanism for determining the guest OS reliably (we can't use the template since there is no guarantee those will match, I do believe if the guest-agent is installed it reports it back somewhere).

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+When VirtualMachineFileBackup is applied:
+* If the target is PVC
+  * If the PVC does not exist, create it
+  * Hotplug the volume
+  * Using VirtualMachineGuestCommand
+    * Format the volume if needed
+    * Mount the backup filesystem
+    * Backup the specified files or directories
+    * Unmount the backup filesystem
+  * Unplug the volume
```
mhenriks: I think it would be better to have the backed up files directly on a FileSystem PVC instead of inside a disk image on a PVC. I think we could accomplish that with the following flow:

1. Hotplug [utility volume](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md) in virt launcher
2. Run rsync client in virt-handler. Connect to guest over ssh + vsock. Write to utility volume

I'm not exactly sure how to configure rsync to use ssh+vsock but it seems to be possible by using `ProxyCommand` with ssh and netcat (or some other proxy)

mhenriks (commented):
I think it would be better if files were backed up directly to a FileSystem PVC and I think that is possible with [utility volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md)

I don't think `rsync` should be the only protocol. What if the guest doesn't have rsync? Maybe support `scp` as well?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+    host:
+      address: backup-01.us-east.internal # ip or name
+```
+
+### Guest Command
+To perform the actual guest operations the controller creates a VirtualMachineGuestCommand, which executes the required commands inside the guest via SSH over VSOCK. We currently do not want to support running arbitrary commands on the guest, so only the `kubevirt-controller` service account has RBAC for creating VirtualMachineGuestCommand.
+
+```yaml
+apiVersion: guestcommand.kubevirt.io/v1alpha1
+kind: VirtualMachineGuestCommand
```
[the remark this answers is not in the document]
arnongilboa: See [previous](https://github.com/kubevirt/enhancements/pull/170#discussion_r2702221127) answer.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+Proposal to support virtual machine file-level backup and restore, applying on specific files or directories. This will give users and data protection partners the ability to restore files or subset of user data into a virtual machine.
+
+## Motivation
+Currently, you need to restore an entire VM or a specific volume separately, and retrieve the data from it. This enhancment will allow the user restore the small set of files needed.
```
Acedus: The motivation for this proposal isn't clear to me.

> Currently, you need to restore an entire VM or a specific volume separately, and retrieve the data from it.

It isn't obvious why this is inherently problematic. Even if we assume this is a limitation, it's unclear why KubeVirt should be responsible for addressing it at the data-plane level.

In all existing tangential areas -- such as VMSnapshot and the upcoming VMBackup feature -- KubeVirt deliberately acts as an orchestrator that relies on third-party implementations; CSI drivers for VMSnapshot, and QEMU/KVM (at least today) for storage-agnostic VMBackup. This separation of concerns has been intentional and effective, and I don’t see a compelling reason to deviate from it here.

Moreover, many backup vendors already provide mature guest-level, file-level backup and file-level restore solutions. If the goal is to enable or improve support for such workflows, it would seem more appropriate for KubeVirt to facilitate their integration rather than reimplementing similar functionality itself.

One possible direction could be a shim or mediation layer. Conceptually it would be similar to something like a CSI in the sense that host-side components (e.g., virt-handler or virt-launcher) would forward user-initiated requests to a guest backup agent.

Acedus (commented):
I have some issues with this VEP as it stands, please see my inline comment.

veps/sig-storage/file-restore.md
arnongilboa: Right, we will fail the operation with an informative error in this case.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+For using this feature the DeclarativeHotplugVolumes and VSOCK feature gates must be enabled.
+For access control, VirtualMachineFileBackup/Restore will have similar RBAC as VirtualMachineSnapshot/Restore, so the same service accounts are authorized to create them.
+
+### Backup
+#### Directory Backup to PVC
+```yaml
+apiVersion: filerestore.kubevirt.io/v1alpha1
+kind: VirtualMachineFileBackup
+metadata:
+  name: filebackup1
```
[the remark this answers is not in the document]
arnongilboa: Sure, namespace scoped. Fixing.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+spec:
+  vmiName: fedora
+  sourcePath: /home/donald # optional, otherwise manual copy
```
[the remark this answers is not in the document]
arnongilboa: Following the same flow until backup filesystem is mounted, and letting the user manually do the actual file copy, and when he's done with it he deletes the CR, which does the umount and volume unplug.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+  target:
+    pvc:
+      name: filebackup1    # optional, otherwise use CR name
+      size: 1Gi            # optional
+  #targetPath: /home/duck  # optional
+  #targetPartition: 1      # optional
+status:
+  phase: InProgress
```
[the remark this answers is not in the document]
arnongilboa: Sure, will add these and think of other relevant ones.

veps/sig-storage/file-restore.md
arnongilboa: Answered [here](https://github.com/kubevirt/enhancements/pull/170#issuecomment-3765320203).

veps/sig-storage/file-restore.md
arnongilboa: Agree, will add some relevant ones.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,219 @@
...
+* As a KubeVirt user, I would like to partially restore my VM data with specific files or directories
+* As a KubeVirt user, I would like to partially backup my VM data with specific files or directories
+* As a backup vendor, I would like to offer file level restore capability similar to vSphere
+* As a user of a Linux VM, I would like to backup and restore a single file on a LUKS encrypted volume in my running VM.
+* As a user of a Linux VM, I would like to backup and restore multiple files in a single directory on a LUKS encrypted volume in my running VM in a consistent manner.
+* As a user of a Windows VM, I would like to backup and restore a single file on a bitlocker encrypted file system in my running VM.
+* As a user of a Windows VM, I would like to backup and restore multiple files in a single directory on a bitlocker encrypted file system in my running VM in a consistent manner.
+
+## Repos
+[KubeVirt](https://github.com/kubevirt/kubevirt)
```
akalenyu: I agree with @Acedus on this one
> Even if we assume this is a limitation, it's unclear why KubeVirt should be responsible for addressing it at the data-plane level.

I think we have to ask ourselves where we draw the line wrt "core" kubevirt features;
As we see in this VEP, there's already some concern with being in this problem domain to begin with
(guest binaries etc) so to me that already implies that we should at least consider separate project

akalenyu (commented):
Thanks for the detailed proposal :pray: 
Some worries about tackling it in the first place/tackling it in kubevirt.

If possible, could you have a small section about existing FLR in the wilderness and what you know about their impl? maybe we should refocus on making those more robust as they are today.

Just throwing it out there, maybe something like https://github.com/csi-addons/spec/ can be suitable for such advanced storage operations?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Definition of Users
+* VM owners
+* Cluster Admins
+* Backup vendors
+
+## User Stories
+* As a KubeVirt user, I would like to partially restore my VM data with specific files or directories
+* As a KubeVirt user, I would like to partially backup my VM data with specific files or directories
+* As a backup vendor, I would like to offer file level restore capability similar to vSphere
```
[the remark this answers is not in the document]
arnongilboa: We are sharing the VEP with backup vendors to gather their feedback and input.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Overview
+Proposal to support virtual machine file-level backup and restore, applying on specific files or directories. This will give users and data protection partners the ability to restore files or subset of user data into a virtual machine.
+
+## Motivation
+Currently, you need to restore an entire VM or a specific volume separately, and retrieve the data from it. This enhancment will allow the user restore the small set of files needed.
+
+## Goals
+- Provide a declarative API allowing user to partially backup and restore VM data.
+- Guest OS agnostic solution, supporting Windows (NTFS) in addition to Linux.
```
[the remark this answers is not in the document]
mhenriks: Nos as concerned about ssh running or rsync existing as I am about ssh listening on VSOCK

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+* As a user of a Windows VM, I would like to backup and restore multiple files in a single directory on a bitlocker encrypted file system in my running VM in a consistent manner.
+
+## Repos
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+### Guest-Level Cooperation
+The solution is designed to support file-level backup and restore for both Linux and Windows guests. Windows NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including: access control lists (ACL), alternate data streams (ADS), hard links and junction points, file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are backed up in their encrypted state and properly restored.
```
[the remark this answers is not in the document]
mhenriks: how are keys managed?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,216 @@
...
+* If the target is PVC
+  * If the PVC does not exist, create it
+  * Hotplug the volume
+  * Using VirtualMachineGuestCommand
+    * Format the volume if needed
+    * Mount the backup filesystem
+    * Backup the specified files or directories
+    * Unmount the backup filesystem
+  * Unplug the volume
+* If remote target
```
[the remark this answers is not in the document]
mhenriks: Guest would need ssh key copied over to do remote backup

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Design
+### Guest-Level Cooperation
+The solution is designed to support file-level restore for both Linux and Windows guests. Windows NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including: access control lists (ACL), alternate data streams (ADS), hard links and junction points, file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are backed up in their encrypted state and properly restored.
+
+### Implementation Approach
+* Hotplug volume will be used to attach the VM a backup PVC or a VolumeSnapshot restored PVC
+* SSH over VSOCK will be used to execute guest operations (mount backup volume filesystem, rsync etc.) using a helper VirtualMachineGuestCommand CRD
```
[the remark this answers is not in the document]
arnongilboa: Can we `mkdir`/`mount`/`rsync` with `internal-sftp`?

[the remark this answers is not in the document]
dominikholler: I see, if the PVC required to be mounted directly to the guest VM, mount would be required, but sftp operates at [file level](https://github.com/openssh/openssh-portable/blob/master/sftp.h#L34).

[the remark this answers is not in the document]
dominikholler: > Is it possible to prevent arbitrary commands being executed inside the VM over the ssh connection? (For example, it may be necessary to install an safe rsync wrapper that restricts the command-line and then only allow that wrapper in .ssh/authorized_keys.)

> Some of these questions depend on the user's security requirements, but often users do not want to give unrestricted ssh access because they consider it a security risk.

Full ack! This is why I would vote for doing all the [permission demanding work in a help pod](https://github.com/kubevirt/enhancements/pull/170#discussion_r2736965476), and restricting the acess to sftp.


> Regarding ssh, does this mean something needs to install the OpenSSH server inside the VM and configure it to listen over vsock?

> The host connects to sshd with which user name (root?) and key? How is the ssh key managed? Is it unique for each VM?

Ack, would be great to have answers to those great quesions in the VEP.

dominikholler (changes requested):

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,218 @@
...
+    key: ssh-privatekey
+    name: my-ssh-key-secret
+  port: 2222
+  useTLS: false
+```
+
+## SSH over VSOCK
+Guest SSHD needs to be configured to listen for incoming connections on a VSOCK. In Linux this can be done by `systemd-ssh-generator`, `socat` etc. Windows supports VSOCK via `virtio-win`, so we need to install viosock driver, viosock-tcp bridge service, OpenSSH server and Cloudbase-Init. `virt-handler` connects the guest `sshd` with restore-user, which is either a user with specific `sudo` permissions for file operations, or `root`. We need to decide if this restore-user is common or unique for each VM. The user and public key are created in the guest via VM spec cloudInit userData. We do not want to give unrestricted ssh access, so the private key is stored in a k8s Secret that only the `kubevirt-handler` service account has RBAC to access. To prevent arbitrary commands being executed inside the VM over the ssh connection we will only support execution of a specifically-named safe wrapper installed on the guest, doing the actual restore (via `rsync` etc), restricting the command-line, and allow only that wrapper in .ssh/authorized_keys.
+
+## Alternatives
```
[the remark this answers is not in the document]
arnongilboa: How would you open the encrypted volume in the additional pod before you mount its file system?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,208 @@
...
+
+### File-Level Backup
+We also provide a file-level backup functionality. It is useful when the VM user knows what he wants to backup. File-level restore is not dependent on it at all.
+
+When VirtualMachineFileBackup is applied:
+* If the target is PVC
+  * If the PVC does not exist, create it
+  * Hotplug the volume
+  * Using VirtualMachineGuestCommand
+    * Format the volume if needed
```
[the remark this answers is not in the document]
stefanha: My understanding is that rsync skips the delta algorithm by default when copying between local files. While it is possible to explicitly enable the delta algorithm on the command-line, that still involves reading the source and destination files and that is expensive for large files. So there is no way to make the backup really efficient with rsync itself.

Efficient solutions are more specialized, like dm-era, btrfs diff, or inotify-based changed block tracking. It makes sense to support a general-purpose solution like rsync or cp, but it may be necessary to support specialized changed block tracking solutions for large file use cases.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,182 @@
...
+### Guest-Level Cooperation
+The solution is designed to support file-level restore for both Linux and Windows guests. Windows NTFS filesystems require guest cooperation to properly handle filesystem-specific features, including: access control lists (ACL), alternate data streams (ADS), hard links and junction points, file metadata (attributes and timestamps).
+
+Additionally, encrypted volumes (such as BitLocker for Windows) require guest-level access to ensure files are backed up in their encrypted state and properly restored.
+
+### Implementation Approach
+* Hotplug volume will be used to attach the VM a backup PVC or a VolumeSnapshot restored PVC
+* A helper VirtualMachineGuestCommand CRD will be created by the controller for executing allowed guest commands using the specified transport.
+* SSH over VSOCK will be the reference transport, while the API will allow adding support for other transports as well (SSH etc.).
+* The guest OS-specific operations will be warpped in a restricted script or binary. We will provide reference scripts, but it can be implemented by a backup vendor guest agent as well.
```
stefanha: "warpped" -> "wrapped"

veps/sig-storage/file-restore.md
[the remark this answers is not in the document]
arnongilboa: Cool. We currently leave the guest-file-restore-tool open, as it will be wrapped in a trusted guest script/binary. rsync was just an initial suggestion.

mhenriks:
I don't think KubeVirt should be in the file restore game. It's role should be providing the infrastructure primitives that enable backup vendors to do their job well, not reimplementing what they already do

The `VirtualMachineGuestCommand` CRD is really the core proposal here, with file restore being the first consumer. KubeVirt has historically stayed at the hypervisor layer, with narrow exceptions for things like guest agent and cloud-init. A general-purpose guest command execution framework over SSH+VSOCK is a fundamentally different category and deserves its own design discussion separate from file restore

But before we consider `VirtualMachineGuestCommand` can we hear from some backup vendors? What are their pain points? What are the gaps in KubeVirt's infrastructure that make file-level restore harder?

arnongilboa:
> I don't think KubeVirt should be in the file restore game. It's role should be providing the infrastructure primitives that enable backup vendors to do their job well, not reimplementing what they already do

The backup vendors main job is volume-level backup and restore. File-level restore to running VMs is performed by  vendors in different approaches such as helper pod (Veeam Kasten), guest agent (Commvault), etc.

Our contribution here is that unlike the existing vendor-specific solutions, we introduce a common declarative API specifically for file-level restore from a backup PVC in the VM namespace. All we provide is the infrastructure: hotplugging the backup volume to the VM, and passing the relevant command to a guest wrapper doing the actual file-level restore. We don't force using any specific guest tool, but provide reference scripts for Linux and Windows guests. The API allows specific file/directory restore, and also manual restore via any user preferred guest tool. 
   
> The `VirtualMachineGuestCommand` CRD is really the core proposal here, with file restore being the first consumer. KubeVirt has historically stayed at the hypervisor layer, with narrow exceptions for things like guest agent and cloud-init. A general-purpose guest command execution framework over SSH+VSOCK is a fundamentally different category and deserves its own design discussion separate from file restore

Right, it's the core, and since the file restore is built on top of it, it was placed here to get reviewed together with the first reference customer justifying it.

> But before we consider `VirtualMachineGuestCommand` can we hear from some backup vendors? What are their pain points? What are the gaps in KubeVirt's infrastructure that make file-level restore harder?

We initiated discussions with several backup vendors, and we'll update according to their feedback. Afaiu the pain points are restoring files from encrypted backup volume, and that there is no standard kubevirt file restore declarative API.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,236 @@
...
+metadata:
+  name: my-secret
+  namespace: ns1
+data:
+  key: c3...
+type: Opaque
+```
+
+## SSH over VSOCK
+Guest SSHD needs to be configured to listen for incoming connections on a VSOCK. In Linux this can be done by `systemd-ssh-generator`, `socat` etc. Windows supports VSOCK via `virtio-win`, so we need to install viosock driver, viosock-tcp bridge service and OpenSSH server. `virt-handler` connects the guest `sshd` with restore-user, which is either a user with specific `sudo` permissions for file operations, or `root`. We need to decide if this restore-user is common or unique for each VM. We do not want to give unrestricted ssh access, so the private key is stored in a k8s Secret that only the `kubevirt-handler` service account has RBAC to access. To prevent arbitrary commands being executed inside the VM over the ssh connection we will only support execution of a specifically-named trusted file restore helper in the guest, doing the actual restore (`mount`, `rsync` etc), restricting the command-line, and allow only that helper in `.ssh/authorized_keys`.
```
Acedus: > We do not want to give unrestricted ssh access, so the private key is stored in a k8s Secret that only the `kubevirt-handler` service account has RBAC to access.

I'm a bit confused, is this secret the one specified by the user as part of `VirtualMachineGuestCommandConfig`?

Acedus: AFAIK tenant secret read permissions are a no go, how do we plan on handling that?

veps/sig-storage/file-restore.md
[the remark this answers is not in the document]
mhenriks: > We put a key file in the VM encrypted volume that we restore to from its encrypted backup PVC/VolumeSnapshot.

So "we" meaning some KubeVirt code somehow gets the encryption key for a disk and store that key in a regular file on the filesystem? We use this key to unlock the restored disk? 

Can you describe this process in detail in the VEP?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,229 @@
...
+spec:
+  vmiName: fedora
+  source:
+    snapshot:
+      name: snap1
+```
+
+### Guest Command
+To perform the actual guest operations the controller creates a VirtualMachineGuestCommand, which executes the required commands inside the guest via the specified transport (e.g. SSH over VSOCK). We currently do not want to support running arbitrary commands on the guest, so only the `kubevirt-controller` service account has RBAC for creating VirtualMachineGuestCommand. We will define an allow-list of possible acceptable commands. The controller will take the name of the command and its arguments (e.g. paths and names of files), and execute only the acceptable commands. Arguments will be sanitized as well.
+
```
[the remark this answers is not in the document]
mhenriks: I'm with @vladikr in that a declarative API is not ideal executing commands. 

Not using the subresource API would require setting up a lot of plumbing to handle auth and routing to the appropriate virt-handler though. 

Would love to avoid this all together. I still don't fully understand the requirement that these commands execute over vsock+ssh. To support VMs without networking? How common is that? Why not ssh over the network? At least initially and see if there is demand for vsock?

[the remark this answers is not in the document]
arnongilboa: (Back from holiday) I agree with your feedback, and will update the VEP accordingly, dropping the guest command CRD and using an internal RPC instead. The ssh reference will be generalized, mentioning we can start initially with ssh over the network, and add vsock support if there is demand.

veps/sig-storage/file-restore.md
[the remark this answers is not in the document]
arnongilboa: Encrypted volume key management is in the responsibility of the guest file restore helper, looking for them in a specific location. There is no KubeVirt code dealing with the encryption keys, it's stays in the guest. Added this note to the VEP.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+### File-Level Restore
+virt-controller:
+* Watch `VirtualMachineFileRestore`
+* If the source is a VolumeSnapshot, restore it to a PVC in the VM namespace
+* If the source is a PVC in another namespace, clone it to a PVC in the VM namespace
+* Hotplug the volume (read-only)
+* Label `VirtualMachineFileRestore` with the target node (e.g. `kubevirt.io/target-node=node01`).
+* When the restore completes or the CR is deleted:
+  * Unplug the volume
+  * Delete the PVC if temporarily created
```
Acedus: In which scenario would a PVC be temporarily created?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+* The guest OS-specific operations will be wrapped in a restricted file-restore guest helper script or binary. We will provide reference scripts, but the helper can be implemented by a backup vendor as well.
+* Encrypted volume key management is in the responsibility of the guest file restore helper, looking for them in a specific location (e.g. next to the helper, or in the VM encrypted volume that we restore to from its encrypted backup PVC/VolumeSnapshot).
+
+### File-Level Restore
+virt-controller:
+* Watch `VirtualMachineFileRestore`
+* If the source is a VolumeSnapshot, restore it to a PVC in the VM namespace
+* If the source is a PVC in another namespace, clone it to a PVC in the VM namespace
+* Hotplug the volume (read-only)
+* Label `VirtualMachineFileRestore` with the target node (e.g. `kubevirt.io/target-node=node01`).
```
Acedus: Why is this needed?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+* When the restore completes or the CR is deleted:
+  * Unplug the volume
+  * Delete the PVC if temporarily created
+
+virt-handler:
+* Watch `VirtualMachineFileRestore` using a filtered informer to process only restores to local VMIs.
+* SSH into the guest and execute the restore helper with the relevant arguments.
```
Acedus: Is there a reason for executing the file restore helper inside the guest? For example, can't encrypted volumes be unlocked via a key provided in the form of a secret for example? Alternatively, can we prepare a temporary volume that only contains the directory structure with the relevant files to restore via a separate utility pod with the sole purpose of doing that, and once that prep work is done hotplug and copy whatever's required into the guest?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+    remote:
+      name: s3_backup
+      bucket: buck1
+  sourcePath: /home/donald # optional
+  #targetPath: /home/duck  # optional
+```
+This reference example assumes the guest helper uses `rclone` and the remote was already configured via `rclone config`.
+
+## Guest command execution
+`virt-handler` connects to the guest `sshd` as restore-user, a user restricted to `sudo` execution of only the trusted file-restore helper (via `.ssh/authorized_keys` and `/etc/sudoers.d`). `virt-operator` generates an SSH key pair and stores it in a Secret. The public key can be added to the guest either manually or propagated by `virt-controller` patching the VM `accessCredentials.sshPublicKey` with `propagationMethod: qemuGuestAgent` so it is injected via qemu-guest-agent. `virt-handler` mounts the private key Secret at `/etc/kubevirt/ssh/id_rsa` and uses it for file-restore SSH connections. If we find that private key compromise is an issue, we can create an ephemeral key pair per SSH connection.
```
Acedus: > If we find that private key compromise is an issue, we can create an ephemeral key pair per SSH connection.

Consider using SSH certificates instead? https://man.openbsd.org/ssh-keygen.1#CERTIFICATES

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+### Guest file restore helper CLI
+For covering the mentioned file restore flows, a guest helper CLI should support:
+* Restore (automatic):\
+  `filerestore.sh restore --serial <SERIAL> --mount-path <PATH> --source-path <PATH>`\
+* Restore (manual - mount only):\
+  `filerestore.sh restore --serial <SERIAL> --mount-path <PATH>`
+* Cleanup (unmount and remove mount point):\
+  `filerestore.sh cleanup --mount-path <PATH>`
+
+`SERIAL` is the string that identifies the hotplugged disk inside the guest.
```
Acedus: I assume we'll be providing that serial?

Acedus (commented):
Looking better! A few more questions.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+Currently, you need to restore an entire VM or a specific volume separately, and retrieve the data from it. This enhancement will allow the user to restore the small set of files needed.
+
+## Goals
+- Provide a declarative API allowing users to partially restore VM data.
+- A guest-OS-agnostic solution, supporting Windows (NTFS) in addition to Linux.
+- A backup-vendor-agnostic solution.
+
+## Non Goals
+- Creating backup PVCs or volume snapshots, indexing their content, browsing them offline to locate specific items, and making them accessible in the appropriate namespace, are all the responsibility of the backup vendor.
+- Although reference guest-side file restore helpers will be provided, backup vendors are encouraged to provide their own customized and optimized helpers.
```
noamasu: > backup vendors are encouraged to provide their own customized and optimized helpers.

this is a bit confusing to me considering this is under "Non Goals" ... we do or do not encourage vendors to provide customized helpers?

if we do, how would we know how to run it? should this be (somehow) defined in the `VirtualMachineFileRestore` CR? is there a clear definition of how such customized helper communicates?

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+spec:
+  target:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: "fedora"
+  source:
+    pvc:
+      name: filebackup1
+      #namespace: ns2
+  sourcePath: /home/donald # optional
```
noamasu, line 102: just making sure i understand, sourcePath is on top of the defined --mount-path?

for example if --mount-path of the hotplugged volume is /mnt/vol, then the full path inside the guest will be /mnt/vol/home/donald?

veps/sig-storage/file-restore.md
arnongilboa: In restoring from a snapshot or from a PVC in another namespace.

veps/sig-storage/file-restore.md
```diff
@@ -0,0 +1,238 @@
...
+  #targetPath: /home/duck  # optional
+```
+This reference example assumes the guest helper uses `rclone` and the remote was already configured via `rclone config`.
+
+## Guest command execution
+`virt-handler` connects to the guest `sshd` as restore-user, a user restricted to `sudo` execution of only the trusted file-restore helper (via `.ssh/authorized_keys` and `/etc/sudoers.d`). `virt-operator` generates an SSH key pair and stores it in a Secret. The public key can be added to the guest either manually or propagated by `virt-controller` patching the VM `accessCredentials.sshPublicKey` with `propagationMethod: qemuGuestAgent` so it is injected via qemu-guest-agent. `virt-handler` mounts the private key Secret at `/etc/kubevirt/ssh/id_rsa` and uses it for file-restore SSH connections. If we find that private key compromise is an issue, we can create an ephemeral key pair per SSH connection.
+
+### Guest file restore helper CLI
+For covering the mentioned file restore flows, a guest helper CLI should support:
```
noamasu, line 189: as stated in line 17
> A guest-OS-agnostic solution, supporting Windows (NTFS) in addition to Linux.

and line 22
> reference guest-side file restore helpers will be provided 

filerestore.sh is a "simple" helper we provide? how?

to be OS-agnostic, how would such given helper will look like for windows? exe file? i think more info on the helper for windows would be helpful.

veps/sig-storage/file-restore.md
arnongilboa: As already discussed here, we do not want to involve kubevirt in guest OS specifics, such as unlocking a Windows Bitlocked backup volume, or NTFS specifics as mentioned in the Guest-Level Cooperation section. Creating a temporary recovery-dedicated volume is sure a valid application of the solution, but not for the Windows Bitlocked NTFS backup volume.

veps/sig-storage/file-restore.md
arnongilboa: Only for making virt-handler life easier using filtered informer to process only restores to local VMIs. Can we do it cleaner?

noamasu (commented):
Thanks Arnon! added few questions/comments.
also, It looks like several important review threads (disk signature / LVM / RAID, and how much of NTFS the reference helper actually guarantees) are still mostly in comments... I think it would it help to add a short “Known limitations & mitigations” subsection of some sort into the VEP so we do not have to mine the conversation

veps/sig-storage/file-restore.md
arnongilboa: Yep,  the controller sets the Disk serial when it hotplugs the volume.

veps/sig-storage/file-restore.md
arnongilboa: We already presented a POC with both Linux and Windows guest helpers. For Windows we used a Powershell script similar to the Linux bash script with few adjustments. Both scripts align to the CLI spec and implement the logic mentioned, and therefore the kubevirt side is kept guest-OS-agnostic.

veps/sig-storage/file-restore.md
arnongilboa: Right, and it will be restored to `/home/donald`.

veps/sig-storage/file-restore.md
arnongilboa: Rephrased this non-goal line to make it clearer, and also mentioned it in the "Guest command execution" section.

amitmitkar:
I had a few comments regarding this VEP and requests for possible future enhancements.

This is a good first step towards providing this much needed capability esp. important for data protection use cases.

Here is the issue with this implementation and the enhancements that could help :
- Creating a PVC and putting backed up data into it only to be copied out and placed into the guest VM adds an unnecessary extra hop in the restore path.
- Other hypervisors like VMWare provide a much more [robust guest-file API ](https://developer.broadcom.com/xapis/virtual-infrastructure-json-api/latest/virtual-infrastructure/guest-file-manager/) that allows rich FS level primitives that can enable direct granular file level restore operations into the guests.
- Availability of such an API will avoid the requirement of this additional temporary PVC and help achieve an efficient file level transfer using the [qemu-guest-agent protocol](https://qemu-project.gitlab.io/qemu/interop/qemu-ga-ref.html#). There will be no need to copy data into a PVC first, instead data could be directly placed into the VM, thus avoiding the extra hop.

mhenriks:
As discussed in VEP meeting today: Since the design now uses SSH over the network rather than VSOCK, this doesn't need to live in kubevirt/kubevirt. Nothing here requires virt-handler or virt-controller internals. It could be a standalone operator in the kubevirt org that:

  1. Watches VirtualMachineFileRestore CRs
  2. Uses the existing KubeVirt API to declaratively hotplug the backup volume (read-only)
  3. SSHes into the guest over the network and runs the restore helper
  4. Updates CR status on completion/failure/timeout
  5. Unplugs the volume and cleans up temporary PVCs

  This keeps KubeVirt at the infrastructure layer and gives the project its own release cycle and contributor community, which is better for backup vendor participation anyway.

  It also moves SSH key management out of KubeVirt core. The current design has virt-operator generating keys, virt-controller propagating them via accessCredentials, and virt-handler mounting and using them. That's all three core components taking on SSH credential management for a single feature. An external operator can own its own key lifecycle, and if a CRD is needed, it would make more sense here, letting admins configure which SSH keys to use, rather than for the restore operation itself.

  The guest provisioning story also still needs attention. The design requires a restore-user account, SSH access, sudoers rules, and the helper binary pre-installed in the guest, but there's no mechanism described for getting any of that in place.

arnongilboa:
> I had a few comments regarding this VEP and requests for possible future enhancements.
> 
> This is a good first step towards providing this much needed capability esp. important for data protection use cases.
> 
Thanks for your feedback @amitmitkar!

> Here is the issue with this implementation and the enhancements that could help :
> 
> * Creating a PVC and putting backed up data into it only to be copied out and placed into the guest VM adds an unnecessary extra hop in the restore path.

As you can see, we added "Direct restore from external storage" section, so in the VirtualMachineFileRestore CR you can specify a remote storage instead of PVC/snapshot. In the example we restore a directory from S3 bucket. The guest helper can use configured rclone for that. Can a similar approach work for you?

> * Other hypervisors like VMWare provide a much more [robust guest-file API ](https://developer.broadcom.com/xapis/virtual-infrastructure-json-api/latest/virtual-infrastructure/guest-file-manager/) that allows rich FS level primitives that can enable direct granular file level restore operations into the guests.

For the TP we plan supporting only the minimal API for restoring files and directories. However, when we'll see backup vendors adopting this feature, we will sure extend its API.

> * Availability of such an API will avoid the requirement of this additional temporary PVC and help achieve an efficient file level transfer using the [qemu-guest-agent protocol](https://qemu-project.gitlab.io/qemu/interop/qemu-ga-ref.html#). There will be no need to copy data into a PVC first, instead data could be directly placed into the VM, thus avoiding the extra hop.

Discussing it with the Libvirt developers a while ago, they said qemu-agent-command APIs should not be directly used, as it is strongly discouraged by Libvirt. QGA commands, such as guest-file-*, guest-exec-* etc. are considered host admin backdoors and will be blocked for confidential guests due to security risks. RHEL builds already disable these commands. QGA is also not designed for efficient large-scale data transfer and lacks support for file permissions, xattrs, symlinks, and hard links, making it unsuitable for robust file restore operations.

arnongilboa:
> As discussed in VEP meeting today: Since the design now uses SSH over the network rather than VSOCK, this doesn't need to live in kubevirt/kubevirt. Nothing here requires virt-handler or virt-controller internals. It could be a standalone operator in the kubevirt org

@mhenriks sure, updated accordingly.

amitmitkar:
> > I had a few comments regarding this VEP and requests for possible future enhancements.
> > This is a good first step towards providing this much needed capability esp. important for data protection use cases.
> 
> Thanks for your feedback @amitmitkar!
> 
> > Here is the issue with this implementation and the enhancements that could help :
> > 
> > * Creating a PVC and putting backed up data into it only to be copied out and placed into the guest VM adds an unnecessary extra hop in the restore path.
> 
> As you can see, we added "Direct restore from external storage" section, so in the VirtualMachineFileRestore CR you can specify a remote storage instead of PVC/snapshot. In the example we restore a directory from S3 bucket. The guest helper can use configured rclone for that. Can a similar approach work for you?
>
It could certainly work. This however, still means that the data needs to be stored as files in the s3 location first. A lot of data protection vendors, store data in their own proprietary format. For eg. VM disks are backed up as blocks to leverage CBT. The translation from that into a file level representation usually is done via an offline index or other means at run-time during restores. Therefore, exposing an API to place files directly into the guests helps avoid this overhead of first converting the data from backup format into file level representation, secondly put it down on persistent storage like s3/pvc and then finally copying it out to the destination (using the CR).

 
> > * Other hypervisors like VMWare provide a much more [robust guest-file API ](https://developer.broadcom.com/xapis/virtual-infrastructure-json-api/latest/virtual-infrastructure/guest-file-manager/) that allows rich FS level primitives that can enable direct granular file level restore operations into the guests.
> 
> > * Availability of such an API will avoid the requirement of this additional temporary PVC and help achieve an efficient file level transfer using the [qemu-guest-agent protocol](https://qemu-project.gitlab.io/qemu/interop/qemu-ga-ref.html#). There will be no need to copy data into a PVC first, instead data could be directly placed into the VM, thus avoiding the extra hop.
> 
> Discussing it with the Libvirt developers a while ago, they said qemu-agent-command APIs should not be directly used, as it is strongly discouraged by Libvirt. QGA commands, such as guest-file-_, guest-exec-_ etc. are considered host admin backdoors and will be blocked for confidential guests due to security risks. RHEL builds already disable these commands. QGA is also not designed for efficient large-scale data transfer and lacks support for file permissions, xattrs, symlinks, and hard links, making it unsuitable for robust file restore operations.

Understood. Similar challenges have been addressed by other hypervisor platforms. Would be interesting to see how they've been overcome there and if similar fixes can apply.

arnongilboa:
> It could certainly work. This however, still means that the data needs to be stored as files in the s3 location first. A lot of data protection vendors, store data in their own proprietary format. For eg. VM disks are backed up as blocks to leverage CBT. The translation from that into a file level representation usually is done via an offline index or other means at run-time during restores. Therefore, exposing an API to place files directly into the guests helps avoid this overhead of first converting the data from backup format into file level representation, secondly put it down on persistent storage like s3/pvc and then finally copying it out to the destination (using the CR).

@amitmitkar what about sharing the restored files in your side in a location accessible to the guest so we can `rsync` from it similarly to what we do from local PVC? that way you won't need an intermediate S3/PVC etc. You may even apply the VMFileRestore CR during the extraction so rsync will work in parallel with the file extraction (which is much faster) , and maybe once again in the end to make sure it converged.

amitmitkar:
> > It could certainly work. This however, still means that the data needs to be stored as files in the s3 location first. A lot of data protection vendors, store data in their own proprietary format. For eg. VM disks are backed up as blocks to leverage CBT. The translation from that into a file level representation usually is done via an offline index or other means at run-time during restores. Therefore, exposing an API to place files directly into the guests helps avoid this overhead of first converting the data from backup format into file level representation, secondly put it down on persistent storage like s3/pvc and then finally copying it out to the destination (using the CR).
> 
> @amitmitkar what about sharing the restored files in your side in a location accessible to the guest so we can `rsync` from it similarly to what we do from local PVC? that way you won't need an intermediate S3/PVC etc. You may even apply the VMFileRestore CR during the extraction so rsync will work in parallel with the file extraction (which is much faster) , and maybe once again in the end to make sure it converged.

Interesting idea. How'd reverse authentication work? There is also the limitation that the production VM will need to be exposed to backup infrastructure components. It may not always be possible.

mhenriks:
VEP is now disconnected from KubeVirt core and can evolve at its own leisure. I think we are at a good starting point now.
/lgtm

Acedus (approved):

vladikr:
> VEP is now disconnected from KubeVirt core and can evolve at its own leisure. I think we are at a good starting point now. /lgtm

/approve 
Thank you @arnongilboa and thank you for shaping this up @mhenriks !

Issue #172, opened 2026-01-09
VEP 172: Multi-Architecture Software Emulation


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood

**Current Feature Stage**:

New

**Feature Gate**:

`MultiArchitectureSoftwareEmulation`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
N/A

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.9
* Beta release target (x.y): TBD
* Stable release target (x.y): TBD

* [ ] Alpha
  * [ ] VEP PRs:
    * [ ] Initial VEP proposal (sig-compute/xx-multi-arch-software-emulation)
  * [ ] Code PRs:
    * [ ] Extend HypervisorDomainConfigurator with cross-arch detection
    * [ ] Add MultiArchitectureSoftwareEmulation feature gate
    * [ ] Implement QEMU binary path selection logic
    * [ ] Add emulator binary validation
    * [ ] Support AMD64 → ARM64 emulation
    * [ ] Support ARM64 → AMD64 emulation
  * [ ] Test PRs:
    * [ ] Unit tests for architecture detection and binary selection
    * [ ] Integration tests for domain conversion
    * [ ] E2E tests for cross-arch VM lifecycle
  * [ ] Docs PRs:
    * [ ] User documentation for enabling cross-arch emulation
    * [ ] QEMU binary installation guide (manual)
    * [ ] Performance expectations and use case guidance
    * [ ] Troubleshooting guide

jschintag:
@lyarwood i'm interested in this PR and woul volunteer to review your PR's.

lyarwood:
/assign @jschintag

@jschintag as discussed I'm going to co-assign you to this VEP as well moving forward. I don't think we need anything more in v1.10.0 aside from using the current emulation support to work out additional use cases we need to address. Such as allowing CDI to pick up the correct arch the guest when it doesn't correspond to the host.

jschintag:
> /assign [@jschintag](https://github.com/jschintag)
> 
> [@jschintag](https://github.com/jschintag) as discussed I'm going to co-assign you to this VEP as well moving forward. I don't think we need anything more in v1.10.0 aside from using the current emulation support to work out additional use cases we need to address. Such as allowing CDI to pick up the correct arch the guest when it doesn't correspond to the host.

Sure, though i would also add an e2e test lane to the next steps:
[...]

PR #173, opened 2026-01-09
VEP 172: Cross-Architecture Virtualization

lyarwood:
### VEP Metadata

**Tracking issue**: [#172](https://github.com/kubevirt/enhancements/issues/172)
**SIG label**: /sig compute

### What this PR does

This VEP proposes a `CrossArchitectureVirtualization` feature gate to enable KubeVirt to run VMs for guest architectures that differ from the host. The feature supports two emulation modes, auto-detected at runtime via libvirt capabilities:

- **Software emulation** via QEMU TCG: available on any host with cross-architecture QEMU binaries installed. Suitable for development, testing, and CI/CD workloads.
- **Hardware-accelerated virtualization** via KVM extensions (e.g. SAE for ARM64 guests on s390x): available when the host platform provides hardware support. Performance characteristics TBD pending upstream stack stabilization and benchmarking.

The emulation mode is transparent to the user — they enable the feature gate and create cross-architecture VMs. When hardware acceleration is available it is preferred; otherwise software emulation is used as a fallback.

Key design points:

- Single `CrossArchitectureVirtualization` feature gate covers both software and hardware paths
- Independent of `useEmulation` — native KVM VMs and cross-arch VMs coexist
- Architecture converter pipeline instantiated for **guest** architecture, not host
- Multi-tier node scheduling preference: native arch > hardware cross-arch (KVM) > software cross-arch (TCG)
- `SoftwareEmulation` / `HardwareEmulation` VMI conditions distinguish the active mode
- Hardware cross-arch KVM support auto-detected via libvirt capabilities and advertised as `kubevirt.io/cross-arch-kvm-<arch>` node labels

- Alpha targets v1.9.0 with software emulation (TCG) for amd64 ↔ arm64 only
- Hardware acceleration infrastructure lands in Alpha as a no-op (no hardware available yet)
- Hardware acceleration validation deferred to Beta or later, dependent on upstream kernel/QEMU/libvirt SAE support
- Performance claims for hardware emulation are intentionally left as TBD

### DIFF
--- a/veps/sig-compute/172-cross-arch-virtualization/vep.md
+++ b/veps/sig-compute/172-cross-arch-virtualization/vep.md
@@ -1,0 +1,1305 @@
+# VEP #172: Cross-Architecture Virtualization
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes introducing a `CrossArchitectureVirtualization` feature gate to
+enable KubeVirt to run virtual machines for guest architectures that differ
+from the host architecture. The feature supports two emulation modes,
+auto-detected at runtime:
+
+1. **Software emulation** via QEMU's TCG (Tiny Code Generator): available on
+   any host with the appropriate cross-architecture QEMU system binary
+   installed. Suitable for development, testing, and CI/CD workloads (10-100x
+   performance overhead).
+
+2. **Hardware-accelerated virtualization** via KVM extensions: available when
+   the host platform provides hardware support for cross-architecture
+   virtualization (e.g., SAE for ARM64 guests on s390x hosts).
+   Performance characteristics are not yet known and will be determined
+   once the upstream stack stabilizes and hardware is available for
+   benchmarking.
+
+The emulation mode is determined automatically based on libvirt capabilities —
+users enable the feature gate and create cross-architecture VMs without needing
+to specify which emulation backend to use. When hardware acceleration is
+available it is preferred; otherwise software emulation is used as a fallback.
+
+Currently, KubeVirt's `useEmulation` configuration only supports software
+emulation as a fallback for guests running the same architecture as the host
+when `/dev/kvm` is unavailable. This enhancement extends cross-architecture
+execution as a first-class capability.
+
+## Motivation
+
+Cloud providers and development teams often need to run workloads on different
+architectures for testing, CI/CD pipelines, or supporting heterogeneous
+environments. Currently, KubeVirt requires matching architectures between the
+host and guest, limiting its flexibility in multi-architecture scenarios.
+
+Cross-architecture emulation would enable:
+
+- Development and testing of ARM64 workloads on AMD64 infrastructure (and vice versa)
+- Migration scenarios where temporary cross-arch execution is needed
+- Cost optimization by using available hardware regardless of architecture
+- Enhanced CI/CD capabilities for multi-architecture container images
+- Running ARM64 workloads on s390x infrastructure using hardware-accelerated
+  cross-architecture virtualization
+- Consolidating workloads across architectures on a single platform
+
+## Goals
+
+- Enable VMs to run on nodes with different architectures using either software
+  or hardware-accelerated virtualization
+- Auto-detect hardware cross-architecture KVM support via libvirt capabilities
+  and prefer it over software emulation
+- Operate independently of `useEmulation`, allowing native KVM VMs and
+  cross-arch emulated VMs to coexist
+- Ensure proper architecture detection and QEMU binary selection
+- Set VMI conditions that distinguish between hardware-accelerated and
+  software-emulated cross-architecture execution
+- Maintain backward compatibility with existing emulation behavior
+
+## Non Goals
+
+- Performance optimization of software emulation (expected to be significantly
+  slower)
+- Supporting all possible architecture combinations (initial focus on amd64,
+  arm64, s390x)
+- Migration of running VMs across architectures
+- Automatic installation or management of QEMU binaries on nodes
+- Kernel or QEMU development for hardware acceleration support (upstream
+  responsibility)
+
+## Definition of Users
+
+This feature is intended for:
+
+- Developers requiring multi-architecture testing environments
+- CI/CD systems building and testing multi-architecture workloads
+- Organizations with heterogeneous cluster architectures
+- Users performing architecture transition planning
+- Organizations running s390x infrastructure that want to execute ARM64
+  workloads using hardware-accelerated cross-architecture virtualization
+
+## User Stories
+
+1. As a developer, I want to test my ARM64 VM image on an AMD64 development cluster without requiring dedicated ARM64 hardware. I would like my ARM-specific VM.yaml (with `spec.architecture: arm64`) to schedule on AMD64 workers and run using software emulation.
+
+2. As a CI/CD engineer, I want to run automated tests for multiple architectures on a single Kubernetes cluster to simplify infrastructure.
+
+3. As a platform engineer, I want to ensure that cross-architecture VMs are only scheduled when the required QEMU binaries are available on the nodes.
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -0,0 +1,1208 @@
...
+        }
+    }
+    return supported
+}
+```
+
+The detected architectures are advertised as node labels:
+
+```
+kubevirt.io/cross-arch-kvm-aarch64=true
```
jschintag, line 449: Would it make sense for scheduling to also add a label for supported software emulated arches?
Like
```
kubevirt.io/vm-arch-aarch64=true
```
set for all nodes that can run arm64 VMs (native, hardware or software). This way it could be set as a nodeSelector to ensure that the scheduling behaviour remains pending when there are no nodes that can run the VM.
With the current proposal it would instead schedule the VM, leading to a permanent CrashLoop, as the VM is restarted after each error.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -0,0 +1,1208 @@
...
+    message: "VM running with hardware-accelerated cross-architecture emulation (host=s390x, guest=arm64)"
+```
+
+These conditions are mutually exclusive.
+
+### QEMU Binary Requirements
+
+#### Node Prerequisites
+
+Nodes that will run cross-architecture VMs must have the appropriate QEMU system binaries installed:
```
jschintag: I don't think they need to be installed on the nodes, i think we need to include them into the virt-launcher image.
This should run independent from what is installed on the nodes (except kernel modules).

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -0,0 +1,1208 @@
...
+
+The implementation uses a multi-tier preference approach:
+
+1. **Remove the hard `kubernetes.io/arch` node selector** so the pod *can*
+   schedule on any node
+2. **Add a preferred node affinity** for the guest architecture with maximum
+   weight (100) so the scheduler *prefers* native-architecture nodes
+3. **Add a preferred node affinity** for hardware cross-arch KVM support with
+   weight 50, using the `kubevirt.io/cross-arch-kvm-<arch>` node label, so
+   hardware-accelerated cross-arch nodes are preferred over software-only nodes
```
jschintag, line 714: Same as before, there should be a final nodeSelector to ensure we do not schedule VMs on nodes where they can't run.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -0,0 +1,1208 @@
...
+      requests:
+        memory: 1Gi
+  volumes:
+  - name: containerdisk
+    containerDisk:
+      # Important: Use architecture-specific image
+      image: quay.io/containerdisks/fedora:40-aarch64
+```
+
+**Note on ContainerDisk Images**: The user must specify architecture-appropriate container images. KubeVirt does not automatically select multi-arch images. For ARM64 VMs, use ARM64 container disk images (e.g., `:*-aarch64` tags).
```
jschintag, line 964: While this should work, it would also i think log an error since currently the images are marked per-architecture and therefore produce a warning when pulling an arm64 image on amd64.

jschintag (commented):
I finished reading through the proposal and it looks great. I will look at the draft PR next.

What i am missing in the proposal is user documentation in https://github.com/kubevirt/user-guide.
There is already mention about Performance documentation, support matrixes and installing qemu binaries.
I would propose to add user-facing documentation as a goal for the Alpha stage.

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: Good point, thanks. Without a hard constraint the VM would schedule on any node and CrashLoop when the required emulation support is unavailable — that's a much worse UX than Pending.

I've added a `kubevirt.io/vm-arch-<arch>=true` node capability label (new section 6) set by virt-handler for every architecture a node can run (native + cross-arch via libvirt capabilities). This is used as a **hard node selector** in the scheduling section, while the existing native-arch and hardware-KVM preferences remain as soft affinities.

The scheduling priority is now: hard constraint (`kubevirt.io/vm-arch-<guest-arch>=true`) ensures the VM only lands on capable nodes, then soft preferences steer toward native (weight 100) > hardware cross-arch (weight 50) > software cross-arch.

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: You're right — QEMU runs inside the virt-launcher pod, so the binaries need to be in the virt-launcher image, not installed on the host directly. The only host-level dependency is kernel modules (e.g., KVM).

I've reframed this section: renamed "Node Prerequisites" to "virt-launcher Image Requirements" and updated the phased approach so that Alpha includes the cross-arch QEMU binaries directly in the virt-launcher image build (which is what the PoC already does). The old "Container Images" subsection is now "Image Size Considerations" focused on the Beta/Future sidecar/initContainer optimization.

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: Agreed — addressed together with your earlier comment. The `kubevirt.io/vm-arch-<guest-arch>=true` label is now a required node selector here, so VMs stay Pending when no capable node exists instead of CrashLooping.

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: Good point. The container runtime will log a warning when the image's platform metadata doesn't match the node architecture. The pull succeeds and the disk image is correct for the guest, but it generates noise.

I've added a "Note on container runtime warnings" paragraph in the API examples section documenting this as expected behavior for Alpha. We can revisit in Beta if it becomes a usability issue.

lyarwood:
Thanks for the thorough review @jschintag! I've pushed updates addressing all your comments:

1. **Scheduling safety**: Added `kubevirt.io/vm-arch-<arch>=true` node capability labels (new section 6) and a hard node selector so VMs stay Pending on incapable nodes instead of CrashLooping. The existing native-arch and hardware-KVM preferences remain as soft affinities.

2. **QEMU binaries in virt-launcher**: Reframed "Node Prerequisites" → "virt-launcher Image Requirements" — binaries go in the virt-launcher image, not on the host. Updated the phased approach so Alpha includes them in the image build directly (matching the PoC).

3. **ContainerDisk pull warnings**: Added a note documenting the cosmetic CRI-O/containerd warning as expected Alpha behavior.

4. **User-guide documentation**: Added `kubevirt/user-guide` to the Repos section and user-facing documentation as an Alpha graduation criteria item (covering feature gate enablement, supported arch combinations, QEMU requirements, performance limitations, and example VMI specs).

jschintag:
Thank you for the update @lyarwood.
From my perspective the proposal looks good. I will take a look at the PoC (likely tomorrow) before i give the LGTM.

jschintag:
Okay i had a look at the PR and when reviewing it 2 new questions popped into my head:
- What about VMs with devices that are not supported in emulation e.g. graphics?
- And somewhat related, would it make sense to allow per-VM configuration of emulation behavior:
    - Disable emulation
    - prefer hardware emulation
    - require hardware emulation

This would allow users to ensure production apps run on native system, whereas the slow emulation can be used to test/build new versions in the same cluster.

lyarwood:
Thanks @jschintag, good questions. Addressing both:

**Unsupported devices in emulation**

The VEP already handles this for graphics — the `GraphicsDomainConfigurator` silently disables video/VNC when software emulation is active because the minimal cross-arch QEMU binary lacks virtio-gpu (see "Graphics Device" section). For other devices, if the cross-arch QEMU binary doesn't support them, libvirt will reject the domain XML at startup and the error is surfaced as a VMI event.

I've added an "Unsupported Device Handling" subsection under Validation to make this clearer: we silently adapt where possible (graphics) and fail clearly when a device is genuinely unavailable. The user-guide documentation will list known device limitations per emulation mode.

**Per-VM emulation policy**

Good suggestion. Currently the feature gate is cluster-wide — when enabled, the scheduler *prefers* native execution (weight 100) but allows cross-arch fallback. There's no way for a user to say "this VM must never use emulation" or "this VM requires hardware acceleration".

I think this is best suited for Beta once we have Alpha feedback on real usage patterns. I've added it as a Beta graduation consideration:

> Consider per-VM emulation policy (e.g., a VMI-level field to disable cross-arch emulation, prefer hardware acceleration, or require hardware acceleration for specific workloads)

For Alpha, production VMs that must stay on native hardware can use an explicit `nodeSelector` for `kubernetes.io/arch` to override the relaxed scheduling, which effectively opts out of cross-arch emulation on a per-VM basis.

jschintag (approved):
/lgtm

Thank you for the VEP @lyarwood

vladikr:
/approve
Looks good to me as well!

Issue #176, opened 2026-01-13
VEP 0176: Redfish support for managing virtual machines


MarSik:
**Primary contact (assignee)**:

/assign MarSik

**Current Feature Stage**:

New

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:

TODO

**Timeline**:

Currently incubating.

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

The request at this moment is to allow moving a kubevirt-redfish project under the kubevirt project umbrella to get access to CI facilities needed to perform the necessary code cleanups and refactorings.

> [!IMPORTANT]
> Please keep this description up to date.

0xFelix:
Is this still on track for 1.9? However, I don't see it being bound to the KubeVirt release schedule anyway.

MarSik:
Yes, I just updated the PR. But you are right, it is not bound to the release schedule. There are no artifacts planned in 1.9.

vladikr:
/remove-lifecycle stale
I'm not 100% sure what happened here. 
@MarSik is that on track? Can you please update?

PR #177, opened 2026-01-13
VEP0176: kubevirt-redfish inclusion into kubevirt org

MarSik:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/176
**SIG label**:  /sig compute

### What this PR does

Please accept the inclusion of the kubevirt-redfish project under the kubevirt umbrella. This project provides the functionality of the Redfish protocol that can be used to integrate kubevirt VMs into higher level baremetal management systems.

There is not plan to release this as a core kubevirt project yet, the request is to accept the project as an incubating feature and make it ready first.

### DIFF
--- a/veps/sig-compute/0176_redfish_support/redfish.md
+++ b/veps/sig-compute/0176_redfish_support/redfish.md
@@ -1,0 +1,190 @@
+# VEP #0176: Redfish support for managing virtual machines
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+<!--
+Provide a brief overview of the topic)
+-->
+
+Please accept the inclusion of the [kubevirt-redfish](https://github.com/v1k0d3n/kubevirt-redfish/) project under the kubevirt umbrella. This project provides
+the functionality of the [Redfish protocol](https://www.dmtf.org/standards/redfish) that can be used to integrate kubevirt VMs into higher level
+baremetal management systems.
+
+## Motivation
+
+<!--
+Why this enhancement is important
+-->
+
+Both developers and system administrators require the ability to deploy applications or systems in local virtual environments like bare-metal ones.
+Orchestration systems like Spacewalk or OpenShift installers commonly use Redfish and Boot ISO approaches to deploy the operating system or application the user requested in the high level UI.
+
+## Goals
+
+<!--
+The desired outcome
+-->
+
+- Provide essential BMC functionality such as rebooting, changing boot devices, and mounting virtual media via the Redfish protocol against VMs backed by KubeVirt
+- Enable deployment tools like Metal3 and installers (OpenShift ABI or IPI) to treat the virtual machines in the same way they treat bare metal nodes
+
+## Non Goals
+
+<!--
+Why this enhancement is important Limitations to the scope of the design
+-->
+
+## Definition of Users
+
+<!--
+Who is this feature set intended for
+-->
+
+- A cluster administrator - The administrator of the hypervisor cluster where the VMs are running
+- A VM owner - The owner of the specific VMs on the hypervisor cluster. Most likely also a the administrator of a newly deployed cluster that uses the VMs.
+
+## User Stories
+
+<!--
+List of user stories this design aims to solve
+-->
+
+- As a cluster administrator I want to be able to enable redfish protocol support for VM running on the cluster.
+- As a cluster administrator I want to control who (credentials) can access the redfish endpoints per namespace.
+- As a VM owner I want to be able to control which VMs will be exposed to the redfish protocol.
+- As a VM owner (virtual cluster administrator), I want to deploy a cluster using Metal3 or bare-metal installers on KubeVirt VMs.
+
+## Repos
+
+<!--
+List of repose this design impacts
+-->
+
+A new repository should be created. I am proposing the name `kubevirt/redfish-controller`.
+
+## Design
+
+<!--
+This should be brief and concise. We want just enough to get the point across
+-->
+
+The redfish protocol implementation runs as a single service providing endpoints to manage all eligible VirtualMachines.
+Namespaces are represented as Chassis and VirtualMachines are the machines under their respective Chassis.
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
I wonder if renaming the project might make sense. `kubevirt/kubevirt-redfish` sounds slightly repetitive to me.

enp0s3 (commented):
@MarSik Hi! Would it make sense to integrate redfish support for KubeVirt into metalKube?

0xFelix (commented):
In general, looks good to me. Added two questions.

veps/sig-compute/0176_redfish_support/redfish.md
```diff
@@ -0,0 +1,174 @@
...
+A new repository should be created. I am proposing the name `kubevirt/redfish-controller`.
+
+## Design
+
+<!--
+This should be brief and concise. We want just enough to get the point across
+-->
+
+The redfish protocol implementation runs as a single service providing endpoints to manage all eligible VirtualMachines.
+Namespaces are represented as Chassis and VirtualMachines are the machines under their respective Chassis.
```
starbops: The single-endpoint design where one Redfish service manages all VMs across namespaces is an interesting choice. I’d like to understand the target use cases better, because they significantly affect whether this architecture fits.

In a typical bare-metal environment, there’s a 1:1 relationship between a BMC endpoint and a physical server. Tools like Metal3/Ironic register each BareMetalHost with its own BMC address pointing to a specific system. The Redfish spec itself doesn’t mandate this, but it’s a commonly-used pattern, and most BMC clients are built around that assumption.

It looks that the current design is more like an aggregator of multiple managed systems. That said, if the primary use case that we're going to fulfill can consume this pattern, then it should be fine.

starbops (commented):
Thank you, @MarSik, for making the proposal.

Just a few comments regarding the design choices made by the kubevirt-redfish project.

I also skimmed through the project codebase. I am curious about the project goals. Should we anticipate this, either the existing kubevirt-redfish or the upcoming redfish-controller, being the solution for a lightweight Redfish API wrapper, rather than providing general-purpose BMC emulators for KubeVirt virtual machines?

Lastly, do we have any plan for the conformance/interoperability testing? AIUI, Metal3/Ironic published a Redfish interoperability profile that we can validate using the Redfish interop validator.

https://book.metal3.io/bmo/supported_hardware.html#redfish-interoperability

I suppose we can discuss more on the community Slack channel, if you'd like to.

veps/sig-compute/0176_redfish_support/redfish.md
```diff
@@ -0,0 +1,174 @@
...
+## Alternatives
+
+<!--
+Outline any alternative designs that have been considered)
+-->
+
+An older request was approved in https://github.com/kubevirt/community/blob/main/design-proposals/kubevirtbmc.md but never materialized.
+
+This alternative implementation known as kubevirt-redfish provides only the Redfish implementation that is becoming the industry standard and was only planned as Stage 3
+in the original design proposal. However thanks to this, only one service and pod are needed, because the namespace and VM separation is handled via HTTP url nesting.
```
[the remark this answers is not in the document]
MarSik: You are right, my "never materialized" was about the inclusion to kubevirt. The original VEP was approved, but the implementation was not moved to kubevirt. I actually do not know why.

veps/sig-compute/0176_redfish_support/redfish.md
MarSik: There is still a 1:1 relationship, the only difference is that redfish uses the full url for machine identification not a dedicated IP like the older and traditional IPMI. That allows serving the multiple endpoints from a single service.

The benefit of having a single service that serves all the VMs is in two places: resource consumption (both cpu/memory and potentially IP addresses) and also no need for another service to manage starting and stopping of the BMC pods when VMs are started/stopped.

One additional reason rto separate VM and redfish creation related to the above point is that the cluster admin and VM namespace owner/user might be separate (automated or human) users.

Acedus (commented):
This is awesome (I was looking for something exactly like this a while back and I'm really excited to see this)!

BTW, the current kubevirt-redfish CR has cluster-wide permissions to get/create/update/delete DVs, PVCs, Pods, Services, CMs and Secrets, is all of that really necessary or can it be scoped down to the namespace in which kubevirt-redfish is deployed?

veps/sig-compute/0176_redfish_support/redfish.md
MarSik: The redfish allows access to reboot the VMs for example, but not for creating them or updating hardware configuration. That provides an extra bit of separation between the hypervisor cluster admin and the VM user. Traditionally the user would have access to the VM object or the namespace.

0xFelix (commented):
I'd love to see this being moved to the KubeVirt org, even if its current state is not perfect yet.

Can you add some sort of future plans to this as well? Like explaining what the current state is and roughly what still is planned for it?

veps/sig-compute/0176_redfish_support/redfish.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Scalability
+
+<!--
+Overview of how the design scales)
+-->
+
+The current design only uses one pod, service and endpoint for all eligible VMs. Control of multiple VMs is handled by changing the HTTP request URL. This is cheap and should scale well.
```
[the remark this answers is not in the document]
MarSik: Replicas are still not part of the project atm, however I did update the code to not keep state in memory. All the tracking is now part of the shared etcd state (via labels on various objects) and this allows recovery after pod restarts or rescheduling.

0xFelix:
Ping @MarSik: Do you want to address this so we can get it in? VEP freeze is on Thursday (although we don't need to follow as closely with this one).

MarSik:
@0xFelix 

> Ping @MarSik: Do you want to address this so we can get it in? VEP freeze is on Thursday (although we don't need to follow as closely with this one).

Yep, I just updated the content to provide more detailed current state and the graduation plan.

veps/sig-compute/0176_redfish_support/redfish.md
```diff
@@ -0,0 +1,190 @@
...
+- Code managed via peer review, clean and maintainable.
+
+### Beta
+
+- Support for high availability in form of leader election or distributed request processing.
+- Documentation and best practices.
+
+### GA
+
+- Integration with HCO and documented kubevirt deployment flows.
```
0xFelix: We need to think about that but otherwise it looks good to me. Can you make this optional?

MarSik: Done.

0xFelix (approved):
/lgtm

Thanks!

Issue #178, opened 2026-01-15
VEP 178: Integration of virt-lint validations


nertpinx:
**Primary contact (assignee)**: @nertpinx

/assign @nertpinx

**Current Feature Stage**: New

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

enp0s3:
Following an offline discussion we decided to defer the delivery to 1.10

PR #182, opened 2026-01-19
Add proposal document for VEP 178: virt-lint integration

nertpinx:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/178
**SIG label**: /sig compute

### What this PR does
Adds first proposal for VEP 178: virt-lint integration.

This is my first VEP, so take it with a grain of salt, but I'm very open to discussions.

### DIFF
--- a/veps/sig-compute/178-virt-lint-integration/virt-lint-integration.md
+++ b/veps/sig-compute/178-virt-lint-integration/virt-lint-integration.md
@@ -1,0 +1,155 @@
+# VEP #178: virt-lint integration
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This VEP allows using [virt-lint](https://gitlab.com/MichalPrivoznik/virt-lint)
+during VMI boot-up (and later possibly VM creation) to run various validators to
+check anything of interest that admins or users should be warned about.
+
+## Motivation
+
+The integration of virt-lint allows for sharing of various policies between
+projects as well as the separation between technical part of the codebase and
+the policy/validation part.  That consequently allows for separate updates and
+settings of the validators.
+
+The bigger idea is that the validators can be shared between other projects that
+are interested in checking various scenarios while each project can subscribe
+only to those they are interested in.
+
+## Goals
+
+The virt-lint validators will be shipped as part of kubevirt container image(s)
+-- for starters the virt-launcher container -- and all VMIs will be checked
+before starting.  Validation errors (fatal for virt-lint) will be reported in
+the logs and warnings will be forwarded as events to be seen in the web UI and
+in VMI's description.
+
+In the future there might be another virt-lint API that supports hinting,
+i.e. replying with suggestions, but since that was (at least to my knowledge)
+only raised in a non-technical discussion, it is something to be designed.
+Being an user of virt-lint, of course, gives leverage in the design discussion
+of such APIs.
+
+## Non Goals
+
+- Any modification of the definition.
+- Preventing VMI creation.
+
+## Definition of Users
+
+- VM owners
+- Cluster admins
+- KubeVirt developers
+
+## User Stories
+
+- As a VM owner I have a configuration that cannot, for some reason, be matched
+  by the scheduler (e.g. not a hard limit, not implemented yet) and I want to be
+  notified if there is a potential (non-fatal) performance, security, or any
+  other impact of such configuration.
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/178-virt-lint-integration/virt-lint-integration.md
```diff
@@ -0,0 +1,155 @@
...
+- VM owners
+- Cluster admins
+- KubeVirt developers
+
+## User Stories
+
+- As a VM owner I have a configuration that cannot, for some reason, be matched
+  by the scheduler (e.g. not a hard limit, not implemented yet) and I want to be
+  notified if there is a potential (non-fatal) performance, security, or any
+  other impact of such configuration.
```
victortoso: Do you have an example of such case?

victortoso (commented):
More than once I wished we had something like virt-lint after the sidecar hooks are called...

veps/sig-compute/178-virt-lint-integration/virt-lint-integration.md
```diff
@@ -0,0 +1,155 @@
...
+In the future there might be another virt-lint API that supports hinting,
+i.e. replying with suggestions, but since that was (at least to my knowledge)
+only raised in a non-technical discussion, it is something to be designed.
+Being an user of virt-lint, of course, gives leverage in the design discussion
+of such APIs.
+
+## Non Goals
+
+- Any modification of the definition.
+- Preventing VMI creation.
```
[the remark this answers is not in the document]
nertpinx: Well it can be added, that's true, but in my mind I did not think of it that way. Good point for discussion or an idea for changing the VEP.

veps/sig-compute/178-virt-lint-integration/virt-lint-integration.md
```diff
@@ -0,0 +1,155 @@
+# VEP #178: virt-lint integration
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
```
iholder101, line 26: Can you please add this new section to the VEP?

https://github.com/kubevirt/enhancements/blob/b199ea2cb179077ed149f522bb93cd346bce4454/veps/NNNN-vep-template/vep.md?plain=1#L5-L19

iholder101 (commented):
Thanks @nertpinx!

Using virt-lint with a shared domain XML validation is an interesting idea. I'd like to see it being explored further.

My main concern is embedding virt-lint directly in virt-launcher. That's the a critical component in KubeVirt, hence pulling in a Go->cgo->C->Rust chain plus Lua/Python interpreters into its boot path is risky - a segfault in these layers would crash virt-launcher and kill the VM. virt-lint being v0.0.1 with no stable API or official packaging makes this harder to justify.

I think this is actually a perfect use-case for the plugin mechanism we're working on ([VEP 190](https://github.com/kubevirt/enhancements/pull/191)). virt-lint maps naturally to a domain hook plugin - receive the DomainSpec, validate, return unmodified, emit warnings. Running it in its own plugin container isolates the dependencies, avoids pulling Rust/custom RPMs into KubeVirt's build, and lets virt-lint mature on its own schedule. This way we can test and use it out-of-tree and maybe reconsider integrating it into KubeVirt in the future when we have more confidence.

I'd suggest rewriting it as a follow-up VEP that builds on the plugin framework. WDYT?

nertpinx:
Sorry for not noticing this sooner, I don't use GH that often.

> Thanks @nertpinx!
> 
> Using virt-lint with a shared domain XML validation is an interesting idea. I'd like to see it being explored further.
> 
> My main concern is embedding virt-lint directly in virt-launcher. That's the a critical component in KubeVirt, hence pulling in a Go->cgo->C->Rust chain plus Lua/Python interpreters into its boot path is risky - a segfault in these layers would crash virt-launcher and kill the VM. virt-lint being v0.0.1 with no stable API or official packaging makes this harder to justify.
> 
> I think this is actually a perfect use-case for the plugin mechanism we're working on ([VEP 190](https://github.com/kubevirt/enhancements/pull/191)). virt-lint maps naturally to a domain hook plugin - receive the DomainSpec, validate, return unmodified, emit warnings. Running it in its own plugin container isolates the dependencies, avoids pulling Rust/custom RPMs into KubeVirt's build, and lets virt-lint mature on its own schedule. This way we can test and use it out-of-tree and maybe reconsider integrating it into KubeVirt in the future when we have more confidence.
> 
> I'd suggest rewriting it as a follow-up VEP that builds on the plugin framework. WDYT?

If the Domain Hook gets the XML, then it could, theoretically, work. I just moved on to something else in the meantime since there was almost no feedback on this issue.

@zippy2 Could you have a look and see if that is viable?

Issue #186, opened 2026-01-22
VEP 186: Host Devices Assignment with IOMMUFD


luckyh:
**Primary contact (assignee)**:

/assign @luckyh

**Current Feature Stage**:

New

**Feature Gate**:

`IOMMUFD`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

luckyh:
Closed in favor of https://github.com/kubevirt/enhancements/issues/266 .

PR #187, opened 2026-01-22
VEP 186: Host Devices Assignment with IOMMUFD

luckyh:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/186
**SIG label**: /sig compute

### What this PR does
Add first proposal for VEP 186: Host Devices Assignment with IOMMUFD

### DIFF
--- a/veps/sig-compute/186-hostdev-iommufd/hostdev-iommufd.md
+++ b/veps/sig-compute/186-hostdev-iommufd/hostdev-iommufd.md
@@ -1,0 +1,258 @@
+# VEP #186: Support Host Devices Assignment with IOMMUFD
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [X] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This proposal advocates for the adoption of IOMMUFD as an alternative
+VFIO backend to augment KubeVirt's host device assignment capabilities.
+
+IOMMUFD is a new, general-purpose user API in the Linux kernel intended
+to replace existing driver-specific IOMMU implementations, such as
+`VFIO_IOMMU_TYPE1`. With the integration in Qemu/KVM virtualization stack,
+IOMMUFD now can be used to support device passthrough for VMs.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


xpivarc:
@luckyh virt-launcher will have all the info and that is
Did the plugin provide `/dev/iommu`? Did the device has the `/dev/vfio/<group>` or `/dev/vfio/devices/<devid>` pattern? And this is enough to know if we should enable the new API or not

ajcaldelas:
> Updated the proposal for adopting the latest change of libvirt (since 12.2.0), that is, now the per-domain configuration `<iommufd enabled='yes' .../>` is primarily used together with `<driver iommufd='no'>` for configuring the vfio backend [1].
> 
> Apart from that, it expects kubevirt to handle the configuration of the iommufd device instead of libvirt. Due to this requires the `SYS_RESOURCE` capability the operation has to be performed in virt-handler, therefore, a new mechanism is needed to be introduced for passing file descriptors between handler and launcher.
> 
> [1] https://libvirt.org/formatdomain.html#host-device-iommufd

Am I understanding this correctly?  Basically we need either a webhook or expand virt-handler to do some fd-passing. Currently virt-handler doesn't have this capability correct?

luckyh:
> Am I understanding this correctly? Basically we need either a webhook or expand virt-handler to do some fd-passing. Currently virt-handler doesn't have this capability correct?

@ajcaldelas the original idea came up to my mind is that, we could let virt-hander open the `/dev/iommu` device node, configure the opened fd (which is about to set the per-process memlock accounting mode in order to ensure the proper rlimit setting for VMs, and this requires `SYS_RESOURCE`), and then pass the fd to virt-launcher.

However, @vladikr gives another proposal, that is, introducing a new device plugin [1] for handling the above procedure. Based on that, it needs to use either the virt-controller or a webhook to inject the iommufd resource request for the device assignments to VMs.

```
resources:
  limits:
    devices.kubevirt.io/iommufd: "1"
``` 

[1] https://github.com/kubevirt/iommufd-device-plugin

fossedihelm:
> > Am I understanding this correctly? Basically we need either a webhook or expand virt-handler to do some fd-passing. Currently virt-handler doesn't have this capability correct?
> 
> @ajcaldelas the original idea came up to my mind is that, we could let virt-hander open the `/dev/iommu` device node, configure the opened fd (which is about to set the per-process memlock accounting mode in order to ensure the proper rlimit setting for VMs, and this requires `SYS_RESOURCE`), and then pass the fd to virt-launcher.
> 
> However, @vladikr gives another proposal, that is, introducing a new device plugin [1] for handling the above procedure. Based on that, it needs to use either the virt-controller or a webhook to inject the iommufd resource request for the device assignments to VMs.
> 
> ```
> resources:
>   limits:
>     devices.kubevirt.io/iommufd: "1"
> ```
> 
> [1] [kubevirt/iommufd-device-plugin](https://github.com/kubevirt/iommufd-device-plugin?rgh-link-date=2026-04-08T05%3A38%3A49Z)

@luckyh The iommu-device-plugin was created as an external plugin.
The idea is to include it in virt-handler.
@ajcaldelas FYI.
Please let's continue this in #267

Issue #190, opened 2026-01-29
VEP 190: Structured KubeVirt Plugins


iholder101:
**Primary contact (assignee)**:

/assign @iholder101 
/assign @vladikr 

**Current Feature Stage**:

**Feature Gate**:

- `KubevirtPlugins`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
All other SIGs.

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/191.

**Timeline**:

* Alpha release target (x.y): v1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/191.
  * [ ] Code PRs:
  * [ ] Docs PRs:

fabiand:
How does this proposal add anything that we can not do today?

https://github.com/michalskrivanek/kubevirt-hugepages-addon
https://github.com/openshift-cnv/kubevirt-rawio-addon
https://github.com/openshift-cnv/kubevirt-force-rw-disk-addon

Are plugins that cover different ends of KubeVirt, including SCC, domxml changes, node changes ...

vladikr:
> How does this proposal add anything that we can not do today?
> 
> https://github.com/michalskrivanek/kubevirt-hugepages-addon https://github.com/openshift-cnv/kubevirt-rawio-addon https://github.com/openshift-cnv/kubevirt-force-rw-disk-addon
> 
> Are plugins that cover different ends of KubeVirt, including SCC, domxml changes, node changes ...

Hi @fabiand 

Let me try to answer...

With the old sidecar model, the cluster admin didn't have a way to permit only a specific hook.
It also only allowed to patch raw libvirt XML inside the pod.
 - No node-level operations
 - you couldn’t touch virt-launcher pod resources or specs (without webhooks)
 - couldn’t set VMI conditions, changes were untracked and unverified.

With the new approach we get :
 - The admin can approve specific plugins
 - structured DomainSpec modifications (with CEL or controlled sidecar)
 - Node hooks in virt-handler that will aloow now run privileged operations at specific lifecycle points (like PreVMStart, etc...)
 - and native K8s admission integration to mutate pod resources, set VMI conditions, add labels/annotations cleanly through standard Mutating/Validating policies - but all in one place.

With the new model, we're trying to make the plugins a first-class object, declarative, out-of-tree extension point(s)
[...]

PR #191, opened 2026-01-29
VEP 190: Kubevirt structured plugins

iholder101:
### VEP Metadata

**SIG label**: /sig compute

### What this PR does
Introduces a proposal for Kubevirt structured plugins.

VEP in a human-readable format: https://github.com/iholder101/kubevirt-enhancements/blob/vep/kubevirt-structured-plugins/veps/sig-compute/190-kubevirt-structured-plugins/vep.md

### DIFF
--- a/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
+++ b/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
@@ -1,0 +1,550 @@
+# VEP #190: Kubevirt Structured Plugins
+
+Owners:
+- @iholder101
+- @vladikr
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9. 
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Table of contents
+
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non Goals](#non-goals)
+- [Definition of Users](#definition-of-users)
+- [User Stories](#user-stories)
+- [Repos](#repos)
+- [Design](#design)
+  - [The Domain Hook](#the-domain-hook)
+    - [Domain Hook modes](#domain-hook-modes)
+    - [Domain Hook API examples](#domain-hook-api-exmaples)
+  - [The Node Hook](#the-node-hook)
+    - [Node hook points](#node-hook-points)
+    - [Node hook API examples](#node-hook-api-examples)
+  - [Admission Policies and Webhooks](#admission-policies-and-webhooks)
+    - [Admission options](#admission-options)
+    - [API example:](#api-example)
+  - [Plugin metadata](#plugin-metadata)
+- [API Examples (full version)](#api-examples-full-version)
+- [Alternatives](#alternatives)
+- [Possible future enhancements](#possible-future-enhancements)
+  - [Versioning and Upgrade Path](#versioning-and-upgrade-path)
+  - [multi-plugin support](#multi-plugin-support)
+- [Scalability](#scalability)
+- [Update/Rollback Compatibility](#updaterollback-compatibility)
+- [Functional Testing Approach](#functional-testing-approach)
+- [Implementation History](#implementation-history)
+- [Graduation Requirements](#graduation-requirements)
+  - [Alpha](#alpha)
+  - [Beta](#beta)
+  - [GA](#ga)
+
+## Overview
+
+KubeVirt currently has a gated feature that allows for virtual machine (VM) customization through a sidecar model that modifies raw libvirt XML.
+While functional, this approach is very limited (see below), and untrivial to create complex features with.
+
+This proposal is building on the philosophy of frameworks like NRI (Node Resource Interface).
+It introduces a structured hooking-based plugin mechanism to replace the existing "sidecar hook" model.
+The new mechanism will enable a safer, holistic, maintainable and powerful way of integrating out-of-tree functionality
+into KubeVirt.
+
+Not only will this plugin mechanism be able to modify the domain XML in a more structured way, but this proposal adds
+node-level hooks to perform operations by virt-handler, allowing privileged modifications to resources like
+housekeeping CPUs, PRlimit, or node sockets.
+
+Additionally, we propose integrating Kubernetes' Mutating Admission Policy to extend customization to pod and vmi
+specifications, enhancing the flexibility of virtual machine deployments.
+
+## Motivation
+
+Often, KubeVirt features must be implemented through external extensions rather than in-tree changes.
+See user stories below for detailed cases where this is needed.
+
+The current hooking mechanism in KubeVirt allows sidecars to directly modify libvirt XML.
+While this is helpful, this is an extremely partial and fragile solution that does not allow for
+complex features to be implemented completely out of tree.
+For example, it lacks the ability to perform node-level (possibly privileged) operations,
+change the virt-launcher's resources, modify the VMI's conditions, and more.
+
+This proposal suggests to deprecate and eventually remove the old Sidecar hook approach.
+
+Instead, we will be introducing a complete and full plugin mechanism that will allow for complex features
+to be implemented entirely out of tree, while being able to perform node-level operations,
+modify objects like pods, VMs and VMIs via admission techniques and much more.
+This will be achieved by introducing a `Plugin` Custom Resource Definition (CRD) that will serve as a central
+place to capture all plugin-related components, which will let the admin both reason about a plugin easily
+and to permit only whatever's acceptable by him via admission policies.
+
+## Goals
+
+The high level goal is to allow complex features to be implemented out-of-tree while integrating well with KubeVirt.
+
+For example, this includes:
+- Support a structured way of modifying the `DomainSpec` via plugins connected at well-defined hook points.
+- Support two plugin modes: simple JSON/CEL-based and advanced sidecar-based.
+- Support modifying Kubernetes objects (e.g. VM, VMI, virt-launcher pod, other pods, etc) via admission policies or webhooks.
+  this can be helpful to modify properties like resources, conditions (e.g. migratability), etc.
+- Track and depend on Kubernetes Admission Policies and Webhooks as part of a plugin,
+  providing a central place to reason about objects modifications and validations.
+- Support node-level (possibly privileged) operations in virt-handler via trusted plugins in well-defined hook points.
+- Deprecate and eventually remove the `Sidecar` feature gate.
+
+## Non Goals
+
+- Allowing arbitrary, unvalidated XML modifications.
+- Provide a helper library/framework for simplifying Plugin development (this should be addressed in a follow-up VEP).
+
+## Definition of Users
+
+- KubeVirt Developer: aims to develop features for KubeVirt.
+- Cluster administrator: in charge of managing clusters in production.
+
+## User Stories
+
+- As a cluster admin, I would like to extend or modify KubeVirt in my clusters, but I am not interested in forking
+  KubeVirt or rebuilding it due to the maintenance overhead.
+- As a KubeVirt developer, I would like to design a feature for advanced VM owners, so that they can test it quickly
+  and provide feedback. When it is in good shape, I would like to file a mature VEP with a POC and feedback from
+  real users.
+- As a KubeVirt developer of a very complex feature, I understand that a VEP would be challenging to review and comprehend.
+  To help the community understand my intentions, I wish to build a high-quality POC that demonstrates a limited-scoped
+  working feature to serve as a base ground for an enhancement proposal discussion.
+- As a KubeVirt developer, I want to be able to develop a feature that the core
+  KubeVirt community does not aim to maintain for at least one of these reasons:
+  - A niche specialized behavior is needed that does not make sense upstream.
+  - Too complex / dangerous / for most use-cases.
+  - No expertise in this subject among the community members (no one to maintain it).
+
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+The architecture of the new Plugin mechanism will consist of these different components:
+**domain hook**, **node hook**, **mutating policies** and **plugin metadata**.
+
+Let's describe them one by one,
+then explain the overall CRD structure that centralizes all plugin-related configuration and components.
+
+### The Domain Hook
+
+The proposed mechanism focuses on modifying the `DomainSpec`, a Go struct generated from the VirtualMachineInstance
+(VMI) spec, which serves as the internal representation of a VM's desired configuration before conversion to libvirt XML.
+Hooks will be executed after the initial `DomainSpec` generation but before the final XML is created.
+
+The current KubeVirt `DomainSpec` struct only includes a subset of the fields available in the full libvirt `DomainSpec`.
+To allow the plugin complete control over the generated XML (similar to the legacy Sidecar Hook),
+it will be replaced it with the complete, fully-expanded libvirt DomainSpec struct.
+
+#### Domain Hook modes
+
+The domain hook would support two types of hooks:
+
+**Simple hooks**: These are JSON or CEL based hooks.
+These hooks are valuable for simple XML changes that do not require complex logic.
+These hooks are easily deployable, eliminating the need to write code, provide a container image,
+running another sidecar container, etc.
+
+**Advanced (plugin-based) hooks**: In this mode, a sidecar container would run inside `virt-launcher`.
+This sidecar container would listen to a defined socket, waiting for requests by virt-launcher.
+In pre-defined hook points (e.g. before the domain XML is applied) virt-launcher would send a request to the socket,
+providing the `DomainSpec` struct generated by virt-launcher alongside a `DomainHookContext` struct that will contain
+extra information.
+
+Currently, we think of a single hook point, right after the domain XML is generated but before it's handed to libvirt.
+Moving forward we can support more hook points if needed.
+
+#### Domain Hook API examples
+
+<ins>Simple Hooks</ins>:
+
+JsonPatch based simple hook:
+```yaml
+domainHooks:
+  - type: JsonPatch
+    operations:
+      - op: add
+        path: /spec/devices/disks
+        value:
+          name: monitoring-disk
+          disk:
+            bus: virtio
+```
+
+CEL based simple hook:
+```yaml
+domainHooks:
+  - type: ApplyConfiguration
+    expression: |
+      object.spec.devices.disks + [
+        {
+          "name": "monitoring-disk",
+          "disk": {
+            "bus": "virtio"
+          }
+        }
+      ]
+```
+
+<ins>Advanced Hooks</ins>:
+
+```yaml
+domainHooks:
+  - type: Plugin
+    socketPath: "/var/run/kubevirt-plugin/my-cool-plugin/my-plugin-socket.sock"
+```
+
+Note: The socket would need to reside under `/var/run/kubevirt-plugin/<my-plugin-name>`,
+which will be defined as a shared volume with the compute container.
+The plugin's container will have visibility only to its plugin folder in an isolated way.
+
+### The Node Hook
+
+To support node-level customizations, hooks can also define operations triggered by the virt-handler pod.
+These are invoked via gRPC (or ttRPC, as used by NRI) calls from virt-handler to a plugin server
+(a DaemonSet running on each node).
+The communication is via a socket (its path is defined by the `Plugin` CR).
+The existence of a socket marks the readiness of the plugin - similarly to NRI plugins.
+
+The node hooks will not be running by virt-handler directly, but instead will be triggered by virt-handler via an RPC
+mechanism to be run inside the plugin's DaemonSet.
+This approach gives the cluster-admin the flexibility to decide (via validating admission policies) what are the acceptable
+resources and capabilities that the plugin DaemonSet will be defined with.
+It also delegates the responsibility of managing resources to the plugin author.
+
+These hooks are invoked at specific points in virt-handler’s controllers.
+Node hooks cannot modify `DomainSpec` or XML directly but can influence runtime behavior.
+Plugins can implement both domain and node hooks for cohesive functionality.
+
+Similarly to the domain hooks, `virt-handler` communicates with the relevant socket at pre-defined hook points.
+It provides a `NodeHookContext` struct to provide context to the plugin's DaemonSet.
+
+#### Node hook points
+
+- **PreVMStart**: Before VM launch (e.g., setup node devices/network; aligns with vmUpdateHelperDefault).
+- **PostVMStart**: After VM is running (e.g., verify node resources).
+- **PreMigrationSource**: Before migration from source node (e.g., prepare sockets; aligns with migrateVMI in migration-source.go).
+- **PostMigrationSource**: After migration completes on source (e.g., cleanup; aligns with handleSourceMigrationProxy).
+- **PreMigrationTarget**: Before migration to target node (e.g., setup target sockets; aligns with prepareMigrationTarget in migration-target.go).
+- **PostMigrationTarget**: After migration completes on target (e.g., repair/verify; aligns with finalizeMigration).
+- **OnVMStop**: When VM stops (e.g., cleanup node resources; aligns with processVmUpdate on shutdown).
+- **PostVMStop**: Called periodically.
+
+The plugin would have to explicitly mention which hook points it wishes to register for.
+
+#### Node hook API examples
+
+```yaml
+nodeHooks:
+  - mode: Plugin
+    socket: /var/run/my-node-socket.sock
+    permittedHooks:
+    - PreVMStart
+    - PostMigrationSource
+selector:
+  matchLabels:
+    use-plugin: my-plugin
+  matchFields:
+    - key: metadata.name
+      operator: Prefix
+      value: vm-
+  conditions: # CEL-based
+    - "vmi.status.phase == 'Running'"
+    - "vmi.status.conditions.exists(c, c.type == 'Ready' && c.status == 'True')"
+    - "vmi.status.conditions.exists(c, c.type == 'LiveMigratable' && c.status == 'True')"
+```
+
+### Admission Policies and Webhooks
+
+In order to provide pod or VMI level modifications and validations,
+the new plugin mechanism would be able to track and depend on Kubernetes
+[Mutating Admission Policies](https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/)
+and [Validation Admission Policies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/).
+By utilizing these admission policies, we would support providing a declarative policy to modify and/or validate VMIs
+and pods which would be performed on the API server's side.
+
+The fact that admission policies/webhooks are listed in the `Plugin` CR has the following advantages:
+1. It provides a centric place to holistically understand every plugin component.
+2. It allows the KubeVirt plugin mechanism to track these objects and consider the plugin ready only when they are
+deployed and ready for action.
+
+#### Admission options
+
+This VEP proposes to use admission mutations and validations according to the amount of complexity needed.
+It is recommended to use the least complex method whenever possible.
+
+These admission policies/webhooks will be tracked by KubeVirt's plugin mechanism,
+but a different component (e.g. the job/controller that implements the plugin) will be responsible to deploy and delete them.
+
+<ins>Admission policies</ins>:
+
+Admission policies can be used to mutate or validate an object using a declarative model that runs on the API server.
+This is preferred over admission webhooks whenever possible.
+
+For these cases, the plugin can reference admission policies by name and namespace.
+The plugin mechanism will verify that the referenced policies are ready before considering
+the plugin to be fully ready to use.
+
+<ins>Admission webhooks</ins>:
+
+Since admission policies are purely CEL (or JSONPatch) based, they are more limited than admission webhooks.
+For example, they cannot make API calls or perform a chain of statements - only a single expression.
+Therefore, in complex scenarios, an admission webhook would also be supported.
+Similarly to admission policies, the plugin can reference webhooks by name.
+The plugin mechanism will verify that the referenced webhooks are ready before considering
+the plugin to be fully ready to use.
+
+#### API example:
+
+```yaml
+spec:
+  mutatingAdmissionPolicies:
+    - name: "my-mutating-policy-1"
+  validatingAdmissionPolicies:
+    - name: "my-validating-policy-1"
+    - name: "my-validating-policy-2"
+  mutatingAdmissionWebhooks:
+    - name: "my-mutating-webhook-1"
+    - name: "my-mutating-webhook-2"
+  validatingAdmissionWebhooks:
+    - name: "my-validating-webhook-1"
+```
+
+### Plugin metadata
+
+Each plugin would have to provide some basic metadata about itself.
+
+The main idea here goes beyond specifying a unique name and identity to the plugin,
+but also set the ground for providing an upgrade path, versioning, multi-plugin support, dependency between plugins and more.
+See [possible plans for the future](#possible-future-enhancements) below.
+
+These are the basic fields every plugin would need to populate:
+- name:  a unique identifier for the plugin.
+- Version: Specifies the plugin’s version.
+- failureStrategy: Defines the behavior when a plugin fails. `Fail` (default) blocks the operation, `Ignore` logs the error and continues.
+- timeout: The maximum duration to wait for a plugin response before considering it failed (e.g. `30s`, `2m`). Defaults to `30s`.
+
+## API Examples (full version)
+
+```yaml
+apiVersion: plugin.kubevirt.io/v1alpha1
+kind: Plugin
+metadata:
+  name: monitoring-hook
+spec:
+  PluginMetadata:
+    name: "best-plugin-ever"
+    version: "0.0.1"
+    failureStrategy: Fail
+    timeout: 30s
+
+  domainHooks:
+  # Simple hook: JsonPatch
+  - type: JsonPatch
+    operations:
+      - op: add
+        path: /spec/devices/disks
+        value:
+          name: monitoring-disk
+          disk:
+            bus: virtio
+  # Simple hook: CEL-based ApplyConfiguration
+  - type: ApplyConfiguration
+    expression: |
+      object.spec.devices.disks + [
+        {
+          "name": "extra-disk",
+          "disk": {
+            "bus": "virtio"
+          }
+        }
+      ]
+  # Advanced hook: Plugin sidecar
+  - type: Plugin
+    socketPath: "/var/run/kubevirt-plugin/my-cool-plugin/my-plugin-socket.sock"
+
+  mutatingAdmissionPolicies:
+    - name: "my-mutating-policy-1"
+  validatingAdmissionPolicies:
+    - name: "my-validating-policy-1"
+    - name: "my-validating-policy-2"
+  mutatingAdmissionWebhooks:
+    - name: "my-mutating-webhook-1"
+    - name: "my-mutating-webhook-2"
+  validatingAdmissionWebhooks:
+    - name: "my-validating-webhook-1"
+
+  nodeHooks:
+    - mode: Plugin
+      socket: /var/run/my-node-socket.sock
+      permittedHooks:
+      - PreVMStart
+      - OnVMStop
+      - PostVMStop
+
+  selector:
+    matchLabels:
+      use-plugin: my-plugin
+    matchFields:
+      - key: metadata.name
+        operator: Prefix
+        value: vm-
+    conditions: # CEL-based
+      - "vmi.status.phase == 'Running'"
+      - "vmi.status.conditions.exists(c, c.type == 'Ready' && c.status == 'True')"
+      - "vmi.status.conditions.exists(c, c.type == 'LiveMigratable' && c.status == 'True')"
+```
+
+## Alternatives
+
+One of the obvious alternatives to this approach is forking KubeVirt in order to provide out-of-tree functionality.
+
+While a fork is possible, the maintenance price of using it is huge,
+mainly because the core KubeVirt is not aware of the made changes,
+hence the added features can go through rough rebase conflicts, which is bad,
+or be completely broken by new KubeVirt logic, which is worse.
+
+## Possible future enhancements
+
+These are ideas that we are not currently aiming to implement.
+They are left here for a reference for future discussions.
+
+* Version Incompatibility: XML changes may break with updates to KubeVirt or libvirt.
+Plugins declare version constraints (kubeVirtVersionCondition, libvirtVersionCondition) and upgrade paths,
+ensuring compatibility across versions.
+This is an example to be considered:
+```yaml
+spec:
+  PluginMetadata:
+    name: "best-plugin-ever"
+    version: "0.0.1"
+    failureStrategy: Fail
+    timeout: 30s
+    kubeVirtVersionCondition: "version > 1.5"
+    libvirtVersionCondition: "version > 9.0.0"
+    upgradePaths:
+      - KubeVirtVersion: 1.7
+        pluginImageTag: 1.7
+      - KubeVirtVersion: 1.8
+        pluginImageTag: 1.8
+    dependsOn:
+      - "another-cool-plugin"
+    preservePaths: "domainspec.devices.disks"
+```
+* Better Auditing: Changes are not tracked, making debugging difficult.
+Consider for modifications to be logged via Kubernetes events for traceability.
+* Support multi-plugins: dependencies, conflict detection, etc.
+
+### Versioning and Upgrade Path
+
+With the following fields the plugin system will enable to properly define version constraints and a planned upgrade path:
+- kubeVirtVersionCondition: a CEL based condition constraint for KubeVirt's version.
+- libvirtVersionCondition: a CEL based condition constraint for Libvirt's version.
+- upgradePaths: This field will be a list of values mapping a KubeVirt version to an image tag.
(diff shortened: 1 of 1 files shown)


lyarwood (commented):
Nice ideas, just some quick comments.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+
+Note: The socket would need to reside under `/var/run/kubevirt-plugin`,
+which will be defined as a shared volume with the compute container.
+
+### The Node Hook
+
+To support node-level customizations, hooks can also define operations that run in virt-handler on the node.
+These are invoked via gRPC (or, ttRPC, that’s used by NRI) calls from virt-handler to a plugin server
+(a DaemonSet running on each node). 
+node hooks run in virt-handler on the host node, allowing operations that affect node resources.
```
Acedus: Does that mean that the plugin can only work with a predetermined set of properties similarly to NRI? What exactly does the virt-handler run?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+These hooks are invoked at specific points in virt-handler's controllers.
+Node hooks cannot modify `DomainSpec` or XML directly but can influence runtime behavior.
+Plugins can implement both domain and node hooks for cohesive functionality.
+
+Similarly to the domain hooks, `virt-handler` will communicate with the relevant socket at pre-defined hook points.
+It would provide a `NodeHookContext` struct, and ask for a function to run by virt-handler.
+
+#### Node hook points
```
Acedus: What are the idempotency requirements? If a migration fails halfway and retries, will the `PreMigrationTarget` hook be called again? How should the plugin handle that?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+    socketPath: "/var/run/kubevirt-plugin/my-plugin-socket.sock"
+```
+
+Note: The socket would need to reside under `/var/run/kubevirt-plugin`,
+which will be defined as a shared volume with the compute container.
+
+### The Node Hook
+
+To support node-level customizations, hooks can also define operations that run in virt-handler on the node.
+These are invoked via gRPC (or, ttRPC, that’s used by NRI) calls from virt-handler to a plugin server
```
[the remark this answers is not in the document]
Acedus: Cont. How do we mitigate risks such as unix socket hijacking or identity spoofing?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+(a DaemonSet running on each node). 
+node hooks run in virt-handler on the host node, allowing operations that affect node resources.
+These hooks are invoked at specific points in virt-handler's controllers.
+Node hooks cannot modify `DomainSpec` or XML directly but can influence runtime behavior.
+Plugins can implement both domain and node hooks for cohesive functionality.
+
+Similarly to the domain hooks, `virt-handler` will communicate with the relevant socket at pre-defined hook points.
```
Acedus: How do we handle failures when the plugin server is unreachable or exceeds its timeout budget (is there one to begin with)? How are synchronous blocking calls managed to prevent cascading latency or resource starvation within the node-level handler?

Acedus (commented):
Super cool idea, mostly concerned with node hook (maybe a follow-up VEP?)

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+
+## User Stories
+
+- As an enterprise, it prioritizes rapid delivery over backward-compatibility, community support and architectural
+  stability, but is not interested in forking KubeVirt due to the huge maintenance overhead.
+  It wishes to deliver fast by developing a Plugin.
+- As an enterprise, it wants to design a feature for an advanced customer that wants to test it quickly and provide
+  feedback.
+  With a quick feedback loop, the company is able to make quick changes.
+  After the customer is satisfied, the company can file a mature VEP with a POC and feedback from real users.
```
dankenigsberg: I don't think that this is a user story, it is more of a business case (which does not belong here).

```suggestion
- As a cluster admin, I would like to extend or modify KubeVirt in my clusters, but I am not not interested in forking KubeVirt or rebuilding it due to the maintenance overhead.
- I would like to design a feature for my advanced VM owners, so that they can test it quickly and provide
  feedback. When it is in good shape, I would like to file a mature VEP with a POC and feedback from real users.
```

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+- As a developer of a very complex feature, I understand that a VEP would be challenging to review and comprehend.
+  To help the community understand my intentions, I wish to build a high-quality POC that demonstrates a limited-scoped
+  working feature to serve as a base ground for an enhancement proposal discussion.
+- As a feature developer, I want to be able to develop a feature that the core
```
dankenigsberg:
```suggestion
- As a KubeVirt developer, I want to be able to develop a feature that the core 
```

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+- [Functional Testing Approach](#functional-testing-approach)
+- [Implementation History](#implementation-history)
+- [Graduation Requirements](#graduation-requirements)
+  - [Alpha](#alpha)
+  - [Beta](#beta)
+  - [GA](#ga)
+
+## Overview
+
+KubeVirt currently allows for virtual machine (VM) customization through a sidecar model that modifies raw libvirt XML.
```
dankenigsberg:
```suggestion
KubeVirt currently has a gated feature that allows for virtual machine (VM) customization through a sidecar model that modifies raw libvirt XML.
```

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+
+This proposal addresses these issues by introducing a `Plugin` Custom Resource Definition (CRD) that allows:
+- Structured modifications to the `DomainSpec` in a safe way.
+- Node-level privileged operations to run via virt-handler.
+- Integrating Kubernetes' Mutating Admission Policy, we further enable pod-level and vmi-level customizations,
+  providing a comprehensive solution for VM and pod configuration.
+
+## Goals
+
+- Introduce a `Plugin` CRD that centralizes all plugin-related configuration and components.
```
dankenigsberg: This is not a goal, but how the VEP suggests to implement the goals of the users.
```suggestion
```

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+- Support a structured way of modifying the `DomainSpec`.
+- Support two hook modes: simple JSON/CEL-based and advanced plugin-based.
+- Ensure hooks are validated, version-compatible, and auditable via Kubernetes events.
+- Support modifying pod attributes like resources, migratability, etc.
+- Support Mutating Admission Policy to allow vmi or pod level modifications.
+- Support node-level privileged operations in virt-handler.
+- Provide a way of versioning the plugin and provide KubeVirt version contraints.
+- Provide a way to plan an upgrade path in advance.
+- Support multi-plugins: dependencies, conflict detection, etc.
```
dankenigsberg:
```suggestion
- Support multi-plugins: dependencies, conflict detection, etc.
- Deprecate and eventually remove the `sidecar` feature gate.
```

dankenigsberg (commented):
This vep is very much appreciated; pardon for this preliminary review.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: My general answer is - the plugin developer has a lot of power and a lot of responsibility :smile:. Plugins are indeed dangerous.

But to be more specific, I think things like timeouts and failure strategies should be configurable at the Plugin CRD level. For example, `failureStrategy` could be `Fail` (block the operation) or `Ignore` (log and continue). The admin can then enforce policies like "I don't accept plugins with more than 120s timeout" or "I only accept plugins with `failureStrategy: Ignore`" via a VAP.

Great point. I'll add these configurables to the Plugin CRD.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,555 @@
...
+  providing a comprehensive solution for VM and pod configuration.
+
+## Goals
+
+- Introduce a `Plugin` CRD that centralizes all plugin-related configuration and components.
+- Support a structured way of modifying the `DomainSpec`.
+- Support two hook modes: simple JSON/CEL-based and advanced plugin-based.
+- Ensure hooks are validated, version-compatible, and auditable via Kubernetes events.
+- Support modifying pod attributes like resources, migratability, etc.
+- Support Mutating Admission Policy to allow vmi or pod level modifications.
```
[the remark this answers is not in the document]
iholder101: Yes, and we plan to use the Kubernetes mechanisms. I think this is phrased in a confusing way, I've rephrased. Thanks!

iholder101:
Thanks @lyarwood @Acedus and @dankenigsberg for your reviews! Much appreciated!

I believe I've addressed the comments. You're welcome to take another look!

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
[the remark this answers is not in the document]
iholder101: > NRI plugins actually probe the NRI socket, not the other way around, if we re-queue the task how do we guarantee plugin ordering (e.g., depends-on)? Do other plugins not run until the missing one completes?

You're right about NRI doing it the other way around, thanks for enlightening me!

Unlike NRI though, we have the Plugin CR for discovery and configuration. That being said, readiness is per-node (DaemonSet pods start at different times), so I still suggest using socket existence as the per-node readiness signal. IOW the plugin must only create its socket once it's fully ready to serve.

If a plugin's socket doesn't exist on a given node, or fails to respond, it should "fail". What happens then depends on that plugin's `failureStrategy`. The same applies to its dependents - if a plugin is skipped because a dependency failed, its own `failureStrategy` determines the outcome. Each plugin independently answers "what happens if I don't run," regardless of the reason.

> how would the virt-handler authenticate a plugin registration?

The admin can decide, via VAPs, to restrict the socket path. He can then rely on filesystem permissions to ensure only the right process can create a socket at that path. This is the same trust model used by Kubernetes device plugins and NRI IIUC.

What do you think?
Would appreciate your opinion :)

iholder101:
A few changes made:

[Change](https://github.com/kubevirt/enhancements/compare/fbe9665a41aa4e48fb9781fcdc4f99cbce78a74e..a527dc10d5bb84bc5ea7a9a7e2ee7d412e408b7d) for removing the "compact" admission policies. @lyarwood, @dankenigsberg and others thought it's pretty complex and the gain is pretty low. I tend to agree, and we can always add them later if we fell they're necessary.

[Change](https://github.com/kubevirt/enhancements/compare/a527dc10d5bb84bc5ea7a9a7e2ee7d412e408b7d..caa5063bf67467c8c789cc0023304793f8d16352) for consistent terminology - use `Plugin` CR instead of calling it a `VirtualMachineHook`, which is the older and more confusing name which is now removed.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,526 @@
...
+  Admission Policies for pod/VMI-level customizations.*
+
+## Goals
+
+- Support a structured way of modifying the `DomainSpec` via plugins connected at well-defined hook points.
+- Support two plugin modes: simple JSON/CEL-based and advanced sidecar-based.
+- Ensure plugins are validated, version-compatible, and auditable via Kubernetes events.
+- Support modifying pod attributes like resources, migratability, etc.
+- Track and depend on Kubernetes Admission Policies and Webhooks as part of a plugin, providing a unified
+  experience for VMI and pod level modifications and validations.
```
dankenigsberg: I don't think that this is KubeVirt responsibility or expertise. As a cluster-admin, I am well-versed in deploying software on the cluster and tracking it.

KubeVirt must handle plugin errors gracefully, but not track plugin availability (that would be raceful anyway)

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,526 @@
...
+- Support node-level privileged operations in virt-handler via plugin hook points.
```
dankenigsberg:
```suggestion
- Let trusted plugins perform privileged node-level operations in virt-handler.
```

dankenigsberg (changes requested):

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,524 @@
...
+running another sidecar container, etc.
+
+**Advanced (plugin-based) hooks**: In this mode, a sidecar container would run inside `virt-launcher`.
+This sidecar container would listen to a defined socket, waiting for requests by virt-launcher.
+In pre-defined hook points (e.g. before the domain XML is applied) virt-launcher would send a request to the socket,
+providing the `DomainSpec` struct generated by virt-launcher alongside a `DomainHookContext` struct that will contain
+extra information.
+
+Currently, we think of a single hook point, right after the domain XML is generated but before it's handed to libvirt.
+Moving forward we can support more hook points if needed.
```
bgartzi, line 173: Would in this case also make sense to integrate the domain XML mods hook in the target pre-migration hook list?

Some sidecars need to modify the domainXML based on pod specifics (e.g. a device attached to it), which are subject to change in the migration target. Right now other ways around this are used, such as creating symlinks in the hook-sidecar socket directory.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,524 @@
...
+<ins>Advanced Hooks</ins>:
+
+```yaml
+domainHooks:
+  - type: Plugin
+    socketPath: "/var/run/kubevirt-plugin/my-plugin-socket.sock"
+```
+
+Note: The socket would need to reside under `/var/run/kubevirt-plugin`,
+which will be defined as a shared volume with the compute container.
```
bgartzi: Are these shared volumes isolated from each other? I mean, would each hook container get its own socket shared dir as in https://github.com/kubevirt/kubevirt/pull/14623?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,524 @@
...
+        }
+      ]
+```
+
+<ins>Advanced Hooks</ins>:
+
+```yaml
+domainHooks:
+  - type: Plugin
+    socketPath: "/var/run/kubevirt-plugin/my-plugin-socket.sock"
```
bgartzi: Where is the hook server container image reference specified?

And I might be stuck in how the sidecar mechanism works nowadays, but doesn't the virt-launcher have enough information to find the socket in the shared volume by itself?

bgartzi (commented):
Leaving a few comments.

mhenriks:
@vladikr @iholder101 

I'd love to use the plugin model for SCSI raw I/O support (https://github.com/kubevirt/enhancements/issues/259). A plugin could handle the domain XML mutation (rawio="yes" on LUN disks) and the pod security context mutation (adding CAP_SYS_RAWIO to the bounding set), but there's a gap around Linux capability propagation.

  KubeVirt propagates capabilities to QEMU via a chain of ambient capabilities and file capabilities:

  1. File capability on virt-launcher-monitor — set at build time via bazeldnf xattrs in
  https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/cmd/virt-launcher/BUILD.bazel#L72-L80
  2. Ambient cap: virt-launcher-monitor → virt-launcher —
  https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/cmd/virt-launcher-monitor/virt-launcher-monitor.go#L105
  3. Ambient cap: virt-launcher → virtqemud —
  https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/pkg/virt-launcher/virtwrap/util/libvirt_helper.go#L241

  Today these are hardcoded to CAP_NET_BIND_SERVICE. To support a new capability like CAP_SYS_RAWIO, all three points need updating

Would the plugin model be able to support declaring additional Linux capabilities that get wired into this chain? Without that, any feature requiring a new capability on the QEMU process would need core KubeVirt changes and couldn't be implemented purely as a plugin.

vladikr:
> @vladikr @iholder101
> 
> I'd love to use the plugin model for SCSI raw I/O support (#259). A plugin could handle the domain XML mutation (rawio="yes" on LUN disks) and the pod security context mutation (adding CAP_SYS_RAWIO to the bounding set), but there's a gap around Linux capability propagation.
> 
> KubeVirt propagates capabilities to QEMU via a chain of ambient capabilities and file capabilities:
> 
> 1. File capability on virt-launcher-monitor — set at build time via bazeldnf xattrs in
>    https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/cmd/virt-launcher/BUILD.bazel#L72-L80
> 2. Ambient cap: virt-launcher-monitor → virt-launcher —
>    https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/cmd/virt-launcher-monitor/virt-launcher-monitor.go#L105
> 3. Ambient cap: virt-launcher → virtqemud —
>    https://github.com/kubevirt/kubevirt/blob/a8a6cef466778dd0bfbe58c44a7b4d004cbceb15/pkg/virt-launcher/virtwrap/util/libvirt_helper.go#L241
> 
> Today these are hardcoded to CAP_NET_BIND_SERVICE. To support a new capability like CAP_SYS_RAWIO, all three points need updating
> 
> Would the plugin model be able to support declaring additional Linux capabilities that get wired into this chain? Without that, any feature requiring a new capability on the QEMU process would need core KubeVirt changes and couldn't be implemented purely as a plugin.

@mhenriks Yes, I'll take a step back. I needed some time to think about this... I think I agree with your approach now. 

Overall, my thinking was that if we do it in a plugin, we won't need to elevate the virt-launchers capabilities. 
However, we can solve the 1 and make 2 dynamic ( All we need to do, from my point of view, is to add CAP_RAWIO to the pod via `securityContext.capabilities.add` and the change the
```
cmd.SysProcAttr = &syscall.SysProcAttr{
        AmbientCaps: []uintptr{unix.CAP_NET_BIND_SERVICE},
    }
```
to infer the caps dynamically, either via /proc/self/caps or some env var )

But SCC is still the limiting factor here. :facepalm: 

I agree, with that said, plugin doesn't make sence and this IS a core API.

Sorry for pushing back too quickly.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,526 @@
...
+  experience for VMI and pod level modifications and validations.
+- Support node-level privileged operations in virt-handler via plugin hook points.
+- Provide a way of versioning the plugin and provide KubeVirt version constraints.
+- Provide a way to plan an upgrade path in advance.
+- Support multi-plugins: dependencies, conflict detection, etc.
+- Deprecate and eventually remove the `Sidecar` feature gate.
+
+## Non Goals
+
+- Allowing arbitrary, unvalidated XML modifications.
```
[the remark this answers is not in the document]
vladikr: I agree, and I think this is especially critical in light of the ongoing multi-hypervisor work

I respectfully disagree with the approach of "I'm a trusted cluster-admin, so let me shoot myself in the foot"
From my point of view, this weakens the value of the structured plugin mechanism and of the KubeVirt API in general.

When a VM is created via KubeVirt API, the users expect the same guarantees they get from any other Kubernetes API, such as validation, defaults, reconciliation, and consistent behavior, especially for core features like live-migration, but also hotplug, and others...
Yes, virt-launcher may be unprivileged, but the control plane is not. It makes operational promises based on the declared spec.

Letting plugins bypass that entirely turns KubeVirt into a best-effort launch mechanism rather than a reliable orchestration platform.
It effectively becomes some kind of a wrapper around "whatever the plugin feels like doing".
This breaks those operational promises. That's not an API anymore.

The whole idea of this structured plugin mechanism is to move away from the raw XML chaos that makes debugging and upgrading even more difficult.

In my view, we need this extensibility mechanism to be backend aware and safe by design if we want this to be a trustworthy foundation for complex features. 
The platform must have the final say on whether a configuration is valid and migratable.

If we don't enforce this now, we aren't just allowing users to shoot themselves in the foot, then we are basically trading the integrity of the entire platform for the sake of a freedom that eventually leads to unsupportable clusters."

This is the essence of another work that is taking place related to multi-hypervisors https://github.com/kubevirt/enhancements/pull/122

vladikr:
I am approving this VEP as a pragmatic and necessary first step. 
However, I want to be clear that I see this as the start of a much broader evolution toward a safe, structured extension model.

My vision for this feature remains a unified 'Plugin' CRD that allows users to extend KubeVirt in the right places, in a way that a cluster admin can actually understand and maintain. 
To get there, I envision this expanding into a framework that includes:

 - Structured DomainSpec modification which are backend-aware 
 - Beyond just the VM, we need to bring in node level hooks via virt-handler. That should be our very next step, and pod-level customization via Mutating Admission Policies.
 - There should be a viable upgrade path. Plugins should be well-defined 'features' with version constraints, not just scripts injected into the side.

I also want to address the discussion regarding whether we really need strict validation if virt-launcher is unprivileged.
I replied [here](https://github.com/kubevirt/enhancements/pull/191#discussion_r3097310396); but tl;dr is that I think we should avoid tth the approach of 'letting the user shoot themselves in the foot'. 
From my point of view, that approach weakens the entire KubeVirt API.
/approve

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Great question.

The advanced domain hook (sidecar-based) is designed to handle these cases. Domain hooks are re-invoked on the migration target during domain reconstruction, and the `DomainHookContext` struct that's passed to the hook will carry migration-specific metadata - e.g. whether this is during a migration, whether this is the source or target side, etc. This gives the hook author full context to adjust the domain based on target-pod specifics (like device paths) without needing a separate hook point.

The simpler hook types are currently just static transformations that apply the same way regardless of any other context.

That being said, this is a good point to keep in mind. This VEP will most probably have more alpha stages in which we will revisit the API multiple times, and we might indeed want to come up with more hook points for the simple domain hook ask well.

Another alternative to more hook points could be to use CEL-based conditions to determine what should be applied. With this approach we'll need to somehow inject the right variables from the DomainHookContext into the CEL context so the CEL code could reference it (still demands some research). Then the CEL code itself can express something like "if migration source, do X, else if migration target, do Y, otherwise do Z".

TLDR: I think that's a great thought. Let's start simple and revisit it in the next cycle. Does that work?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Great catch!
You're right, it should be isolated. Changed the path and added a note below.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: The hook server container image is not specified in the Plugin CR - it's the plugin author's responsibility to deploy the sidecar via a Mutating Admission Policy (MAP), which gives them full control over the container spec (image, resources, security context, etc.). The Plugin CR references the MAP but stays focused on the hook contract.

Regarding socket discovery - virt-launcher can auto-discover sockets by scanning the shared volume, same as today. The `socketPath` in the Plugin CR serves a different purpose: it maps each socket to a specific hook definition. This matters when a single plugin exposes multiple advanced hooks, each with its own socket - virt-launcher needs to know which socket corresponds to which hook.

It could also become important for plugin dependencies in the future - determining the order in which hooks should be called.

Makes sense?

iholder101:
Changes made (all very minor):

- Added the "Target release" metadata section added to the template [[1](https://github.com/kubevirt/enhancements/blob/fb63951038940ae364a1b837dd02445985baa403/veps/NNNN-vep-template/vep.md?plain=1#L5-L18)]: https://github.com/kubevirt/enhancements/compare/440032d10011ba51bad50a1938871b25530c8e70..b15b0b5083d71989ad19f784b5a845a1c3402598
- Rebased:
https://github.com/kubevirt/enhancements/compare/b15b0b5083d71989ad19f784b5a845a1c3402598..ba68510d02e2885c6776619e56092b22ab64c36f.
- Mounting isolated socket dirs to each plugin, see [this comment](https://github.com/kubevirt/enhancements/pull/191/changes#r3106574120):
https://github.com/kubevirt/enhancements/compare/ba68510d02e2885c6776619e56092b22ab64c36f..54407a1f57ce78a5d7b23e37f5beaa40ce262a78

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
bgartzi: Sure it works, I just wanted to bring/share the idea and confirm it could make sense. Thanks!

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
bgartzi: > The hook server container image is not specified in the Plugin CR - it's the plugin author's responsibility to deploy the sidecar via a Mutating Admission Policy (MAP), which gives them full control over the container spec (image, resources, security context, etc.). The Plugin CR references the MAP but stays focused on the hook contract.

I see, missed that part.

> 
> Regarding socket discovery - virt-launcher can auto-discover sockets by scanning the shared volume, same as today. The `socketPath` in the Plugin CR serves a different purpose: it maps each socket to a specific hook definition. This matters when a single plugin exposes multiple advanced hooks, each with its own socket - virt-launcher needs to know which socket corresponds to which hook.

I read the proposal again. I think I was missing the fact that we need a way to differentiate hooks (e.g. domain vs node) and their related sockets.

> 
> It could also become important for plugin dependencies in the future - determining the order in which hooks should be called.
> 
> Makes sense?

Yes, thank you!

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+<ins>Advanced Hooks</ins>:
+
+```yaml
+domainHooks:
+  - type: Plugin
+    socketPath: "/var/run/kubevirt-plugin/my-cool-plugin/my-plugin-socket.sock"
+```
+
+Note: The socket would need to reside under `/var/run/kubevirt-plugin/<my-plugin-name>`,
+which will be defined as a shared volume with the compute container.
```
Acedus, line 216: Not sure I understand, which compute container belonging to what pod in this context? If it's indeed shared, how would mount propagation settings look like?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non Goals](#non-goals)
+- [Definition of Users](#definition-of-users)
+- [User Stories](#user-stories)
+- [Repos](#repos)
+- [Design](#design)
+  - [The Domain Hook](#the-domain-hook)
+    - [Domain Hook modes](#domain-hook-modes)
+    - [Domain Hook API exmaples](#domain-hook-api-exmaples)
```
Acedus: s/exmaples/examples/g

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+
+### Versioning and Upgrade Path
+
+With the following fields the plugin system will enable to properly define version constraints and a planned upgrade path:
+- kubeVirtVersionCondition: a CEL based condition constraint for KubeVirt's version.
+- libvirtVersionCondition: a CEL based condition constraint for Libvirt's version.
+- upgradePaths: This field will be a list of values mapping a KubeVirt version to an image tag.
+  - For example: `KubeVirtVersion: 1.7 -> pluginImageTag: 1.7, KubeVirtVersion: 1.8 -> pluginImageTag: 1.8`, etc.
+  - In the above example, picture the v1.7 image existing and used in production, while the 1.8 image does not yet exist.
+However, when v1.8 is realsed, an image with a 1.8 tag would be added to a remote registry, allowing the Plugin owner
```
Acedus: s/realsed/released

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+    - name: "my-mutating-webhook-2"
+  validatingAdmissionWebhooks:
+    - name: "my-validating-webhook-1"
+```
+
+### Plugin metadata
+
+Each plugin would have to provide some basic metadata about itself.
+
+The main idea here goes beyond specifying a unique name and identify to the plugin,
```
Acedus: s/identify/identity

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+The plugin would have to explicitly mention which hook points it wishes to register for.
+
+#### Node hook API examples
+
+```yaml
+nodeHooks:
+  - mode: Plugin
+    socket: /var/run/my-node-socket.sock
+    permittedHooks:
+    - PreStartVm
```
Acedus: s/PreStartVm/PreVMStart

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+This approach gives the cluster-admin the flexibility to decide (via validating admission policies) what are the acceptable
+resources and capabilities that the plugin DaemonSet will be defined with.
+It also delegates the responsibility of managing resources to the plugin author.
+
+These hooks are invoked at specific points in virt-handler’s controllers.
+Node hooks cannot modify `DomainSpec` or XML directly but can influence runtime behavior.
+Plugins can implement both domain and node hooks for cohesive functionality.
+
+Similarly to the domain hooks, `virt-handler` communicates with the relevant socket at pre-defined hook points.
+It provides a `NodeHookContext` struct, and requests a function for virt-handler to execute.
```
Acedus: I believe that following our discussion, this should be changed considering the DS executes the logic itself. Can we have a better concrete definition of what data flows between the virt-handler and a plugin as well as what the plugin responses can be?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+
+## Implementation History
+
+<!--
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Graduation Requirements
```
Acedus, line 518: We should probably fill out graduation criteria for this, even though the VEP covers quite a big scope and is heavily subject to change.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+#### Node hook points
+
+- **PreVMStart**: Before VM launch (e.g., setup node devices/network; aligns with vmUpdateHelperDefault).
+- **PostVMStart**: After VM is running (e.g., verify node resources).
+- **PreMigrationSource**: Before migration from source node (e.g., prepare sockets; aligns with migrateVMI in migration-source.go).
+- **PostMigrationSource**: After migration completes on source (e.g., cleanup; aligns with handleSourceMigrationProxy).
+- **PreMigrationTarget**: Before migration to target node (e.g., setup target sockets; aligns with prepareMigrationTarget in migration-target.go).
+- **PostMigrationTarget**: After migration completes on target (e.g., repair/verify; aligns with finalizeMigration).
+- **OnVMStop**: When VM stops (e.g., cleanup node resources; aligns with processVmUpdate on shutdown).
+- **PostVMStop**: Called periodically.
```
Acedus, line 249: Could you elaborate on this? What does called periodically mean in that context? Run `PostVMStop` on all matching VMs every set interval?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,533 @@
...
+
+These are the basic fields every plugin would need to populate:
+- name:  a unique identifier for the plugin.
+- Version: Specifies the plugin’s version.
+- failureStrategy: Defines the behavior when a plugin fails. `Fail` (default) blocks the operation, `Ignore` logs the error and continues.
+- timeout: The maximum duration to wait for a plugin response before considering it failed (e.g. `30s`, `2m`). Defaults to `30s`.
+
+## API Examples (full version)
+
+```yaml
```
Acedus, line 347: I wonder if we can create further orthogonality and allow reuse by having a `DomainHook` CR, a `NodeHook` CR and a `Plugin` CR which can reference them to control the orchestration for their use. Just food for thought.

Acedus (commented):
Looking good, some more questions and a few nits.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: The "compute container" is the main container inside the `virt-launcher` pod - it's the one that runs the QEMU/libvirt process for the VM. In the advanced (plugin-based) hook mode, a sidecar container is injected into the same `virt-launcher` pod alongside the compute container.

The shared volume would be a simple `emptyDir` volume mounted into both containers - no special mount propagation is needed. For example:
- The `emptyDir` volume is mounted at `/var/run/kubevirt-plugin/<plugin-name>` in both the compute container and the plugin sidecar.
- The plugin sidecar creates and listens on a Unix socket within that path.
- The compute container (virt-launcher) connects to that socket to invoke the domain hook.

Since both containers share the same pod and the volume is a standard `emptyDir`, default mount propagation (`None`) is sufficient.

Does that make sense?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: > I believe that following our discussion, this should be changed considering the DS executes the logic itself.

Nice catch! I've changed the text to align with the DS execution model.

> Can we have a better concrete definition of what data flows between the virt-handler and a plugin as well as what the plugin responses can be?

I didn't want to delve into the exact fields in the "context" structs and leave it for the implementation PR to focus on the main design. Does it make sense to discuss the exact fields in there?

Generally though, due to the conversation in this PR, I tend to start slowly and expand in the next alpha rounds. To keep this thin, I guess a simple "succeeded" or "failed" answer should be provided, just to let the virt-handler know if it should skip it or re-queue it (according to the `failureStrategy`).

Is there anything specific you'd like to discuss regarding that?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Yes, exactly.
Similar to `PostVMStart`, but after it's stopped.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
Acedus: Fine by me, we can discuss it in the impl PR.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: That's an interesting idea!

However, people have pushed back even for the single Plugin CR :) I tend to keep this thin for the first round.

I'll add it to the `Possible future enhancements` sections so we can revisit it on next rounds!

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Nice catch! Thanks!

iholder101:
Thanks for the review @Acedus!
I think all is addressed :) PTAL

Acedus (approved):
/lgtm
/hold

Giving other reviewers a chance to lgtm as well.

iholder101:
> /lgtm /hold
> 
> Giving other reviewers a chance to lgtm as well.

Thanks a lot @Acedus! Your reviews were especially fruitful, appreciate it!

Let's keep this held until Monday. I'll unhold by then if there are no further comments.

iholder101:
/hold cancel

Thanks for everyone involved!

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,550 @@
...
+  matchLabels:
+    use-plugin: my-plugin
+  matchFields:
+    - key: metadata.name
+      operator: Prefix
+      value: vm-
+  conditions: # CEL-based
+    - "vmi.status.phase == 'Running'"
+    - "vmi.status.conditions.exists(c, c.type == 'Ready' && c.status == 'True')"
+    - "vmi.status.conditions.exists(c, c.type == 'LiveMigratable' && c.status == 'True')"
```
enp0s3, line 272: @iholder101 What will happen when the hook point was reached but the conditions weren't met?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,550 @@
...
+Moving forward we can support more hook points if needed.
+
+#### Domain Hook API examples
+
+<ins>Simple Hooks</ins>:
+
+JsonPatch based simple hook:
+```yaml
+domainHooks:
+  - type: JsonPatch
```
enp0s3, line 182: @iholder101 Why would we need to propose both CEL and JSONPatch? Why not using only one of them?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -0,0 +1,550 @@
...
+
+The proposed mechanism focuses on modifying the `DomainSpec`, a Go struct generated from the VirtualMachineInstance
+(VMI) spec, which serves as the internal representation of a VM's desired configuration before conversion to libvirt XML.
+Hooks will be executed after the initial `DomainSpec` generation but before the final XML is created.
+
+The current KubeVirt `DomainSpec` struct only includes a subset of the fields available in the full libvirt `DomainSpec`.
+To allow the plugin complete control over the generated XML (similar to the legacy Sidecar Hook),
+it will be replaced it with the complete, fully-expanded libvirt DomainSpec struct.
+
+#### Domain Hook modes
```
enp0s3, line 157: @iholder101 The domain that the hook is operating on, is it the full libvirt DomainSpec as provided by libvirt-go, or is it the KubeVirt API DomainSpec?

enp0s3 (commented):
@iholder101 Hi, thanks for the valuable VEP! I have some questions below. Besides that, are there any intentions to create a plugin controller? For example if I have a plugin that consists of a MAP and a VAP, what will happen when I refer the Plugin to a non existing ones (I might had a typo in the Plugin spec when referring the the MAP/VAP I've created)

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: The hook point will be considered as disabled, i.e. it will be skipped

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: You're right!

In the implementation PR https://github.com/kubevirt/kubevirt/pull/17790 I'm actually dropping JSON Patch. Will make a VEP update to reflect this.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: > is it the full libvirt DomainSpec as provided by libvirt-go

Yes indeed. This way the plugin author can make use of domain XML features KubeVirt doesn't officially supports. In https://github.com/kubevirt/kubevirt/pull/17897 there's a logic to convert our limited domain XML struct to the full libvirt one and vice-versa.

iholder101:
> For example if I have a plugin that consists of a MAP and a VAP, what will happen when I refer the Plugin to a non existing ones (I might had a typo in the Plugin spec when referring the the MAP/VAP I've created)

This will cause the plugin to become "not ready", which means it will currently be skipped or failed (according to the failure policy).

enp0s3:
> > For example if I have a plugin that consists of a MAP and a VAP, what will happen when I refer the Plugin to a non existing ones (I might had a typo in the Plugin spec when referring the the MAP/VAP I've created)
> 
> This will cause the plugin to become "not ready", which means it will currently be skipped or failed (according to the failure policy).

@iholder101 Thanks for the quick reply!
The "not ready" condition will be written in the `Plugin` CR status by the virt-handler plugin controller?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
enp0s3: @iholder101 Do we have this behavior defined in the VEP?

iholder101: Thanks for raising this into attention. I'll ensure to add this as part of the VEP update I'll create!

iholder101:
> > > For example if I have a plugin that consists of a MAP and a VAP, what will happen when I refer the Plugin to a non existing ones (I might had a typo in the Plugin spec when referring the the MAP/VAP I've created)
> > 
> > 
> > This will cause the plugin to become "not ready", which means it will currently be skipped or failed (according to the failure policy).
> 
> @iholder101 Thanks for the quick reply! The "not ready" condition will be written in the `Plugin` CR status by the virt-handler plugin controller?

I think this sounds good, although I'm not sure yet exactly what'll be included in this cycle. But thanks, will definitely keep this in mind!

Issue #192, opened 2026-01-30
VEP 191: Decentralized Live Migration LM network proxy support


awels:
**Primary contact (assignee)**:

/assign @awels 

**Current Feature Stage**:

New

**Feature Gate**:

DecentralizedLiveMigration

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig core
/sig network

**Enhancement link**:

**Timeline**:

* Alpha release target (1.9):
* Beta release target (1.10?):
* Stable release target (1.11?):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

During testing of Decentralized Live Migration we encountered environments we had not considered before which require some additional functionality for decentralized live migration. In particular the follow environment:
* There are more than 2 clusters lets say N clusters that are all connected through the same LM network.
* The amount of available IPs on this network is limited, there is no DHCP and it is hard to get blocks of ip addresses for the virt-handlers.
* Each cluster needs to be configured so it can migrate VMs from any cluster to any other cluster, so all clusters need to be able to communicate with each other.

The current setup of using the LM as the Decentralized Live Migration network is problematic here since we need a large number of IP addresses on this network. Namely for each cluster we need X + Y IP addresses where X is the number of virt-handlers, and Y is the number of synchronization controllers. And we need that many PER cluster. So it can become unwieldy quickly.

> [!IMPORTANT]
> Please keep this description up to date.

xpivarc:
No VEP linked

awels:
Added VEP link

vladikr:
@awels can you please update the status here?

awels:
> [@awels](https://github.com/awels) can you please update the status here?

I am not entirely sure what I am missing so I can update it. Basically the point of this is to update the original VEP to include this new datapath field so we can control how the proxy operates. In particular this is important if we want to use the pod network for the proxy -> proxy traffic (allowing us to use a service and Ingress/Route). Are you talking about the tracking drop down to the right?

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

[...]

PR #198, opened 2026-02-03
Cross cluster live migration proxy design proposal

awels:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/192
**SIG label**: /sig storage

### What this PR does
This VEP introduces a proxy as part of the synchronization controller in order to reduce the number of needed IP addresses for cross cluster live migration. It also introduces a mechanism that allows one to specify nodeSelectors for the synchronization controllers to give better control over where the synchronization controller runs.

### DIFF
--- a/veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
+++ b/veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
@@ -1,0 +1,153 @@
+# VEP #191: Decentralized Live Migration LM network proxy support 
+
+## VEP Status Metadata
+
+### Target releases
+
+<!--
+A PR must update this section during the planning phase of a given release in order to track it.
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
+For example, during the planning phase for version v1.123, do **not** target beta for v.124 in advance.
+-->
+
+- This VEP targets alpha for version: 1.9
+- This VEP targets beta for version: 1.10
+- This VEP targets GA for version: 1.11
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add a proxy to the synchronization controllers that forwards cross-cluster live migration traffic onto the cross-cluster live migration network. Doing so reduces the number of IP addresses required for cross-cluster live migration.
+
+## Motivation
+
+During testing of Decentralized Live Migration we encountered environments we had not considered before, which require additional functionality. In particular:
+
+* **Multi-cluster, single LM network:** More than two clusters are connected through the same live migration (LM) network.
+* **Constrained addressing:** The number of available IP addresses on this network is limited; there is no DHCP, and obtaining address blocks for virt-handlers is difficult.
+* **Full mesh:** Each cluster must be able to migrate VMs to and from any other cluster, so every cluster must be reachable over the LM network.
+
+Using the LM network directly for Decentralized Live Migration is problematic in this setting because it requires a large number of IP addresses. For each cluster we need *(virt-handlers + synchronization controllers)* addresses—i.e., one per virt-handler and one per synchronization controller—and that cost is incurred in every cluster. For example, with 10 clusters of 10 worker nodes (10 virt-handlers each) and 2 synchronization controllers per cluster, we need 10 × (10 + 2) = **120 IP addresses** on the cross-cluster live migration network. Without DHCP, assigning and maintaining these addresses is unwieldy.
+
+A further issue is operational: when nodes are added or removed, any mechanism that assigns IP addresses to virt-handlers must also update addressing on the cross-cluster live migration network, increasing configuration churn and failure modes.
+
+## Goals
+
+* **Reduced addressing:** With the proxy in place, the cross-cluster live migration network needs only one IP per synchronization controller. Total addresses = (synchronization controllers per cluster) × (number of clusters). Example: 2 controllers per cluster × 10 clusters = 20 addresses.
+* **Dedicated cross-cluster network:** Support configuration of a cross-cluster live migration network that is separate from the in-cluster live migration network.
+* **Placement control:** Allow admins to choose whether synchronization controller pods run on control-plane or worker nodes. Control-plane nodes often lack access to the cross-cluster live migration network; worker nodes typically have it. Support an optional node selector so admins can target the (possibly small) set of nodes that have cross-cluster live migration network access.
+
+## Non Goals
+
+* Allow proxying of any traffic besides cross cluster live migrations.
+* Proxying cross namespace migrations. Those can go over the internal migration network.
+
+## Definition of Users
+
+* Cluster Admins
+* Admins managing several clusters
+
+## User Stories
+
+* As a KubeVirt admin managing multiple clusters on a network with limited IP addresses, I want to perform cross-cluster live migration without building and maintaining a complex network that requires many addresses.
+* As a KubeVirt admin, I want to configure a dedicated cross-cluster live migration network that is separate from the in-cluster live migration network.
+* As a KubeVirt admin, I want to choose whether synchronization controller pods run on control-plane nodes or worker nodes.
+* As a KubeVirt admin, I want to use a node selector to place synchronization controllers on specific nodes—for example, the subset of nodes that have access to the cross-cluster live migration network.
+
+## Repos
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Current state migration network
+The current live migration network design poses a problem when there are many clusters connected through a single live migration network. Especially if there are constraints on that network in how it is configured. In particular if we need to manually assing IP addresses to virt-handlers and synchronization controllers. And there is no DHCP availabe on the network. It is technically possible to design a `NetworkAttachmentDefinition` for each cluster that gives a range of IP addresses to a certain cluster, but it becomes a lot harder if the pool of available IP addresses is not a range. Also the number of IP addresses can become problematic if the number of virt-handlers per cluster is large enough. The current state is illustrated in figure 1:
+
+![figure 1](l2_4cluster.png)
+
+
+### Proposed state migration network
+The proposed proxy will allow the same connectivity on an L2 network, but instead of every virt-handler being able to communicate with every other virt-handler directly, it will go through a proxy. The proxy will communicate with a proxy on the other cluster. This reduces the number of IP addresses needed from the number of virt-handlers + the number of synchronization controllers times the number of clusters to just the number of synchronization controllers times the number clusters. The proposed state is illustrated in figure 2:
+
+![figure 2](proxy_cluster.png)
+
+As you can see from the figure, the synchronization controller has two network interfaces. One connected to the 'in cluster' network (10.244.0.0/16), and one connected to the shared L2 network (172.16.0.0/16). The shared network is marked L2 in the figure, but it can also be an L3 network, as long as synchronization controllers can connect to each other it will work.
+
+In order to define a specific cross cluster live migration network, there will need to be a new API field in the migration struct of the `KubeVirt` CR. This will be an optional additional field called `crossClusterNetwork`. It will be the name of a `NetworkAttachmentDefinition` just like the live migration network is now.
+
+### Current state synchronization controller placement
+Currently the synchronization controllers are marked as control plane pods, and are thus scheduled on the control plane nodes. This works if the control plane nodes have access to the live migration network. This is not guaranteed to be the case.
+
+### Proposed state synchronization controller placement
+
+## API Examples
+
+### Cross cluster live migration network
+Add an additional field to `kubevirt.spec.configuration.migrations` called `crossClusterNetwork`. If enabled this causes the synchronization controller pods to be started with an additional network besides whatever the live migration network is.
+
+Additional cross cluster migration network
+```yaml
+...
+migrations
+  ...
+  crossClusterNetwork: crosscluster-network
+  ...
+```
+
+### Specify node placement for synchronization controllers
+
+If only certain nodes have access to the cross cluster live migration network, and those nodes are not control plane nodes it will be necessary to specify the `nodePlacement` for the synchronization controllers. In order to specify the node placement for synchronization controller an additional field `kubevirt.spec.synchronizationPlacement` will be added. This will be a normal nodePlacement just like the ones defined in `kubevirt.spec.workloads.nodePlacement`.
+
+Additional node placement
+```yaml
+...
+spec
+  synchronizationPlacement
+    nodeSelector:
+    - label: selector
+...
+```
+
+## Alternatives
+
+Instead of the proxy approach, we considered a mechanism that would allow specifying a list of IP addresses assignable to virt-handlers. This could work in principle, but it would not remove the scaling problem: whenever new nodes are added, the list would need to be updated to include IPs for those nodes. Moreover, such a mechanism would have to be implemented at the CNI level, and feasibility would depend on the CNI in use.
+
+## Scalability
+
+One of the issues with the current implementation is scalability: as more virt-handlers join the L2/L3 network, the number of required IP addresses grows linearly. The new design addresses this by requiring only as many IPs as there are synchronization controllers (2 by default). The trade-off is higher network utilization. Migration traffic is carried over three segments in sequence—the source cluster migration network, the cross-cluster live migration network, and the target cluster migration network. This the same data traverses each segment, increasing total traffic on the migration path.
+
+## Update/Rollback Compatibility
+
+**Update:** No impact. This is net-new functionality and is only active when the `crossClusterNetwork` field is set.
+
+**Rollback:** If a cluster is rolled back to an older version while `crossClusterNetwork` is set, that field is ignored because older versions do not define it. No special handling is required. 
+
+## Functional Testing Approach
+
+Testing real cross-cluster live migration is difficult with the current kubevirtci setup, so we will simulate it using cross-namespace live migration. We will define a dedicated cross-cluster live migration network (separate from the default live migration network), then run a cross-namespace live migration over it. These tests will extend the existing decentralized live migration namespace functional tests.
+
+## Implementation History
+
+
+
+## Graduation Requirements
+
+### Alpha
+It will be hidden behind a feature gate called `CrossClusterMigrationProxy`. If the feature gate is not enabled the `crossClusterLiveMigration` network field will be rejected by the webhook. This will allow users to opt-in or out at will without affecting any decentralized live migrations.
+
+### Beta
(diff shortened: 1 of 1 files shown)


vladikr:
@awels did you manage to find any dedicated reviewers for this work?

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
jean-edouard, line 1: I don't understand... I see as many `10.244` IPs on this diagram as there are on the other one.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Graduation Requirements
+
+### Alpha
+Tied to decentralized live migration. It will not be hidden behind a feature gate since, it is only enabled if the `crossClusterLiveMigration` network is defined. This will allow users to opt-in or out at will without affecting any decentralized live migrations.
+
+### Beta
+Tied to decentralized live migration, as it is an extension of decentralized live migration
```
jean-edouard: Makes sense, but it does mean that for Decentralized Migration to graduate to Beta, Migration Proxy also needs to qualify.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+
+![figure 1](l2_4cluster.png)
+
+
+### Proposed state migration network
+The proposed proxy will allow the same connectivity on an L2 network, but instead of every virt-handler being able to communicate with every other virt-handler directly. It will go through a proxy. The proxy will communicate with a proxy on the other cluster. This reduces the number of IP addresses needed from the number of virt-handlers + the number of synchronization controllers times the number of clusters to just the number of synchronization controllers times the number clusters. The proposed state is illustrated in figure 2:
```
maiqueb:
```suggestion
The proposed proxy will allow the same connectivity on an L2 network, but instead of every virt-handler being able to communicate with every other virt-handler directly, it will go through a proxy. The proxy will communicate with a proxy on the other cluster. This reduces the number of IP addresses needed from the number of virt-handlers + the number of synchronization controllers times the number of clusters to just the number of synchronization controllers times the number clusters. The proposed state is illustrated in figure 2:
```

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Cross cluster live migration network
+Add an additional field to `kubevirt.spec.configuration.migrations` called `crossClusterNetwork`. If enabled this causes the synchronization controller pods to be started with an additional network besides whatever the live migration network is.
+
+Additional cross cluster migration network
+```yaml
+...
+migrations
+  ...
+  crossClusterNetwork: crosscluster-network
```
maiqueb, line 104: maybe the example could refer a namespace

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+...
+```
+
+## Alternatives
+
+Instead of the proxy approach, we considered a mechanism that would allow specifying a list of IP addresses assignable to virt-handlers. This could work in principle, but it would not remove the scaling problem: whenever new nodes are added, the list would need to be updated to include IPs for those nodes. Moreover, such a mechanism would have to be implemented at the CNI level, and feasibility would depend on the CNI in use.
+
+## Scalability
+
+One of the issues with the current implementation is scalability: as more virt-handlers join the L2/L3 network, the number of required IP addresses grows linearly. The new design addresses this by requiring only as many IPs as there are synchronization controllers (2 by default). The trade-off is higher network utilization. Migration traffic is carried over three segments in sequence—the source cluster migration network, the cross-cluster live migration network, and the target cluster migration network—so the same data traverses each segment, increasing total traffic on the migration path.
```
maiqueb:
```suggestion
One of the issues with the current implementation is scalability: as more virt-handlers join the L2/L3 network, the number of required IP addresses grows linearly. The new design addresses this by requiring only as many IPs as there are synchronization controllers (2 by default). The trade-off is higher network utilization. Migration traffic is carried over three segments in sequence: the source cluster migration network, the cross-cluster live migration network, and the target cluster migration network. Thus, the same data traverses each segment, increasing total traffic on the migration path.
```

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+
+Testing real cross-cluster live migration is difficult with the current kubevirtci setup, so we will simulate it using cross-namespace live migration. We will define a dedicated cross-cluster live migration network (separate from the default live migration network), then run a cross-namespace live migration over it. These tests will extend the existing decentralized live migration namespace functional tests.
+
+## Implementation History
+
+<!--
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
```
maiqueb: we can delete this, right ?

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,139 @@
...
+
+![figure 1](l2_4cluster.png)
+
+
+### Proposed state migration network
+The proposed proxy will allow the same connectivity on an L2 network, but instead of every virt-handler being able to communicate with every other virt-handler directly. It will go through a proxy. The proxy will communicate with a proxy on the other cluster. This reduces the number of IP addresses needed from the number of virt-handlers + the number of synchronization controllers times the number of clusters to just the number of synchronization controllers times the number clusters. The proposed state is illustrated in figure 2:
+
+![figure 2](proxy_cluster.png)
+
+As you can see from the figure, the synchronization controller has two network interfaces. One connected to the 'in cluster' network, and one connected to the shared L2 network. The network is marked L2 in the figure, but it can also be an L3 network, as long as synchronization controllers can connect to each other it will work.
```
maiqueb: maybe explain here a little bit about the IPs in each network, so we can make better sense of the img added above.

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
maiqueb: yes, but - afaiu - the IPs used for cross cluster live migration are within the `172.16.X.Y` range.

Still, they all the same IP - `172.16.X.100`. Is this expected, @awels ? 

Plus, the sync controller in cluster C has the IP `10.244.4.20`, which should be in cluster D subnet. I think you need to correct the diagram.

maiqueb (changes requested):
Very well put.

I think you just need to correct the second img, and clarify those IPs. Everything else looks good to me.

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
awels: The AI that generated the image probably didn't do it right. Let me see if I can manually correct it.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: I don't think so, the proxy would be an additional feature gate IMO. Similar to how hotplug volume and declarative hotplug volume works.

jean-edouard: Doesn't that go against your `It will not be hidden behind a feature gate` statement above?

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
jean-edouard: Please follow the [KubeVirt AI Contribution Policy](https://github.com/kubevirt/community/blob/main/ai-contribution-policy.md)

awels: Well I can't draw worth a damn, so I had the AI draw the image. Does that require the AI contribution policy thing? I described what I wanted and it generated an image for it (poorly I might add)

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Good point, I had forgotten I wrote that. That makes it kind of hard to decouple the two. We definitely want to move decentralized live migration to beta. This is just an extension. But the extension should be alpha. Not sure how to do that.

jean-edouard: You can just have a new feature gate of the alpha feature. The second will depend on the first but that's fine.

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
jean-edouard: Just follow the policy, not sure if it has anything specific for your scenario. Either way it's usually just about adding a line to commit messages / PRs, nothing crazy.

awels: Now I have to remember what I used to generate it...

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Okay I will remove the verbage about the feature gate

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: The cross cluster network will be in the namespace kubevirt is in, no other namespace will work.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Done

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Done

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Done

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
awels: No the diagram is correct, I am illustrating that the internal cluster networks can be identical. They don't have to be. I should probably expand that a little bit. For the 172.16.x.100. I should probably replace X with 1/2/3/4 for each synchronization controller like the other diagram does.

jean-edouard: Then I don't understand this proposal... Which is fine because I trust the opinion of sig-network for all the network-specific bits.
However, it would be nice to add a simplified explanation so network-illiterate people like me can get the general idea.

awels: Okay I see that for cluster C I had the wrong number, I fixed it, and also fixed the X.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Okay I added the CIDR for the various networks. Hopefully that makes it clearer what I mean.

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
jean-edouard: > Then I don't understand this proposal... Which is fine because I trust the opinion of sig-network for all the network-specific bits. However, it would be nice to add a simplified explanation so network-illiterate people like me can get the general idea.

Now, after some offline chats, I think I understand this proposal. I summarized it below.
Feel free to use any parts of it, but don't feel pressured to do so, I'm just sharing my own understanding with others and my future self.
More importantly, please correct me if I'm wrong!!
Here it is:

This proposal adds support for a migration network dedicated to cross-cluster live migrations (CCLMs).

CCLMs currently leverage the existing dedicated migration network feature, originally intended for same-cluster migrations.
The downside of the migration network feature is that it causes every virt-handler to connect to it, and therefore get an IP on it.

With the new network option introduced by this proposal, a network dedicated to cross-cluster migrations can be configured.
Only synchronization-controller pods will connect to this new CCLM migration network, not virt-handler pods, therefore requiring only 2 IPs per cluster (or whatever the number of synchronization-controller replicas is set to).
Since migrations are still virt-launcher to virt-launcher (proxied by their local virt-handler pods), a new proxy will be introduced between the synchronization-controller and the virt-handler pods on their respective clusters.

We will keep using the original dedicated migration network when no CCLM network is defined.

Here's the path of a migration without a CCLM network (current/fallback design), from `cluster 1 node 1` to `cluster 2 node 3`:
virt-launcher c1n1 - **virt-handler c1n1 - dedicated migration network - virt-handler c2n3** - virt-launcher c2n3

With this new feature, migrations will follow this path:
virt-launcher c1n1 - virt-handler c1n1 - **synchronization-controller c1 - CCLM migration network - synchronization-controller c2** -  virt-handler c2n3 - virt-launcher c2n3

I attempted a couple rough draft diagrams (red lines mean proxy). They're not all that different from yours, they just speak to me more than your IP-based ones. Happy to share the document if you'd like to finish and use them.

mignet
cclmnet

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Update:** No impact. This is net-new functionality and is only active when the `crossClusterNetwork` field is set.
+
+**Rollback:** If a cluster is rolled back to an older version while `crossClusterNetwork` is set, that field is ignored because older versions do not define it. No special handling is required. 
+
+## Functional Testing Approach
+
+Testing real cross-cluster live migration is difficult with the current kubevirtci setup, so we will simulate it using cross-namespace live migration. We will define a dedicated cross-cluster live migration network (separate from the default live migration network), then run a cross-namespace live migration over it. These tests will extend the existing decentralized live migration namespace functional tests.
+
+## Implementation History
```
maiqueb, line 140: do we need this empty section ?

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
maiqueb: OK, that's clean. Nice; tks.

maiqueb (approved):
Nice. I think this is a nice improvement to the current flow.

Tks.

vladikr:
@jean-edouard do you think this is ready?

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
awels: Once I get the PRs up, I will create an update to link to the implementation PRs. I have most of the code done at this point, just running through a variety of testing scenarios before posting it.

jean-edouard:
> @jean-edouard do you think this is ready?

I'm still unclear on the feature-gate situation. Every graduations step is marked as `Tied to decentralized live migration`, but from [discussions above](https://github.com/kubevirt/enhancements/pull/198#discussion_r3040240642) I understood that we wanted this feature to have its own gate, so it could graduate later than the main decentralized LM feature.
If so, this document needs to talk about the new feature gate.

On a more subjective note, I still find the document and diagrams quite hard to understand (also [discussed above](https://github.com/kubevirt/enhancements/pull/198#discussion_r3040232424)). But I don't see any other comment to that effect so maybe it's just me, in which case great!

veps/sig-compute/191-decentralized-migration-proxy/proxy_cluster.png
awels: So I like yours as well, however yours doesn't talk about the difference between a dedicated internal migration network and a pod network. I guess from the flow perspective there is no difference. But from a practical perspective, we need to ensure that the synchronization controller is connected to both the CCLM network, and whatever network we are using for the migration network inside a cluster. I guess me using some concrete examples for ip addresses in the cluster are confusing if you don't think those terms.

Also one of the main points I wanted to make is that this works for many clusters, so my example had more than 2 clusters to show case the reduction in usage of IP addresses in the CCLM network.

jean-edouard: Yeah the above is largely incomplete, it was just me trying to get an idea of what this proposal looks like, bounce back my understanding of it to you to make sure I got it right, and as a bonus maybe help others understand it too. Also I now have a non-AI template that I'm happy to share in case you'd like to use it for additional diagrams.

And yes, I'm sure network details help people understand this, but they do have the opposite effect on me :sweat_smile: 

Anyway, all is good, unless the above tells you I misunderstood things, and as long as others find the current document easy enough to understand.

awels: It is just a different way of viewing the same thing, prioritizing a different view. But it is showing the exact same thing. So your understanding is correct. The question is how would I incorporate it into the document without confusing people. I would like everyone to understand what I am saying

awels:
@jean-edouard I updated the document to make the feature gates more explicit, take another look please? And if you happy give lgtm.

veps/sig-compute/191-decentralized-migration-proxy/migration-proxy.md
```diff
@@ -0,0 +1,135 @@
+# VEP #191: Decentralized Live Migration LM network proxy support 
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
```
iholder101: Can you rebase and use the new section? https://github.com/kubevirt/enhancements/blob/main/veps/NNNN-vep-template/vep.md#target-releases

awels: Done

jean-edouard (approved):

vladikr:
/approve
Thank you!

enp0s3 (commented):
@awels Hi. Can you please elaborate more about the limitations of infrastructure providers to configure additional host-level L2 overlay network and assign a custom multi-cluster IPAM to this overlay?

awels:
The biggest problem presented to me, which prompted the proxy proposal is the following layout:
1. I have many clusters 10+ on the same L2 network.
2. I do NOT have HDCP on this L2 network.
3. I do NOT have contiguous blocks of IP addresses available on this network. Basically all the IP addresses are statically assigned.
4. There is no CNI that allows one to dynamically assign static IP addresses to a deployment/daemonset.

Give the restrictions above. I can use something like whereabouts and exclude everything I don't want to get the IP address I am looking for. However this exclude list will be very long and very hard to maintain for a large number of IP addresses. The proxy as explained in the VEP reduces the number of required IP addresses on this L2 network to 2 per cluster, instead of 2 + number of virt-handlers per cluster.

This makes the management of the IP addresses much more manageable. All of this was explained in the VEP.

enp0s3:
@awels Thank you for your patience. I did understand these limitations and the proposed solution. It seems like I can't properly express my concern. I wonder why couldn't they make the effort to adopt their environment to our need. Instead of solving the IP shortage and the network functionality on their side, it is required from us to maintain more code in order to support environments with limited resources. 

I hope our project would benefit from being compatible with their limits, in the end we are the ones that need to maintain the code.

Issue #200, opened 2026-02-10
VEP 200: virt-handler pools


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood 

**Current Feature Stage**:

New

**Feature Gate**:

`VirtHandlerPools`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig control-plane

**Enhancement link**:

**Timeline**:

* Alpha release target `v1.9.0`
* Beta release target `v1.10.0`
* Stable release target `1.11.0`

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

lyarwood:
PoC implementation is available as a draft PR: https://github.com/kubevirt/kubevirt/pull/17597

The implementation covers:

- **API types**: `WorkerPoolConfig`, `WorkerPoolSelector`, `WorkerPoolVMLabels` added to `KubeVirtSpec.WorkerPools`, gated behind `WorkerPools` feature gate (Alpha)
- **Pool matcher**: VMI-to-pool matching by device names and/or VMI labels with OR semantics and first-match-wins ordering
- **virt-operator**: Creates additional virt-handler DaemonSets per pool, applies anti-affinity to the primary virt-handler, blocks pool removal while matched VMIs are running
- **virt-controller**: Selects pool-specific launcher image, merges pool nodeSelector into pod affinity, annotates matched pods with pool name, detects outdated VMIs on pool config changes
- **Validation webhook**: Rejects invalid pool configs, blocks pool removal with running matched VMIs, warns on overlapping selectors/nodeSelectors
- **Functional tests**: Webhook validation, DaemonSet lifecycle, anti-affinity, label-based VMI matching with pod annotation, pool removal blocking

Issue #97, opened 2025-10-08
VEP 97: Introduce abstraction to enable alternative hypervisors


vyta:
**Primary contact (assignee)**:

/assign @harshitgupta1337 

**Current Feature Stage**: New

**Feature Gate**: ConfigurableHypervisor

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig $SIG

**Enhancement link**: 
https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/hypervisor-abstraction.md

**Timeline**:

* Alpha release target (x.y): 1.8
* Beta release target (x.y): 1.9
* Stable release target (x.y): 1.10

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/98
  * [ ] Code PRs: 
      * https://github.com/kubevirt/kubevirt/pull/16326
      * https://github.com/kubevirt/kubevirt/pull/16635
      * https://github.com/kubevirt/kubevirt/pull/16818
      * https://github.com/kubevirt/kubevirt/pull/16836
      * https://github.com/kubevirt/kubevirt/pull/16905
      * https://github.com/kubevirt/kubevirt/pull/16927
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

lyarwood:
@harshitgupta1337 Hey, the VEP currently lists the following as something you wanted to look into for Beta:

> KubeVirt should support multiple hypervisors in the same cluster. A VMI-level hypervisor field should be introduced to let the user choose which hypervisor to use for creating the VMI.

Did you have any plans to address this in v1.9.0?

I'm actually looking to tackle the deployment aspect of this in #200 and wouldn't mind extending this or introducing a follow up VEP specifically for the multi-hypervisor use case.

I'm planning on talking about this more next Tuesday at the v1.9.0 unconference, if you could attend it would be great to talk about this more then!

PR #202, opened 2026-02-10
VEP #200: Worker Pools

lyarwood:
**Tracking issue**: https://github.com/kubevirt/enhancements/issues/200

## Summary

This VEP proposes adding support for deploying multiple virt-handler DaemonSets to serve heterogeneous node pools with different virt-handler and virt-launcher container images.

### Key Design Points

- **Standalone WorkerPool CRD**: Pool configuration is defined via a cluster-scoped `WorkerPool` CRD in `pools.kubevirt.io/v1alpha1`, following the `MigrationPolicy` pattern. This avoids polluting the KubeVirt CR and enables RBAC scoping for pool management.
- **Transparent to VM users**: Pool selection is entirely admin-configured. VM users create VMIs as normal — requesting devices, setting labels, etc. — and the correct virt-launcher image is selected automatically.
- **Device and label selectors**: Administrators configure pools with `selector.deviceNames` (matching VMI GPU/HostDevice device names) and/or `selector.vmLabels` (matching VMI labels), evaluated with OR semantics.
- **Automatic node placement**: When a VMI matches a pool, the pool's `nodeSelector` is merged into the virt-launcher pod's required node affinity, ensuring it lands on a node served by that pool's virt-handler.
- **Custom handler and launcher images**: Each pool can override virt-handler and/or virt-launcher images for its nodes. At least one image override is required.
- **Safe pool deletion**: Deleting a `WorkerPool` CR is blocked via a finalizer while matched VMIs are still running. Administrators must drain impacted nodes before the pool can be deleted.

### Relationship to Other Work

- Per-pool hypervisor backend configuration ([VEP #97](https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/hypervisor-abstraction.md)) via an optional `hypervisor` field in `WorkerPoolSpec` is targeted for Beta, enabling mixed-hypervisor clusters (e.g., KVM on some nodes, MSHV on others).

### Alternatives Considered

- Embedded in KubeVirt CR (pollutes CR, no RBAC scoping, coupled lifecycle)
- Namespace-based separation (operational overhead)
- Per-VMI image override annotations (security concerns)
- External webhook-based image mutation (fragile, no handler customization, not integrated into KubeVirt lifecycle)
- KubeVirt Structured Plugins ([VEP #190](https://github.com/kubevirt/enhancements/pull/191)) — MAPs cannot perform the cross-object VMI lookup required for image selection, so a full webhook would still be needed; the Plugin CR only adds indirection
- nodeSelector-only VMI matching (leaks infrastructure details to users)

### Presentation

- [Unconference slides (v1.9.0)](https://lyarwood.gitlab.io/presentations/kubevirt/unconference/v1.9.0/virt-handler-pools/presentation.html)

### DIFF
--- a/veps/sig-compute/200-worker-pools/worker-pools.md
+++ b/veps/sig-compute/200-worker-pools/worker-pools.md
@@ -1,0 +1,621 @@
+# VEP #200: Worker Pools
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes adding support for deploying multiple virt-handler DaemonSets
+to serve heterogeneous node pools with different virt-handler and virt-launcher
+container images. Pool selection is transparent to VMI users — administrators
+configure pools with label and device selectors, and KubeVirt automatically
+matches VMIs to the appropriate pool based on the VMI's existing labels and
+device requests.
+
+## Motivation
+
+Many organizations operate heterogeneous Kubernetes clusters with specialized
+node pools for different workloads. Examples include:
+
+- **GPU nodes**: Require virt-launcher images with GPU drivers and libraries
+- **FPGA nodes**: Need specialized images with FPGA support libraries
+- **Secure enclaves**: Run hardened images with additional security configurations
+
+Currently, KubeVirt deploys a single virt-handler DaemonSet that runs the same
+virt-handler and virt-launcher images across all nodes. This forces operators to
+either:
+
+1. Run separate KubeVirt installations for each node pool (operational overhead)
+2. Build a single image containing all specialized components (image bloat)
+3. Use external webhooks to mutate VMI pods (fragile, external dependency)
+
+This VEP proposes folding launcher and handler image selection into KubeVirt
+natively, removing the need for external webhooks or separate installations and
+giving administrators a single configuration surface for heterogeneous clusters.
+
+Worker pools also provide a path for consuming custom handler and launcher
+images built from forked release branches that carry patches not yet merged
+upstream, without requiring changes to KubeVirt's default images.
+
+## Goals
+
+- Enable operators to deploy additional virt-handler DaemonSets targeting
+  specific node pools
+- Allow custom virt-handler and virt-launcher images per node pool
+- Automatically and transparently select the appropriate virt-launcher image
+  for VMIs based on admin-configured selectors that match VMI device requests
+  and/or labels
+- Maintain backward compatibility with existing single virt-handler deployments
+- Require no VMI-level changes from end users — pool selection is entirely
+  admin-configured
+
+## Non Goals
+
+- Runtime image switching for running VMIs
+- Automatic detection of node capabilities
+- Cross-pool live migration with image transformation
+- Multi-hypervisor support in Alpha: the hypervisor abstraction layer
+  ([VEP #97](../hypervisor-abstraction.md)) introduces per-cluster hypervisor
+  configuration. Worker pools are a natural mechanism for extending this
+  to per-pool hypervisor backends (e.g., KVM on some nodes, MSHV on others),
+  but this is deferred to Beta. See the Beta graduation requirements for
+  details.
+
+## Definition of Users
+
+- **Cluster administrators**: Configure worker pools by creating `WorkerPool`
+  CRs with selectors that match VMIs by device names and/or labels.
+  Administrators control which VMIs get which launcher images without end
+  user involvement.
+- **VMI users**: Create VMIs as normal — requesting devices, setting labels, etc.
+  Pool selection happens automatically and transparently.
+
+## User Stories
+
+### Story 1: GPU Node Pool (Admin)
+
+As a cluster administrator using the
+[NVIDIA GPU Operator](https://github.com/NVIDIA/gpu-operator), I want VMIs
+that request Tesla T4 GPUs to automatically use a GPU-optimised virt-launcher
+image, without requiring VMI users to know about pools or set special labels.
+
+The GPU Operator labels nodes with product-specific labels such as
+`nvidia.com/gpu.product=Tesla-T4`. KubeVirt's own device plugin registers
+per-model GPU resources via `permittedHostDevices` in the KubeVirt CR (e.g.,
+`nvidia.com/TU104GL_Tesla_T4`). The pool's `nodeSelector` targets nodes with
+the matching product label, while the `vmiSelector.deviceNames` matches VMIs
+requesting that GPU model.
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - WorkerPools
+    permittedHostDevices:
+      pciHostDevices:
+        - pciVendorSelector: "10DE:1EB8"
+          resourceName: "nvidia.com/TU104GL_Tesla_T4"
+---
+apiVersion: worker.kubevirt.io/v1alpha1
+kind: WorkerPool
+metadata:
+  name: gpu-pool
+spec:
+  virtLauncherImage: registry.example.com/kubevirt/virt-launcher:v1.0.0-gpu
+  nodeSelector:
+    nvidia.com/gpu.product: Tesla-T4
+  vmiSelector:
+    deviceNames:
+      - "nvidia.com/TU104GL_Tesla_T4"
+```
+
+### Story 2: Multi-Pool Configuration (Admin)
+
+As a cluster administrator, I want different virt-launcher images for Tesla T4
+GPU and Intel Arria 10 FPGA workloads, each matched automatically by their
+KubeVirt device name.
+
+```yaml
+apiVersion: worker.kubevirt.io/v1alpha1
+kind: WorkerPool
+metadata:
+  name: gpu-pool
+spec:
+  virtLauncherImage: registry.example.com/kubevirt/virt-launcher:v1.0.0-gpu
+  nodeSelector:
+    nvidia.com/gpu.product: Tesla-T4
+  vmiSelector:
+    deviceNames:
+      - "nvidia.com/TU104GL_Tesla_T4"
+---
+apiVersion: worker.kubevirt.io/v1alpha1
+kind: WorkerPool
+metadata:
+  name: fpga-pool
+spec:
+  virtLauncherImage: registry.example.com/kubevirt/virt-launcher:v1.0.0-fpga
+  nodeSelector:
+    fpga.intel.com/present: "true"
+  vmiSelector:
+    deviceNames:
+      - "intel.com/fpga-arria10"
+```
+
+### Story 3: Custom Handler and Launcher Images (Admin)
+
+As a cluster administrator, I want to deploy custom virt-handler and
+virt-launcher images on Tesla T4 GPU nodes because both components require
+changes to support the hardware — the handler needs a custom device manager
+plugin and node labeller, while the launcher needs GPU driver libraries.
+
+```yaml
+apiVersion: worker.kubevirt.io/v1alpha1
+kind: WorkerPool
+metadata:
+  name: gpu-pool
+spec:
+  virtHandlerImage: registry.example.com/kubevirt/virt-handler:v1.0.0-gpu
+  virtLauncherImage: registry.example.com/kubevirt/virt-launcher:v1.0.0-gpu
+  nodeSelector:
+    nvidia.com/gpu.product: Tesla-T4
+  vmiSelector:
+    deviceNames:
+      - "nvidia.com/TU104GL_Tesla_T4"
+```
+
+### Story 4: Label-Based Pool (Admin)
+
+As a cluster administrator, I want to assign a custom virt-launcher to VMIs
+labelled for a specific workload class, regardless of their device requests.
+
+```yaml
+apiVersion: worker.kubevirt.io/v1alpha1
+kind: WorkerPool
+metadata:
+  name: secure-pool
+spec:
+  virtLauncherImage: registry.example.com/kubevirt/virt-launcher:v1.0.0-hardened
+  nodeSelector:
+    security-zone: restricted
+  vmiSelector:
+    vmiLabels:
+      matchLabels:
+        workload-class: secure
+```
+
+### Story 5: VMI User Creates a GPU VMI (Transparent)
+
+As a VMI user, I create a VMI requesting an NVIDIA Tesla T4 GPU. I do not need
+to know about pools, set special node selectors, or reference any pool
+configuration. The cluster administrator has already configured the pool to
+match `nvidia.com/TU104GL_Tesla_T4` devices, and the correct virt-launcher
+image is selected automatically.
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: gpu-vm
+spec:
+  domain:
+    resources:
+      requests:
+        memory: 4Gi
+    devices:
+      gpus:
+        - name: gpu1
+          deviceName: nvidia.com/TU104GL_Tesla_T4
+      disks:
+        - name: containerdisk
+          disk:
+            bus: virtio
+  volumes:
+    - name: containerdisk
+      containerDisk:
+        image: registry.example.com/my-gpu-vm:latest
+```
+
+### Story 6: Rolling Node OS Upgrade (Admin)
+
+As a cluster administrator performing a rolling upgrade from one OS version
+to another (e.g., RHEL 9 to RHEL 10), I need to run different virt-handler
+and virt-launcher images on upgraded nodes because the new OS introduces
+breaking libvirt/QEMU compatibility changes. During the rollout, the cluster
+temporarily contains both old and new OS nodes.
+
+Some breaking changes may also remove features entirely — for example,
+RHEL 10 removes support for mediated devices (mdev/vGPU). VMIs relying on
+mdev must remain on RHEL 9 nodes with a virt-launcher built against the
+older libvirt/QEMU stack that still supports mdev. In this case the pool
+is not temporary: RHEL 9 nodes and their pool must be retained as long as
+mdev workloads exist.
+
+I label upgraded nodes with `os-version: rh10`, create a WorkerPool targeting
+them, and progressively label VMIs with `runtime: rh10` to migrate them to
+the new runtime stack. VMIs that depend on features removed in RHEL 10
+(such as mdev) remain on the default RHEL 9 handler and are not labelled.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/200-virt-handler-pools/virt-handler-pools.md
```diff
@@ -0,0 +1,331 @@
...
+- Avoids race conditions during scheduling
+
+### Component Changes
+
+#### virt-operator
+
+- Creates additional DaemonSets (`virt-handler-<name>`) for each entry in `virtHandlerPools`
+- Applies the pool's `nodeSelector` as pod scheduling constraints
+- Adds `kubevirt.io/handler-pool: <name>` label to additional DaemonSets
+- Injects `RequiredDuringSchedulingIgnoredDuringExecution` node affinity on the primary virt-handler with `NotIn` expressions to avoid nodes matching any pool's nodeSelector
```
r4start: Perhaps `virt-operator` should also add a PodAntiAffinity rule which should prevent a case when one node is eligible for several pools.

It seems that it's better to have a DaemonSet stuck in Pending rather than several virt-handler's on one node.

veps/sig-compute/200-virt-handler-pools/virt-handler-pools.md
```diff
@@ -0,0 +1,331 @@
...
+| `{cpu: "intel"}` | `{gpu: "true"}` | No |
+| `{}` (empty) | `{gpu: "true"}` | No |
+
+**Rationale for nodeSelector matching over node detection:**
+
+- Allows image selection before pod scheduling
+- Provides predictable behavior - users know which image will be used by looking at the VMI spec
+- Avoids race conditions during scheduling
+
+### Component Changes
```
ksimon1: Will there be any change in virt-handler component?

veps/sig-compute/200-virt-handler-pools/virt-handler-pools.md
```diff
@@ -0,0 +1,331 @@
+# VEP #200: Virt-Handler Pools
```
ksimon1: Would it make sense to define in this VEP possible validation rules? E.g. like:
Duplicate pool names
Overlapping nodeSelectors
Empty nodeSelector
...

veps/sig-compute/200-virt-handler-pools/virt-handler-pools.md
lyarwood: I don't think so directly to handler itself at this time.

veps/sig-compute/200-virt-handler-pools/virt-handler-pools.md
```diff
@@ -0,0 +1,331 @@
...
+
+**Downgrade:**
+
+- Additional DaemonSets are deleted when feature gate is disabled
+- Running VMIs served by additional handler pools continue running
+- New VMIs use default images
+
+**Version Skew:**
+
+All virt-handler images must be from the same KubeVirt version. Using mismatched versions is unsupported and may cause undefined behavior.
```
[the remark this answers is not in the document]
lyarwood: Yes we can do some basic semver validation but I'd also like a way of disabling this as ultimately the admin/user might have their own versioning scheme for their images.

lyarwood:
> > Thanks for the initial review folks, @xpivarc has highlighted privately that this effort overlaps with upcoming multi-hypervisor work and as such we need to ensure we can address both use cases with a single solution.
> 
> @lyarwood, maybe we can reduce the scope of this VEP to virt-handler pools. What do you think?

Launcher is a dependency of handler ultimately so what ever solution we eventually land on should be capable of switching both out IMHO.

> I’ve been [prototyping](https://github.com/r4start/kubevirt/tree/gradual-ds-releases-issue-16681) the proposal. The diff is already large, and I haven’t finished yet. Moreover, I haven’t started implementing virt-launcher support.

Thanks, https://github.com/kubevirt/kubevirt/pull/16692 was my previous attempt but again given the need to handle the multi-hypervisor case I think this design is going to change and likely to include more changes in the VMI itself.

lyarwood:
> My notes from the [v1.9.0 unconference](https://unconference.kubevirt.io/-3cpJ5bITgWJNchS-m_jeg?edit).
> 
> ```
> * [lyarwood] VEP 200 - virt-handler pools
>   * [lyarwood] https://blog.yarwood.me.uk/presentations/kubevirt/unconference/v1.9.0/virt-handler-pools/presentation.html - Slides
>   * [lyarwood] https://github.com/kubevirt/enhancements/pull/202 - VEP PR
>   * [lyarwood] This might allow actual canary upgrades through sepereate DaemonSets
>   * [lyarwood] Concerens
>     * Updates/upgrades
>     * Scheduling aspects
>     * Why this isn't possible through an external webhook?
>   * [lyarwood] Suggestions
>     * Image pull runtime class k8s feature
>     * Can this be a kubevirt plugin if we extend that implementation to support injecting virt-handler/virt-launcher images somehow?
>         * [lyarwood] Looks like podHooks was already included but not documented, might be able to support the launcher injection usecase.
>   * [lyarwood] Actions
>     * Revisit the plugin VEP handling this use case
>     * Maybe think about the canary upgrade use case as a seperate feature?
> ```

As discussed at the unconference I'm going to close this VEP in favour of turning this into a https://github.com/kubevirt/enhancements/pull/191 based plugin for now.

lyarwood:
Rebased and updated the VEP with the following changes:

- Added VEP Status Metadata section with v1.9 alpha target per the template
- Added VEP #190 (Structured Plugins) as a rejected alternative — MAPs cannot perform cross-object lookups required for VMI inspection during virt-launcher pod admission, so a full webhook would still be needed
- Added VEP #210 context — worker pools remain necessary for consuming custom images from forked release branches beyond the CS10 transition
- Pool removal is now blocked while matched VMIs are running; administrators must drain impacted nodes first
- Added overlapping nodeSelector validation warning
- Clarified that virt-handler itself requires no changes
- Reduced implementation detail throughout (removed function names, kubebuilder markers, specific test file paths) to keep the design high level
- Made the VEP less AIE-specific and more generic about the custom handler/launcher use case

lyarwood:
@vladikr would you mind taking another look at this after our discussion around kubevirt plugins not being a great fit for the use case.

fossedihelm:
@lyarwood Generally, I agree with the need. I have several doubts about overlapping cases, and, in general, these questions:
- What happens if a VM matches one worker pool with a device and another with labels?
- Should we have to introduce a priority field? With alphabetical order as a fallback?
- [_Thinking(typing) out loud_] How about a new CRD? like MigrationPolicies?

lyarwood:
FWIW I'm going to file an exception request shortly for this VEP as I was out sick last week and didn't have an opportunity to land ths.

> @lyarwood Generally, I agree with the need. I have several doubts about overlapping cases, and, in general, these questions:
> 
> * What happens if a VM matches one worker pool with a device and another with labels?

Thanks @fossedihelm , the VEP suggests a first matcher wins approach, so if the device matches first the VM would end up on that pool regardless of the label.

> * Should we have to introduce a priority field? With alphabetical order as a fallback?

Priority in the sense of where VMs are scheduled?

> * [_Thinking(typing) out loud_] How about a new CRD? like MigrationPolicies?

Yes that is something we could do here to reduce polluting the KubeVirt CR during Alpha and potentially beyond that. I'll repsin today with this.

alaypatel07:
RH10 introduces breaking libvirt/qemu compatibility changes that require new virt-launcher images. 

During a rolling node OS upgrade, the cluster temporarily contains both older and RH10 worker nodes. This creates the same operational problem as permanent heterogeneous node pools: virt-launcher and virt-handler images must be selected based on the node compatibility domain, otherwise VMIs can land on nodes with an incompatible runtime stack.

Common problems:
1. Two separate node types
2. Needs different virt-launcher (potentially virt-handler) images

Additional problems:
1. Due to rolling update, one node pool will grow and other will shrink
2. VMI spec does not have a marker ahead of time which node pool the vmi is going to, it will be determined after scheduling.

@vladikr @lyarwood curious if this problem can be handled as part of this VEP?

lyarwood:
Thanks @alaypatel07, great question — I think this is already covered by the current design.

The flow for a rolling node OS upgrade would be:

1. Label the RH10 nodes (e.g. `os-version: rh10`) and create a `WorkerPool` targeting them with `nodeSelector: {os-version: rh10}` and a `vmLabels` selector (e.g. `runtime: rh10`). The pool specifies the RH10-compatible handler and launcher images.
2. The pool's virt-handler DaemonSet schedules on RH10 nodes, and anti-affinity removes the primary handler from those nodes — each node runs exactly one handler.
3. As the admin is ready to migrate workloads, they label VMIs with `runtime: rh10`. On next restart or live migration, those VMIs are matched to the pool and land on RH10 nodes with the correct launcher image.
4. As the rollout progresses, more nodes are upgraded and more VMIs are labelled until the migration is complete.
5. Once all nodes are on RH10, the pool is deleted and the default KubeVirt images are updated to the RH10 versions.

For your second additional problem — the VMI not having a marker ahead of time — I think it's reasonable to ask administrators to label workloads they want to migrate to the new pool. This keeps the matching explicit and auditable, and avoids the complexity of post-scheduling image selection.

The pool's `nodeSelector` being merged into the virt-launcher pod's node affinity ensures that a matched VMI can only land on nodes served by that pool, so there's no risk of a VMI ending up on an incompatible node.

lyarwood:
@alaypatel07 I've also added this rolling node OS upgrade use case as Story 6 in the VEP. It covers both the temporary rollout scenario you described and a longer-lived variant — for example, RHEL 10 removes mdev/vGPU support, so VMIs relying on mdev must remain on RHEL 9 nodes with a compatible virt-launcher indefinitely, not just during the transition.

vladikr:
/approve
I'm fine with trying out this approach for alpha.

0xFelix (commented):
These are my notes skimming over this VEP. I wasn't familiar with it until now, so please excuse me if some of these topics already have been discussed. 
 
* **Dynamic Resource Allocation:** How does DRA (Dynamic Resource Allocation) factor into this design? 
* **Selector Standards:** Should we consider using standard `nodeSelector` and `vmiSelector` instead of custom implementations? Similarly, shouldn't `VMLabels` and `MatchLabels` utilize a standard Kubernetes `LabelSelector`?
* **Scheduling and Resource Conservation:** If a VMI does not match any specific pool, will it default to being scheduled anywhere? We should clarify if there is a need for exclusive worker pools that only accept matching VMIs to conserve specialized resources for the workloads that truly require them.
* **Predictability of Pool Assignment:** Is there a risk that worker pool matching will come as a surprise to VMI owners? Much like the old `VMIPresets`, it seems difficult to predict when a worker pool will apply until it actually happens. How can we improve this visibility?
* **KubeVirt as a Scheduler:** If the node affinity on the `virt-launcher` Pod matches any node within the worker pool, it effectively shifts KubeVirt into a scheduler-like role. Is this the intended architectural direction?
* **VEP Terminology Consistency:** The VEP currently mixes "VM" and "VMI" terminology. Assuming the scope of this VEP only applies to the instance level, we should ensure the document strictly uses "VMI".
* **Custom Image Compatibility:** How are users expected to build these customized images? More importantly, how do we guarantee versioning and compatibility with the rest of the KubeVirt control plane (e.g., ensuring gRPC communication remains stable between a default `virt-handler` and a custom `virt-launcher` or custom `virt-handler` and the regular `virt-api`)?

lyarwood (commented):
Thanks for taking the time to look through this @0xFelix!

> How does DRA (Dynamic Resource Allocation) factor into this design?

DRA is definitely something the selector could match on in the future. For alpha the focus is on the existing device plugin resource names and labels, but the `selector` structure is designed to be extensible — adding a `resourceClaims` field later would be straightforward. I'll add a note about this as a potential future extension.

> Should we consider using standard `nodeSelector` and `vmiSelector` instead of custom implementations? Similarly, shouldn't `VMLabels` and `MatchLabels` utilize a standard Kubernetes `LabelSelector`?

`nodeSelector` already uses the standard `map[string]string` consistent with PodSpec. For `vmLabels`, good catch on the naming — I'll rename to `vmiLabels`/`WorkerPoolVMILabels` since the matching scope is VMI (see terminology point below). On using `metav1.LabelSelector` with `matchExpressions` support — that's worth considering, I can look at switching to that for more flexibility. On renaming `selector` to `vmiSelector` — I'm not a huge fan of that but I can see the argument that it's more specific about what's being matched. Happy to hear other opinions.

> If a VMI does not match any specific pool, will it default to being scheduled anywhere? We should clarify if there is a need for exclusive worker pools that only accept matching VMIs to conserve specialized resources for the workloads that truly require them.

Unmatched VMIs use the default virt-launcher image and the primary virt-handler has anti-affinity to exclude pool nodes, so each node runs exactly one handler. For device-based pools, resource scarcity naturally prevents non-GPU VMIs from landing on GPU nodes. For label-based pools the admin controls which VMIs match. There is currently a gap where an unmatched VMI could theoretically land on a pool node — the default launcher would be used but the pool's custom virt-handler would serve it. I'll look at whether we should add anti-affinity for pool nodes to unmatched VMI pods to close this for alpha, and update the VEP to clarify the behaviour either way.

> Is there a risk that worker pool matching will come as a surprise to VMI owners? Much like the old `VMIPresets`, it seems difficult to predict when a worker pool will apply until it actually happens. How can we improve this visibility?

Fair comparison, but the key difference from VMIPresets is that worker pools don't mutate the VMI spec at all — the VMI the user created is exactly what runs. Pool selection only affects which virt-launcher container image wraps the VMI, which is an infrastructure detail below the VMI abstraction. Admins are ultimately responsible for pool configuration, and I can't think of a reason a VM owner would need to care about which handler/launcher image is being used. That said, the pool name annotation on the virt-launcher pod does provide visibility when needed.

> If the node affinity on the `virt-launcher` Pod matches any node within the worker pool, it effectively shifts KubeVirt into a scheduler-like role. Is this the intended architectural direction?

The pool's `nodeSelector` is merged into the virt-launcher pod's required node affinity — a standard Kubernetes scheduling constraint, not custom scheduling logic. The Kubernetes scheduler still makes the final scheduling decision. KubeVirt already sets node affinity constraints today through various node placement features on the VMI spec; this extends that to admin-configured pool configuration rather than user-specified VMI fields.

> The VEP currently mixes "VM" and "VMI" terminology. Assuming the scope of this VEP only applies to the instance level, we should ensure the document strictly uses "VMI".

Fair point — I'll clean up the VM/VMI terminology throughout and rename `vmLabels`/`WorkerPoolVMLabels` to `vmiLabels`/`WorkerPoolVMILabels` in the API.

> How are users expected to build these customized images? More importantly, how do we guarantee versioning and compatibility with the rest of the KubeVirt control plane (e.g., ensuring gRPC communication remains stable between a default `virt-handler` and a custom `virt-launcher` or custom `virt-handler` and the regular `virt-api`)?

This works the same way custom images do today — virt-operator already allows overriding handler and launcher images globally, and users who build custom images are responsible for version compatibility. Worker pools don't change that model, they just provide a per-pool configuration surface. The version skew policy documents the constraint. Build process documentation would be good user-facing docs but is out of scope for the VEP itself.

0xFelix:
> nodeSelector already uses the standard map[string]string consistent with PodSpec. For vmLabels, good catch on the naming — I'll rename to vmiLabels/WorkerPoolVMILabels since the matching scope is VMI (see terminology point below). On using metav1.LabelSelector with matchExpressions support — that's worth considering, I can look at switching to that for more flexibility. On renaming selector to vmiSelector — I'm not a huge fan of that but I can see the argument that it's more specific about what's being matched. Happy to hear other opinions.

Why not rename the field to `vmiSelector`, drop the prefix in `WorkerPoolSelector` from `VMLabels` and use a `LabelSelector` instead of the custom `WorkerPoolVMLabels` type?

> For device-based pools, resource scarcity naturally prevents non-GPU VMIs from landing on GPU nodes.

Does it? How? A VMI without GPUs can schedule on a GPU node as well? Taking up CPU and memory?

> the default launcher would be used but the pool's custom virt-handler would serve it

We should avoid that. If the VMI was not matched by a `WorkerPool`, it should run on standard virt-launcher+virt-handler. Otherwise we might have the mentioned surprise to the VMI owner. :)

lyarwood:
> Why not rename the field to `vmiSelector`, drop the prefix in `WorkerPoolSelector` from `VMLabels` and use a `LabelSelector` instead of the custom `WorkerPoolVMLabels` type?

Agreed, I've updated the VEP. The API now uses `vmiSelector` as the field name, `Selector` as the type (dropping the `WorkerPool` prefix since the package already scopes it), and `metav1.LabelSelector` instead of the custom `WorkerPoolVMLabels` type.

> Does it? How? A VMI without GPUs can schedule on a GPU node as well? Taking up CPU and memory?

You're right, I was wrong here. A non-GPU VMI doesn't request GPU resources so resource scarcity is irrelevant to the scheduler — it can absolutely land on a GPU node if there's CPU and memory available.

> We should avoid that. If the VMI was not matched by a `WorkerPool`, it should run on standard virt-launcher+virt-handler. Otherwise we might have the mentioned surprise to the VMI owner. :)

Agreed. I've updated the VEP so that when a VMI does not match any pool, virt-controller adds anti-affinity for all pool `nodeSelector` labels to the virt-launcher pod (using `NotIn` expressions). This ensures unmatched VMIs only land on nodes served by the primary virt-handler.

alaypatel07:
> DRA is definitely something the selector could match on in the future. For alpha the focus is on the existing device plugin resource names and labels, but the selector structure is designed to be extensible — adding a resourceClaims field later would be straightforward. I'll add a note about this as a potential future extension.

I am not sure if I follow this from DRA perspective.  The devices field in workerpools match the nodes that have a specific device as extended resource, eg. `nvidia.com/foo-gpu`. In case of DRA, whether a device is available on a node is governed by ResourceSlice object and not ResourceClaims. ResourceSlice object is much more nuanced than the extended resource string `nvidia.com/foo-gpu`, hence it would be good to have an appendix to set the future plan of integrating with DRA APIs.

lyarwood:
The `vmiSelector.deviceNames` field matches **VMIs** based on their device requests (e.g., `spec.domain.devices.gpus[].deviceName`), not nodes. Node-level targeting is handled separately by `nodeSelector`.

So for DRA, a future extension would add a new field to `vmiSelector` (e.g., `resourceClaims`) that matches VMIs using DRA-based device requests. ResourceSlice is the node-side view and would be covered by `nodeSelector` labels, the same way GPU node labels work today.

vladikr:
Hey, thanks everyone for the continued discussion

When I approved this for alpha, my view was that the main goal was a clean way to deliver different virt-handler images onto specific node pools. 
The launcher override was mainly needed for the handler on those nodes to label them with the capabilities supported by that virt-launcher image/stack.
That was never intended to decide which launcher image would be used for the VMI labding on that node... 
I don't think it's the right mechanism for that - at least not in its current form.

I certainly agree that this VEP overlaps with the multi-hypervisor work in VEP97, and we should avoid ending up with two similar mechanisms. 

I think that reducing the scope to virt-handler pools (as @r4start  suggested), would treat the launcher image as a required component for the handler's node labeller.
This would help us defer the alignment with VEP97 for the next step.

We have already discussed ways to reduce the coupling between the virt-handler and the need virt-launcher image for capabilities management as part of the VEP97 with @harshitgupta1337 
One way is to extract the node-labeller functionality out of the virt-handler / virt-launcher init container entirely. 
The labeller could produce labels (or structured hints) as a payload that the handler can consume. 
This would let specialized handlers avoid pulling in a specific launcher image just for labeling, and would give each hypervisor backend its own clean extension point for capability discovery.

Let me know how I can help unblock this and move things forward.

alaypatel07:
> One way is to extract the node-labeller functionality out of the virt-handler / virt-launcher init container entirely.
The labeller could produce labels (or structured hints) as a payload that the handler can consume.
This would let specialized handlers avoid pulling in a specific launcher image just for labeling, and would give each hypervisor backend its own clean extension point for capability discovery.

+1 this will be a helpful change that can additionally help with the OS upgrade issues I mentioned earlier. https://github.com/kubevirt/enhancements/pull/202#issuecomment-4389475864

Issue #201, opened 2026-02-10
VEP 201: Enhanced VMI status reason


MeenaAlfons:
**Primary contact (assignee)**:

/assign @MeenaAlfons 

**Current Feature Stage**:

New

**Feature Gate**:

`EnhancedReason`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (1.9?):
* Beta release target (1.10?):
* Stable release target (1.11?):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

## Summary

`vmi.Status.Reason` already exists but it is only set for one case `NodeUnresponsive` and `vmi.Status.Phase` is simultaneously set to `Failed`.

`vmi.Status.Reason` should be set everytime `vmi.Status.Phase` is set, especially for final phases `Succeeded` and `Failed`.

`vmi.Status.Phase` could reach a final state for many reasons including:
- domain state changes, like `Shutdown`, `Destroyed`, `Crash`, etc
- migrations: `Migrated`, `MigrationFailed`, etc
- other failure: `NodeUnresponsive`, `VirtLauncherUnresponsive`, `VirtLauncherCrashed`, etc

This enhancement is focussing on domain state changes where the reason is extracted from `domain.Status.Reason` including the following cases:

| phase      | reason      | 
|------------|-------------|
| `Suceeded` | `Shutdown`  |
| `Suceeded` | `Destroyed` |
| `Failed`   | `Crash`     |

## Use cases

Displaying a reason for why the VM is in it's current state is important for end users as well as for debugging.

We are currently missing proper value for `Reason`. This enhancements proposes setting `Reason` to proper values.

0xFelix:
I'm open to review this. Is it on track for 1.9?

PR #203, opened 2026-02-10
VEP 201: Enhanced VMI status reason

MeenaAlfons:
### VEP Metadata

**Tracking issue**: #201 
**SIG label**: For example, /sig compute

### What this PR does
Introduce a proposal for enhanced `vmi.Status.Reason`

👉 [Readable version](https://github.com/MeenaAlfons/enhancements/blob/reason-and-shutdown-reason/veps/sig-compute/201-enhanced-vmi-status-reason/enhanced-vmi-status-reason.md)


0xFelix:
Can you squash commits? From what I can tell it looks good so far.

veps/sig-compute/201-enhanced-vmi-status-reason/enhanced-vmi-status-reason.md
```diff
@@ -0,0 +1,212 @@
...
+  
+
+## API Examples
+
+<!--
+Tangible API examples used for discussion
+-->
+
+Notice that:
+- `Reason` currently already exists as `string`. It's proposed to change it to `VirtualMachineInstanceReason` with proper `const`s
```
0xFelix, line 109: Can we keep it as string? Other reasons we have are `string` too. I'd prefer to avoid changing this existing field.

0xFelix:
@MeenaAlfons VEP freeze is on Thursday, do you still want to get this into 1.9?

veps/sig-compute/201-enhanced-vmi-status-reason/enhanced-vmi-status-reason.md
```diff
@@ -0,0 +1,212 @@
...
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+- [kubevirt/enhancements](https://github.com/kubevirt/enhancements) (this VEP)
+
+## Design
+
+<!--
+This should be brief and concise. We want just enough to get the point across
+-->
+
+- Set `vmi.Status.Reason` each time `vmi.Status.Phase` is set.
```
EdDev, line 96: I commenting post the VEP being closed due to no activity in case it is re-opened.

Usage of state type fields like `Phase` have been long realized as problematic as they introduce the need to maintain a machine state which fits an imperative system and not a declarative one.

Kubernetes has moved to `conditions` as a way to express the status in a declarative manner with simpler processing for consumers.

Therefore, if there is a need to express a "reason" for a condition in which the VM is in, please use an existing or new condition. Existing generic statuses (like "Phase") should be eventually deprecated in favor of specific conditions that reflect well the status with relevant reasons for both controllers and users.

Issue #207, opened 2026-02-16
VEP 207: Pause liveness probes


ksimon1:
**Primary contact (assignee)**:
/assign @ksimon1

**Current Feature Stage**:
New

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/208

**Timeline**:
* Alpha release target (x.y): 1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/208
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/17664
  * [ ] Docs PRs: https://github.com/kubevirt/user-guide/pull/990

> [!IMPORTANT]
> Please keep this description up to date.

xpivarc:
Issue is empty, no work tracked against 1.9

ksimon1:
@xpivarc I added all relevant links.

vladikr:
/remove-lifecycle stale

@ksimon1 is there any work you are planing for this cycle?

ksimon1:
@vladikr no, we finish work in 1.9

PR #208, opened 2026-02-16
VEP 207: Add proposal for manual GuestAgentPing probe control

ksimon1:
### VEP Metadata
**Tracking issue**: #207
**SIG label** /sig compute

### What this PR does
This proposal introduces a lightweight, annotation-based mechanism to manually pause GuestAgentPing probes at runtime using VMI annotations. By returning synthetic success responses, it prevents unwanted Pod restarts during maintenance operations such as guest OS updates.

The chosen architecture (Alternative 1) resolves these issues with minimal complexity. It explicitly avoids introducing a new HTTP proxy server in virt-launcher and leaves existing Kubernetes Pod probe specifications completely unmodified. Other explored approaches are preserved in the document under the Alternatives section for historical context.

Code was assisted by Cursor AI.

### DIFF
--- a/veps/sig-compute/207-pauseProbes/pause-probes.md
+++ b/veps/sig-compute/207-pauseProbes/pause-probes.md
@@ -1,0 +1,253 @@
+# VEP: Probe Proxy for Manual Probe Control
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal enables manual pausing of GuestAgentPing probes via VMI annotations to prevent Pod restarts during maintenance (e.g., guest OS updates). The preferred approach leverages the existing `SyncVirtualMachine` gRPC call, which already sends the full VMI JSON (including annotations) to `virt-launcher` on every reconcile. When `kubevirt.io/pause-guest-agent-probes` annotation is set, `virt-launcher` reads it from the VMI received in `SyncVirtualMachine` and short-circuits `GuestPing` calls to return immediate success. No new RPC, proxy server, ConfigMap, or new resources required.
+
+## Motivation
+
+KubeVirt probes translate to Pod probes executed by kubelet. **This affects VMs using GuestAgentPing as a liveness probe** — failures cause Pod restart and destroy the running VM (readiness probe failures only affect Service endpoints).
+
+1. **Guest OS Updates / Reboots**: Guest agent unavailable during OS updates. Especially Windows updates can take hours with multiple reboots — unpredictable duration exceeds any reasonable `failureThreshold`. Requires explicit pause mechanism.
+
+**Why GuestAgentPing only?** Keeping the scope narrow to a single probe type minimizes implementation risk and complexity. Alternative approaches covering all probe types are documented below but require significantly more complex implementations (see Alternatives 3 and 4).
+
+## Goals
+
+- Manual pause/unpause of GuestAgentPing probes via VMI annotations (no Pod restart)
+- Maintain strict backward compatibility (Pod probe specs unchanged)
+- Internal actions which might trigger probe pause (e.g. live migration) have higher priority than annotation.
+
+## Non Goals
+
+- Pausing non-GuestAgentPing probes (TCP, UDP, HTTP)
+- Replacing Kubernetes probe mechanisms
+- Custom health check logic or complex conditions
+- Multi-VMI control or HTTP endpoints
+
+## User Stories
+
+- Manually pause via annotation before guest OS updates (especially long Windows updates) and manually remove the annotation when OS update finishes.
+- Query pause status via annotation if paused by user
+
+## Use Cases
+
+**Supported:**
+1. **Guest OS Updates**: Manual annotation before updates (critical for unpredictable Windows updates, might be used for other OSes updates too)
+
+**Unsupported:**
+1. Permanent disabling (remove probe from spec instead)
+2. Custom probe responses
+3. Non-GuestAgentPing probes
+4. Auto-set annotation during internal operations (fsfreeze, snapshot, backup, ...)
+
+## Repos
+
+kubevirt/kubevirt
+
+### Why Annotations Instead of a Spec API Field
+
+While probe pausing can be viewed as a desired state (e.g., for planned maintenance windows or other OSes), using annotations provides operational flexibility: it allows runtime control without VMI spec changes and maintains backward compatibility. For permanent probe disabling, users should remove the probe from the spec instead of keeping the annotation set indefinitely.
+
+### Observability
+
+- **Info** level: Log state transitions (paused ↔ unpaused) with VMI name
+
+
+### Pausing Probes via Annotation
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-vm
+  annotations:
+    kubevirt.io/pause-guest-agent-probes: "true" 
+spec:
+  domain:
+    devices:
+      disks:
+      - disk:
+          bus: virtio
+        name: containerdisk
+    resources:
+      requests:
+        memory: 1Gi
+  livenessProbe:
+    guestAgentPing: {}
+    initialDelaySeconds: 120
+    periodSeconds: 20
+  volumes:
+  - containerDisk:
+      image: quay.io/containerdisks/fedora
+    name: containerdisk
+```
+
+### Manually Pausing Probes
+
+```bash
+# Add annotation to pause GuestAgentPing probes on running VMI
+kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes=true
+
+# Check probe pause status by inspecting the VMI annotation
+kubectl get vmi my-vm -o jsonpath='{.metadata.annotations.kubevirt\.io/pause-guest-agent-probes}'
+# Output: true
+
+# Remove annotation to unpause probes
+kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes-
+```
+
+## Relationship to Upstream Kubernetes KEP
+
+[KEP #5002: Introduce hot disable of probes](https://github.com/kubernetes/enhancements/issues/5002) is in **Draft stage** (April 2026). Even if implemented, KubeVirt needs its own VM/VMI ↔ Pod interface since the KEP is Pod-focused. The `kubevirt.io/pause-guest-agent-probes` annotation is scoped to GuestAgentPing probes. If the KEP reaches GA and KubeVirt expands to all probe types, a new generic annotation can be introduced alongside migration tooling.
+
+## Design
+
+All alternatives below use the same user-facing interface: VMI annotations control probe pausing.
+
+The preferred alternative is number 1 (Annotation via existing `SyncVirtualMachine` — GuestAgentPing only). Alternatives 2–4 are included to compare the full spectrum of options explored. Alternatives 1 and 2 are narrow in scope (GuestAgentPing only) but extremely lightweight. Alternatives 3 and 4 extend the gRPC and file-based approaches to cover all VMI probe types by leveraging `virt-probe` as the common entry point.
+
+### Alternative 1: Annotation via Existing `SyncVirtualMachine` (Preferred) ✅
+
+Reads the pause annotation from the VMI JSON already delivered by the existing `SyncVirtualMachine` gRPC call. Targets **GuestAgentPing only**. No new RPCs, servers, ConfigMaps, or resources.
+
+`SyncVirtualMachine` sends a `VMIRequest` containing `VMI.vmiJson` — the full JSON-marshaled `VirtualMachineInstance` including `metadata.annotations`. virt-launcher already receives this on every reconcile loop iteration, so the annotation is delivered without any protocol changes.
+
+**Flow**:
+```
+┌───────────────────────────────────────────────────────────────┐
+│                    virt-handler                               │
+│                                                               │
+│   VMI annotation:                                             │
+│   kubevirt.io/pause-guest-agent-probes: "true"                │
+│           │                                                   │
+│           ▼                                                   │
+│   syncVirtualMachine() reconcile loop                         │
+│   → client.SyncVirtualMachine(vmi)   (existing call)          │
+└───────────────────────────────────────────────────────────────┘
+                         │
+                         │ (existing gRPC over Unix socket)
+                         │ (VMIRequest with full VMI JSON incl. annotations)
+                         ▼
+┌───────────────────────────────────────────────────────────────┐
+│                    virt-launcher Pod                          │
+│                                                               │
+│   ┌─────────────────────────────────────────────────────┐     │
+│   │  SyncVirtualMachine handler                         │     │
+│   │  → read annotation from VMI JSON                    │     │
+│   │  → domainManager.guestAgentPaused.Store(true/false) │     │
+│   └─────────────────────────────────────────────────────┘     │
+│                                                               │
+│   ┌─────────────────────────────────────────────────────┐     │
+│   │  GuestPing (called by kubelet liveness probe)       │     │
+│   │  → if guestAgentPaused: return nil (success)        │     │
+│   │  → else: execute guest-ping via QEMU agent          │     │
+│   └─────────────────────────────────────────────────────┘     │
+│                                                               │
+│   ┌─────────────────────────────────────────────────────┐     │
+│   │  VM / QEMU Guest Agent                              │     │
+│   └─────────────────────────────────────────────────────┘     │
+└───────────────────────────────────────────────────────────────┘
+```
+
+**Pros:**
+- Zero protocol changes — reuses existing `SyncVirtualMachine` call and `VMIRequest` message
+- Level-triggered: state is refreshed on every reconcile, no stale in-memory state possible
+- Rollback-safe: old virt-handler sends VMI without the annotation → virt-launcher reads absence as `false` → probes auto-unpause
+- Fast propagation (seconds via reconcile loop)
+- No ConfigMaps, no API calls from virt-launcher
+- Thread-safe (`atomic.Bool`)
+- Minimal code: annotation check in existing handler, 1 field, 1 function change
+
+**Cons:**
+- GuestAgentPing only (no HTTP/TCP/Exec)
+
+[Reference implementation](https://github.com/kubevirt/kubevirt/compare/main...ksimon1:kubevirt:pause-probes?expand=1)
+
+### Alternative 2: File-Based Shared State
+
+GuestAgentPing only. virt-handler writes sentinel file to EmptyDir via `/pods/<pod-uid>/volumes/kubernetes.io~empty-dir/public/probe-paused`. virt-launcher stats file on each probe — exists = success.
+
+**Pros:** No gRPC changes, survives restarts (file persists), minimal code, no API load  
+**Cons:** GuestAgentPing only, relies on undocumented kubelet path, file cleanup required
+
+### Alternative 3: HTTP Proxy + gRPC State
+
+Covers **all probe types** via HTTP proxy (:9500) in virt-launcher. virt-controller rewrites all probes to `httpGet` with original details in headers (`X-KV-Type`, `X-KV-Port`, etc.). Paused flag via gRPC.
+
+**Pros:** All probe types, gRPC state delivery, no API calls  
+**Cons:** New HTTP server (security hardening needed), extra HTTP hop, probe rewriting complexity, port reservation
+
+### Alternative 4: HTTP Proxy + File State
+
+Same as Alternative 3, but paused flag via sentinel file (Alternative 2 approach) instead of gRPC. Proxy stats file on each request.
+
+**Pros:** All probe types, no gRPC changes, survives restarts, no API calls  
+**Cons:** Same as Alt 3 (HTTP server, probe rewriting, extra hop) + `stat` syscall + kubelet path dependency
+
+
+### Alternatives Comparison
+
+| Aspect | Alt 1: SyncVM ✅ | Alt 2: File | Alt 3: Proxy+gRPC | Alt 4: Proxy+File |
+|--------|-----------------|-------------|-------------------|-------------------|
+| **Scope** | GuestAgentPing | GuestAgentPing | All types | All types |
+| **virt-controller** | No change | No change | Probe rewrite | Probe rewrite |
+| **virt-handler** | No change | +file write | +RPC call | +file write |
+| **virt-launcher** | +annotation read +atomic.Bool | +os.Stat | +HTTP server +RPC | +HTTP server |
+| **Proto changes** | None | None | +1 RPC | None |
+| **Pod spec** | No change | No change | All → httpGet | All → httpGet |
+| **Overhead/probe** | atomic read | stat syscall | HTTP hop | HTTP hop + stat |
+| **State delivery** | Level-triggered | Level-triggered | Edge-triggered | Level-triggered |
+| **Rollback safety** | Auto-unpause | Auto-unpause | Stays paused | Auto-unpause |
+| **Restart recovery** | Next reconcile | Persists | Next reconcile | Persists |
+| **New server** | No | No | Yes (:9500) | Yes (:9500) |
+| **Complexity** | Very Low | Very Low | Medium | Medium |
+
+## Update/Rollback Compatibility
+
+- **Upgrade**: Fully backward compatible. Updated virt-handler sends VMI with annotation via existing `SyncVirtualMachine`; old virt-launcher ignores the unknown annotation. Probes work normally.
+- **Rollback**: Old virt-handler continues calling `SyncVirtualMachine` with the full VMI JSON. Since it doesn't set the annotation, virt-launcher reads its absence as `false` — probes auto-unpause on the next reconcile. No manual cleanup or pod restart needed.
+- **Feature Flag**: None. Annotation is opt-in — no effect unless set to `"true"`.
+
+## Functional Testing Approach
+
+**Unit/Integration/E2E**: Test coverage depends on selected alternative (see graduation criteria for required tests).
+
+## Implementation History
+
+- 2026-02-16: Initial VEP proposal created
+- 2026-03-17: Updated based on PR review feedback (dominikholler, 0xFelix): clarified liveness vs readiness probe scope, added automatic pausing for internal operations, addressed multi-launcher migration behavior, added observability requirements, removed feature gate, addressed gRPC overhead and EmptyDir alternatives
+- 2026-03-24: Removed live migration from scope (addressed separately by [kubevirt/kubevirt#17235](https://github.com/kubevirt/kubevirt/pull/17235)). Added Windows updates as a motivation. VEP scope adjusted to focus on annotation-based pausing for backup/snapshot, guest OS updates, and Windows updates.
+- 2026-04-16: Added section on upstream [Kubernetes KEP #5002](https://github.com/kubernetes/enhancements/issues/5002) (hot disable of probes) and documented the migration path.
+- 2026-04-21: Updated based on enp0s3 feedback: clarified why GuestAgentPing-only scope (customer requirements + reduced implementation risk), expanded annotation vs. API field rationale to acknowledge pause as desired state, fixed diagram formatting with code fences.
+- 2026-04-24: Removed internal operations (fsfreeze, snapshot, backup) from supported use cases. Auto-pause during internal operations is now explicitly unsupported — annotation is user-driven only. Replaced dedicated `SetGuestAgentPaused` gRPC RPC with reading annotation from the VMI JSON already delivered by `SyncVirtualMachine` — zero protocol changes, level-triggered state, and rollback-safe (old virt-handler sends VMI without annotation → auto-unpause).
+
+## Graduation Requirements
+
+**Alpha:**
+- [ ] Annotation pause/unpause works end-to-end
+- [ ] Logging at Info/Debug levels
+- [ ] Unit tests for paused/unpaused paths
+- [ ] E2E: set annotation → success without guest contact → remove → resume
+
+**Beta:**
+- [ ] Guest OS update workflow docs
+
+**GA:**
+- [ ] Stable for ≥2 releases, no regressions
+- [ ] Documentation exists
+
+
+## References
+
+- [Kubernetes Probes Documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
+- [KubeVirt Probes Documentation](https://kubevirt.io/user-guide/user_workloads/liveness_and_readiness_probes/)
+- [Kubernetes KEP #5002: Introduce hot disable of probes](https://github.com/kubernetes/enhancements/issues/5002)


dominikholler (changes requested):

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,485 @@
...
+## Update/Rollback Compatibility
+
+- **Upgrade**: The feature is strictly backward compatible. Running VMs created prior to the update will continue executing `GuestAgentPing` probes normally. The `SetGuestAgentPaused` RPC is called by the updated `virt-handler` on every reconcile; if `virt-launcher` has not yet been updated, gRPC gracefully returns an "unimplemented" error which `virt-handler` logs as a warning and ignores — probes continue working as before.
+
+- **Rollback**:
+  - If rolling back to a KubeVirt version without this feature, `virt-handler` will stop sending `SetGuestAgentPaused` calls.
+  - `virt-launcher` will keep whatever in-memory state it had until the Pod restarts, at which point the paused flag defaults to `false` and probes resume normally.
+  - No Kubernetes resources (ConfigMaps, etc.) need to be cleaned up — there are none.
+
+- **Feature Flag**: A feature gate `GuestAgentProbePause` (or `DynamicProbeControl`) controls this functionality:
```
0xFelix: Is the guest agent ping probe still behind a FG? I'd try to push for no additional FG tbh. Annotations are a very soft API anyway.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,485 @@
...
+- Providing HTTP control endpoints for runtime management from within the Pod (annotation-only control)
+
+## Definition of Users
+
+- **VM User**: A person who runs VMs on KubeVirt and wants their VMs to remain stable during maintenance operations
+- **Cluster Administrator**: A person who manages the KubeVirt infrastructure and needs to control probe behavior during cluster-wide operations
+
+## User Stories
+
+- As a VM user, I want my VM to not be restarted during live migration due to probe failures
```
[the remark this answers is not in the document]
ksimon1: Live migration is now removed from the VEP

ksimon1:
@tiraboschi eventhough the k8s KEP will be accepted and implemented, we will need some interface between VM/VMI object and pod (the KEP is focusing only on pods). For this purpose we will have annotation and custom implementation, which can be if the KEP will be implemented changed to the k8s implementation

ksimon1:
@0xFelix, @enp0s3, @vladikr, @iholder101, @lyarwood can you please review this VEP

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal introduces a lightweight mechanism to dynamically control GuestAgentPing probes at runtime. By utilizing VMI annotations, administrators can pause these probes during maintenance operations such as backup, guest OS updates (including long-running Windows updates), or filesystem freeze operations. The preferred approach (Alternative 1) extends the existing gRPC channel between `virt-handler` and `virt-launcher` with a single new `SetGuestAgentPaused` RPC. When the user sets the `kubevirt.io/pause-probes` annotation on a VMI, `virt-handler` communicates this to `virt-launcher` on the next reconcile. `virt-launcher` stores the state in an atomic boolean and short-circuits `GuestPing` calls to return immediate success while the flag is set. This prevents probe failures from triggering unwanted Pod restarts without requiring a proxy server, a ConfigMap, or any new Kubernetes resources.
```
0xFelix: It should pause automatically during a freeze, an administrator should not have to add an annotation in that case? IIUC manual intervention should only be required in the case of maintenance, where a guest would stop responding for reasons unknown to KubeVirt?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+
+## Motivation
+
+Currently, KubeVirt probes are directly translated into Pod probes and executed by the Kubernetes kubelet against the VM or guest agent. This creates several challenges during standard administrative operations.
+
+**Important**: this issue primarily affects VMs that use GuestAgentPing as a **liveness probe**. When used as a readiness probe, probe failures only remove the VM from Service endpoints — the Pod is not restarted. The scenarios below therefore focus on liveness probe failures, which cause kubelet to forcefully restart the Pod and destroy the running VM.
+
+1. **Backup / Snapshot / Filesystem Freeze**: When taking VM snapshots or backups, the guest may need to be quiesced (`fsfreeze`). While the guest agent is frozen, GuestAgentPing probes will fail and kubelet will restart the Pod. Internal KubeVirt operations that freeze the guest (e.g. VirtualMachineSnapshot, `fsfreeze` via the guest agent) should automatically pause probes for the duration of the freeze.
```
0xFelix: Can you make this point more clear in the overview?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+
+2. **Guest OS Updates / Reboots**: During in-guest OS updates that require a reboot, the guest agent is unavailable for a period. Liveness probe failures during this window cause a forceful Pod restart, preventing the update from completing. Windows updates are a particularly severe case — they can take a very long time (sometimes hours), involving multiple reboots and extended periods of OS reconfiguration during which the guest agent is unavailable. Because the duration is unpredictable and can far exceed any reasonable `failureThreshold` setting, users need an explicit mechanism to pause probes for the entire update window.
```
0xFelix: That is the main use case

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+
+## Goals
+
+- Allow dynamic pausing and unpausing of GuestAgentPing probes via VMI annotations without requiring a Pod restart.
+- Return synthetic success from the GuestPing handler when probes are paused, causing virt-probe to exit 0.
```
0xFelix: Not really an expected outcome but an implementation detail.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+- Enable runtime control via VMI annotations to maintain a single source of truth.
+- Internal KubeVirt operations that make the guest agent temporarily unavailable (`fsfreeze`, snapshot) should automatically set and clear the annotation so that no manual user intervention is required.
```
0xFelix: Can't they simply call the proposed gRPC method? The annotation should be for users only? Also what would happen if annotation (set to false) and e.g. a fsfreeze operation contradict each other?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+- Maintain strict backward compatibility by leaving existing Kubernetes Pod probe specifications completely untouched.
+
+## Non Goals
+
+- Pausing of probes other than guest-agent-based probes (e.g. TCP, UDP, HTTP, generic Exec). Only GuestAgentPing is in scope.
+- Replacing Kubernetes native probe mechanisms
+- Providing custom health check logic beyond pause/unpause
+- Modifying probe results based on complex conditions (beyond pause state)
```
0xFelix: Isn't fsfreeze (originates internally) a more complex condition  than "user requested it"?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+- Providing HTTP control endpoints for runtime management from within the Pod (annotation-only control)
+
+## Definition of Users
+
+- **VM User**: A person who runs VMs on KubeVirt and wants their VMs to remain stable during maintenance operations
+- **Cluster Administrator**: A person who manages the KubeVirt infrastructure and needs to control probe behavior during cluster-wide operations
+
+## User Stories
+
+- As a VM user, I want GuestAgentPing liveness probes to be automatically paused when internal KubeVirt operations (fsfreeze, snapshot) make the guest agent temporarily unavailable, so that no manual intervention is required
```
0xFelix: Is it part of this VEP or already done?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+- As a VM user, I want to manually pause GuestAgentPing probes by adding an annotation to my VMI before initiating a guest OS update (including long-running Windows updates), and have probes resume when I remove it after the update completes
+- As a cluster administrator, I want to pause GuestAgentPing probes on specific VMIs during scheduled maintenance windows (e.g. OS patching, Windows updates)
```
0xFelix: Aren't these the same?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+- As a cluster administrator, I want to be able to query the current probe pause status of a VM/VMI via its annotation
+
+## Use Cases
+
+### Supported Use Cases
+
+1. **Filesystem Freeze / Snapshot**: When KubeVirt freezes the guest filesystem (e.g. for VirtualMachineSnapshot or an explicit `fsfreeze` call), the pause annotation is automatically set for the duration of the freeze. This prevents GuestAgentPing liveness probes from failing while the guest agent is unresponsive.
```
0xFelix: This would require a patch/sync of the VMI and fsfreeze would need to wait for the reconciliation to happen before it could safely freeze the VM introducing delay?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+2. **Guest OS Updates / Reboots**: A VM user or automation tooling manually sets the pause annotation before initiating an in-guest OS update that requires a reboot, and removes it after the guest agent is available again. This is especially important for Windows updates, which can take hours with multiple reboots — the unpredictable duration makes `failureThreshold` tuning impractical.
+
+
+### Unsupported Use Cases
+
+1. **Permanent probe disabling**: This feature is meant for temporary pausing, not permanent disabling. For permanent disabling, remove the probe from the VMI spec.
```
0xFelix: If you don't remove the annotation, then you could consider it permanent?

0xFelix: Also, if a user set this annotation it should have precedence over internal events.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+### Why Annotations Instead of a Spec API Field
+
+An annotation was chosen over a formal API field (e.g. a `paused` boolean in `guestAgentPing`) because the pause state is inherently operational and transient — it does not describe the desired probe configuration but rather a temporary runtime override. Using an annotation keeps the VMI spec clean and declarative. Importantly, the mechanism that delivers the annotation value to `virt-launcher` (gRPC or file) is equally capable of propagating a spec field, so this is not a technical limitation. If a future decision promotes the annotation to a first-class API field, the internal plumbing does not need to change.
+
+### Observability
+
+When GuestAgentPing probes are paused, `virt-launcher` must log the event so that operators can confirm the annotation actually influenced behavior:
+
+- Log at **Info** level when the paused state transitions (paused → unpaused or unpaused → paused), including the VMI name.
+- Log at **Debug** level each time a GuestPing call is short-circuited due to the paused flag, to allow fine-grained troubleshooting without flooding logs during normal operation.
```
0xFelix: Isn't that a bit excessive even for debug?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,500 @@
...
+
+Even if the Kubernetes KEP is accepted and implemented, KubeVirt will still need its own interface between the VM/VMI object and the Pod, because the KEP focuses exclusively on Pods and has no awareness of the KubeVirt virtualization layer. The annotation-based mechanism described in this VEP serves that purpose. The annotation is intentionally named `kubevirt.io/pause-probes` (generic, not GuestAgentPing-specific) so that if the upstream KEP reaches GA and KubeVirt expands scope to all probe types, the user-facing API does not need to change. Should the upstream KEP be implemented, the internal implementation can be migrated from the current custom approach (gRPC + `atomic.Bool` short-circuit in `virt-launcher`) to the native Kubernetes probe-disable API, while the VMI annotation contract remains unchanged. This gives KubeVirt a clear migration path without blocking on upstream progress.
```
0xFelix: +1

veps/sig-compute/207-pauseProbes/pause-probes.md
michalskrivanek: but why is it needed, we should know when the fsfreeze is in progress, so why not pause automatically without messing with annotations like we try in https://github.com/kubevirt/kubevirt/pull/17235 for other cases?

0xFelix: Exactly, that's what I was trying to say.

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: For sure it can call the gRPC method. Easiest way is to document, that fsfreeze operation has higher/lower priority, than annotation value

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: There is a fix pending https://github.com/kubevirt/kubevirt/pull/16777

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: yes, you can consider it as permanent. Documentation should make it clear that this is not for permanent disablement (eventhough we know some users will not listen :) ).

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: I removed the debug section, because the info level for this function is probably enough

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: For sure that is the correct way - and once https://github.com/kubevirt/kubevirt/pull/16777 is merged, we can drop this point completelly.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Goals
+
+- Dynamic pause/unpause of GuestAgentPing probes via VMI annotations (no Pod restart)
+- Internal operations (`fsfreeze`, snapshot) auto-set/clear annotation
+- Maintain strict backward compatibility (Pod probe specs unchanged)
+
+## Non Goals
+
+- Pausing non-GuestAgentPing probes (TCP, UDP, HTTP)
```
enp0s3, line 31: Can you please add a few words in the motivation section why it is not needed in other probe types?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,249 @@
...
+
+kubevirt/kubevirt
+
+## Design
+
+All alternatives below use the same user-facing interface: VMI annotations control probe pausing.
+
+### Why Annotations Instead of a Spec API Field
+
+Pause state is operational and transient, not desired configuration — annotation keeps VMI spec clean.
```
enp0s3: I would argue about that statement. I think that a paused probe is a desired state of the VM. IMO its not an ad-hoc operation like reset or start/stop. Its a permanent state of the VM. Please let me know if I am missing something.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,622 @@
...
+
+### Pausing Probes via Annotation
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-vm
+  annotations:
+    kubevirt.io/probes-paused: "true"  # Probes will return success without executing
```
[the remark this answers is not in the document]
enp0s3: @ksimon1 Actually if we decide to go with the API approach then k8s client is not an issue - you can send SyncVMI with the VMI object to the launcher.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Alternatives
+
+The preferred alternative is number 1 (gRPC Channel Extension — GuestAgentPing only). Alternatives 2–4 are included to compare the full spectrum of options explored. Alternatives 1 and 2 are narrow in scope (GuestAgentPing only) but extremely lightweight. Alternatives 3 and 4 extend the same two approaches (gRPC and file-based, respectively) to cover all VMI probe types by leveraging `virt-probe` as the common entry point.
+
+### Alternative 1: gRPC Channel Extension (virt-handler → virt-launcher) (Preferred) ✅
+
+Extends **existing** gRPC channel with `SetGuestAgentPaused` RPC. Targets **GuestAgentPing only**. No new servers, ConfigMaps, or resources.
+
+**Flow**:
```
enp0s3, line 124: I think that you need to quote the diagram with ``

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: Added

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: fixed

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal enables dynamic pausing of GuestAgentPing probes via VMI annotations to prevent Pod restarts during maintenance (backup, guest OS updates, filesystem freeze). The preferred approach extends the existing gRPC channel with a `SetGuestAgentPaused` RPC. When `kubevirt.io/pause-probes` annotation is set, `virt-handler` notifies `virt-launcher`, which short-circuits `GuestPing` calls to return immediate success. No proxy server, ConfigMap, or new resources required.
+
+## Motivation
+
+KubeVirt probes translate to Pod probes executed by kubelet. **This affects VMs using GuestAgentPing as a liveness probe** — failures cause Pod restart and destroy the running VM (readiness probe failures only affect Service endpoints).
```
0xFelix, line 17: In theory, using other probes than guest agent ping (TCP/HTTP) would result in the same issue if used as liveness probe. But generally speaking, it is not recommended to do so?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+1. **Guest OS Updates / Reboots**: Guest agent unavailable during updates. Windows updates can take hours with multiple reboots — unpredictable duration exceeds any reasonable `failureThreshold`. Requires explicit pause mechanism.
+
+2. **Backup / Snapshot / Filesystem Freeze**: `fsfreeze` can make guest agent unresponsive, causing liveness probe failures. Internal operations (VirtualMachineSnapshot) should auto-pause probes.
+
+**Why GuestAgentPing only?** Keeping the scope narrow to a single probe type minimizes implementation risk and complexity. Alternative approaches covering all probe types are documented below but require significantly more complex implementations (see Alternatives 3 and 4).
```
0xFelix, line 21: GuestAgentPing is also the probe suited best as liveness probe. TCP and HTTP are more suited for readiness probes, as they signal the availability of a specific service.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Goals
+
+- Dynamic pause/unpause of GuestAgentPing probes via VMI annotations (no Pod restart)
+- Internal operations (`fsfreeze`, snapshot) auto-set/clear annotation
```
0xFelix: I still don't think that internal operations should result in an annotation being set on a VMI. That should be left to users only.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+- Pausing non-GuestAgentPing probes (TCP, UDP, HTTP)
+- Replacing Kubernetes probe mechanisms
+- Custom health check logic or complex conditions
+- Multi-VMI control or HTTP endpoints
+
+## User Stories
+
+- Auto-pause probes during internal operations (fsfreeze, snapshot) — no manual intervention
+- Manually pause via annotation before guest OS updates (especially long Windows updates)
+- Query pause status via annotation
```
0xFelix: If set by user, set by internal operation, or both?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Use Cases
+
+**Supported:**
+1 **Guest OS Updates**: Manual annotation before updates (critical for unpredictable Windows updates)
+2. **Filesystem Freeze / Snapshot**: Auto-set annotation during VirtualMachineSnapshot/`fsfreeze`
```
0xFelix: Why set the annotation in this case?

0xFelix: Make virt-launcher log that the probe was disabled should be sufficient?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Pausing Probes via Annotation
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-vm
+  annotations:
+    kubevirt.io/pause-probes: "true"
```
0xFelix: What if internal operation requires it but a users sets this to `false`? Internal operations should have precedence.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+kubectl get vmi my-vm -o jsonpath='{.metadata.annotations.kubevirt\.io/pause-probes}'
+# Output: true
+
+# Remove annotation to unpause probes
+kubectl annotate vmi my-vm kubevirt.io/pause-probes-
+
+# Verify probes are active again (annotation should be absent)
+kubectl get vmi my-vm -o jsonpath='{.metadata.annotations.kubevirt\.io/pause-probes}'
+# Output: (empty - annotation removed)
+```
```
0xFelix: I'd drop this example. It's the exact outcome of the above command when it succeeds.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Relationship to Upstream Kubernetes KEP
+
+[KEP #5002: Introduce hot disable of probes](https://github.com/kubernetes/enhancements/issues/5002) is in **Draft stage** (April 2026). Even if implemented, KubeVirt needs its own VM/VMI ↔ Pod interface since the KEP is Pod-focused. The generic `kubevirt.io/pause-probes` annotation enables future expansion to all probe types when/if the KEP reaches GA, with migration from the current gRPC approach to native Kubernetes API without changing the VMI annotation contract.
+
+## Alternatives
+
+The preferred alternative is number 1 (gRPC Channel Extension — GuestAgentPing only). Alternatives 2–4 are included to compare the full spectrum of options explored. Alternatives 1 and 2 are narrow in scope (GuestAgentPing only) but extremely lightweight. Alternatives 3 and 4 extend the same two approaches (gRPC and file-based, respectively) to cover all VMI probe types by leveraging `virt-probe` as the common entry point.
+
+### Alternative 1: gRPC Channel Extension (virt-handler → virt-launcher) (Preferred) ✅
```
0xFelix: Move it to the design then?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+│                                                               │
+│   ┌─────────────────────────────────────────────────────┐     │
+│   │  VM / QEMU Guest Agent                              │     │
+│   └─────────────────────────────────────────────────────┘     │
+└───────────────────────────────────────────────────────────────┘
+```
+
+**Pros:**
+- Reuses existing gRPC channel (no new infrastructure)
+- Fast propagation (seconds via reconcile loop)
```
0xFelix, line 165: Setting annotations would still cause delays for internal operations?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+| **virt-launcher** | +RPC +atomic.Bool | +os.Stat | +HTTP server +RPC | +HTTP server |
+| **Pod spec** | No change | No change | All → httpGet | All → httpGet |
+| **Overhead/probe** | atomic read | stat syscall | HTTP hop | HTTP hop + stat |
+| **Restart recovery** | Re-apply | Persists | Re-apply | Persists |
+| **New server** | No | No | Yes (:9500) | Yes (:9500) |
+| **Complexity** | Very Low | Very Low | Medium | Medium |
+
+## Update/Rollback Compatibility
+
+- **Upgrade**: Fully backward compatible. Updated virt-handler calls new RPC; old virt-launcher returns "unimplemented" (logged, ignored). Probes work normally.
```
0xFelix: What happens when rolling back to an older version but keeping a launcher that has the probe disabled?

0xFelix (commented):
I still think that internal operations should not be required to set the annotation. It causes overhead and possibly confusion for users.

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: The pause probe functionality is ment not to be permanent. It should be enabled just for the time of update, and then turned off. If user needs to disable probes, then the user should remove it from the spec.
I reworded the explanation to be more accurate.

0xFelix:
See also https://github.com/kubevirt/kubevirt/pull/17235

For live migrations it does what you propose in this VEP (internal operation) without requiring an annotation.

ksimon1:
> See also [kubevirt/kubevirt#17235](https://github.com/kubevirt/kubevirt/pull/17235)
> 
> For live migrations it does what you propose in this VEP (internal operation) without requiring an annotation.

@0xFelix see https://github.com/kubevirt/enhancements/pull/208/changes#diff-082aefbf5fb35234066c2cf7d98508ab6cf59006259732f44e516df9fc7ad5a4R229. Live migration use case was removed almost a month ago

0xFelix:
My point is that it accomplishes the same what this VEP proposes on a high level without using an annotation for an internal operation.

ksimon1:
> My point is that it accomplishes the same what this VEP proposes on a high level without using an annotation for an internal operation.

With live migration, we can determine when the migration begins and when it ends. During an operating system update, we cannot determine this. Therefore, the administrator sets an annotation and deletes it again once the update is complete

0xFelix:
> - Internal operations (`fsfreeze`, snapshot) auto-set/clear annotation

I am referring to internal operations. Your proposal suggests that certain internal operations should trigger setting of the annotation.

ksimon1:
> > * Internal operations (`fsfreeze`, snapshot) auto-set/clear annotation
> 
> I am referring to internal operations. Your proposal suggests that certain internal operations should trigger setting of the annotation.

Yes, for internal operations this solution might be used too.
The fsfreeze is mentioned here, because there is a bug (https://github.com/kubevirt/kubevirt/pull/16777) which can cause probe failure. Hovewer as with the live migration issue, we can predict it will be fixed sooner or later. Then this point will not be relevant for this VEP and I can completelly drop the internal operation section.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+This proposal enables dynamic pausing of GuestAgentPing probes via VMI annotations to prevent Pod restarts during maintenance (backup, guest OS updates, filesystem freeze). The preferred approach extends the existing gRPC channel with a `SetGuestAgentPaused` RPC. When `kubevirt.io/pause-probes` annotation is set, `virt-handler` notifies `virt-launcher`, which short-circuits `GuestPing` calls to return immediate success. No proxy server, ConfigMap, or new resources required.
```
tiraboschi: `kubevirt.io/pause-probes` sounds too generic, we are pausing only `guest-agent-ping` based ones, other probes types are not handled.
`kubevirt.io/pause-guest-agent-probes` maybe?

veps/sig-compute/207-pauseProbes/pause-probes.md
tiraboschi: I absolutely agree on this.
Let's add an example:
- as a user I want to perform a complex guest OS upgrade involving many steps and I don't want to get my VM killed in the middle by the guest-agent liveness probe so I manually set the annotation (step 0)
- exactly because it's a risky operation, I took a snapshot as step 1 but when the snapshot completes, the annotation is automatically cleared
- I start my additional tasks thinking that i'm still protected by the annotation but I'm not and the VM get killed

veps/sig-compute/207-pauseProbes/pause-probes.md
tiraboschi: Could be worth to set a limiting timestamp to let the user pause for a specific amount of time (eventually refreshing it) minimizing the risk to forget about that indefinitely?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+    name: containerdisk
+```
+
+### Dynamically Pausing Probes
+
+```bash
+# Add annotation to pause GuestAgentPing probes on running VMI
+kubectl annotate vmi my-vm kubevirt.io/pause-probes=true
+
+# Check probe pause status by inspecting the VMI annotation
```
tiraboschi, line 100: Is it worth to explicitly reflect the actual probe status in the status of the VMI?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,254 @@
...
+kubectl annotate vmi my-vm kubevirt.io/pause-probes-
+
+# Verify probes are active again (annotation should be absent)
+kubectl get vmi my-vm -o jsonpath='{.metadata.annotations.kubevirt\.io/pause-probes}'
+# Output: (empty - annotation removed)
+```
+
+## Relationship to Upstream Kubernetes KEP
+
+[KEP #5002: Introduce hot disable of probes](https://github.com/kubernetes/enhancements/issues/5002) is in **Draft stage** (April 2026). Even if implemented, KubeVirt needs its own VM/VMI ↔ Pod interface since the KEP is Pod-focused. The generic `kubevirt.io/pause-probes` annotation enables future expansion to all probe types when/if the KEP reaches GA, with migration from the current gRPC approach to native Kubernetes API without changing the VMI annotation contract.
```
tiraboschi: The discussion about that KEP  was on going on https://docs.google.com/document/d/1qdwAXvvyHINf4h_taC7EnDNjsyt-u9wwzJt1oJf97Cs/edit?tab=t.0

They proposed two alternatives:
1. a proper API:
```
livenessProbe:
      +  Suspend: true/false
```
2. an annotation based approach similar to this

That KEP is still under discussion, but aligning to the same interface once that lands is almost mandatory for the sake of consistency.

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: Good catch. I changed the implementation detail to load the information about the annotation from `SyncVirtualMachine` call

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: changed

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: I removed completelly the internal operations and kept only the OS upgrage use case and only the manual user intervention.

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: I would start with an annotation in kubevirt. Once K8S introduces this feature, we can very easily convert it to any solution

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: By user

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: We can add a field if it there will be demand for it

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: Since I dropped the internal operations from this VEP, it should not cause any delay

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,251 @@
...
+
+- Pausing non-GuestAgentPing probes (TCP, UDP, HTTP)
+- Replacing Kubernetes probe mechanisms
+- Custom health check logic or complex conditions
+- Multi-VMI control or HTTP endpoints
+
+## User Stories
+
+- Manually pause via annotation before guest OS updates (especially long Windows updates)
+- Query pause status via annotation
```
0xFelix: How to query if internal operation lead to pause? Should be refined to: Query if paused by user?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Use Cases
+
+**Supported:**
+1. **Guest OS Updates**: Manual annotation before updates (critical for unpredictable Windows updates)
```
0xFelix: Frame it more broadly? To allow a maintenance window in which for whatever reason the guest agent might not respond. Can be Linux too.

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,251 @@
...
+
+**Unsupported:**
+1. Permanent disabling (remove probe from spec instead)
+2. Custom probe responses
+3. Non-GuestAgentPing probes
+4. Auto-set annotation during internal operations (fsfreeze, snapshot, backup, ...)
```
0xFelix, line 50: Internal operations should have precedence over the user annotation though.

0xFelix (commented):
The title of this document is a little off ;)

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,251 @@
...
+**Alpha:**
+- [ ] Annotation pause/unpause works end-to-end
+- [ ] Logging at Info/Debug levels
+- [ ] Unit tests for paused/unpaused paths
+- [ ] E2E: set annotation → success without guest contact → remove → resume
+
+**Beta:**
+- [ ] Guest OS update workflow docs
+
+**GA:**
```
enp0s3, line 244: @ksimon1 Can you please add documentation as requirement for GA?

enp0s3 (commented):
@ksimon1 Hi, I have one comment, besides that LGTM. Regarding the merged migration solution for pause probes, I think we can replace it with the suggested approach once it will be GAed.

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: This vep is not focused on internal operations - this point is in **Unsupported:** section

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: Added

ksimon1:
> @ksimon1 Hi, I have one comment, besides that LGTM. Regarding the merged migration solution for pause probes, I think we can replace it with the suggested approach once it will be GAed.

I think we can implement warning when update is triggered and some VMs are paused

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: updated

ksimon1:
> The title of this document is a little off ;)

fixed :+1:

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: Reworded.

veps/sig-compute/207-pauseProbes/pause-probes.md
0xFelix: There's still an overlap with the existing functionality in the code base, so it should be compatible with that and not break it.

ksimon1: I added the note as one of the goal - https://github.com/kubevirt/enhancements/pull/208/changes#diff-082aefbf5fb35234066c2cf7d98508ab6cf59006259732f44e516df9fc7ad5a4R27

0xFelix (approved):
/lgtm

Thanks!

enp0s3 (commented):
/lgtm

@ksimon1 Thanks!

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,253 @@
...
+```bash
+# Add annotation to pause GuestAgentPing probes on running VMI
+kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes=true
+
+# Check probe pause status by inspecting the VMI annotation
+kubectl get vmi my-vm -o jsonpath='{.metadata.annotations.kubevirt\.io/pause-guest-agent-probes}'
+# Output: true
+
+# Remove annotation to unpause probes
+kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes-
```
geetikakay, line 105: will  kubevirt.io/pause-guest-agent-probes false work ?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,253 @@
...
+  - containerDisk:
+      image: quay.io/containerdisks/fedora
+    name: containerdisk
+```
+
+### Manually Pausing Probes
+
+```bash
+# Add annotation to pause GuestAgentPing probes on running VMI
+kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes=true
```
geetikakay, line 98: will this annotation have same impact on both probes? if yes, for liveness it helps by preventing restarts but on other hand readiness will think it is usable and can route the traffic, which is not desired scenario i think?

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,253 @@
...
+
+kubevirt/kubevirt
+
+### Why Annotations Instead of a Spec API Field
+
+While probe pausing can be viewed as a desired state (e.g., for planned maintenance windows or other OSes), using annotations provides operational flexibility: it allows runtime control without VMI spec changes and maintains backward compatibility. For permanent probe disabling, users should remove the probe from the spec instead of keeping the annotation set indefinitely.
+
+### Observability
+
+- **Info** level: Log state transitions (paused ↔ unpaused) with VMI name
```
geetikakay, line 62: do we want to consider adding them in events or an alert if this annotation persist for long hours?

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: if previously was true, then yes it will enable guest agent probe again

geetikakay: kubectl annotate vmi my-vm kubevirt.io/pause-guest-agent-probes=false --overwrite --this should work to unpause?

veps/sig-compute/207-pauseProbes/pause-probes.md
ksimon1: I was thinking about having alert when upgrade is undergoing, but the warning for long paused VM might be handy too

ksimon1:
@vladikr, @xpivarc, @lyarwood  would you please review this VEP?

vladikr:
@harshitgupta1337 fyi 
I don't think there is a conflict, just fyi

Looks good to me. Thank you @ksimon1
/approve

veps/sig-compute/207-pauseProbes/pause-probes.md
```diff
@@ -0,0 +1,253 @@
+# VEP: Probe Proxy for Manual Probe Control
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
```
iholder101, line 9: @ksimon1 can you please add this section from the VEP template in a follow-up PR?
https://github.com/kubevirt/enhancements/blob/main/veps/NNNN-vep-template/vep.md#target-releases

PR #214, opened 2026-02-19
Copy and adjust the feature lifecycle policy from k/community

iholder101:
### What this PR does

The complete policy in a human readable format:
https://github.com/iholder101/kubevirt-enhancements/blob/docs/feature-lifecycle/docs/feature-lifecycle.md

**Context**

About a year and a half ago, the feature-lifecycle policy was merged to the kubevirt/community repository.
The policy can be found here: https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md.
The PR to introduce it can be found here: https://github.com/kubevirt/community/pull/251.

We have dedicated a lot of energy and time in order to merge this policy. As can be seen, the original PR had **around 400 comments** and took **around 6 months** to merge. If that's not enough, as the PR description says, it was only a follow-up work to a draft version I've made which itself carried many conversations and comments.

The policy was a huge step forward at the time.
However, a lot changed since then. When the proposal was merged, Kubevirt didn't have a structured process for introducing new features other than the vague and now [deprecated design proposal](https://github.com/kubevirt/community/blob/main/design-proposals/README.md). Since then, we have introduced this `kubevirt/enhancements` repository and the VEP process, which was a great step forward, allowing a structured, transparent and inclusive process.

This PR copies the policy to this repository and makes some adjustments to keep it up-to-date according to the VEP process.

**The structure of this PR**

Since the policy was already agreed upon after **huge amount of discussion**, my goal in this PR is to mostly copy it over while making minor changes that we should discuss without re-reviewing the whole policy from scratch.

For this reason, the PR is structured in the following way:
- The first commit (`copy feature-lifecycle.md`) -> copies the policy without any changes.
- The next 3 commits (`remove-*`, `Minor proposal->documentation terminology adjustements`) -> remove out-of-date parts from the original proposal.
- The next commits (`Enhance the feature gate section` and onwards) -> make the actual "interesting" changes.

This structure helps the reviewer focus on **what was changed**, without having to **re-review** the proposal from scratch.

### DIFF
--- a/docs/feature-lifecycle.md
+++ b/docs/feature-lifecycle.md
@@ -303,7 +303,7 @@ there may be scenarios where the policy described here will not be a fit.
 
 Therefore, it should be acceptable to have exceptions from time to time
 given a very good reasoning and an agreement from 2/3 of the project
-maintainers (also known as "approvers").
+maintainers.
 
 ### Miscellaneous
 


iholder101:
I believe I've addressed all comments. Thanks for your reviews @EdDev @jean-edouard!
PTAL

docs/feature-lifecycle.md
```diff
@@ -0,0 +1,320 @@
...
+proven to be actually in use (and useful) before graduating.
+
+> **Note**: Once a feature graduates, it is included in a
+> General Availability (GA) release with its functionality available
+> to all users. GA features need to comply with [semver](https://semver.org/)
+> which add constraints on their ability to change (including deprecation).
+
+### Kubevirt's release cycle
+
+Each Kubevirt release cycle is split to 3 different phases:
```
jean-edouard:
```suggestion
Each Kubevirt release cycle is split into 3 different phases:
```

docs/feature-lifecycle.md
```diff
@@ -0,0 +1,320 @@
...
+
+| Stage          | Period range    | F.Gate | Removal Availability       |
+|----------------|-----------------|--------|----------------------------|
+| Alpha          | 1 to 2 releases | YES    | Between **minor** releases |
+| Beta           | 1 to 3 releases | YES    | Between **minor** releases |
+| GA             | -               | NO     | Between **major** releases |
+
+Through Alpha and Beta feature releases, a FG must be set in order
+for the feature to function.
+By default, no FG is specified, therefore the feature is disabled.
```
jean-edouard: To be updated once Beta FGs are enabled by default

docs/feature-lifecycle.md
```diff
@@ -0,0 +1,320 @@
...
+- Removal: Feature removal involves removing the core functionality
+  of a feature and its exposed API.
+  - The core implementation can be removed in two steps:
+    - The FG is removed by assuring it is never reported as set
+      (i.e. even if it is left by the operator configured, internally
+       it is ignored).
+      At this stage, the core implementation will follow the FG conditions
+      and therefore from the outside the feature is inactive.
+    - In case there are no side effects, the core implementation code can
+      be removed.
```
jean-edouard: What do you mean 2 steps? 2 PRs? In which case I don't really see the point...

docs/feature-lifecycle.md
```diff
@@ -0,0 +1,320 @@
...
+    Kubernetes has not removed fields, it just kept them around with the
+    warning that they have been deprecated and no longer available.
+
+### Exceptions
+While the project strives to maintain a stable contract with its users,
+there may be scenarios where the policy described here will not be a fit.
+
+Therefore, it should be acceptable to have exceptions from time to time
+given a very good reasoning and an agreement from 2/3 of the project
+maintainers (also known as "approvers").
```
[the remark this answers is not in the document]
vladikr: I agree that these are two different roles. There is a disconnect here. Let's not solve it here.

Not all Maintainers are root approvers or any approvers. However, the job is to represent the project, resolve conflicts, work out some of the procedural stuff, etc... (@aburdenthehand help me here)
Some root approvers can and should also be maintainers, but that’s a separate discussion we can have elsewhere.

This enhancements repo works differently (similar to k8s). The approvers here are responsible for overseeing the VEP process, resolving conflicts when they arise, and guiding proposals toward a resolution. It’s less about “deciding the better technical approach” and more about making sure the process itself doesn’t get stuck. I guess in other words, the approvers are here to facilitate VEPs, not just vote on technical preferences.
We should have no problem proceeding as long as no SIG or other KubeVirt approvers raise a strong objection (and historically that has been extremely rare in this repo). Anyone who wants to become an approver here should be ready to handle these kinds of situations.

Getting back to the text itself :)
I think the decision needs to be taken by the people who would also take responsibility in case this goes south.  

In my view, whether to give an extension depends on several factors, but the key one is whether the delay would hurt our ability to stabilize the release after the merge. Part of that responsibility is also helping the author finish the work (and being ready to revert if needed).

Issue #215, opened 2026-02-20
VEP 211: vhost-user backed hotplug volumes


keinsword:
**Primary contact (assignee)**:

/assign @keinsword

**Current Feature Stage**:

New

**Feature Gate**:

`VhostUserHotplugVolumes` (proposed name)

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

- VEP PR: https://github.com/kubevirt/enhancements/pull/TBD
- VEP doc path (planned): `veps/sig-storage/211-vhost-user-hotplug-volumes/vhost-user-hotplug-volumes.md`

**Timeline**:

- Alpha release target (x.y): TBD (proposed: 1.7)
- Beta release target (x.y): TBD (proposed: 1.8)
- Stable release target (x.y): TBD (proposed: 1.9)

- [ ] Alpha
  - [ ] VEP PRs:
    - [ ] https://github.com/kubevirt/enhancements/pull/TBD (initial VEP)
  - [ ] Code PRs:
    - [ ] https://github.com/kubevirt/kubevirt/pull/TBD (API + validation + feature gate)
    - [ ] https://github.com/kubevirt/kubevirt/pull/TBD (converter/manager/virt-handler implementation)
    - [ ] https://github.com/kubevirt/kubevirt/pull/TBD (tests)
  - [ ] Docs PRs:
    - [ ] https://github.com/kubevirt/user-guide/pull/TBD (feature docs)

**Additional context**:

### Motivation
KubeVirt hotplug orchestration exists, but a first-class upstream path for vhost-user-backed hotplug block volumes is missing.

A downstream implementation that we came up with at [Nebius](https://nebius.com/) (developed for [NBS](https://github.com/ydb-platform/nbs)-backed volumes) demonstrated that this is feasible by:
- preserving existing KubeVirt hotplug control flow,
- translating selected hotplug disks to libvirt `vhostuser` disk XML,
- and using directory bind-mount semantics so QEMU reconnects when backend sockets are recreated.

### Problem statement
Current hotplug assumptions are mainly file/block payload oriented. Socket-backed block backends require:
- explicit vhost-user disk source representation,
- correct attach-time translation for hotplug deltas,
- robust socket path propagation across helper pod/backend restarts.

### Proposed scope for upstream
- Introduce first-class API semantics for vhost-user-backed hotplug block disks.
- Gate with a new feature gate until maturity.
[...]

PR #218, opened 2026-02-23
VEP 215: vhost-user backed disks support

keinsword:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/215
**SIG label**: /sig storage

### What this PR does
Introducing VEP for vhost-user backed disks support.

None.

### DIFF
--- a/veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
+++ b/veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
@@ -32,7 +32,7 @@ This enhancement proposes first-class upstream support for PVC-backed vhost-user
 
 This proposal models vhost-user as a `VolumeSource`. The intent is to keep the storage identity and lifecycle at the volume layer, while still presenting the disk to the guest as a regular virtio block device whose dataplane is provided over a Unix socket (for example, `vhost-user-blk`).
 
-Hotplug support and full live migration support are not part of the initial contract proposed here. The first-class outcome targeted by this VEP is a clear API and runtime contract for non-hotplug VM disks backed by a PVC and exposed to QEMU through vhost-user.
+This VEP covers both regular attachment and declarative hotplug for VM disks backed by a PVC and exposed to QEMU through vhost-user. Live migration behavior is expected to follow the same PVC-backed volume semantics KubeVirt already relies on for other storage types.
 
 ## Motivation
 
@@ -53,6 +53,7 @@ Upstream support is needed to:
 - Represent vhost-user as a `VolumeSource`.
 - Map the API to libvirt/QEMU domain configuration in a well-defined way.
 - Introduce validation rules for supported vhost-user disk options.
+- Support declarative hotplug of vhost-user volumes through the existing declarative volume hotplug model.
 - Guard the feature behind a feature gate during Alpha and Beta.
 - Keep non-vhost disk behavior unchanged.
 
@@ -62,8 +63,7 @@ Upstream support is needed to:
 - Requiring a specific sidecar implementation.
 - Redesigning the whole disk subsystem.
 - Supporting standalone vhost-user sources with no PVC/PV behind them.
-- Delivering full live migration orchestration for vhost-user disks in the initial implementation.
-- Supporting hotplugging of such volumes in the first implementation increment.
+- Adding vhost-user-specific live migration orchestration or provider contracts beyond existing PVC-backed volume semantics.
 
 ## Definition of Users
 
@@ -77,6 +77,7 @@ Upstream support is needed to:
 1. As a cluster operator, I want to define a PVC-backed vhost-user disk using an explicit KubeVirt API contract.
 2. As a VM user, I want this disk to appear as a regular virtio disk in guest while backend connectivity is handled through vhost-user.
 3. As a contributor, I want API validation and converter behavior to be explicit and testable.
+4. As an operator, I want the same volume source to work for declarative hotplug so VM spec changes can add or remove vhost-user volumes consistently with other declarative hotplug storage types.
 
 ## Repos
 
@@ -95,21 +96,20 @@ Upstream support is needed to:
 - Keep storage identity and lifecycle at the volume layer.
 - Keep guest-facing disk model aligned with existing virtio expectations.
 - Keep runtime mapping deterministic and observable.
-- Avoid exposing KubeVirt-private mount paths in the user-facing API.
 
 ### Scope split
 
 Primary scope (this VEP):
 
 - General upstream support for PVC-backed vhost-user disks.
-- API representation of vhost-user source, socket location, number of queues, and reconnect timeout.
+- API representation of vhost-user source, socket location, number of queues, reconnect timeout, and declarative hotplug intent.
 - Runtime conversion to libvirt `type='vhostuser'` disk XML.
-- Validation and tests for core non-hotplug behavior.
+- Declarative hotplug support using the existing spec-driven hotplug flow.
+- Validation and tests for both regular and declarative-hotplug behavior.
 
 Secondary scope (follow-on work):
 
-- Hotplug support using the same volume-based API model.
-- Live migration support with explicit target-side backend/socket handoff semantics.
+- Validation of live migration behavior using the existing PVC-backed volume model.
 
 ### Proposed model
 
@@ -119,6 +119,7 @@ The proposed API uses a dedicated `VolumeSource`:
 - `volumes[].vhostUser.socket.path`: path to the Unix socket, relative to the root of the mounted PVC.
 - `volumes[].vhostUser.reconnectTimeoutSeconds`: how long QEMU should keep retrying reconnects.
 - `volumes[].vhostUser.queues`: optional queue count.
+- `volumes[].vhostUser.hotpluggable`: indicates that the volume may participate in declarative hotplug flows.
 
 The matching disk remains a regular virtio disk attachment:
 
@@ -132,6 +133,7 @@ This means the API contract is effectively:
 - socket location relative to volume root: `socket.path`
 - reconnect timeout: `reconnectTimeoutSeconds`
 - optional queue count: `queues`
+- declarative hotplug intent: `hotpluggable`
 
 The runtime mapping should produce:
 
@@ -156,6 +158,7 @@ type VhostUserVolumeSource struct {
     Socket                  VhostUserSocket `json:"socket"`
     Queues                  *uint           `json:"queues,omitempty"`
     ReconnectTimeoutSeconds *uint           `json:"reconnectTimeoutSeconds,omitempty"`
+    Hotpluggable            bool            `json:"hotpluggable,omitempty"`
 }
 
 type VhostUserSocket struct {
@@ -164,6 +167,8 @@ type VhostUserSocket struct {
 ```
 
 The matching disk continues to use existing disk structs. No vhost-user-specific field is added to `Disk`; instead, vhost-user behavior is inferred from the matching volume source and constrained by disk validation rules.
+
+For declarative hotplug, `hotpluggable` follows the same role that it already has for PVC/DataVolume-backed storage: it marks the volume as eligible for spec-driven add/remove handling when `DeclarativeHotplugVolumes` feature gate is enabled.
 
 ### Disk field applicability and validation
 
@@ -184,6 +189,8 @@ Fields that do not apply to vhost-user disks should be rejected by validation ra
 
 If additional disk options are found to be unsupported by the libvirt/QEMU vhost-user path, they should be treated the same way: explicitly documented and rejected during admission.
 
+For declarative hotplug, validation must also ensure that a hotpluggable `vhostUser` volume is PVC-backed, uses a matching `virtio` disk entry, and only participates in the declarative VM-spec-driven hotplug path.
+
 ### Why `VolumeSource`
 
 The design uses `VolumeSource.vhostUser`.
@@ -194,14 +201,17 @@ This aligns with the implementation constraints and intended storage model:
 - the socket path is resolved relative to the mounted PVC root inside `virt-launcher`, and
 - KubeVirt already reasons about storage identity and lifecycle primarily at the volume layer.
 
-This is also a good fit for future work around migration semantics, even though this proposal does not yet implement dedicated live migration support.
+This is also a good fit for migration semantics because it keeps vhost-user disks anchored to the same PVC-backed storage model KubeVirt already uses for other volumes.
 
-### Feature gate
+It is also compatible with the declarative hotplug direction in KubeVirt, because the source of truth remains `spec.volumes` and the hotplug controller can reason about `vhostUser` volumes the same way it already reasons about other PVC-backed hotpluggable sources once the source type is added to the declarative hotplug checks.
 
-Proposed gate name: `VhostUserDisks`.
+### Feature gates
 
-- Alpha/Beta: gate required.
-- GA: gate removed.
+Proposed gate name for this feature: `VhostUserVolumes`.
+
+- Alpha/Beta: `VhostUserVolumes` gate required.
+- Declarative hotplug behavior additionally depends on `DeclarativeHotplugVolumes` while that feature remains gated upstream.
+- GA: `VhostUserVolumes` gate removed.
 
 ### Reconnect semantics
 
@@ -211,11 +221,9 @@ At runtime, reconnect is always enabled in the emitted libvirt XML. The API does
 
 ### Socket path semantics
 
-The API does not expose a pod-visible absolute socket path.
+`socket.path` is interpreted relative to the root of the mounted PVC backing the volume. KubeVirt resolves that relative path to the launcher-visible path internally when rendering the libvirt domain.
 
-Instead, `socket.path` is interpreted relative to the root of the mounted PVC backing the volume. KubeVirt resolves that relative path to the launcher-visible path internally when rendering the libvirt domain.
-
-This avoids leaking KubeVirt-private mount paths such as `/run/kubevirt-private/...` into user manifests.
+Validation must reject absolute paths, symlinks, and path traversal outside the mounted volume root, so the contract remains volume-relative and does not permit references to arbitrary filesystem locations.
 
 ### Shared memory requirement
 
@@ -230,11 +238,19 @@ For VMIs that use vhost-user disks, KubeVirt configures the domain memory backin
 </memoryBacking>
 ```
 
+### Declarative hotplug note
+
+This VEP includes declarative hotplug support for `vhostUser` volumes. The intent is that VM or VMI spec changes can add or remove a hotpluggable `vhostUser` volume using the same declarative reconciliation model as other hotpluggable storage sources.
+
+This requires extending the declarative hotplug source checks to recognize `VolumeSource.vhostUser`, teaching the hotplug controller to treat `claimName` as the backing PVC identity, and carrying the same disk-side validation rules into hotplug reconciliation.
+
 ### Migration note
 
-This proposal does not add dedicated live migration support for vhost-user disks.
+This proposal does not introduce any vhost-user-specific migration API, controller flow, or readiness contract.
 
-It reuses existing volume-based migratability checks, but it does not yet define or implement the target-side backend/socket handoff needed for vhost-user as a fully supported migratable storage type.
+Migration is expected to follow the same rules KubeVirt already applies to PVC-backed volumes. If the backing PVC is considered live-migratable by existing KubeVirt checks, and the storage provider exposes the configured `socket.path` when the volume is mounted in the target launcher pod, then migration should work without additional KubeVirt-side orchestration.
+
+The socket lifecycle remains the responsibility of the storage provider or CSI integration, not KubeVirt. This VEP therefore does not define provider internals for migration; it only relies on the existing expectation that a mounted PVC exposes the data and endpoints required by the workload.
 
 ## API Examples
 
@@ -264,7 +280,36 @@ spec:
       queues: 4
 ```
 
-### Example 2: Resulting libvirt disk intent
+### Example 2: VM with a declaratively hotpluggable vhost-user volume
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: vm-vhost-user-hotplug
+spec:
+  runStrategy: Always
+  template:
+    spec:
+      domain:
+        devices:
+          disks:
+          - name: data0
+            disk:
+              bus: virtio
+      volumes:
+      - name: data0
+        vhostUser:
+          claimName: pvc-data0
+          socket:
+            path: nbs.sock
+          reconnectTimeoutSeconds: 1
+          hotpluggable: true
+```
+
+When `DeclarativeHotplugVolumes` is enabled, adding or removing this volume from the VM template is expected to follow the same declarative reconciliation model used for other hotpluggable storage volumes.
+
+### Example 3: Resulting libvirt disk intent
 
 ```xml
 <disk type='vhostuser' device='disk'>
@@ -286,7 +331,7 @@ spec:
 ### Alternative B: disk-level backend fields
 
 - Pros: smaller API delta and easy to explain as a disk transport override.
-- Cons: storage identity and lifecycle remain anchored elsewhere, which makes the PVC-backed socket contract and future migration semantics harder to express cleanly.
+- Cons: storage identity and lifecycle remain anchored elsewhere, which makes the PVC-backed socket contract and future declarative hotplug and migration semantics harder to express cleanly.
 - Rejected in favor of `VolumeSource.vhostUser`.
 
 ### Alternative C: standalone vhost-user source with no PVC/PV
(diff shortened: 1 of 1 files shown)


vladikr:
@keinsword wouldn't VEP190 #190 help in this case? You could deliver this whole feature as a plugin (@iholder101 fyi)

mhenriks (commented):
Thanks for the VEP @keinsword! I'm sure there will be a very good back and forth on this. Here are some random thoughts on initial glance:

How can we get live migration support?

Would a new VolumeSource provide a better API specifically WRT live migration?

How do these disks interact with existing APIs like snapshot/clone/backup and even tools like velero?

Is hotplug a primary or secondary concern? VEP says it's secondary but goes into quite a bit of detail? What do we want for initial release?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,305 @@
...
+- Add first-class API semantics for vhost-user-backed disks.
+- Map the API to libvirt/QEMU domain configuration in a well-defined way.
+- Introduce validation rules for supported vhost-user disk options.
+- Guard the feature behind a feature gate during Alpha and Beta.
+- Keep non-vhost disk behavior unchanged.
+- Provide a clean path to support vhost-user disk hotplug as a secondary step.
+
+## Non Goals
+
+- Standardizing backend-specific control planes or CSI endpoint lifecycle behavior.
```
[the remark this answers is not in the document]
keinsword: Live migration support is outside of the scope of this VEP's goals. However, it's definitely being considered as an extension in the next iterations.

> live migration would require the socket to be on a RWX filesystem PVC and the csi driver would have to make sure the socket exists on source and target nodes.

You're right here, and this is where we've hit a limitation with the storage provider we're currently using, i.e. it does not allow a given disk to be present in two RWX PVC filesystems simultaneously. But that could easily be worked around once CSI driver that is being shipped by that provider supports the recent enhancements in CSI driver spec that allow for changing volume attributes.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,305 @@
...
+  domain:
+    devices:
+      disks:
+      - name: data0
+        disk:
+          bus: virtio
+        backend:
+          type: vhostUserBlk
+          unixSocket:
+            path: /var/run/kubevirt-private/vhost-user/data0.sock
```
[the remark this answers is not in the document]
keinsword: Good catch, will fix that.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,305 @@
+# VEP #215: vhost-user backed disks support
```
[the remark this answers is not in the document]
keinsword: Fair, forgot to add it. Will be fixed.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,305 @@
...
+  - name: data0
+    persistentVolumeClaim:
+      claimName: pvc-data0
+```
+
+### Example 2: Optional AddVolume request using same backend semantics
+
+```yaml
+apiVersion: subresources.kubevirt.io/v1
+kind: AddVolumeOptions
```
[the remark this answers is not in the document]
keinsword: Will take that into account once hotplug is in scope, thank you!

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,305 @@
...
+
+- Hotplug requests can use the same vhost-user disk semantics as regular disks (bus/backend/reconnect/queues), with API and CLI validation adapted accordingly.
+- `virt-controller` helper pod rendering must preserve mounts for already-attached filesystem hotplug PVCs so helper pod restarts do not silently drop active disks.
+- `virt-handler` must propagate socket-backed disks as directory mounts (not single socket-file bind mounts) to avoid stale inode pinning and allow reconnect after backend socket recreation.
+- `virt-launcher` attach-time conversion must produce `vhostuser` disk XML for hotplug deltas, while `SyncVMI` source identity must include path-based sources to avoid repeated attach loops.
+- Attach-time handling must account for `vhost-user-blk` limitations (for example cache/IOMode/error-policy behavior) consistently with non-hotplug vhost-user disks.
+
+So hotplug should reuse the same first-class vhost-user disk contract, with only hotplug-specific orchestration hardening on top.
+
+## API Examples
```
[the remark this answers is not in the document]
keinsword: Added into the new revision, thank you.

keinsword:
@mhenriks thanks a lot for the review!

Addressed some of your questions in replies to the corresponding review comments. 

But as a short recap here:
> How can we get live migration support?

Reliable support of live migration is outside of the scope of this first increment, but we're definitely aiming for it in the next iterations. Though I do not foresee any major blockers here. 

> Would a new VolumeSource provide a better API specifically WRT live migration?

Indeed, and I've redone the VEP doc to reflect that, especially considering that we aim for live migration support and hotplugging. 

> Is hotplug a primary or secondary concern? VEP says it's secondary but goes into quite a bit of detail?

Again, hotplug is out of scope for now. Though in our downstream it's already implemented and works in production clusters, we don't want to overload this initial increment with additional logic for now. 

And I realize that we're approaching the freeze date of Apr 30th very fast, so do you think this VEP still has a chances to squeeze in?
I'm ready to submit an implementation PR today/tomorrow to support what's been stated in this VEP and that would probably make things easier in terms of review/discussions. WDYT?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+    // ... existing fields omitted
+    VhostUser *VhostUserVolumeSource `json:"vhostUser,omitempty"`
+}
+
+type VhostUserVolumeSource struct {
+    ClaimName               string          `json:"claimName"`
+    Type                    VhostUserDiskType `json:"type,omitempty"`
+    Socket                  VhostUserSocket `json:"socket"`
+    Queues                  *uint           `json:"queues,omitempty"`
+    ReconnectTimeoutSeconds *uint           `json:"reconnectTimeoutSeconds,omitempty"`
```
mhenriks, line 160: `Hotpluggable` field can be added when supported

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+}
+
+type VhostUserSocket struct {
+    Path string `json:"path"`
```
mhenriks, line 165: I think it's probably best to have this configurable but we could possibly support a default like `/disk.socket` or similar

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+### Feature gate
+
+Proposed gate name: `VhostUserDisks`.
+
+- Alpha/Beta: gate required.
+- GA: gate removed.
+
+### Reconnect semantics
+
+The API exposes only `reconnectTimeoutSeconds`.
```
mhenriks, line 218: How are errors/reconnects handled with this disk type? Currently, we detect when a vm is paused because an io error and exposed via a condition. Does that path need to be updated for vhost-user?

Similarly do we know when disconnected? And should we expose that to users?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+
+Instead, `socket.path` is interpreted relative to the root of the mounted PVC backing the volume. KubeVirt resolves that relative path to the launcher-visible path internally when rendering the libvirt domain.
+
+This avoids leaking KubeVirt-private mount paths such as `/run/kubevirt-private/...` into user manifests.
+
+### Shared memory requirement
+
+`vhost-user-blk` requires shared guest memory.
+
+The PoC therefore enables shared `memfd`-backed memory for VMIs that use vhost-user disks, similarly to the existing virtiofs shared-memory handling.
```
mhenriks: Can you give a little more detail here?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
mhenriks: Also need some limitations like no `..` or other symlinks

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
mhenriks: Also is there a default value we use? Or defer to libvirt when empty?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+
+```go
+type VolumeSource struct {
+    // ... existing fields omitted
+    VhostUser *VhostUserVolumeSource `json:"vhostUser,omitempty"`
+}
+
+type VhostUserVolumeSource struct {
+    ClaimName               string          `json:"claimName"`
+    Type                    VhostUserDiskType `json:"type,omitempty"`
```
mhenriks: Is this ever not `blk`? If so, do we need this field at all?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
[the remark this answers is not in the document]
mhenriks: So if PVC is RWX assume live migration support?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `VhostUserDisks` guards all new behavior.
+- [ ] API fields for PVC-backed vhost-user disks are defined and validated.
+- [ ] Converter/runtime support for non-hotplug vhost-user-backed disks is merged.
+- [ ] Unit tests for API + converter behavior are merged.
+- [ ] Initial integration tests for non-hotplug lifecycle are merged.
```
mhenriks: Is there a provisioner available (or some hacks) to test basic functionality in kubevirtci?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
```diff
@@ -0,0 +1,375 @@
...
+
+`vhost-user-blk` requires shared guest memory.
+
+The PoC therefore enables shared `memfd`-backed memory for VMIs that use vhost-user disks, similarly to the existing virtiofs shared-memory handling.
+
+### Migration note
+
+This proposal does not add dedicated live migration support for vhost-user disks.
+
+It reuses existing volume-based migratability checks, but it does not yet define or implement the target-side backend/socket handoff needed for vhost-user as a fully supported migratable storage type.
```
mhenriks: Ultimately we expect the responsibility of managing the socket to be outside of the scope of kubevirt right?

mhenriks (commented):
> And I realize that we're approaching the freeze date of Apr 30th very fast, so do you think this VEP still has a chances to squeeze in?
I'm ready to submit an implementation PR today/tomorrow to support what's been stated in this VEP and that would probably make things easier in terms of review/discussions. WDYT?

@keinsword I commit to make myself available to help you get this in for VEP freeze. I think the api looks pretty good but still have to iron out some details. No need to rush on implementation PR. Very curious to hear how we can test basic functionality in kubevirtci and interactions with other APIs like vmsnapshot/backup

cc @akalenyu @awels

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: > I think it's probably best to have this configurable but we could possibly support a default like /disk.socket or similar

I agree that `socket.path` should remain configurable. The socket name is part of the storage backend contract, so different providers may expose different filenames. We could add a default such as `disk.socket`, but that would effectively define a hidden provider convention. My preference is to keep socket.path required initially, and only add a default later if we see a common convention emerge or if CSI/provider integrations agree on one.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: In the current approach, reconnect behavior is delegated to QEMU/libvirt through the vhost-user reconnect stanza. We do not yet expose a dedicated VMI condition for vhost-user backend disconnected/reconnecting state. Existing IO error detection may still report VM pause conditions if QEMU pauses due to an IO error, but vhost-user disconnects that are being retried may not necessarily surface through that path.
I don’t think we should claim dedicated disconnected/reconnecting observability yet. We may add this as a Beta graduation/task item: test which libvirt/QEMU events are emitted for vhost-user backend disconnect, reconnect, and reconnect timeout, then either confirm existing IO-error handling is sufficient or add a VMI condition/event for vhost-user backend connectivity.

keinsword: > Also is there a default value we use? Or defer to libvirt when empty?

If the user does not specify `reconnectTimeoutSeconds`, the current implementation defaults it to 1.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: Good catch, will mention that in the doc:

`vhost-user-blk` requires the QEMU process and the external vhost-user backend to access the same guest memory. Without shared guest memory, libvirt rejects the domain configuration with an error similar to `'vhostuser' requires shared memory`.

For VMIs that use vhost-user disks, KubeVirt configures the domain memory backing as shared and memfd-backed:

```xml
<memoryBacking>
  <source type='memfd'/>
  <access mode='shared'/>
</memoryBacking>
```

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: Good point. In the current proposal this is only blk, so the field is not carrying useful user intent. I’ll remove it from the initial API and let vhostUser imply vhost-user-blk semantics. If another vhost-user storage variant appears later, we can add a discriminator with a concrete use case.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: Yes. KubeVirt should not manage the socket/backend lifecycle directly. That should be owned by the storage provider/CSI integration. I’ll reword this.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
keinsword: Guys that develop [NBS](https://github.com/ydb-platform/nbs) do have a lightweight setup that might be integrated into kubevirtci.
For Alpha, I think we can cover validation, XML generation, PVC mount wiring, and socket path resolution with unit/functional tests, and add full guest IO e2e in Beta. Would that be acceptable?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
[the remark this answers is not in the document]
keinsword: I would not assume migration support from RWX alone. Future migration support would require a KubeVirt-facing readiness contract that the socket is usable on the target node, but not a standardization of how providers implement that internally. Until that is standardized/tested, this VEP should not mark vhost-user disks as migratable solely based on PVC access mode.

keinsword:
> @keinsword I commit to make myself available to help you get this in for VEP freeze. I think the api looks pretty good but still have to iron out some details. No need to rush on implementation PR. Very curious to hear how we can test basic functionality in kubevirtci and interactions with other APIs like vmsnapshot/backup

@mhenriks thank you, very much appreciate that!

Please find my replies to your comments above. Also, updated the VEP doc according to some of your findings.

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
mhenriks: Can you add a note about path validation?

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
mhenriks: I think this is probably okay but just hit me that I don't see any featuregate mentioned. Let's make sure this feature is gated

veps/sig-storage/215-vhost-user-backed-disks-support/vhost-user-backed-disks-support.md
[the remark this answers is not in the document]
mhenriks: I don't think KubeVirt should need any special readiness contract for vhost-user sockets. The expectation should be the same as any other PVC-backed volume: when the PVC is mounted in the target pod, the socket is available. That's the storage provider's / CSI driver's responsibility, not KubeVirt's.

If the PVC is RWX and the storage provider ensures the socket exists on the target node when the volume is mounted there, migration should just work without any KubeVirt-side orchestration

Maybe there is no provisioner that supports it, but I think KubeVirt should assume the volume is live-migratable simply if it is RWX

mhenriks (commented):
Getting there!

keinsword:
Hi @mhenriks,

Thanks for your review!

So after a discussion with colleagues, given that we're now working towards updating kubevirt in our clusters from 1.1 to 1.6, and the fact that declarative volume hotplug is supported since 1.6, I've decided to have hotplug support as one of the goals of this VEP. Shouldn't be a big deal as we're already doing this through the new PVC-backed volume source. Updated the doc respectively. 

Also, according to your last review:
- Made a note of socket path validation under `Socket path semantics`
- Added integration tests requirement as a Beta graduation criteria, as discussed previously
- Made a note of live migration support according to your last comment on that topic in `Migration note`, and added migration e2e's to Beta as well (though that may probably be a bit ambitious)

Pleas have a look :)

mhenriks:
@keinsword I'm hoping to get more reviews on this as we approach the final hours before VEP freeze. Admittedly, I am more concerned about  testing support now that you are proposing hotplug for alpha. Can you comment on the feasibility of adding a supported provisioner to kubevirtci in that timeframe?

Also, curious if you have looked at https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/190-kubevirt-structured-plugins/vep.md? Does this improve things for you? Obviously a dedicated API would be better but maybe this can help in the meantime?

akalenyu:
> @keinsword I'm hoping to get more reviews on this as we approach the final hours before VEP freeze. Admittedly, I am more concerned about testing support now that you are proposing hotplug for alpha. Can you comment on the feasibility of adding a supported provisioner to kubevirtci in that timeframe?
> 
> Also, curious if you have looked at https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/190-kubevirt-structured-plugins/vep.md? Does this improve things for you? Obviously a dedicated API would be better but maybe this can help in the meantime?

Hey, sorry for jumping in late. Thank you for this detailed proposal (and your contributions in general!)

Couple of things I'm worried about here:
- Maintenance surface - new volume type that is fundamentally shaking up the kubevirt volume structure (single disk.img/blockdev). This is inherently going to be non trivial to integrate with other features (migration came up, hotplug, CBT, ...)
- Testing dependency - bringing NBS (lightweight as may be) is a nontrivial external dependency for CI, both in runtime and code base
- Community demand - right now we know of Nebius (and that is already working there downstream IIUC)

With that, I do think VEP 190 is worth evaluating as the path forward. Once we prove out community demand, we could consider in-tree support. If multiple orgs end up using the plugin and the API stabilizes, that's a much stronger signal for eventual upstreaming.

I do also a feel somewhat confused about the whole matrix of disk image operations wrt to this CSI driver.
For example, how is the data (fedora cloud disk image, for instance) end up on the volume?
These volumes cannot be populated using CDI/volume populator, correct?

What about getting creative to present this volume similarly to other PVCs?
Since the data behind the socket is a real disk, there might be a way to present it with small/no performance penalty to the cri-o container?
Or alternatively, can't the target be exposed as something else and passed through (kubevirt supports this)?
Reference to somewhat similar discussion: https://github.com/kubevirt/kubevirt/issues/12370

These are just some of the things that to me point to this being a deeply vendor-specific storage architecture, not a general-purpose disk transport that belongs in KubeVirt's core API.
I may be wrong though, but that is my impression

btw
/cc @stefanha 
I saw you gave a talk about this a while ago in FOSDEM :pray: Just curious if you have thoughts here

keinsword:
Hi @mhenriks @akalenyu, and thank you both for sparing your time for a meaningful review!

@mhenriks 
> Admittedly, I am more concerned about testing support now that you are proposing hotplug for alpha. Can you comment on the feasibility of adding a supported provisioner to kubevirtci in that timeframe?

The concern is fair: by moving declarative hotplug into Alpha I increased the testing and maintenance surface too much for this stage. Instead, we indeed may re-target that to Beta. For kubevirtci, I do not want to make NBS a required Alpha dependency. I’ll investigate a generic vhost-user-blk backend, ideally [qemu-storage-daemon](https://www.qemu.org/docs/master/tools/qemu-storage-daemon.html), or [SPDK](https://spdk.io/doc/vhost.html), or another small generic backend, for Beta/e2e coverage.

@akalenyu
> I do think VEP 190 is worth evaluating as the path forward.

I agree it is worth evaluating as an interim path. While It could help prove demand and reduce downstream patching, declarative hotplug support would be a tricky part. VEP 190 currently describes a domain hook "after initial DomainSpec generation before XML is handed to libvirt". For declarative hotplug, we'd need either:
- a guarantee that the domain hook also runs for live hotplug disk deltas, or
- an additional hook point like PreDiskAttach / PreHotplugVolumeAttach, or
- a small upstream change that lets plugins participate in disk attach conversion.

I’ll add VEP 190 to Alternatives with that tradeoff.

> Since the data behind the socket is a real disk, there might be a way to present it with small/no performance penalty to the cri-o container?

Could you elaborate a bit more on this one?

> I do also a feel somewhat confused about the whole matrix of disk image operations wrt to this CSI driver.

I’ll add a compatibility matrix covering CDI/import/populator, VM snapshot/backup, CBT/incremental backup/blockjobs, live migration, hotplug, and clone behavior. The short version is that KubeVirt should not claim generic support for features that libvirt/QEMU do not provide for vhost-user disks, so unsupported disk options should be rejected or explicitly documented.

> this being a deeply vendor-specific storage architecture, not a general-purpose disk transport that belongs in KubeVirt's core API

I agree that the NBS backend lifecycle is vendor-specific and should stay outside KubeVirt core. But `vhost-user-blk` itself is not Nebius-specific: it is an upstream QEMU virtio block transport, and libvirt already models it as `disk type='vhostuser'`, so there's nothing there that we as a vendor invented here.
As I pointed out above, we can also test the generic integration with either qemu-storage-daemon or SPDK, instead of requiring NBS in kubevirtci.

/cc @awels
Would love to have your review here as well 🙏

mhenriks:
@keinsword what do you say we iterate on this further and push alpha to 1.10? Maybe we can flush out the testing story and support hotplug for the initial release?

Let's keep the discussion flowing! I'm really interested in making this a reality!

Very good feedback on VEP 190. I encourage you to also share your thoughts with the developers as that feature evolves.

KreanXie:
> Maintenance surface - new volume type that is fundamentally shaking up the kubevirt volume structure (single disk.img/blockdev). This is inherently going to be non trivial to integrate with other features (migration came up, hotplug, CBT, ...)

I agree with @akalenyu’s point that treating a vhost-user socket disk as a dedicated volume type is probably not ideal. Beyond the concerns raised in this discussion, it’s also hard to converge on a standard approach across different storage vendors.

That said, we could add a small amount of built-in support for vhost-user disks—for example, making it possible to mount a vhost-user disk without having to write and maintain a hook sidecar. As for features like live migration, hotplug, and export, those could be supported incrementally as follow-up patches.

I also have a PR (https://github.com/kubevirt/kubevirt/pull/17208) that does something similar, adding export support for vhost-user disks (for reference only).

By the way, I really appreciate this VEP👍，vhost-user is also a pain point in our environment.

keinsword:
@akalenyu @mhenriks 

Thank you so much for the valuable feedback on this one!
After a careful consideration, I'll move forward with this one using VEP #190. I'll work with @iholder101 on hotplug feasibility, as it clearly requires additional node hooks to be implemented.

Issue #222, opened 2026-03-03
VEP 222: VSOCK network namespace confinement


0xFelix:
**Primary contact (assignee)**:

/assign 0xFelix

**Current Feature Stage**: Alpha

**Feature Gate**: `VSOCK`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig network

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 0.59
* Beta release target (x.y): 1.9.0
* Stable release target (x.y):

* [ ] Beta
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/223
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #223, opened 2026-03-03
Introduce VEP 222: VSOCK network namespace confinement

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/pull/222
**SIG label**: /sig compute

### What this PR does

This introduces VEP 222: VSOCK network namespace confinement

### DIFF
--- a/veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
+++ b/veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
@@ -1,0 +1,398 @@
+# VEP #222: VSOCK network namespace confinement
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone /
+release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir
+  in [kubevirt/enhancements](https://github.com/kubevirt/enhancements)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+VSOCK (`AF_VSOCK`) is a socket family for communication between a virtual
+machine and its host, identified by a Context ID (CID) rather than an IP
+address.
+
+Linux 7.0
+introduces [VSOCK network namespace support](https://stefano-garzarella.github.io/posts/2026-02-11-vsock-netns/),
+allowing VSOCK sockets to be confined within network namespaces. This VEP
+proposes leveraging this kernel feature in KubeVirt to isolate each
+`VirtualMachine`'s (VM) VSOCK device within its Pod's network namespace,
+improving security by preventing unauthorized VSOCK access from other
+processes on nodes, including those running in other Pods.
+
+## Motivation
+
+The current VSOCK implementation in KubeVirt has several limitations:
+
+1. **No isolation of VM VSOCK devices**: VSOCK is a host-guest protocol - guests
+   can only connect to CID 2 (the host) and cannot directly reach other guests.
+   However, the host-to-guest (H2G) direction is unconfined: in `global` mode,
+   the `vhost-vsock` kernel module exposes all guest CIDs to the host's entire
+   VSOCK socket layer, without requiring elevated privileges. VSOCK traffic is
+   invisible to Kubernetes NetworkPolicies, which operate on IP-based traffic,
+   so there is no way to restrict this access with standard Kubernetes tooling.
+
+2. **Global CID space**: CIDs must be globally unique across all VMs on a node.
+   The `virt-controller` maintains a CID allocator to prevent collisions, adding
+   complexity with in-memory state. It also complicates cross-cluster live
+   migration, since CIDs must be coordinated between source and target clusters
+   to avoid conflicts.
+
+### Attack surface of unconfined VSOCK
+
+**Pod-to-VM attack**: A compromised or malicious Pod on the node can open an
+`AF_VSOCK` socket without any elevated privileges and scan CIDs starting from 3
+to discover VMs. For each discovered VM, the Pod can connect to VSOCK services
+running inside the guest - including SSH if systemd v256 auto-started `sshd` on
+VSOCK. The VM owner has no way to prevent or even detect the connection.
+
+**VM-to-host attack**: A guest can connect to CID 2 (the host) on any VSOCK
+port. If the host runs services on VSOCK - such as the `sshd` instance that
+systemd v256 starts automatically when `openssh-server` is installed - the
+guest can reach them, enabling lateral movement from VM to node.
+
+Both directions bypass IP-based firewalls and IDS/IPS, since VSOCK traffic
+never touches the IP networking stack. Malware such as
+[BRICKSTORM](https://seclists.org/oss-sec/2025/q4/298) already exploits
+unconfined VSOCK for covert C2 channels on VMware, and the same techniques
+apply to KVM.
+
+To mitigate the lack of isolation, KubeVirt provides mutual TLS authentication
+via a gRPC service (`System.CABundle`) that runs permanently on each node to
+deliver CA certificates to guests.
+
+With kernel-level VSOCK namespace support, the network namespace itself becomes
+the trust boundary, making the global CID allocator unnecessary and providing
+stronger isolation by default. The TLS identity verification remains valuable
+as an additional layer - proving the peer is `virt-handler` - but no longer
+needs to run permanently on the host.
+
+## Goals
+
+- Improve VSOCK security by isolating each VM's VSOCK device within its Pod's
+  network namespace, so that only processes within the same Pod can access it
+- The change should be transparent to users of the KubeVirt VSOCK API - existing
+  workflows continue to work without modification
+- Support both `global` and `local` VSOCK modes during alpha and beta, so
+  clusters that have not yet adopted Linux 7.0 continue to function
+- Warn operators when VSOCK is running in `global` mode (beta onwards) to
+  encourage migration to `local` mode
+- Remove `global` mode support at GA, requiring kernel namespace confinement
+
+## Non Goals
+
+- Supporting cross-network-namespace VSOCK access to VMs
+- Adding new API fields or feature gates - the existing `VSOCK` feature gate
+  is reused
+- Providing a knob to switch between `global` and `local` mode - the mode is
+  determined by the kernel sysctl on nodes
+
+## Definition of Users
+
+- **VM owners**: Users running VMs with VSOCK enabled who benefit from improved
+  isolation without configuration changes
+- **Guest application developers**: Developers of applications inside VMs
+  that communicate over VSOCKs
+- **Cluster administrators**: Operators who deploy and configure KubeVirt on
+  clusters with varying kernel versions
+
+## User Stories
+
+- As a **VM owner**, I want VSOCK connectivity to continue working transparently
+  regardless of whether namespace confinement is active, with no changes to my
+  `VirtualMachine` definitions.
+
+- As a **guest application developer**, I want to expose a service over VSOCK
+  and have external clients connect to it through the KubeVirt API (subject to
+  RBAC), knowing that namespace confinement ensures only authorized access via
+  the API proxy.
+
+- As a **guest application developer**, I want to run a sidecar container
+  alongside my VM that communicates with the guest over VSOCK within the shared
+  Pod network namespace, relying on namespace confinement for isolation. TLS is
+  still advised if mutual authentication between the sidecar and the guest is
+  required.
+
+- As a **cluster administrator**, I want `virt-handler` to clearly log when
+  VSOCK is running in `global` mode on a node, so I can identify nodes that
+  need kernel configuration updates before GA.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Kernel Feature
+
+Linux 7.0 adds VSOCK network namespace support via two sysctls
+([`child_ns_mode`](https://docs.kernel.org/admin-guide/sysctl/net.html#child-ns-mode)
+and [`ns_mode`](https://docs.kernel.org/admin-guide/sysctl/net.html#ns-mode)):
+
+- `child_ns_mode` controls the mode assigned to newly created child network
+  namespaces. It is read when a child namespace is created and is immutable once
+  set to a non-default value:
+  - `global` (default): Children inherit global mode - single CID space, all
+    VSOCK sockets visible across all network namespaces on the host
+  - `local`: Children get isolated VSOCK - sockets can only communicate within
+    the same network namespace
+- `ns_mode` (read-only) reflects the VSOCK mode of the current network namespace
+
+The init namespace always operates in `global` mode - its `ns_mode` cannot be
+changed. Setting `child_ns_mode = local` on the init namespace causes all
+subsequently created child network namespaces to inherit `local` mode. Once a
+child namespace is created, its mode is immutable. A node reboot is required to
+reset `child_ns_mode` back to `global`.
+
+### Change 1: Detect VSOCK mode at `virt-handler` startup
+
+When the VSOCK feature gate is enabled, `virt-handler` reads
+`/proc/sys/net/vsock/child_ns_mode` at startup and registers the
+`devices.kubevirt.io/vhost-vsock` device plugin:
+
+- If the sysctl exists and is set to `local`, `virt-handler` operates in
+  `local` mode.
+- If the sysctl does not exist or is set to `global`, `virt-handler` operates
+  in `global` mode. Starting with beta, this also emits a warning log.
+  Optionally, a metric (e.g. `kubevirt_virt_handler_vsock_global_mode`) can
+  be exposed so operators can set up monitoring alerts for nodes that still
+  need migration.
+- At GA, `virt-handler` will require `local` mode and refuse to register the
+  device plugin when the sysctl is missing or set to `global`.
+
+Configuring the sysctl is the responsibility of the cluster administrator,
+using system tooling such as `systemd-sysctl`, `MachineConfig`, or the node
+tuning operator.
+
+### Change 2: CID allocation
+
+The dynamic CID allocator in `virt-controller` and the `VSOCKCID` field in
+`VirtualMachineInstanceStatus` are kept. Since `virt-controller` operates at
+the cluster level, it cannot distinguish between node modes and must continue
+allocating unique CIDs for all VMs.
+
+On `local`-mode nodes, `virt-launcher` reads its own `ns_mode`, ignores the
+allocated CID, and substitutes a fixed CID (e.g. `3`, the minimum valid guest
+CID) in the libvirt domain XML, since each network namespace has its own CID
+space. On `global`-mode nodes, the allocated CID is used as before.
+
+The dynamic CID allocator can be removed at GA when `global` mode support is
+dropped and all nodes are guaranteed to run in `local` mode.
+
+### Change 3: Namespace-aware VSOCK dialing
+
+With `local` mode, `virt-handler` can no longer dial VSOCK from the host
+network namespace. Instead, it enters the Pod's network namespace before
+dialing. In `global` mode, entering the Pod namespace before dialing is
+harmless - the socket remains reachable. The dial path enters the Pod
+namespace unconditionally and reads the Pod's `ns_mode` to determine the
+CID. This also handles upgrade scenarios where stale Pods retain `global`
+mode on a node that has since been configured for `local` mode:
+
+1. Use the existing `podIsolationDetector` to discover the `virt-launcher`
+   Pod's PID
+2. Enter the Pod's network namespace using the existing `netns.New(pid).Do()`
+   pattern
+3. Read `/proc/sys/net/vsock/ns_mode` inside the namespace. If the Pod's
+   `ns_mode` is `local`, use the fixed CID; if `global`, use
+   `vmi.Status.VSOCKCID`
+4. Call `vsock.Dial(cid, port)` inside the network namespace
+5. Return the connection
+
+This change applies to all VSOCK connections regardless of TLS. For plain
+connections (without `?tls=true`), the flow ends here - the raw VSOCK
+connection is returned. For TLS-authenticated connections, Change 4 adds
+additional steps before the dial.
+
+### Change 4: On-demand VSOCK CA service
+
+The `System.CABundle` gRPC service currently runs permanently on CID 2
+(port 1) in the host network namespace to deliver KubeVirt CA certificates to
+guest agents for mutual TLS verification. In `global` mode, this continues to
+work as before - guests can reach the host namespace service. In `local` mode,
+guests cannot cross the namespace boundary to reach it, so `virt-handler`
+starts the service on-demand inside the Pod's network namespace when a
+TLS-authenticated VSOCK connection is requested.
+
+The permanent service continues to run regardless of the detected mode.
+In `global` mode it serves all guests as before. In `local` mode it is
+unreachable (guests cannot cross the namespace boundary) but harmless - it
+ensures backward compatibility with stale `global`-mode Pods that may remain
+on a node that has since been configured for `local` mode. On-demand listeners
+are not used for `global`-mode Pods because all namespaces share the same
+VSOCK space in that mode, and concurrent listeners on CID 2:port 1 would
+conflict.
+
+When a VSOCK connection request with `?tls=true` arrives from `virt-api`
+for a VM in a `local`-mode namespace:
+
+1. `virt-handler` enters the Pod's network namespace
+2. Starts a temporary `System.CABundle` gRPC listener on CID 2 (the host CID
+   within that namespace)
+3. Dials the guest's VSOCK server and initiates the TLS handshake.
+   `virt-handler` is the TLS client; the guest application is the TLS server.
+   During the handshake, the guest application can connect to CID 2 to fetch
+   the CA bundle and use it to verify `virt-handler`'s client certificate.
+   This happens as part of the TLS negotiation - no explicit synchronization
+   between `virt-handler` and the guest is needed
+4. Tears down the CID 2 listener after the handshake completes or fails
+
+## API Examples
+
+No API changes are required. Existing specifications with
+`AutoattachVSOCK: true` continue to work as before.
+
+## Alternatives
+
+### Remove the TLS infrastructure entirely
+
+Namespace confinement could replace TLS entirely, since only processes within
+the same Pod can reach the VM's VSOCK device. This would simplify the codebase
+but loses mTLS verification that the peer is `virt-handler` rather than a
+malicious process within the Pod - relevant for confidential computing where
+pod-level access should not imply VM-level access. Rejected in favor of making
+the CA service on-demand.
+
+### Enforce `local` mode from `virt-handler`
+
+Have `virt-handler` write `child_ns_mode = local` itself at startup instead of
+requiring external configuration. This would be secure by default but kernel
+sysctl configuration is not `virt-handler`'s responsibility - dedicated node
+configuration tooling (`systemd-sysctl`, `MachineConfig`, node tuning operator)
+is more appropriate for this. Rejected to keep `virt-handler` focused on its
+role and leave system-level tuning to the right tools.
+
+### Drop `VSOCKCID` from the API
+
+The `VSOCKCID` field in `VirtualMachineInstanceStatus` could be removed once
+`global` mode is dropped at GA, since the CID would be fixed. This was
+rejected because removing a field from the API is a breaking change. Keeping
+the field is harmless and preserves API compatibility.
+
+### Require `local` mode from the start
+
+Instead of supporting both modes, `virt-handler` could require `local` mode
+from alpha and refuse to register the device plugin on nodes without support.
+This was rejected because Linux 7.0 - the first kernel with VSOCK namespace
+support - is still very new and not yet widely deployed. Requiring it
+immediately would force all clusters to upgrade their kernel before using
+VSOCK with KubeVirt. A phased approach - support both modes, warn on `global`,
+then remove it at GA - gives operators time to adopt Linux 7.0 while still
+driving migration to the more secure mode.
+
+### New feature gate for namespace confinement
+
+A separate `VSOCKNetNS` feature gate could independently control namespace
+confinement. This was rejected in favor of extending the existing `VSOCK` gate -
+namespace confinement is an improvement to the VSOCK feature, not a separate
+feature.
+
+## Scalability
+
+No scalability concerns. The sysctl is read once per node at `virt-handler`
+startup. The network-namespace-aware dial adds one `setns` syscall per VSOCK
+connection, which is negligible.
+
+## Update/Rollback Compatibility
+
+### Upgrade
+
+No node configuration changes are required for upgrade. `virt-handler` reads
+`/proc/sys/net/vsock/child_ns_mode` at startup and operates in whichever mode
+is configured. Nodes with `local` mode get namespace-isolated VSOCK; nodes
+without the sysctl or with `global` mode continue to work as before. Starting
+with beta, `global`-mode nodes emit a warning.
+
+If old `virt-launcher` Pods remain (depending on `spec.workloadUpdateStrategy`
+in the KubeVirt CR), they retain their inherited mode and dynamically-assigned
+CIDs. The namespace-aware dial (Change 3) handles both modes, since entering
+a global-mode network namespace before dialing is harmless.
+
+### Rollback
+
+Since `virt-handler` does not modify the sysctl, rollback does not require
+node reboots. The old `virt-handler` dials VSOCK from the host network
+namespace, which cannot reach into local-mode namespaces. If `local` mode was
+configured, the administrator must reboot nodes to reset the sysctl back to
+`global` (it is write-once) before the old VSOCK behavior works again.
+
+### GA upgrade
+
+The GA release will remove `global` mode support. Administrators must
+configure `child_ns_mode = local` on all nodes before upgrading to GA. Nodes
+without `local` mode will no longer register the device plugin, and VMs with
+VSOCK enabled will not be scheduled there. This requirement will be
+communicated in release notes for the preceding beta releases.
+
+## Functional Testing Approach
+
+- **Unit tests**: Mock `netns.Do()` and `vsock.Dial()` to verify namespace-aware
+  dialing logic; test sysctl detection with mock filesystem
+- **Functional tests**: VM with VSOCK enabled - verify VSOCK connectivity
+  through the API works end-to-end
+- **Isolation test**: Verify that a process outside a VM's Pod network namespace
+  cannot connect to the VM's VSOCK device when `local` mode is active
+- **Global mode test**: Verify that VSOCK continues to work on nodes without
+  `child_ns_mode` support (global mode fallback)
+- **Existing tests**: Run `tests/vmi_vsock_test.go` to ensure no regressions
+
+## Implementation History
+
+<!--
+To be filled as implementation progresses.
+-->
+
+## Graduation Requirements
+
+### Alpha (v1.9.0)
+
+The VSOCK feature is currently Alpha. This VEP adds namespace confinement
+support while remaining in Alpha:
+
+- [ ] `virt-handler` detects `child_ns_mode` at startup and operates in
+  `local` or `global` mode accordingly
+- [ ] Fixed CID used in `local` mode, dynamically-allocated CID in `global`
+  mode
+- [ ] VSOCK REST proxy enters Pod network namespace before dialing
+  (unconditionally, both modes)
+- [ ] gRPC CA service on-demand per-Pod-namespace in `local` mode; permanent
+  service kept for `global` mode
+- [ ] Unit and functional tests covering both modes
+- [ ] Documentation updated with namespace confinement behavior and kernel
+  requirements
+
+### Beta
+
+- [ ] `virt-handler` emits a warning log when operating in `global` mode
+- [ ] (optional) `virt-handler` exposes a metric for `global` mode, enabling
+  monitoring alerts
+- [ ] Evaluate whether extended support for `global` mode is needed beyond GA
+  (e.g. for clusters unable to adopt Linux 7.0 in the near term)
+- [ ] Evaluate whether guests actively polling the permanent `System.CABundle`
+  service in `global` mode requires extended support beyond GA
+- [ ] Stable across at least one release with no regressions
+- [ ] Documentation updated with deprecation notice for `global` mode
+- [ ] Feature gate enabled by default
+
+### GA
+
+- [ ] `virt-handler` requires `child_ns_mode = local` and does not register
+  the VSOCK device plugin on nodes without support
+- [ ] Dynamic CID allocator removed
+- [ ] Permanent `System.CABundle` gRPC service removed
+- [ ] `global` mode code paths removed
+- [ ] Stable across multiple releases
+- [ ] No reported regressions in VSOCK connectivity


veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,318 @@
...
+
+Linux 7.0 introduces [VSOCK network namespace support](https://stefano-garzarella.github.io/posts/2026-02-11-vsock-netns/)
+, allowing VSOCK sockets to be confined within network namespaces. This VEP 
+proposes leveraging this kernel feature in KubeVirt to isolate each
+`VirtualMachine`'s (VM) VSOCK device within its Pod's network namespace, 
+improving security by preventing cross-VM VSOCK access.
+
+Currently, KubeVirt's VSOCK implementation operates in a global CID space
+where all VMs' VSOCK devices are reachable from any network namespace on the
+host.
```
lyarwood: Can you also reintroduce what VSOCK actually is?

rmohr (changes requested):

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Motivation
+
+The current VSOCK implementation in KubeVirt has several limitations:
+
+1. **No isolation between `VirtualMachines`**: Any process on the host or in any
+   Pod can potentially connect to any VM's VSOCK device using its CID. There is
+   no network-level boundary between VMs.
```
enp0s3: @0xFelix Can you please elaborate more on this limitation? Currently we have TLS-based identity verification, can you please give an example how an arbitrary process can break this mechanism?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,378 @@
...
+
+2. **Global CID space**: CIDs must be globally unique across all VMs on a node.
+   The `virt-controller` maintains a CID allocator to prevent collisions, adding
+   complexity.
+
+3. **TLS-based identity verification**: Because VSOCKs have no network namespace
+   boundary, KubeVirt runs a gRPC service (`System.CABundle`) permanently on
+   each node to provide CA certificates to guests for mutual TLS
+   authentication.
```
enp0s3: @0xFelix Here its not clear to me the limitations this mechanism introduces, can you please give an example?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,318 @@
...
+When the VSOCK feature gate is enabled, `virt-handler` attempts to write `local`
+to `/proc/sys/net/vsock/child_ns_mode` at startup:
+
+- **Success** (including when already set to `local`): Log that local VSOCK
+  namespace mode is active. All Pods created after this point will have VSOCK
+  confined to their network namespace.
+- **Sysctl not found**: Log that the kernel does not support VSOCK namespace
+  confinement. Continue with implicit global mode.
+- **Write failure** (e.g. permissions): Log the error. Continue with global
+  mode.
```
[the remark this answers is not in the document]
enp0s3: @0xFelix @lyarwood Hey folks. Can you please give an example of a potential attack vector that the isolation solves? As I see it if an attacker got privileged access to the node he can access the network namespace as well. For the privileged pod use-case, the attacker can access /dev/vsock from the pod, but for that we have the TLS solution.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,378 @@
...
+
+- As a **guest application developer**, I want to expose a service over VSOCK
+  and have external clients connect to it through the KubeVirt API (subject to
+  RBAC), knowing that namespace confinement ensures only authorized access via
+  the API proxy.
+
+- As a **guest application developer**, I want to run a sidecar container
+  alongside my VM that communicates with the guest over VSOCK within the shared
+  Pod network namespace, without needing TLS-based authentication since the
+  network namespace boundary guarantees isolation.
```
enp0s3: Here I don't understand the use-case. How would the agent inside the guest know that it can trust the sidecar container?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Enhanced the motivation on your request.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Right, TLS is not a limitation!

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: The isolation is stronger, but to be 100% sure you would still need some kind of mutual authentication.

0xFelix:
@enp0s3 Updated the VEP to drop `vsock/child_ns_mode` handling from virt-handler. Per this proposal only if a given node has set the `child_ns_mode` to `local` then `virt-handler` will enable VSOCKs on it going forward.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+- As a **VM owner**, I want VSOCK connectivity to continue working transparently
+  regardless of whether namespace confinement is active, with no changes to my
+  `VirtualMachine` definitions.
+
+- As a **guest application developer**, I want to expose a service over VSOCK
+  and have external clients connect to it through the KubeVirt API (subject to
+  RBAC), knowing that namespace confinement ensures only authorized access via
+  the API proxy.
+
+- As a **guest application developer**, I want to run a sidecar container
```
rthallisey, line 123: This implies a 1:1, client to guest relationship.  What about cases where host daemons require a 1:n relationship?  For example, a host metric service that aggregates metrics from other host pods.

0xFelix: That is indeed a use case that would no longer be possible unless the host daemon enters the Pod's network namespace and advertises its service there. However, I want to emphasize that confined VSOCKs will greatly improve the security of the platform and are well worth this tradeoff. Please see the references I added to this VEP, unconfined VSOCKs are already being actively exploited today.

rthallisey: Ok - there's probably still a way to enabled 1:n securely.  Maybe instead of doing direct service -> guest vsock, the service uses virt-handler as a proxy for the guest vsock.

0xFelix: If you are thinking of a pull model where the host initiates the connection, that is possible with the `vsock` subresource on VMIs.

rthallisey: Ya I'm assuming a pull model.

The issue with using a subresource is it will go through the api server.  While that's part of the appeal (auth) it's also not ideal for control-plane scale.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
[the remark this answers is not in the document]
0xFelix: @enp0s3 Sorry I've missed your comment here, see the references to BRICKSTORM and shai-hulud I've added to the proposal. This will prevent lateral movement between guests, if the host is compromised of course this won't help much.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Agree! It's not ideal for scale, basically the same limitation other subresource APIs like `vnc` and `portforward` are subject to as well. It's definitely a topic to explore in the future.

enp0s3: @rthallisey Hi, I assume that you are referring to the DCGM architecture, what are the implications of switching from VSOCK to TCP mode in your case?

rthallisey: > I assume that you are referring to the DCGM architecture
>
Yes - using the vsock subresource would not work well for the metrics use case, e.g. DCGM exporter.

Vsock provides a way for DCGM expoter to operate as if DCGM was running locally on the host.  This simplifies the guest network requirements, which often involve dedicated networks for TCP connections leaving the guest.

I'm still trying to think of other ways to meet the security requirements while retaining the 1:N use case for metrics...

enp0s3: @rthallisey I think that maybe we should support both modes for Alpha, then default to local with an option to fallback, and GA by removing the global support.

0xFelix: Keeping support for both modes in Alpha/Beta brings a lot of additional complexity. For example the CID allocator lives in virt-controller, so we cannot simply rely on the sysctl setting of individual nodes and I would prefer to avoid adding yet another configurable to KV just to drop / noop it at a later stage during the graduation.

When enforcing `local` mode, the only thing that really breaks is the 1:N use case mentioned by Ryan. Since this feature is still in Alpha, I would deem that acceptable.

0xFelix: I've written up an alternative proposal that keeps support for both modes. PTAL and let me know what you think. I would really love to proceed with local mode only though.

https://github.com/kubevirt/enhancements/blob/0b3cd6dd64ed9f005ee1dd5c1f539ff064110fd9/veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md

0xFelix: /cc @lyarwood

0xFelix:
Okay I need to rewrite this a little, there was a general misunderstanding in which communication directions are possible (VM to VM not possible, but Pod to VM is possible)

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Motivation
+
+The current VSOCK implementation in KubeVirt has several limitations:
+
+1. **No isolation between `VirtualMachines`**: Any process on the host or in any
+   Pod can potentially connect to any VM's VSOCK device using its CID. There is
+   no network-level boundary between VMs. Since VSOCK does not require elevated
+   privileges - any VM owner can enable `AutoattachVSOCK` - a guest can scan the
+   full CID range (nmap-style) to discover other VMs on the same node, connect
```
stefano-garzarella: `vhost-vsock` doesn't allow VM <-> VM communication, so usually inside a guest the CID is handled in this way:
- `>= 3` talk with a nested guest
- `2` talk with the host
- `1` loopback (test only)
- `0` reserved (vmware used this to talk with hypervisor, but we didn't implement this in virtio for now)

That said, there are some virtio-vsock device implementations (vhost-user-vsock, Nitro Enclaves, etc.) that allows VM <-> VM communication, and we also introduced a flag in the address (i.e. `VMADDR_FLAG_TO_HOST`) to support this scenario, but in all the cases I know, the sibling communication is opt-in (and not supported so far by `vhost-vsock` at all to avoid any security issue, firewall, etc.).

So I'd say that VMs can't discover other VMs in any case (before Linux 7.0, or after in `global` netns).

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+With kernel-level VSOCK namespace support, the network namespace itself becomes
+the trust boundary, making the global CID allocator unnecessary and providing
+stronger isolation by default. The TLS identity verification remains valuable
+as an additional layer - proving the peer is `virt-handler` - but no longer
+needs to run permanently on the host.
+
+## Goals
+
+- Improve VSOCK security by isolating each VM's VSOCK device within its Pod's
+  network namespace, so that VMs cannot access each other's VSOCK
```
stefano-garzarella: Ditto about VMs, but this remains true for any other process running in the host. I mean without `local` namespace, any process in the host can access any VMs, which is not true anymore with `local` namespaces.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+VSOCK (`AF_VSOCK`) is a socket family for communication between a virtual
+machine and its host, identified by a Context ID (CID) rather than an IP
+address.
+
+Linux 7.0
+introduces [VSOCK network namespace support](https://stefano-garzarella.github.io/posts/2026-02-11-vsock-netns/),
```
stefano-garzarella, line 29: Not sure if it can be useful to mention, but we added some documentation also in the kernel docs:
- https://docs.kernel.org/admin-guide/sysctl/net.html#ns-mode
- https://docs.kernel.org/admin-guide/sysctl/net.html#child-ns-mode

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+
+Linux 7.0 adds VSOCK network namespace support via the sysctl
+`/proc/sys/net/vsock/child_ns_mode`:
+
+- `global` (default): Legacy behavior - single CID space, all VSOCK sockets
+  visible across all network namespaces on the host
+- `local`: Child network namespaces get isolated VSOCK - sockets can only
+  communicate within the same network namespace
+
+The mode is set on a parent network namespace and inherited by child network
```
stefano-garzarella: Just a note, that I also clarified in the blog post, since we changed it, but merged with 7.0, so it's part of the final release. We added a `write-once` behaviors on `child_ns_mode`, so once it's set, it can't be changed anymore. (see https://lore.kernel.org/netdev/20260223-vsock-ns-write-once-v3-0-c0cde6959923@meta.com/)

This should work fine with this proposal.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,342 @@
...
+namespaces at creation time. Once inherited, the mode is immutable for that
+namespace.
```
stefano-garzarella: The mode of a network namespace is defined by `/proc/sys/net/vsock/ns_mode` (read-only). I honestly don't know kubevirt architecture, so not sure if it can be useful or not, but since I didn't see it mentioned, I'd like to clarify.

stefano-garzarella (commented):
@0xFelix I left some comments from the AF_VSOCK perspective. Hope they can help ;-)

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Updated to reflect that, thanks for the input!

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Thanks! Yes, I kept that in mind.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Setting this on the host in the init netns (is it called like that?) would mean all child nses created on that host would be in local mode. Is that correct? That's what I was aiming for.

stefano-garzarella: Nope, `ns_mode` can't be written in any netns, including `initns`. It's just the sysctl that allows a process to understand the mode of the netns in which it is running on.

To do what you mentioned, you just need to set `child_ns_mode` in `initns` (e.g. by systemd).

This also means that `initns` will be always `global`, but this is not an issue, since if you set `child_ns_mode` of `initns` to `local`, all the children will be `local`, so at the end, even if `initns` is `global`, all the other `netns` will be local, so can't communicate with it.

0xFelix: Yes, sorry, that's what I've meant. Setting the `child_ns_mode` of `initns` to `local`.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,359 @@
...
+to discover VMs. For each discovered VM, the Pod can connect to VSOCK services
+running inside the guest - including SSH if systemd v256 auto-started `sshd` on
+VSOCK. The VM owner has no way to prevent or even detect the connection.
+
+**VM-to-host attack**: A guest can connect to CID 2 (the host) on any VSOCK
+port. If the host runs services on VSOCK - such as the `sshd` instance that
+systemd v256 starts automatically when `openssh-server` is installed - the
+guest can reach them, enabling lateral movement from VM to node.
+
+Both directions are invisible to standard monitoring tools (`ss`/`netstat`,
```
stefano-garzarella: AFAIK `ss --vsock` is supported since `iproute2-4.15.0` (released in 2018), not sure about the other tools TBH.

stefano-garzarella (commented):
LGTM (for the AF_VSOCK perspective), left just a small comment about `ss` supporting AF_VSOCK, but I don't think it is so relevant, you got the point.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
stefano-garzarella: Yeah! This comment was just to point out that there is another sysctl (`ns_mode`) available, that maybe can be used to double-check that settings are the one expected inside a pod.

e.g. the daemon want to be sure that in the `netns` just created, where the VM will run, the `ns_mode` is set `local` before using vsock, otherwise rise an error. But again, I don't know the kubevirt architecture at all.

0xFelix: Makes sense, we can add such a check to `virt-handler`. It fits the KubeVirt architecture. Thanks!

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Thanks, I changed the wording a little.

0xFelix:
@stefano-garzarella Thank you very much, very valuable feedback!

enp0s3:
/lgtm

@0xFelix Thank you!

enp0s3:
@vladikr @xpivarc Can you please have a look?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+as an additional layer - proving the peer is `virt-handler` - but no longer
+needs to run permanently on the host.
+
+## Goals
+
+- Improve VSOCK security by isolating each VM's VSOCK device within its Pod's
+  network namespace, so that only processes within the same Pod can access it
+- The change should be transparent to users of the KubeVirt VSOCK API - existing
+  workflows continue to work without modification
+- Support both `global` and `local` VSOCK modes during alpha and beta, so
```
xpivarc, line 89: Just feedback. This could have been summarized here as "Provide seamless transition to new mode" and then the graduation criteria could be more specific.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+  RBAC), knowing that namespace confinement ensures only authorized access via
+  the API proxy.
+
+- As a **guest application developer**, I want to run a sidecar container
+  alongside my VM that communicates with the guest over VSOCK within the shared
+  Pod network namespace, relying on namespace confinement for isolation. TLS is
+  still advised if mutual authentication between the sidecar and the guest is
+  required.
+
+- As a **cluster administrator**, I want `virt-handler` to clearly log when
```
xpivarc, line 129: Wouldn't metric be better

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+
+When the VSOCK feature gate is enabled, `virt-handler` reads
+`/proc/sys/net/vsock/child_ns_mode` at startup and registers the
+`devices.kubevirt.io/vhost-vsock` device plugin:
+
+- If the sysctl exists and is set to `local`, `virt-handler` operates in
+  `local` mode.
+- If the sysctl does not exist or is set to `global`, `virt-handler` operates
+  in `global` mode. Starting with beta, this also emits a warning log.
+  Optionally, a metric (e.g. `kubevirt_virt_handler_vsock_global_mode`) can
```
xpivarc, line 170: @machadovilaca Is this metric ok?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+  be exposed so operators can set up monitoring alerts for nodes that still
+  need migration.
+- At GA, `virt-handler` will require `local` mode and refuse to register the
+  device plugin when the sysctl is missing or set to `global`.
+
+Configuring the sysctl is the responsibility of the cluster administrator,
```
xpivarc, line 176: +1

xpivarc, line 176: Is there a way for to have only the Pods to be in `local` namespace?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+tuning operator.
+
+### Change 2: CID allocation
+
+The dynamic CID allocator in `virt-controller` and the `VSOCKCID` field in
+`VirtualMachineInstanceStatus` are kept. Since `virt-controller` operates at
+the cluster level, it cannot distinguish between node modes and must continue
+allocating unique CIDs for all VMs.
+
+On `local`-mode nodes, `virt-launcher` reads its own `ns_mode`, ignores the
```
xpivarc, line 187: What is the reason to ignore it ?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+allocated CID, and substitutes a fixed CID (e.g. `3`, the minimum valid guest
+CID) in the libvirt domain XML, since each network namespace has its own CID
+space. On `global`-mode nodes, the allocated CID is used as before.
+
+The dynamic CID allocator can be removed at GA when `global` mode support is
```
xpivarc, line 192: @rmohr Do you see reason for the `global` mode to stay for extended period?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+1. `virt-handler` enters the Pod's network namespace
+2. Starts a temporary `System.CABundle` gRPC listener on CID 2 (the host CID
+   within that namespace)
+3. Dials the guest's VSOCK server and initiates the TLS handshake.
+   `virt-handler` is the TLS client; the guest application is the TLS server.
+   During the handshake, the guest application can connect to CID 2 to fetch
+   the CA bundle and use it to verify `virt-handler`'s client certificate.
+   This happens as part of the TLS negotiation - no explicit synchronization
+   between `virt-handler` and the guest is needed
+4. Tears down the CID 2 listener after the handshake completes or fails
```
xpivarc, line 251: Don't we need the service all the time?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+  monitoring alerts
+- [ ] Stable across at least one release with no regressions
+- [ ] Documentation updated with deprecation notice for `global` mode
+- [ ] Feature gate enabled by default
+
+### GA
+
+- [ ] `virt-handler` requires `child_ns_mode = local` and does not register
+  the VSOCK device plugin on nodes without support
+- [ ] Dynamic CID allocator removed
```
xpivarc, line 394: I would move this after GA

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
...
+- [ ] `virt-handler` emits a warning log when operating in `global` mode
+- [ ] (optional) `virt-handler` exposes a metric for `global` mode, enabling
+  monitoring alerts
+- [ ] Stable across at least one release with no regressions
+- [ ] Documentation updated with deprecation notice for `global` mode
+- [ ] Feature gate enabled by default
+
+### GA
+
+- [ ] `virt-handler` requires `child_ns_mode = local` and does not register
```
xpivarc, line 392: This basically means we force KubeVirt users to consume 7.x Kernel line, right?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,319 @@
...
+   PID
+2. Enter the Pod's network namespace using the existing `netns.New(pid).Do()`
+   pattern
+3. Read the CID from `vmi.Status.VSOCKCID` if set (upgrade case), otherwise
+   use the fixed CID
+4. Call `vsock.Dial(cid, port)` inside the network namespace
+5. Return the connection - the socket FD remains valid after leaving the network
+   namespace
+
+### Change 4: Remove VSOCK gRPC CA service
```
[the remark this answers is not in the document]
xpivarc: @rmohr Was this addressed?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,319 @@
...
+
+1. **No isolation between `VirtualMachines`**: Any process on the host or in any
+   Pod can potentially connect to any VM's VSOCK device using its CID. There is
+   no network-level boundary between VMs.
+
+2. **Global CID space**: CIDs must be globally unique across all VMs on a node.
+   The `virt-controller` maintains a CID allocator to prevent collisions, adding
+   complexity.
+
+3. **TLS-based identity verification**: Because VSOCKs have no network namespace
```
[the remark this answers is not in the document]
xpivarc: @rmohr What about rotating the CA in case of breach/leak?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: @rthallisey Any thoughts, do you want make it requirement for Beta/GA?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: I proposed an optional metric as well

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: If I got your question right, that is achieved by setting the `child_ns_mode` of the initns to `local`. It will lead to the network namespaces of all Pods to have `ns_mode = local`.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: To eventually get rid of the complexity of having to handle individual CIDs. We could use the CID that is assigned in `VSOCKCID` instead, but once we GA this, I'd like to get rid of it. In `local` mode we can always assign a static CID, e.g. `3` to the VM.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: @enp0s3's and my idea was to keep the support for `global` mid-term, to allow users adopting Linux kernels that support this new feature.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Once the application has fetched the CA, it can cache it for the lifetime of the incoming request and therefore does not need the service anymore. By tearing down the service we can free up some resources, otherwise listeners would accumulate fast, since we would have one extra listener per each Pod.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: It means we force users to consume a kernel that supports the VSOCK `child_ns_mode`. It will be backported to CentOS Stream 9 that uses a 5.x kernel for example.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Once we GA and no longer support global mode, there is no need for the CID allocator anymore. Why keep it then?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: I am not sure how many people do look at logs, I would prefer metrics + alerts

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: Is there any other way to achieve this? Like could there be optin by container runtime?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: What about CA rotation?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
rthallisey: >Remove global mode support at GA, requiring kernel namespace confinement
>
I'm still unsure.  IMO, using `global` is a valid use case but it's missing some security hardening.  So I don't think `global` being unsupported is ideal.  Is there a way we can support that use case?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: You have to set `child_ns_mode` in the parent network namespace that is used to create the Pod network namespaces. This can achieved in many ways. In theory the container runtime could it too, but setting it on the initns seems like the simplest to me.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: On the next call to the `vsock` subresource, the CA service would be started again and the application could cache the new CA. That's what I meant by lifetime of an incoming request.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: I'd like to keep the wording like it is right now, but before we move to GA we could still change the graduation requirements to retain support for the `global` mode if absolutely required. I suggest we think about the 1:n use-case more once we have support for both modes, so we can explore other solutions more easily.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc It seems that it cannot be configured per namespace because of the inheritance and the write-once mechanism. IOW while in child namespace the runtime cannot change the mode from local to global.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc It should be a gauge metric, naming can be polished later on.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc Since metrics were already mentioned, is there anything missing here?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc Why should we keep it? Unless you mean that you want to support global mode in GA. Can you please elaborate on your plan?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc Proactive termination upon CA rotation was missing even before the VEP. We aim to preserve the current functionality while transitioning to the confined mode. We can fix this behavior but it can be done independently. One way or another, this VEP makes sure to bring up the CA service before any new dial to the guest.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @xpivarc Is there any added value of keep using the allocated CID in the confined local mode?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
rmohr: +1 to @enp0s3 , the current solution works under the premise that you would fetch the cert on every new connection and implementations would make use of it that way ( see the reference imlementation in our repo).

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
rmohr: +1 to @enp0s3 as well, either we keep global then keep this as well, or remove. But whatever we do, not filling the API field is fine, removing it from the API should probably follow the api deprectaion policy timelines.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
rmohr: I guess moving by default to the most secure "local" mode and explicitly opting in to global may be a way. 1:n is really useful (altough we don't use it right now, so not an issue for us). With the built in TLS negotiation or doing your own TLS independent of virt-handler it is as secure/insecure as normal TLS over normal network.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
rmohr: I think @rthallisey brought up some cases and I am sure some people would find use-cases. We have tests and the code, so it may be ok to keep, but I have no too strong opinion for as long as  the switch to local provides the same identity guarantees identifying virt-handler caller via TLS.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
```diff
@@ -0,0 +1,394 @@
+# VEP #222: VSOCK network namespace confinement
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
```
xpivarc, line 7: Please update the issue as it is talking about Beta... but you correctly state here Alpha

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: I understand that I am just being cautious as users could observe the service is always there and could be actively pooling the service. @0xFelix make this Beta requirement to evaluate if this is needed.

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: @0xFelix make this Beta/GA requirement to evaluate if we need extended support. For me the force to use such new Kernel seems a good reason to keep it around...

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: @stefano-garzarella could you comment?

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
xpivarc: +1 will depend on the need to keep the old way around ...

xpivarc (commented):
/approve
I believe we are good for Alpha
/hold
The comments can be addressed as follow-up, feel free to remove hold

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Issue updated

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Added this to the graduation requirements

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
0xFelix: Added requirement

0xFelix:
@xpivarc Thanks! Added more requirements during beta.

/hold cancel

enp0s3 (approved):

veps/sig-compute/222-vsock-netns-vep/vsock-netns-vep.md
enp0s3: @stefano-garzarella Hi, speaking from container runtime perspective, is it possible to solve the global to local transition on a container creation phase by using double unsharing and intermediate namespace (Rewriting the the mode to local in the intermediate namespace)?

stefano-garzarella: @enp0s3 IIUC you would like to do something like what I did in the blog post examples https://stefano-garzarella.github.io/posts/2026-02-11-vsock-netns/#local-mode

> Local mode
> 
> Setting child_ns_mode to local enables isolation. Since this sysctl is write-once, we use a nested unshare to avoid locking the init_netns mode:
> 
> $ unshare --user --map-root-user --net bash -c \
>     'echo local > /proc/sys/net/vsock/child_ns_mode && \
>      unshare --net nc --vsock -l 1234' &
> $ nc --vsock 1 1234
> Ncat: Connection reset by peer.

@xpivarc @enp0s3 Does that answer your question?

xpivarc: @stefano-garzarella Thanks, so if I got it right there could be mismatch between node and container.

stefano-garzarella: @xpivarc I'm not really into k8s, so I don't know how a node is configured, but if the init process of the node (i.e. systemd) when starting does `echo local > /proc/sys/net/vsock/child_ns_mode` in the init_ns, then for sure all the child netns (so all containers), will be local and can't go back to global. (note, this can be done also later after the boot, once is done in init_ns, for sure all the new netns will be local).

If this is not done, then yes, there could be a mismatch.

0xFelix: There can be a mismatch for containers created before setting the mode to local, yes.

Issue #224, opened 2026-03-06
VEP 223: Metadata Service Enablement


shengnuo:
**Primary contact (assignee)**:

/assign shengnuo

**Current Feature Stage**: Alpha

**Feature Gate**: `MDS`

**Responsible SIGs**:

Primary SIG:
/sig network

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): __TBD__
* Beta release target (x.y): __TBD__
* Stable release target (x.y): 

* [ ] Alpha
  * [ ] VEP PRs: 
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #225, opened 2026-03-06
VEP 224: Metadata Service (MDS) Enablement for KubeVirt VMs

shengnuo:
### VEP Metadata

**Tracking issue**: #224 
**SIG label**: /sig network /sig compute

### What this PR does
Enabling Metadata Service for KubeVirt VMs through veth pair and nftable plumbing in virt-launcher network namespace.

### DIFF
--- a/veps/sig-network/mds-enablement/mds-enablement.md
+++ b/veps/sig-network/mds-enablement/mds-enablement.md
@@ -1,0 +1,247 @@
+# VEP 224: Metadata Service (MDS) Enablement for KubeVirt VMIs
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+This proposal introduces an opt-in mechanism for exposing a link-local Metadata Service (MDS) endpoint to a KubeVirt guest.
+
+The proposition suggests introducing a new [network binding plugin](https://kubevirt.io/user-guide/network/network_binding_plugins/) whose CNI component configures the virt-launcher pod network namespace during pod network setup (for example attaching the metadata address and installing redirection rules so traffic to the well-known link-local IP, default `169.254.169.254/32`, reaches a metadata listener in the pod). An accompanying sidecar image registered on the KubeVirt CR can run the metadata HTTP service.
+This document has two focuses:
+
+a. **Network plumbing** required in the virt-launcher pod namespace to make an MDS endpoint reachable from the guest in a consistent way, across different cluster overlay/network deployments and KubeVirt interface bindings.
+
+b. **Metadata service** exposure in the virt-launcher pod (typically a sidecar) that responds to MDS requests.
+
+## Motivation
+Users running VM workloads on Kubernetes frequently want a cloud-like metadata endpoint for:
+
+- Instance identity and lifecycle signals
+- Bootstrap configuration that must be fetched at runtime (as opposed to static cloud-init)
+- In-cluster services that expect the conventional metadata IP (`169.254.169.254`)
+
+## Goals
+- Provide a mechanism to expose a link-local MDS endpoint using a **network binding plugin** (NAD + CNI + optional sidecar), suitable for a [predefined data path](#supported-use-cases-initial).
+- Provide a mechanism for developers to bring their own metadata service implementations (sidecar image on the binding and/or optional container spec on the VMI).
+
+## Non Goals
+- Define the metadata schema/content served by MDS.
+- Standardize an authentication/authorization model for metadata content.
+- Support arbitrary user-defined privileged containers in virt-launcher pods.
+- Provide IPv6 MDS in the initial phase (future work).
+
+## Definition of Users
+- A user is a person that wants a guest to access a metadata endpoint over a conventional link-local IP.
+- An admin is a person who manages the cluster, defines the platform-level resources and controls the security constraints for KubeVirt VMs.
+- A developer is a person who wants to build a metadata provider/service to run alongside KubeVirt.
+
+## User Stories
+- As a user, I want my guest OS to retrieve its own workload identity metadata from a well-known link-local IP (e.g. `169.254.169.254`).
+- As a developer, I want to plug in my own metadata service implementation that runs as a sidecar in the virt-launcher pod.
+- As a developer, I want metadata service enablement to be explicit.
+- As an admin, I want my VMs to remain in private subnet (no public internet, no cluster API access), while allowing the guest to reach the metadata service.
+- As an admin, I want the MDS packet steering to be constrained so only guest-originated traffic can reach it.
+
+## Use Cases
+
+### Supported Use cases (initial)
+
+1. VMI declares a dedicated network using a registered MDS network binding (Multus `networkName` pointing at an NAD whose plugin chain includes the MDS CNI), with guest networking configured so metadata traffic uses that interface.
+2. Guest consuming metadata via link-local IP `169.254.169.254` on the path the MDS CNI configures (initial implementations may scope to **TCP** to selected ports).
+3. Custom metadata service via sidecar image configured on the KubeVirt CR for that binding.
+
+### Unsupported Use cases (initial)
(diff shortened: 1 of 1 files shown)


Acedus (commented):
Thank you for the in-depth proposal @shengnuo! I think that while technically sound, KubeVirt already has a way to enable something like this in a more streamlined manner. 

https://kubevirt.io/user-guide/network/network_binding_plugins/ - KubeVirt network binding plugins are modular extensions that define exactly how a VM's vNIC is wired through its host pod to the guest OS, as such, it allows for some interesting usecases and I believe this may be one of them.

What we can do is create an MDS CNI plugin that intercepts and routes requests to the MDS address. We can attach the address to the pod's loopback if and create a DNAT rule for any TCP traffic hitting that address to be redirected to localhost. 

As for the metadata service itself, it'll be a simple sidecar plugin that will be specified to be used for that network binding plugin under the kubevirt CR configuration, so maybe something like:
```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
spec:
  configuration:
    network:
      binding:
        metadata-service:
          migration:
            method: link-refresh
          networkAttachmentDefinition: default/kubevirt-metadata-service
          sidecarImage: registry.example/network-metadata-service
```

And the relevant NAD:
```yaml
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: kubevirt-metadata-service.
  namespace: default
spec:
  config: '{ "cniVersion": "1.0.0", "name": "kubevirt-metadata-binding", "plugins": [ { "type": "kubevirt-metadata-binding" } ] }'
```

Then requesting MDS for a VM becomes as simple as:
```yaml
  template:
    spec:
      interfaces:
        - name: default
          masquerade: {}
        - name: metadata-service
          binding:
            name: metadata-service
      networks:
        - name: default
          pod: {}
        - name: metadata-service
          multus:
            networkName: kubevirt-metadata-binding
```

This also has the benefit of developing this VEP completely out-of-tree.

WDYT?

Acedus:
/cc @EdDev @orelmisan 

Your input would be greatly appreciated here.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,351 @@
...
+This document has two focuses
+a. **network plumbing** required in the virt-launcher pod namespace to make an MDS endpoint reachable
+from the guest in a consistent way, across different cluster overlay/network deployments and KubeVirt interface bindings.
+
+b. **metadata service** injection in virt-launcher pod that responds to the MDS requests.
+
+## Motivation
+Users running VM workloads on Kubernetes frequently want a cloud-like metadata endpoint for:
+
+- Instance identity and lifecycle signals
```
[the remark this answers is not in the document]
EdDev: cloud-init is a specific implementation to pass data to guests. I can only guess that MDS is another "protocol" with its own implementation.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,351 @@
...
+- Bootstrap configuration that must be fetched at runtime (as opposed to static cloud-init)
+- In-cluster services that expect the conventional metadata IP (`169.254.169.254`)
```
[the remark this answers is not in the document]
EdDev: @shengnuo , I suggest you provide a ref (link) to MDS, which explains the need.
It will clarify where this needs is coming from.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Goals
+- Provide a mechanism to expose a link-local MDS endpoint using a **network binding plugin** (NAD + CNI + optional sidecar), suitable for a [predefined data path](#supported-use-cases-initial).
```
EdDev, line 28: The Network Binding Plugin is an implementation detail that solved the need introduced in this VEP.
The goals section should try and avoid expressing the implementation part, keeping the need & requirement clean and not locked to a specific solution.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+- Provide a mechanism for developers to bring their own metadata service implementations (sidecar image on the binding and/or optional container spec on the VMI).
+
+## Non Goals
+- Define the metadata schema/content served by MDS.
+- Standardize an authentication/authorization model for metadata content.
+- Support arbitrary user-defined privileged containers in virt-launcher pods.
+- Provide IPv6 MDS in the initial phase (future work).
```
EdDev, line 35: For a general purpose functionality, I think such a limitation may need to be solved by the GA stage.
However, I think this proposal does not require any core Kubevirt changes, therefore, the feature lifecycle may not apply here at all.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Definition of Users
+- A user is a person that wants a guest to access a metadata endpoint over a conventional link-local IP.
```
EdDev, line 38: The user is not defined by his needs, but by its role. E.g. a "guest user" can be an application running in the guest.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+- An admin is a person who manages the cluster, defines the platform-level resources and controls the security constraints for KubeVirt VMs.
```
EdDev, line 39: If you refer to a cluster admin, then please name it as such.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+- A developer is a person who wants to build a metadata provider/service to run alongside KubeVirt.
```
EdDev, line 40: I would use a name like "MDS Author" or "MDS Developer" to provide a clear name. Using a name like "dveloper" is too generic.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Where the logic runs
+
+- **MDS CNI** (binding plugin): configures the pod network namespace during CNI `ADD`/`DEL` on the node.
+- **Sidecar** (binding `sidecarImage`): serves metadata; may participate in DomainXML mutation per binding design.
+- **virt-controller**: merges launcher pod spec (Multus annotations, sidecar container from binding registration).
+
+
+## Virt-controller changes
+For VMIs using the MDS binding: ensure launcher pod annotations and sidecar containers match the **KubeVirt CR** binding registration (NAD, `sidecarImage`, optional
```
EdDev, line 185: I do not understand what is expected to be changed from the existing supported plugin binding.
Please clarify what does not work today and/or missing for you suggestion to work.

veps/sig-network/mds-enablement/mds-enablement.md
```diff
@@ -0,0 +1,247 @@
...
+- **virt-controller**: merges launcher pod spec (Multus annotations, sidecar container from binding registration).
+
+
+## Virt-controller changes
+For VMIs using the MDS binding: ensure launcher pod annotations and sidecar containers match the **KubeVirt CR** binding registration (NAD, `sidecarImage`, optional 
+overhead).
+
+## Status and conditions
+
+Add a VMI condition (name TBD) indicating MDS plumbing readiness, for example:
```
EdDev, line 190: The network binding plugin mechanism was introduced to extend Kubevirt capabilities without requiring core kubevirt changes.

This request is unlikely to be possible at this time. The only way to do this is to create an external controller to update the VMI status.

EdDev (commented):
Thank you for this VEP.

As this VEP is described, it seems like a valid extension to Kubevirt with its pluggable mechanism (network binding plugin).
That means that there is no need to change anything in Kubevirt to enable this capability and work on it can happen out-of-tree to the core project.

If you have a prototype which you will be interested to add to the Kubevirt org, we can discuss its addition as a self contained repository.

If I missed something specific that you think we need to enhance in the Kubevirt project or its ecosystem, please focus us.

For now, I find this VEP a place for brainstorming and discussion, without a need to target it to a specific Kubevirt release.
If you think differently, please let us know.

Issue #229, opened 2026-03-15
[Meta-VEP] VEP 229: Beta features on by default


iholder101:
**Primary contact (assignee)**:
/assign @iholder101 

**Feature Gate**:
Meta-VEP - no feature gate.

**Responsible SIGs**:
Cross-SIG meta-VEP.

**Enhancement link**:
https://github.com/kubevirt/enhancements/blob/main/veps/meta-VEPs/229-beta-features-on-by-default/vep.md.

**Timeline**:
Targeted for immediate effectiveness in v1.9.

* [X] VEP PRs: https://github.com/kubevirt/enhancements/pull/230.
* [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/17405.
* [x] Docs PRs:
	* Blog post: https://github.com/kubevirt/kubevirt.github.io/pull/1027.
	* VEP Template addition: https://github.com/kubevirt/enhancements/pull/273.
	* User-guide: https://github.com/kubevirt/user-guide/pull/995.
	* Add the feature-gate report to release: https://github.com/kubevirt/kubevirt/pull/18376.

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
This one is complete! 🥳 

/close

Issue #16, opened 2025-03-09
VEP 12: Use ImageVolume to mount OCI images as volumes in virt-launcher


Barakmor1:
**Primary contact (assignee)**: @Barakmor1 

/assign @Barakmor1 

**Current Feature Stage**: New.

**Feature Gate**: ImageVolume

**Responsible SIGs**: sig-compute

/sig sig-comp

**Enhancement link**:

**Timeline**:

* [] Alpha (v1.6.0)
* [] Beta (v1.7.0?)
  * [] ImageVolume is GAed in Kubernetes
  * [] Enable feature gate by default
* [] GA (v1.8.0?)
  * [] Remove all code for the old method in virt-handler, virt-controller, and virt-launcher, but continue supporting the old virt-launchers in virt-handler.
* [] Post GA (v1.10.0?)
  * [] Remove any remaining code for the old method in virt-handler once it is safe from upgrade perspective.

**Additional context**:

PoC:
https://github.com/kubevirt/kubevirt/pull/13656

aglitke:
@akalenyu  Can you take a look at this from the sig-storage perspective?

vladikr:
@Barakmor1 Could you please update whether we should track this VEP for 1.7 release

Barakmor1:
Currently, there’s only one remaining requirement on our side for Beta. The other requirements depend on the Beta and GA versions of the ImageVolume feature gate in Kubernetes. Here's the missing piece on our side: https://github.com/kubevirt/kubevirt/pull/14973.

However, I think this raises an important discussion. According to [this KEP](https://github.com/kubernetes/enhancements/pull/5450), ImageVolume is expected to GA in Kubernetes 1.35.

We defined our own Beta requirement to be: once ImageVolume is GA in Kubernetes and the above PR merges, we can mark it as Beta in KubeVirt and enable it by default.

What worries me is the scenario where users run the latest KubeVirt on Kubernetes 1.34. If we enable ImageVolume by default, VMs will crash because the feature is still disabled by default in Kubernetes 1.34.

I'm not sure if this kind of situation was discussed before. Should we support compatibility with older Kubernetes versions where the feature gate is not enabled by default? Or is this an acceptable limitation?

FYI @jean-edouard @iholder101 @akalenyu @xpivarc

iholder101:
Any plans for v1.9?

Barakmor1:
> Any plans for v1.9?

[...]

PR #230, opened 2026-03-15
meta-VEP 229: beta features on by default

iholder101:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/229.
**SIG label**: a cross-sig VEP.

### What this PR does
Introduce a meta-VEP proposing that beta feature gates be enabled by default starting from v1.9, aligning with the Kubernetes KEP process and ensuring wider testing and feedback before GA graduation.

### DIFF
--- a/veps/meta-VEPs/229-beta-features-on-by-default/vep.md
+++ b/veps/meta-VEPs/229-beta-features-on-by-default/vep.md
@@ -1,0 +1,173 @@
+# VEP #229: Beta features on by default
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- ~[ ] (R) Target version is explicitly mentioned and approved~
+- ~[ ] (R) Graduation criteria filled~
+
+## Overview
+
+This is a meta-VEP to enable beta features by default.
+
+The term meta-VEP means this VEP is not for a new feature nor an API change.
+Even more so, little to no code changes are required for this VEP.
+However, after some discussions, it was agreed that this is a good platform to discuss these changes in a transparent
+and public form.
+
+## Motivation
+
+The Beta phase of a feature acts as a "dress rehearsal" before going to GA.
+In this phase the community needs to ensure the feature is stable,
+gain wide user-feedback,
+ensure the API remains stable, etc.
+
+By enabling beta feature gates by default, we "eat our own dog food", and achieve the following:
+- Every beta feature will be tested by our CI with every entered PR.
+- Every beta feature will be tested by each of our developers by default during development.
+- Different beta features will be tested with one another.
+
+_Note: only the feature gate will be enabled by default.
+If a feature is disabled by default via configuration, this is out of scope for this vep._
+
+This will give us "out-of-the-box" wide testing and user feedback,
+turning our contributors to early beta testers by default.
+It also aligns with the KEP process in Kubernetes.
+
+Downstream vendors will obviously have the ability to decide which features are being enabled or disabled in production.
+See below for more information.
+
+## Goals
+
+- Beta features would be enabled by default (see specific policy below).
+- Help graduate features by enabling them for a larger audience earlier in the process.
+- Downstream vendors would be able to list immature yet enabled features, and easily disable those they do not like.
+
+## Non Goals
+
+- Discuss the requirements of graduation to beta.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/229-beta-features-on-by-default/vep.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order for vendors to know the state of KubeVirt's feature gates without having to parse its codebase,
+which is fragile and cumbersome,
+KubeVirt will include a metadata file as part of its release artifacts that will reflect the state of all of its feature gates.
+
+This metadata file will be auto-generated via a simple script which will parse KubeVirt's [active.go file](https://github.com/kubevirt/kubevirt/blob/02d3ea105dd61960f754cd8da2721fe9dd8ff00d/pkg/virt-config/featuregate/active.go).
+I think the specific format can be discussed in the relevant implementation PR,
+but just to give a general idea, it can look something similar to:
+
+```json
```
dankenigsberg: This file is an API that kubevirt promises to keep. Every consumer (e.g HCO) would have to write a (small) piece of code to parse it and use it.  Would you consider creating an importable library that can be vendored by HCO, with a function such as `GetFeatureGates()` which returns this tructure?

0xFelix (commented):
Added these yesterday already and they might be outdated by now, but thanks to GitHub they got stuck pending.

veps/sig-compute/229-beta-features-on-by-default/vep.md
```diff
@@ -0,0 +1,179 @@
...
+
+This means that whenever a VEP owner thinks that an exceptional configuration is required,
+e.g. enabling a beta feature by default on production environments which deploy KubeVirt by using HCO,
+it is the VEP owner's responsibility to add a PR to HCO which adds it to the list of exceptionally configured features.
+
+#### HCO Json-Patches
+
+If a downstream vendor's desired feature gate configuration differs from HCO's defaults        
+and the HCO maintainers decline to accommodate it,                                          
+the vendor can still use JSON patches to override the KubeVirt CR's `DisabledFeatureGates` list directly.
```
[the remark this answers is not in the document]
alaypatel07: I agree with AI tools it becomes even more simple, it will be fair to ask admins to do this.

veps/sig-compute/229-beta-features-on-by-default/vep.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Design
+
+### All Beta features becoming on-by-default
+
+From v1.9 onwards, **every** beta feature gate will be on by default.
+
+This will not be configurable, but auto-enabled for all Beta features with no technical ability to disable by default.
+See [Downstream control](#downstream-control) below for more details regarding how to enable these in production.
+
```
Barakmor1: That would work well as long as we don’t have a beta feature gate that depends on an alpha feature gate, which I believe we don’t.

I think it’s also worth mentioning, here or elsewhere, that for a feature gate to move to beta and be enabled by default, it should work out of the box on the last three Kubernetes versions without requiring special configuration, such as enabling a dependent feature gate that is disabled by default or any other additional settings. Otherwise, we would need an exception, for example [ImageVolume is already beta](https://github.com/kubevirt/enhancements/issues/16) but won’t work on 1.34(still supported) out of the box.

veps/sig-compute/229-beta-features-on-by-default/vep.md
```diff
@@ -0,0 +1,172 @@
...
+HCO would have the liberty to decide which features will be enabled/disabled by default when KubeVirt is deployed via HCO.
+
+This means that whenever a VEP owner thinks that an exceptional configuration is required,
+e.g. enabling a beta feature by default on production environments which deploy KubeVirt by using HCO,
+it is the VEP owner's responsibility to add a PR to HCO which adds it to the list of exceptionally configured features.
+
+#### HCO Json-Patches
+
+If a downstream vendor's desired feature gate configuration differs from HCO's defaults        
+and the HCO maintainers decline to accommodate it,
```
Barakmor1: I might be missing something, but would it make sense to include these in a separate VEP instead?

EdDev (commented):
Thank you.

I am good with making (most) Beta FGs enabled by default, I just ask to be less aggressive and allow exceptions, same as Kubernetes do. See inline thread for more info.

/assign

veps/sig-compute/229-beta-features-on-by-default/vep.md
iholder101: Discussed in another comment above.

Issue #231, opened 2026-03-15
VEP 231: Add VirtualMachine snapshot schedules


SkalaNetworks:
**Primary contact (assignee)**:

/assign @SkalaNetworks

**Current Feature Stage**:

New

**Feature Gate**:

VMSnapshotSchedule

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): TBD
* Beta release target (x.y): TBD
* Stable release target (x.y): TBD

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/232
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/16339/ (Sidenote, this is the original author, but no more activity for about 3 month, so I'll take over with my own PR shortly. It will use the same code.)
  * [ ] Docs PRs: TBD

> [!IMPORTANT]
> Please keep this description up to date.

PR #232, opened 2026-03-15
feat(vep): add virtualmachinesnapshotschedule

SkalaNetworks:
### VEP Metadata

**Tracking issue**: [231](https://github.com/kubevirt/enhancements/issues/231)
**SIG label**: 
/sig compute
/sig storage

### What this PR does
The snapshot API supports doing a singleshot snapshot of a VirtualMachine.
Often times, it might be interesting to put the VM snapshot process on a schedule (e.g everyday at midnight).

Introduce a CRD of type VirtualMachineSnapshotSchedule. The controller would create a new VirtualMachineSnapshot according to the schedule defined in the schedule (CRON expression).

t also handles GCing the snapshots when they get out of the window of the schedule (32nd snapshot gets deleted if retention is set to 31 snapshots)

### DIFF
--- a/veps/sig-compute/vm-snapshot-schedules.md
+++ b/veps/sig-compute/vm-snapshot-schedules.md
@@ -1,0 +1,198 @@
+# VEP #231: Introduce VirtualMachine snapshot schedules
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+<!--
+Provide a brief overview of the topic)
+-->
+
+The Snapshot API allows for point-in-time snapshots of a VM. It still needs a manual intervention to run the snapshot.
+
+The goal of this VEP is to introduce a new way of scheduling snapshots on a VMs, at regular intervals.
+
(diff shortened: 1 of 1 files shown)


akalenyu (commented):
Real preliminary review. Just wondering - could we maybe piggy back on something like https://github.com/backube/snapscheduler instead of having a core kubevirt implementation?

veps/sig-compute/vm-snapshot-schedules.md
```diff
@@ -0,0 +1,198 @@
...
+      mylabel: myvalue
+```
+
+## Alternatives
+
+<!--
+Outline any alternative designs that have been considered)
+-->
+
+- Have an external project that handles scheduled snapshots through an operator
```
awels: My question is, why was this rejected? An external project seems perfect for this scenario?

akalenyu:
Summarizing sig-storage; we tend to think this makes more sense outside of kubevirt core
(but could definitely be under kubevirt org).

SkalaNetworks:
I'll work on 2 things:
- See if we can make snapscheduler more "generic" (issue to maintainers to see if they are ok with that)
- See if we should make a dedicated operator for this

If option 1 is a go, I'll just propose an edit to the Kubevirt docs to point to this project as a possibility for scheduling snapshots

If option 2 is a go, I'll port the existing PR as an operator.

SkalaNetworks:
/remove-lifecycle stale

Will try to present my work on a SIG meeting once I've got it fully tested in production

Project is at 
https://github.com/super-phenix/vmsnapshot-scheduler

Issue #235, opened 2026-03-18
VEP 235: Support for VM Network Bandwidth Limiting


Aneesh-Hegde:
**Primary contact (assignee)**:

/assign @lyarwood 

**Current Feature Stage**:
Alpha (v1.6)

**Feature Gate**:
NetworkBandwidthLimiting

**Responsible SIGs**:

Primary SIG:
/sig network

Additional SIGs (optional):
/sig compute

**Enhancement link**:
https://github.com/kubevirt/kubevirt/pull/17213

**Timeline**:

* Alpha release target (1.6):
* Beta release target (1.7):
* Stable release target (1.8):

* [x] Alpha
  * [ ] VEP PRs:
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/17213
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

Issue #62, opened 2025-06-24
VEP 62: Generalized Migration Priority in KubeVirt


vladikr:
**Primary contact (assignee)**:

/assign $ASSIGNEE

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig $SIG

Additional SIGs (optional):
/sig $SIG

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/73

**Timeline**:

* Alpha release target (x.y): 1.7
* Beta release target (x.y): 1.8
* Stable release target (x.y): 1.9

* [X ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/73
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

fossedihelm:
@vladikr Can you update the description adding the Code PR? Thanks

fossedihelm:
Proposing this for Beta graduation in release-1.8

enp0s3:
@fossedihelm Hi, can you please update the GA section and mention the VEP PR?

fossedihelm:
@enp0s3 I cannot :)  @vladikr can you? Thanks

vladikr:
/remove-lifecycle stale
@fossedihelm what do I need to do? :)

fossedihelm:
@vladikr This is the good part :)
/completed
/close

PR #236, opened 2026-03-18
feat(vep): Add network bandwidth limit for vm

Aneesh-Hegde:
### VEP Metadata

**Tracking issue**: #235  
**SIG label**: 
/sig network
/sig compute

Code PR: https://github.com/kubevirt/kubevirt/pull/17213

### What this PR does

This VEP proposes adding Network Quality of Service (QoS) to KubeVirt to prevent "noisy neighbor" scenarios where a single VM saturates the host's network. It introduces a bandwidth field to the VMI interface specification, allowing users to set average, peak, and burst limits for both inbound and outbound traffic using standard Kubernetes units (e.g., 1Gi).

The technical design involves translating these quantities into Libvirt’s XML configuration via the virt-launcher, supported by a safety utility that prevents integer overflows.

### DIFF
--- a/veps/sig-network/vm-network-bandwidth-limiting.md
+++ b/veps/sig-network/vm-network-bandwidth-limiting.md
@@ -1,0 +1,124 @@
+# VEP #235: VM Network Bandwidth Limiting
+
+## Release Signoff Checklist
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [x] (R) Target version is explicitly mentioned and approved
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version: TBD
+- This VEP targets GA for version: TBD
+
+## Overview
+This VEP proposes adding Network Quality of Service (QoS) support to KubeVirt by enabling bandwidth limiting for Virtual Machine (VM) secondary network interfaces managed by Multus. This allows users to specify bandwidth limits in the VirtualMachine template spec, which `virt-controller` translates into Multus network-selection annotations (`k8s.v1.cni.cncf.io/networks`) on the `virt-launcher` pod, leveraging standard Kubernetes CNI chaining to enforce limits at the pod boundary.
+
+## Motivation
+Currently, KubeVirt VMs attached to secondary networks can consume the full available bandwidth of the host's physical network interface. While administrators can manually configure CNI bandwidth plugins inside Multus `NetworkAttachmentDefinition` (NAD) objects, doing so hardcodes the limit for all VMs using that specific NAD. There is a requirement for a native KubeVirt API to define these limits on a per-VM basis, allowing for dynamic QoS tiers (e.g., via KubeVirt `VirtualMachineInstancetypes`) without proliferating duplicate NADs across the cluster.
+
+## Goals
+- Enable per-interface traffic shaping for secondary VM networks.
+- Provide a native Kubernetes-style API using standard resource units for network QoS.
+- Act as a control-plane abstraction layer: translating KubeVirt API intent into standard Multus CNI JSON annotations.
+- Ensure limits survive Live Migration natively by persisting the intent in the VMI spec.
+
+## Non Goals
+- **Primary Network (Masquerade) Support:** Due to live-migration IP reassignment limitations and the lack of a NAD for the default network, shaping the primary network is out of scope for the Alpha phase.
+- **Dynamic "live" updating:** Because CNI plugins only execute at pod creation (the CNI ADD phase), users cannot dynamically update the bandwidth of a running VM without a restart during the Alpha phase.
+- **Direct Libvirt/tc manipulation:** This VEP strictly relies on the Kubernetes-native Bandwidth CNI plugin at the pod boundary, rather than configuring `tc` rules directly on the tap device via Libvirt.
+
+## Definition of Users
+- **Cluster Administrators**: Responsible for node health and resource fair-share.
+- **Cloud Service Providers**: Delivering tiered networking performance levels to end-users via `Instancetypes`.
+- **Users**: End-users deploying VMs who need predictable network throughput.
+
+## Design
+
+### API Change
+The implementation introduces a `Bandwidth` field within the existing `Interface` struct. The API aligns exactly with the standard Kubernetes Bandwidth CNI plugin capabilities, accepting `Rate` and `Burst` parameters for both ingress and egress:
+
+* **Rate**: The desired average bandwidth rate (e.g., `100M`, `1G`).
+* **Burst**: The maximum amount of data that can be sent in a single spike before the `Rate` limit is enforced.
+
+### Conversion to Multus Annotations
+The `virt-controller` will act as an abstraction layer, injecting the API configuration directly into the pod's Multus annotation. 
+
+When a VM defines a bandwidth limit on a secondary interface, `virt-controller` will translate the parameters and append the bandwidth arguments to the JSON payload of the `k8s.v1.cni.cncf.io/networks` annotation on the `virt-launcher` pod before the pod boots.
+
+### Validation
+To ensure system stability and predictable behavior, the KubeVirt API validation webhook will enforce the following:
+1. **Primary Network Rejection:** The webhook will explicitly reject the `bandwidth` configuration if it is applied to the default pod network (e.g., `masquerade` without a Multus NAD), as primary network shaping is currently unsupported.
+2. **Standard Formatting:** Ensure the rate and burst values conform to standard Kubernetes resource quantity formats and that burst values are mathematically sound relative to the rate.
+
+### Feature Gate
+Initial implementation is considered Alpha. A dedicated `NetworkBandwidthLimiting` feature gate is required to be enabled in the KubeVirt CR to use this API during the experimental phase.
+
+## Scalability
+This feature introduces negligible overhead to the control plane. The `virt-controller` translation occurs only during pod creation. At the host level, standard CNI bandwidth plugins utilize Linux `tc` token bucket filters, which are highly optimized. 
+
+## Update/Rollback Compatibility
+As this is a new API field guarded by a feature gate, there are no backward compatibility issues. Disabling the feature gate will cause the API to ignore the `bandwidth` block on new VM creations. Existing running VMs with applied bandwidth limits will maintain their CNI constraints until restarted.
+
+## Testing
+* **Unit Tests**: API validation webhook logic (ensuring primary networks are rejected) and `virt-controller` annotation generation (ensuring the Multus JSON array is constructed and merged correctly).
+* **E2E Tests**: Spawning a VM with a secondary Multus network and the feature gate enabled, applying a bandwidth limit, and verifying that the `virt-launcher` pod receives the correct Multus annotations and the CNI bandwidth plugin executes successfully.
+
+## Architecture Flow
+```mermaid
+flowchart TD
+    U([User / Instancetype]) --> |"1. Apply VM Spec (limit: 100M)"| API[KubeVirt API Webhook]
+    
+    API --> |"2. Validate & Reject if Primary<br>3. Accept VM Object"| VC[virt-controller]
+    
+    VC --> |"4. Read Base NAD & Extract Bandwidth<br>5. Inject Multus JSON Annotation"| VL[virt-launcher Pod]
+    
+    %% Visual representation of the injected JSON
+    JSON_NOTE["k8s.v1.cni.cncf.io/networks: <br> [{'name': 'base-nad', 'bandwidth': {'ingressRate': '100M'}}]"] -.-> VL
+    style JSON_NOTE fill:#f9f2f4,stroke:#d04437,stroke-width:1px,stroke-dasharray: 5 5
+    
+    VL --> |"6. Pod Boot Sequence"| M((Multus CNI))
+    
+    M --> |"7. Apply Base Network"| B[Base CNI]
+    B --> |"8. Enforce QoS"| BW[Bandwidth CNI Plugin]
+```
+
+## API Examples
+```yaml
+kind: VirtualMachine
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          interfaces:
+          - name: secondary-net
+            bridge: {}
+            bandwidth:
+              ingressRate: "100M"
+              ingressBurst: "10M"
+              egressRate: "50M"
+              egressBurst: "5M"
+      networks:
+      - name: secondary-net
+        multus:
+          networkName: my-bridge-nad
+```
+## Translated Pod Annotation (virt-controller output)
+```yaml
+metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/networks: |
+      [
+        {
+          "name": "my-bridge-nad",
+          "namespace": "default",
+          "bandwidth": {
+            "ingressRate": "100M",
+            "ingressBurst": "10M",
+            "egressRate": "50M",
+            "egressBurst": "5M"
+          }
+        }
+      ]
+```


Aseeef (commented):
Also, please squash your commits into a single commit. Maintainers like to have a clean commit history in the main branch.

I.e. they prefer a final commit history thay looks something like:

a766e92: added vep #3 proposing bla bla
bla.
c772e92: added vep #9 proposing bah bah bah.

f766e92: added vep #62 proposing ba ba ba.

NOT:
a767392: added vep #3 proposing bla bla
bla.
a767392: fixed typo in vep #3
78aabbd: addressed reviewer in vep #3
8aaa55d: addressed reviewer again in vep #3
c770092: added vep #9 proposing bah bah bah.
7018362: fixed typo

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
+# VEP #235: VM Network Bandwidth Limiting
+
+## Release Signoff Checklist
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Target version is explicitly mentioned and approved (Target: v1.6)
```
frenzyfriday: Did you mean `Target: v1.9`? Or if you add the Target releases section you dont have to mention the target here explicitly

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
...
+
+## User Stories
+- As a **cluster administrator**, I want to limit a non-critical VM's egress traffic so that production-critical services on the same node maintain guaranteed network throughput.
+- As a **Cloud Service Provider**, I want to cap the network bandwidth of "Basic Tier" VMs to 100Mbps so that I can differentiate them from "Premium Tier" offerings.
+- As a **user**, I want to define my VM's network speed in standard units (e.g., `1Gi`) within the VMI manifest, without interacting with host-level networking tools.
+
+## Repos
+- `kubevirt/kubevirt`
+- `kubevirt/api`
+- `kubevirt/user-guide`
```
frenzyfriday: Just `kubevirt/kubevirt` is sufficient

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Limitations
+* **Outbound Peak**: Per Libvirt documentation, the `peak` attribute in the `outbound` element is currently ignored because underlying Linux ingress filters perform policing (dropping) rather than classful queueing. `burst` remains fully functional as it defines the token bucket size for the policing filter.
+* **API Validation**: To prevent user confusion regarding the Outbound Peak limitation, the KubeVirt API validation webhook and OpenAPI schema will strictly **reject** any VMI creation or update request that attempts to define a `peak` value within the `outbound` bandwidth configuration.
+* **Network Plugin Independence (SR-IOV / VFs)**: Because this feature relies on applying Linux `tc` rules to the host's `tap` device, it requires a software-defined backend (like `masquerade` or `bridge`). If an interface binds a Virtual Function (VF) directly to the VMI via SR-IOV or hardware passthrough, the traffic bypasses the host's `tc` layer entirely, meaning these bandwidth limits will be ignored.
```
frenzyfriday: You can keep this limitation here as well as add in the Design/Validations section as one of the things to validate

frenzyfriday (changes requested):
Great work, thanks! Please consider the following review.
Also, please update the commit message to something like
```
VEP 235: Support for VM Network Bandwidth Limiting

This VEP proposes adding Network Quality of Service (QoS) to KubeVirt to prevent 
....
safety utility that prevents integer overflows.

Signed...com>
```

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
+# VEP #235: VM Network Bandwidth Limiting
+
```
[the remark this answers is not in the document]
frenzyfriday: looks good now, thanks!

veps/sig-network/vm-network-bandwidth-limiting.md
frenzyfriday: looks good now, thanks!

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
...
+## Release Signoff Checklist
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Target version is explicitly mentioned and approved (Target: v1.6)
+- [ ] (R) Graduation criteria filled
+
+## Overview
+This VEP proposes adding Network Quality of Service (QoS) support to KubeVirt. It enables users to define maximum bandwidth consumption for Virtual Machine Instance (VMI) interfaces, mapping these limits directly to the underlying Libvirt domain configuration via the `virt-launcher`.
```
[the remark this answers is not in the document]
frenzyfriday: looks good now, thanks!

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Non Goals
+- Support for dynamic "live" updating of bandwidth without a VMI restart during the Alpha phase.
+- Implementation of CNI-level shaping (this focuses strictly on the `virt-launcher` / Libvirt layer).
+- Support for Explicit Congestion Notification (ECN) marking to signal network congestion, as this is currently unsupported within the bandwidth QoS configuration of the Libvirt Domain XML API.
+
+## Definition of Users
+- **Cluster Administrators**: Responsible for node health and resource fair-share.
+- **Cloud Service Providers**: Delivering tiered networking performance levels to end-users.
+- **Users**: End-users or application owners deploying VMIs.
```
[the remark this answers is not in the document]
frenzyfriday: Makes sense. Maybe in that case we can keep the `Definition of Users` simpler by removing the Cluster admins? But anyways i have no strong opinion about it

veps/sig-network/vm-network-bandwidth-limiting.md
frenzyfriday: looks good now, thanks!

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,122 @@
...
+The implementation introduces a `Bandwidth` field within the existing `Interface` struct. The API accepts `Inbound` and `Outbound` parameters. Because Kubernetes `resource.Quantity` lacks a native time dimension, the interval for rate-based fields is strictly evaluated as **per second**:
+
+* **Average**: The desired average bandwidth rate, measured in **bytes per second** (e.g., `10Mi` evaluates to 10 Mebibytes per second).
+* **Peak**: The maximum transmission rate allowed while discharging a burst, measured in **bytes per second**.
+* **Burst**: The maximum absolute amount of data, measured in **bytes**, that can be sent at the `Peak` speed in a single spike before the `Average` limit is enforced.
+
+### Conversion to Domain XML
+The `virt-launcher` contains a converter that translates KubeVirt specs into Libvirt domain XML. A new builder function will be added to `configureInterfaces`. This builder will parse the `bandwidth` struct, convert the `resource.Quantity` byte values into the Kilobytes/second (KiB/s) format expected by Libvirt, and populate the empty `<bandwidth>` struct in the XML.
+
+### Validation
```
frenzyfriday, line 49: Will there be a max/min value validation? Or stuff like `peak`>`average`?
But maybe these are low level details and can be skipped in the VEP

veps/sig-network/vm-network-bandwidth-limiting.md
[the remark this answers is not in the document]
frenzyfriday: Also I am wondering what would happen if all the Vm owners in a cluster configure very high bandwidth on their VMs. Technically there are no bw limits right now too, so VM owners can run high bw consuming workloads even right now. But wont having a bw field make it easier for malicious VM owners to exploit this?

veps/sig-network/vm-network-bandwidth-limiting.md
Aneesh-Hegde: I think it's worth explicitly adding to the Validation section so it doesn't get missed during implementation.

Aneesh-Hegde:
@frenzyfriday made the changes and also regarding
> Also I am wondering what would happen if all the Vm owners in a cluster configure very high bandwidth on their VMs. Technically there are no bw limits right now too, so VM owners can run high bw consuming workloads even right now. But wont having a bw field make it easier for malicious VM owners to exploit this?

That is a very legitimate concern. The key distinction here is that these tc parameters are purely for throttling, not reservations. While it's true that requesting an absurdly high limit just means the VM will hit the physical hardware ceiling (which is the exact same behavior as having no limit today), I agree it's an edge case we should be mindful of.

To be completely honest, I'm currently not sure what the best practice is to tackle this natively within KubeVirt. Do you think we should consider adding a maximum ceiling in the validation webhook in a future PR, or is leaving it entirely to external policy engines (like Kyverno) the standard approach here?

Aneesh-Hegde:
> @frenzyfriday made the changes and also regarding
> 
> > Also I am wondering what would happen if all the Vm owners in a cluster configure very high bandwidth on their VMs. Technically there are no bw limits right now too, so VM owners can run high bw consuming workloads even right now. But wont having a bw field make it easier for malicious VM owners to exploit this?
> 
> That is a very legitimate concern. The key distinction here is that these tc parameters are purely for throttling, not reservations. While it's true that requesting an absurdly high limit just means the VM will hit the physical hardware ceiling (which is the exact same behavior as having no limit today), I agree it's an edge case we should be mindful of.
> 
> To be completely honest, I'm currently not sure what the best practice is to tackle this natively within KubeVirt. Do you think we should consider adding a maximum ceiling in the validation webhook in a future PR, or is leaving it entirely to external policy engines (like Kyverno) the standard approach here?

/cc @maiqueb any suggestions on this

EdDev (changes requested):
Thank you for the proposal.

There are several issues with this enhancement request:
- Controlling BW limits is already available through the CNI infrastructure, see [CNI BW](https://www.cni.dev/plugins/current/meta/bandwidth/). Users have already used it successfully for secondary networks.
- The proposed implementation, through libvirt, requires rootful (or net admin) capabilities which the virt-launcher does not have.
- The implications of adding API fields to the VMI spec is high and it is not clear if the need is on a per VM instance, per network or per other criteria (e.g. VM groups). The usage of Network Attachment Definitions is a good example of the level of grouping for specific network granularity.


We can re-evaluate this proposal again once the above points are considered.

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,122 @@
...
+- Support for Explicit Congestion Notification (ECN) marking to signal network congestion, as this is currently unsupported within the bandwidth QoS configuration of the Libvirt Domain XML API.
+
+## Definition of Users
+- **Cluster Administrators**: Responsible for node health and resource fair-share.
+- **Cloud Service Providers**: Delivering tiered networking performance levels to end-users.
+- **Users**: End-users or application owners deploying VMs.
+
+## User Stories
+- As a **Cloud Service Provider**, I want to cap the network bandwidth of "Basic Tier" VMs to 100Mbps so that I can differentiate them from "Premium Tier" offerings.
+- As a **user**, I want to define my VM's network speed in standard units (e.g., 1Gi) within the VM manifest so I can ensure my specific application has the guaranteed throughput it needs to function correctly.
```
maiqueb: I don't understand why does it *have* to be in the vm manifest. That's getting into the "how".

Aneesh-Hegde:
Hey @EdDev ,I’ve been thinking about your suggestion regarding CNI chaining, and I agree that it is the most reasonable path forward.Like it is better UX for QoS ,less blast radius,but will have some architectural tradeoff like:
- We will not have parameter like `burst` and `peak`  that we get from Libvirt XML.
- The annotation applies globally to the pod's primary network interface, rather than strict per-vNIC.
- Because CNI plugins only execute at pod creation, users wouldn't be able to dynamically update the bandwidth of a running VM without a restart.

This also brings up a question like if I implement this, then it would be like an just abstraction layer, so will VEP hold any importance for this.
Would love your feedback on this

maiqueb:
> Hey @EdDev ,I’ve been thinking about your suggestion regarding CNI chaining, and I agree that it is the most reasonable path forward.

I also share the thought that chaining bandwidth CNI is the preferable alternative out there. At least, it is the most kubernetes native. 

> Like it is better UX for QoS ,less blast radius,but will have some architectural tradeoff like:
> 
>     * We will not have parameter like `burst` and `peak`  that we get from Libvirt XML.

Hm, don't you have `burst` available ? In ingress & egress [forms](https://www.cni.dev/plugins/current/meta/bandwidth/#network-configuration-reference). 

> 
>     * The annotation applies globally to the pod's primary network interface, rather than strict per-vNIC.

I don't understand this comment - probably I'm missing something. Why can't you chain bandwith CNI with secondary interfaces ? e.g. bridge 

> 
>     * Because CNI plugins only execute at pod creation, users wouldn't be able to dynamically update the bandwidth of a running VM without a restart.

Correct.

> 
> 
> This also brings up a question like if I implement this, then it would be like an just abstraction layer, so will VEP hold any importance for this. Would love your feedback on this

Aneesh-Hegde:
Hey @maiqueb thanks for the feedback. But I think there is some disconnect.
Let me reframe my implementation plan, as the architecture actually splits into two paths depending on the network type:
1. Secondary Networks (Multus):
For secondary interfaces, the CNI binary is chained via the NAD, and we have full access to ingressBurst and egressBurst using the exact JSON config you linked.

2. Primary Network (Masquerade):
My previous comment about "losing burst" was strictly focused on the default pod network. Because virt-controller cannot write host-level CNI JSON config to `/etc/cni/net.d/`, we have to pass our intent to the Kubelet using the `kubernetes.io/ingress-bandwidth` pod annotation. Since that specific annotation only accepts a single flat rate, we can't cleanly expose a custom burst parameter for the primary network (the Kubelet just calculates a default burst under the hood).

Given the different backend mechanisms, I plan to implement the VEP as a unified abstraction layer. Instead of users managing K8s annotations or Multus JSON, virt-controller will translate a single KubeVirt API field into the correct backend:
- If Primary: Translates to `kubernetes.io/ingress-bandwidth` pod annotations.
- If Secondary: Translates to Multus JSON overrides.

Beyond the UX improvement, keeping this abstraction layer in the KubeVirt API gives us two critical features:
- Fleet Scale: Admins can bake QoS tiers directly into VirtualMachineInstancetypes.
- Live Migration: virt-controller stamps the destination pod before the VM migrates, ensuring limits survive the move.

Does this dual-architecture approach using the VEP as the abstraction layer while letting the CNI handle execution helpful?

maiqueb:
> Hey @maiqueb thanks for the feedback. But I think there is some disconnect. Let me reframe my implementation plan, as the architecture actually splits into two paths depending on the network type:
> 
>     1. Secondary Networks (Multus):
>        For secondary interfaces, the CNI binary is chained via the NAD, and we have full access to ingressBurst and egressBurst using the exact JSON config you linked.
> 
>     2. Primary Network (Masquerade):
>        My previous comment about "losing burst" was strictly focused on the default pod network. Because virt-controller cannot write host-level CNI JSON config to `/etc/cni/net.d/`, we have to pass our intent to the Kubelet using the `kubernetes.io/ingress-bandwidth` pod annotation. Since that specific annotation only accepts a single flat rate, we can't cleanly expose a custom burst parameter for the primary network (the Kubelet just calculates a default burst under the hood).
> 

Thanks for elaborating. 

> 
> Given the different backend mechanisms, I plan to implement the VEP as a unified abstraction layer. Instead of users managing K8s annotations or Multus JSON, virt-controller will translate a single KubeVirt API field into the correct backend:
> 
>     * If Primary: Translates to `kubernetes.io/ingress-bandwidth` pod annotations.
> 
>     * If Secondary: Translates to Multus JSON overrides.
> 
> 
> Beyond the UX improvement, keeping this abstraction layer in the KubeVirt API gives us two critical features:
> 
>     * Fleet Scale: Admins can bake QoS tiers directly into VirtualMachineInstancetypes.
> 
>     * Live Migration: virt-controller stamps the destination pod before the VM migrates, ensuring limits survive the move.
> 
> 
> Does this dual-architecture approach using the VEP as the abstraction layer while letting the CNI handle execution helpful?

It might, but all in all, I think the question should be reversed: is there any need to implement this feature for the cluster default network attachment ? I wonder if attempting to tackle both is not premature optimization. 

While it is the default cluster network attachment, usually, for VMs, that network attachment is not that useful. Can you capture a use case for this VEP on the default cluster network attachment ? That might change the picture (for me).

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,122 @@
...
+Currently, KubeVirt VMs can consume the full available bandwidth of the host's physical network interface. In multi-tenant or high-density environments, a "noisy neighbor" VM can saturate the link, causing high latency or packet loss for other workloads on the same node. There is a requirement for a native KubeVirt API to enforce fair network resource distribution.
+
+## Goals
+- Enable per-interface inbound and outbound traffic shaping for VMs.
+- Provide a native Kubernetes-style API using standard resource units for network QoS.
+- Ensure cluster stability by preventing individual guests from monopolizing host network resources.
+
+## Non Goals
+- Support for dynamic "live" updating of bandwidth without a VM restart during the Alpha phase.
+- Implementation of CNI-level shaping: the CNI bandwidth plugin operates at the pod boundary and cannot shape traffic at the tap device level, meaning the VM itself remains unthrottled on its primary network interface. This feature enforces limits directly on the tap device via virt-handler.
```
Aseeef: Why does it matter if the limit is at the pod boundary or directly at the VM's tap device? The VM still gets throttled no?

Aneesh-Hegde:
Some usecases that I can comeup where controlling the primary network (masquerade) remains critical are:
- The Managed Kubernetes UX: (EKS, GKE, AKS) While platforms like EKS and GKE do technically support Multus, configuring it often requires significant administrative overhead, elevated privileges (`DaemonSets`), and can sometimes complicate cloud provider SLAs. Many platform teams offering internal "VMs-as-a-Service" on standard dev clusters simply don't enable Multus because it's overkill for their needs. For these everyday users, the primary network is the only network they have. Scoping this feature strictly to secondary interfaces leaves them without any QoS controls.
- Live Migration Continuity: If a user defines bandwidth via a Multus NAD, that NAD is naturally re-evaluated on the destination node during the CNI ADD phase. However, for the primary network, if a user relies on manual pod annotations (`kubernetes.io/ingress-bandwidth`) to enforce limits, those limits vanish during live migration because the destination pod boots from the base VMI spec. We need virt-controller (via the VEP API) to explicitly propagate that intent to the destination pod to ensure QoS survives the move.

I think these are some usecases where it can be helpful,but if this feels kind of nit we can then just stick to having abstraction layer for secondary network
@maiqueb

Aseeef (commented):
When I first created https://github.com/kubevirt/kubevirt/issues/17200, in all honestly I wasn't very familiar with CNIs in the native K8s ecosystem for traffic shaping. So after looking at this, I do think the issue I raised is not really a real issue. Apologies for sending you on a wild goose chase. 😅

> Live Migration: virt-controller stamps the destination pod before the VM migrates, ensuring limits survive the move.

While again, I'm not too familiar with the K8s networking ecosystem, from what I found, I don't think this is a valid argument for implementing the bandwidth limit at the KubeVirt/LibVirt level. When configured correctly, limits you configure should automatically survive the migration as they should be a part of the new target VMI spec too (because the VM would be referencing the NAD in the spec for the target VM as copied from the source VM).

maiqueb:
> Some usecases that I can comeup where controlling the primary network (masquerade) remains critical are:
> 
>     * The Managed Kubernetes UX: (EKS, GKE, AKS) While platforms like EKS and GKE do technically support Multus, configuring it often requires significant administrative overhead, elevated privileges (`DaemonSets`), and can sometimes complicate cloud provider SLAs. Many platform teams offering internal "VMs-as-a-Service" on standard dev clusters simply don't enable Multus because it's overkill for their needs. For these everyday users, the primary network is the only network they have. Scoping this feature strictly to secondary interfaces leaves them without any QoS controls.
> 
>     * Live Migration Continuity: If a user defines bandwidth via a Multus NAD, that NAD is naturally re-evaluated on the destination node during the CNI ADD phase. However, for the primary network, if a user relies on manual pod annotations (`kubernetes.io/ingress-bandwidth`) to enforce limits, those limits vanish during live migration because the destination pod boots from the base VMI spec. We need virt-controller (via the VEP API) to explicitly propagate that intent to the destination pod to ensure QoS survives the move.

Live migration is broken in masquerade networks, since both src and dst pods will get a different IP address while the guest has the original IP. I am unsure this reason is enough.

> 
> 
> I think these are some usecases where it can be helpful,but if this feels kind of nit we can then just stick to having abstraction layer for secondary network @maiqueb

That's what **I** would try to do. But, if your motives are strong, it might make sense to also cover the default attachment. Let's see what the rest of the community have to say.

Aneesh-Hegde:
> Live migration is broken in masquerade networks, since both src and dst pods will get a different IP address while the guest has the original IP. I am unsure this reason is enough.

So i think we need to then stick to the secondary network. Will wait for some feedbacks from @EdDev or someone from sig-network. Thanks for your time and feedback :)

veps/sig-network/vm-network-bandwidth-limiting.md
Aneesh-Hegde: Sorry missed your comment ;).
Yes, it true but when I initially read the issue, my mind went straight to ensuring strict per-vNIC (interface-level) granularity. Because a VM can have multiple network interfaces, I was focused on enforcing the limit directly at the tap device level to ensure one interface's limit didn't accidentally throttle another.🙃
However, after the recent discussions on the VEP regarding CNI chaining, I agree that enforcing this at the pod boundary is the much more Kubernetes-native path forward.
Given the pivot in the architecture, what should our next steps be for this specific issue? Should we close this out, or keep it open to track the CNI-based implementation?

Edit: I have made the changes in VEP based on feedbacks received till now. If possible PTAL:).

Aseeef: From what I understood, it sounds like the CNI-based implementation already works so no further work is required. (Or did I miss something?)

maiqueb: how do you request that bandwidth / limit for a particular pod ? ...

what I think we'd have out of the box is the ability of setting bandwidth / limit for **all** pods attached to a network. AFAIU, that doesn't meet the goals (ability of defining the bandwidth per pod/VM).

Aneesh-Hegde: Exactly. Out of the box, admins would have to hardcode limits into static NADs, meaning every VM on that network gets the exact same limit.
To achieve the per-VM granularity requested in the goals without forcing admins to manage dozens of duplicate NADs for different speed tiers, we need this VEP to act as the control-plane abstraction. `virt-controller` will read the per-VM intent from the API and dynamically generate the Multus CNI annotations for the pod.
Furthermore, moving this intent into the KubeVirt API is the only way to ensure these QoS constraints survive a Live Migration. KubeVirt generates the destination pod strictly from the base VMI spec, so `virt-controller` needs to be able to read this API field and explicitly re-stamp the destination pod to prevent the limits from being lost.

Aseeef: Based of what was discussed, and some additional research I did here are my thoughts:

Typically, you'd want fixed QoS groups in most practical scenarios (meaning every VM that requires a unique QoS setting is not a well known use case). Each QoS speed tier is its own group as defined by a NAD.

So for each speed tier, you'd create a NAD with the rate limits, and then attach that NAD to whatever VMs you want to be in the QoS group. So suppose you have a 50mbps network tier, you create 1 NAD for it called `limited-50m` and for each VM that belongs in this tier you put:
```
spec:
  networks:
    - name: limited
      multus:
        networkName: limited-50m    # NAD name in this namespace
  domain:
    devices:
      interfaces:
        - name: limited
          bridge: {} 
```
in the VM spec.

And this way, it would survive migrations as well since NADs are scoped to an entire namespace across all nodes. And of course you can have any number of VMs with in this speed tier.

I don't think this is particularly hard to manage, and I don't understand what the use case for defining this logic in KubeVirt would be at least when considering fixed tier groups. (If instead you have a strong use case for dynamic tiering, that would be a different story though)

Edit: While you can tell KubeVirt to use this NAD defined network as your primary network, this does disconnect you from pod networking of the default primary network interface. And the primary network does come with many limits on how you can only shape traffic with annotations. So that might be a good use case for additional work...

Aneesh-Hegde: > Edit: While you can tell KubeVirt to use this NAD defined network as your primary network, this does disconnect you from pod networking of the default primary network interface. And the primary network does come with many limits on how you can only shape traffic with annotations. So that might be a good use case for additional work...

You're right about the disconnect from the default pod network and the limitations of shaping traffic with annotations. My initial draft for this VEP actually included logic to translate the API into `kubernetes.io/ingress-bandwidth` annotations for the primary network. But as @maiqueb pointed out earlier, live migration is currently broken for masquerade networks anyway due to pod IP reassignment. So, I decided to drop primary network support from the Alpha scope to avoid premature optimization and just focus on secondary (Multus) networks.

>I don't think this is particularly hard to manage, and I don't understand what the use case for defining this logic in KubeVirt would be at least when considering fixed tier groups. (If instead you have a strong use case for dynamic tiering, that would be a different story though)

I agree that for simpler setups with a few fixed tiers, hardcoding the bandwidth into static NADs works fine and naturally survives migration.

The main reason to keep this logic in the KubeVirt API is to support dynamic tiering. It decouples the QoS tier from the network plumbing, which is especially useful for environments using `VirtualMachineInstancetypes`.

To give a practical example: if a cluster has 3 base networks (vlan-dev, vlan-stage, vlan-prod) and wants to offer 3 speed tiers (50M, 100M, 500M), relying solely on static NADs means the admin has to create and maintain 9 separate NADs (dev-50m, dev-100m, etc.). If they add another network or tier, that number multiplies pretty quickly.
With this VEP acting as an abstraction layer:
- The admin only manages 3 base NADs (one per network).
- The QoS tier can be defined directly in an Instancetype (e.g., a "Premium" instancetype automatically applies a 500M limit).
- `virt-controller` reads the API intent and dynamically generates the Multus JSON override for that specific pod.
So while static NADs handle the basic cases, this API provides the abstraction needed for larger environments to prevent a sprawl of duplicate NADs.

Aseeef: Hmm OK, maybe this makes sense then.

Aneesh-Hegde: do you have any feedback on the proposed architecture or edge cases I should address

Aneesh-Hegde: > > Edit: While you can tell KubeVirt to use this NAD defined network as your primary network, this does disconnect you from pod networking of the default primary network interface. And the primary network does come with many limits on how you can only shape traffic with annotations. So that might be a good use case for additional work...
> 
> You're right about the disconnect from the default pod network and the limitations of shaping traffic with annotations. My initial draft for this VEP actually included logic to translate the API into `kubernetes.io/ingress-bandwidth` annotations for the primary network. But as @maiqueb pointed out earlier, live migration is currently broken for masquerade networks anyway due to pod IP reassignment. So, I decided to drop primary network support from the Alpha scope to avoid premature optimization and just focus on secondary (Multus) networks.
> 
> > I don't think this is particularly hard to manage, and I don't understand what the use case for defining this logic in KubeVirt would be at least when considering fixed tier groups. (If instead you have a strong use case for dynamic tiering, that would be a different story though)
> 
> I agree that for simpler setups with a few fixed tiers, hardcoding the bandwidth into static NADs works fine and naturally survives migration.
> 
> The main reason to keep this logic in the KubeVirt API is to support dynamic tiering. It decouples the QoS tier from the network plumbing, which is especially useful for environments using `VirtualMachineInstancetypes`.
> 
> To give a practical example: if a cluster has 3 base networks (vlan-dev, vlan-stage, vlan-prod) and wants to offer 3 speed tiers (50M, 100M, 500M), relying solely on static NADs means the admin has to create and maintain 9 separate NADs (dev-50m, dev-100m, etc.). If they add another network or tier, that number multiplies pretty quickly. With this VEP acting as an abstraction layer:
> 
> * The admin only manages 3 base NADs (one per network).
> * The QoS tier can be defined directly in an Instancetype (e.g., a "Premium" instancetype automatically applies a 500M limit).
> * `virt-controller` reads the API intent and dynamically generates the Multus JSON override for that specific pod.
>   So while static NADs handle the basic cases, this API provides the abstraction needed for larger environments to prevent a sprawl of duplicate NADs.

Hey @maiqueb , what are your thoughts? Edy rightly questioned if this abstraction justifies expanding the API, and is also skeptical about the technical feasibility of the abstraction layer fully working across edge cases.

Are there other use cases where KubeVirt natively needs this? If we feel the maintenance burden and technical risks outweigh the UX benefits, I'm happy to summarize the NAD limitations and close the VEP

maiqueb: > skeptical about the technical feasibility of the abstraction layer fully working across edge cases

What's the source of said skepticism ? 

What edge cases should we track or monitor ? What's so special into each of those edge cases ? 

Requiring ONE NAD for each configuration, when the standard (and implementing CNI) allows for a pod to request params for its own attachment is - IMHO - bad.

Aneesh-Hegde: To clarify, 'edge cases' was just my phrasing for general skepticism regarding the technical feasibility of the translation layer.
Since we agree that the legacy static NAD approach is a bad UX, I'll focus my efforts on the implementation details in the VEP so we can prove out the mechanics. Thanks for clarification.
@EdDev pls let me know if i missed something from the meet.

Aneesh-Hegde:
Hey @frenzyfriday I’ve made some tweaks to the VEP to showcase the implementation details more clearly. I realized I didn't explicitly mention in sig meeting that the design leverages the standard CNI bandwidth plugin by building a dynamic abstraction layer on top of it (essentially an abstraction layer on top of CNI chain). I also verified this approach and use case with Miguel in the PR [comments](https://github.com/kubevirt/enhancements/pull/236#discussion_r3181892917).

Could you take a look and give it an initial review when you have a chance?

veps/sig-network/vm-network-bandwidth-limiting.md
```diff
@@ -0,0 +1,125 @@
...
+- This VEP targets beta for version: TBD
+- This VEP targets GA for version: TBD
+
+## Overview
+This VEP proposes adding Network Quality of Service (QoS) support to KubeVirt by enabling bandwidth limiting for Virtual Machine (VM) secondary network interfaces managed by Multus. This allows users to specify bandwidth limits in the VirtualMachine template spec, which `virt-controller` translates into Multus network-selection annotations (`k8s.v1.cni.cncf.io/networks`) on the `virt-launcher` pod, leveraging standard Kubernetes CNI chaining to enforce limits at the pod boundary.
+
+## Motivation
+Currently, KubeVirt VMs attached to secondary networks can consume the full available bandwidth of the host's physical network interface. While administrators can manually configure CNI bandwidth plugins inside Multus `NetworkAttachmentDefinition` (NAD) objects, doing so hardcodes the limit for all VMs using that specific NAD. There is a requirement for a native KubeVirt API to define these limits on a per-VM basis, allowing for dynamic QoS tiers (e.g., via KubeVirt `VirtualMachineInstancetypes`) without proliferating duplicate NADs across the cluster.
```
EdDev, line 18: > There is a requirement for a native KubeVirt API to define these limits on a per-VM basis, allowing for dynamic QoS tiers (e.g., via KubeVirt `VirtualMachineInstancetypes`) without proliferating duplicate NADs across the cluster.

- AFAIK `VirtualMachineInstancetypes` does not support network interfaces.
- The grouping through Net Attach Def is satisfying all scenarios that I know of. Kubevirt operators are unlikely to manage individual VMs and network interfaces.

IMHO the motivation is not strong enough to pass the minimum criteria of adding API knobs to the VM/VMI.
The existing solution is good enough with all existing known users.

I would like some real ref of an existing user or project that needs this, before we continue here.

EdDev (changes requested):

Aneesh-Hegde:
Closing this VEP for now.
The motivation lacks sufficient backing. The NAD-sprawl argument, while intuitive, has no concrete evidence from real users or projects actually blocked by the current approach. The existing NAD-based solution covers all known use cases.
Happy to revisit if a concrete user story or upstream project requirement emerges.

Issue #227, opened 2026-03-12
VEP 227: Enable Secure Boot for ARM64 Guests


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood

**Current Feature Stage**:

New

**Feature Gate**:

ARM64SecureBoot

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/228

**Dependencies**:

* [VEP #241: Firmware Auto-Selection](https://github.com/kubevirt/enhancements/issues/241) — provides the shared firmware auto-selection infrastructure
* [VEP #210: CentOS Stream 10 Support](https://github.com/kubevirt/enhancements/issues/210) — required for GA (QEMU 10.0+, libvirt with varstore, edk2-aarch64 firmware)

**Timeline**:

* Alpha release target: v1.9.0
* Beta release target:
* Stable release target:

* [ ] Alpha
  * [ ] VEP PRs:
    * [ ] https://github.com/kubevirt/enhancements/pull/228
  * [ ] Code PRs:
    * [ ] https://github.com/kubevirt/kubevirt/pull/17135 (draft, depends on #17263)
    * [ ] https://github.com/kubevirt/kubevirt/pull/17263 (VEP #241 prototype)
    * [ ] https://github.com/kubevirt/project-infra/pull/4820
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #242, opened 2026-03-23
VEP #241: Use libvirt firmware auto-selection for EFI Secure Boot

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/241
**SIG label**: /sig compute

### What this PR does

Proposes using libvirt's firmware auto-selection feature (`<os firmware='efi'>` with `<firmware><feature>` elements) for standard EFI Secure Boot on x86_64, replacing KubeVirt's hardcoded OVMF loader paths.

This delegates firmware loader selection to libvirt based on JSON descriptor files shipped by the `edk2-ovmf` package, eliminating fragile path assumptions and aligning x86_64 with the ARM64 Secure Boot approach (VEP #227). The NVRAM path remains explicitly set by KubeVirt to preserve the existing filename convention and persistent storage compatibility.

Key points:
- Scoped to standard EFI Secure Boot only (not SEV/SNP/TDX or non-SB EFI)
- Guarded by `FirmwareAutoSelection` feature gate (Alpha, default disabled)
- Feature gate passed to virt-launcher via CLI flag (`--firmware-auto-selection`), following the pattern used by `--libvirt-hook-server-and-client`
- NVRAM path explicitly preserved via `PathForNVram()` to avoid filename convention mismatch between KubeVirt and libvirt domain names
- No user-facing API changes
- Includes rationale for excluded cases and future work for CS10 `uefi-vars`
- Implementation PR: https://github.com/kubevirt/kubevirt/pull/17263

### DIFF
--- a/veps/sig-compute/firmware-auto-selection/vep.md
+++ b/veps/sig-compute/firmware-auto-selection/vep.md
@@ -1,0 +1,528 @@
+# VEP #241: Use Libvirt Firmware Auto-Selection for EFI Secure Boot
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Replace KubeVirt's hardcoded OVMF firmware loader paths for EFI Secure Boot
+with libvirt's firmware auto-selection feature. Instead of specifying an
+explicit `<loader>` path in the domain XML, KubeVirt will use
+`<os firmware='efi'>` with `<firmware><feature>` elements, delegating
+firmware loader selection to libvirt based on JSON descriptor files shipped
+by the `edk2-ovmf` package. The NVRAM path remains explicitly set by
+KubeVirt to preserve filename convention and persistent storage
+compatibility.
+
+## Motivation
+
+KubeVirt currently hardcodes firmware file names (e.g. `OVMF_CODE.secboot.fd`,
+`OVMF_VARS.secboot.fd`) and constructs explicit `<loader>` / `<nvram>` paths
+in the domain XML for EFI Secure Boot. This approach has several drawbacks:
+
+1. **Fragile path assumptions**: Firmware file names and locations are
+   distribution-specific and change over time. For example, edk2 packaging has
+   shifted from `/usr/share/OVMF/` to `/usr/share/edk2/ovmf/` and is moving
+   from raw to qcow2 firmware formats.
+
+2. **Duplicated logic**: KubeVirt reimplements firmware selection logic
+   (matching secure boot capability, SMM requirements, architecture) that
+   libvirt already handles via its firmware descriptor system.
+
+3. **Missed optimizations**: Newer firmware builds (e.g. qcow2-format OVMF)
+   offer reduced memory usage and faster startup, but KubeVirt cannot
+   automatically pick them up without code changes.
+
+4. **Inconsistency with ARM64**: VEP #227 (ARM64 Secure Boot) already uses
+   firmware auto-selection for ARM64 guests. Using the same mechanism for
+   x86_64 unifies the code path.
+
+Libvirt's firmware auto-selection feature (available since libvirt 7.2.0) was
+designed to solve exactly this problem. The `edk2-ovmf` package in CentOS
+Stream 9 already ships the required JSON firmware descriptors that declare
+secure boot, enrolled keys, and SMM capabilities.
+
+## Goals
+
+- Use libvirt firmware auto-selection for standard EFI Secure Boot on x86_64.
+- Eliminate hardcoded OVMF firmware paths for the Secure Boot case.
+- Ensure functional equivalence with the current explicit-path approach.
+- Unify the Secure Boot firmware selection mechanism across x86_64 and ARM64.
+
+## Non Goals
+
+- Changing the user-facing API (`spec.domain.firmware.bootloader.efi`).
+- Adopting firmware auto-selection for non-Secure-Boot EFI or confidential
+  computing VM types (SEV/SEV-ES, SEV-SNP, TDX). See the
+  [Why Only Secure Boot?](#why-only-secure-boot) section below for detailed
+  rationale.
+- Removing the explicit firmware path detection code in `efi.go` entirely
+  (it remains needed for the excluded cases above).
+
+## Definition of Users
+
+- **VM Authors**: No change in behavior. The same `secureBoot: true` API
+  field produces the same Secure Boot functionality.
+- **Cluster Admins**: Should be aware that the internal domain XML changes,
+  which could affect custom monitoring or tooling that parses domain XML.
+- **KubeVirt Developers**: Benefit from a simpler, more maintainable firmware
+  selection code path.
+
+## User Stories
+
+- As a VM Author, I want EFI Secure Boot to work the same way as before with
+  no changes to my VM definitions.
+- As a KubeVirt Developer, I want the firmware selection logic to be delegated
+  to libvirt so that KubeVirt doesn't need to track distribution-specific
+  firmware file names and paths.
+- As a Cluster Admin, I want KubeVirt to automatically use the best available
+  firmware format (raw or qcow2) without requiring KubeVirt code changes when
+  the edk2 package is updated.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Current Behavior (Explicit Paths)
+
+For EFI Secure Boot, KubeVirt currently:
+
+1. Probes for firmware files on the filesystem (`OVMF_CODE.secboot.fd`,
+   `OVMF_VARS.secboot.fd`) via `DetectEFIEnvironment()` in `efi.go`.
+2. Constructs explicit `<loader>` and `<nvram>` XML with those paths.
+3. Manages the NVRAM file path at a KubeVirt-controlled location.
+
+Resulting domain XML:
+
+```xml
+<os>
+  <type arch='x86_64' machine='pc-q35-rhel10.2.0'>hvm</type>
+  <loader readonly='yes' secure='yes' type='pflash'>/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd</loader>
+  <nvram template='/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd'>/var/run/kubevirt-private/libvirt/qemu/nvram/vmi_VARS.fd</nvram>
+</os>
+```
+
+### Proposed Behavior (Firmware Auto-Selection)
+
+For standard EFI Secure Boot (no SEV/SNP/TDX), KubeVirt will instead:
+
+1. Set `<os firmware='efi'>` with a `<firmware><feature>` element requesting
+   `enrolled-keys`.
+2. Let libvirt match these requirements against its firmware descriptor files.
+3. Let libvirt resolve the loader path and NVRAM template.
+4. Continue to explicitly set the `<nvram>` path via `PathForNVram()` to
+   preserve NVRAM filename convention and directory compatibility (see
+   [NVRAM Path Handling](#nvram-path-handling) below).
+
+Resulting domain XML (as submitted to libvirt):
+
+```xml
+<os firmware='efi'>
+  <type arch='x86_64' machine='pc-q35-rhel10.2.0'>hvm</type>
+  <firmware>
+    <feature enabled='yes' name='enrolled-keys'/>
+  </firmware>
+  <nvram format='raw'>/var/run/kubevirt-private/libvirt/qemu/nvram/vmi_VARS.fd</nvram>
+</os>
+```
+
+Libvirt resolves this against the `30-edk2-ovmf-x64-sb-enrolled.json`
+firmware descriptor from the `edk2-ovmf` package, which declares:
+
+```json
+{
+  "description": "OVMF with SB+SMM, SB enabled, MS certs enrolled",
+  "features": ["acpi-s3", "enrolled-keys", "requires-smm", "secure-boot", "verbose-dynamic"],
+  "mapping": {
+    "executable": { "filename": "/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd" },
+    "nvram-template": { "filename": "/usr/share/edk2/ovmf/OVMF_VARS.secboot.fd" }
+  }
+}
+```
+
+The resulting active domain XML (after libvirt processes it) contains the
+same loader and NVRAM paths as before, but they were resolved by libvirt
+rather than hardcoded by KubeVirt.
+
+### Why Only Secure Boot?
+
+This VEP intentionally limits the scope to standard EFI Secure Boot. The
+other EFI modes are excluded for specific technical reasons:
+
+#### Non-Secure-Boot EFI
+
+Non-Secure-Boot EFI is excluded from the alpha scope to limit the initial
+change surface, but there is no fundamental technical blocker preventing
+firmware auto-selection for this case.
+
+As clarified by the libvirt maintainer in the VEP review, requesting
+`<feature enabled='no' name='enrolled-keys'/>` correctly matches both
+firmware descriptors with no Secure Boot support compiled in (e.g.
+`51-edk2-ovmf-2m-raw-x64-nosb.json`) and those where Secure Boot is present
+but the NVRAM template doesn't have keys enrolled (e.g.
+`41-edk2-ovmf-2m-raw-x64-sb.json`). The descriptor sorting is intentional,
+so the SEV-capable descriptor (`50-edk2-ovmf-x64-nosb.json`) does not
+introduce ambiguity — descriptors for "regular" builds are sorted before
+those intended for the CVM use case.
+
+This means non-Secure-Boot EFI could be added to firmware auto-selection
+in a subsequent release cycle by requesting `enrolled-keys=no`, following
+the same pattern used by
+[virt-install](https://github.com/virt-manager/virt-manager/pull/1038).
+
+The current explicit-path approach works well for non-Secure-Boot EFI and
+is not affected by the firmware naming/format changes that motivate this VEP
+(since `OVMF_CODE.fd` / `OVMF_VARS.fd` are stable names), so deferring this
+carries no risk.
+
+#### SEV/SEV-ES
+
+KubeVirt uses `OVMF_CODE.cc.fd` (the confidential-computing-specific OVMF
+build) for SEV/SEV-ES guests. This firmware is purpose-built for confidential
+computing with SEV-specific initialization and lacks SMM support.
+
+The CentOS Stream 9 `edk2-ovmf` package does **not** ship a firmware
+descriptor for `OVMF_CODE.cc.fd`. The only descriptor with `amd-sev` /
+`amd-sev-es` features is `50-edk2-ovmf-x64-nosb.json`, which maps to the
+generic `OVMF_CODE.fd` — a different firmware binary. Using auto-selection
+for SEV would silently select the wrong firmware.
+
+#### SEV-SNP and TDX
+
+These use stateless firmware (`type='rom'`, no NVRAM):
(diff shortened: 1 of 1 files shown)


veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,432 @@
...
+`nvram_dir` configuration in `qemu.conf`. KubeVirt's embedded libvirt uses:
+
+- Root: The system default (`/var/lib/libvirt/qemu/nvram/`)
+- Non-root: A copy of the system `qemu.conf` at
+  `/var/run/kubevirt-private/libvirt/qemu.conf`
+
+For root VMs, the paths match. For non-root VMs, verification is needed to
+ensure libvirt's auto-selected NVRAM path aligns with KubeVirt's expectations
+for the volume mount at `PathForNVram()`. If they diverge, the `qemu.conf`
+copy must be updated to set `nvram_dir` explicitly.
```
dasionov: So have you been able to verify libvirt's NVRAM path behavior for non-root VMs when using firmware auto-selection?                                                                                                                                                                
                                                                                                                                                                                                                                                                                 
  Today KubeVirt explicitly sets the NVRAM path via `PathForNVram()`, and this same path is used by `rendervolumes.go` to mount backend storage PVCs for persistent: true EFI VMs. With auto-selection, NVRam is set to nil in the domain XML, so libvirt determines the NVRAM path  
  based on `nvram_dir` in `qemu.conf`. For non-root VMs, `PathForNVram()` returns `/var/run/kubevirt-private/libvirt/qemu/nvram/` while libvirt's default `nvram_dir` is `/var/lib/libvirt/qemu/nvram/`. If they diverge, `qemu.conf` would need to be updated to set `nvram_dir` explicitly so  
  the NVRAM file lands inside the backend storage mount.                                                                                                                                                                                                                         
                  
  Beyond the directory, we should also verify the NVRAM filename convention. KubeVirt constructs it as `<vmi.Name>_VARS.fd`, but libvirt's domain name is `<namespace>_<vmi-name>`. If libvirt generates the filename from the domain name, a VM that previously had NVRAM created   
  via explicit paths could end up with a new file from template instead of reusing its existing one. For example:
                                                                                                                                                                                                                                                                                 
  - Existing NVRAM (explicit path): `my-vm_VARS.fd   `                                                                                                                                                                                                                             
  - Auto-selected NVRAM (domain name): `default_my-vm_VARS.fd`
                                                                                                                                                                                                                                                                                 
  This would silently reset the guest's EFI state (enrolled Secure Boot keys, boot order, etc.) on the first restart after the feature gate is enabled.

dasionov (commented):
Hi @lyarwood VEP looks great to me !
just left a few questions.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Thanks for this excellent catch @dasionov, I've reworked the NVRAM Path Handling section to address both concerns:

1. **NVRAM directory**: Documented the non-root path divergence explicitly and noted that `qemu.conf` must set `nvram_dir` to match `PathForNVram()`.

2. **NVRAM filename convention**: You're right — libvirt derives the filename from the domain name (`<namespace>_<vmi-name>`) which would produce `default_my-vm_VARS.fd` instead of the existing `my-vm_VARS.fd`, silently resetting EFI state.

The solution is to continue explicitly setting the `<nvram>` element in the domain XML via `PathForNVram()` even when using firmware auto-selection for the loader. Libvirt supports combining `<os firmware='efi'>` with an explicit `<nvram>` path — auto-selection applies to the `<loader>` only while the NVRAM path stays under KubeVirt's control.

I've updated the Overview, Proposed Behavior XML example, Code Changes, Update/Rollback Compatibility, and Alpha graduation checklist to reflect this throughout the VEP.

dasionov: > 2. **NVRAM filename convention**: You're right — libvirt derives the filename from the domain name (`<namespace>_<vmi-name>`) which would produce `default_my-vm_VARS.fd` instead of the existing `my-vm_VARS.fd`, silently resetting EFI state.
> 
> The solution is to continue explicitly setting the `<nvram>` element in the domain XML via `PathForNVram()` even when using firmware auto-selection for the loader. Libvirt supports combining `<os firmware='efi'>` with an explicit `<nvram>` path — auto-selection applies to the `<loader>` only while the NVRAM path stays under KubeVirt's control.

Since persisting the NVRAM in the domain isn’t ideal and seems to be becoming a requirement we didn’t originally plan for I’d prefer we avoid introducing it altogether. This feels like the right time to cut it during the VEP phase rather than carry the complexity forward.

What if we modify `PathForNVram()` to return the path expected by libvirt? That way, other consumers will naturally resolve to the correct location (assuming libvirt keeps this behavior stable). With this approach, we could drop NVRAM from the domain definition from day one of the alpha.

From a testing perspective, validating this in the upgrade lane should be straightforward and shouldn’t require much additional setup.

lyarwood: Thanks for the suggestion @dasionov.

I agree that fully delegating NVRAM to libvirt would be cleaner long term, but I think it's additional complexity to take on for the alpha in v1.9.0. Changing `PathForNVram()` to match libvirt's derived path means we also need to handle NVRAM file migration on upgrade — existing VMs have files named `<vmi.Name>_VARS.fd` while libvirt would derive `<namespace>_<vmi-name>_VARS.fd` from the domain name. That migration path (and its rollback) needs careful design and testing.

The explicit `<nvram>` element in the alpha is actually simpler than what we have today — we drop the `template=` attribute since libvirt auto-selects the source, and the file location doesn't change so the upgrade path is zero-effort:

```xml
<os firmware='efi'>
  <firmware>
    <feature enabled='yes' name='secure-boot'/>
    <feature enabled='yes' name='enrolled-keys'/>
  </firmware>
  <nvram>/path/to/my-vm_VARS.fd</nvram>
</os>
```

What if we add dropping the explicit `<nvram>` element as a goal for beta? That gives us time to design the migration properly and validate the libvirt naming convention in e2e tests before relying on it. I'll update the VEP to capture this.

dasionov: Make sense to me, Thanks! :+1:

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,510 @@
...
+    // for booting in EFI mode with SecureBoot disabled
+    code = codeWithSB
+}
+```
+
+If `OVMF_CODE.fd` is not present but `OVMF_CODE.secboot.fd` is, KubeVirt
+uses the Secure Boot code ROM with Secure Boot disabled in the loader
+(`secure='no'`). This is a valid configuration that libvirt's auto-selection
+would not produce — it would instead fail to find a matching non-Secure-Boot
+descriptor.
```
andreabolognani: Setting `secure=no` doesn't really do much I think. If you want to boot an unsigned OS using a build of edk2 that contains Secure Boot support, you just have to pick an NVRAM template that doesn't have the Secure Boot keys enrolled, i.e.

```xml
<os firmware='efi'>
  <firmware>
    <feature enabled='no' name='enrolled-keys'/>
  </firmware>
</os>
```

That will match both the firmware descriptor for a build of edk2 with no Secure Boot support compiled in, e.g. `51-edk2-ovmf-2m-raw-x64-nosb.json` in Fedora, or one where Secure Boot is present but the NVRAM template doesn't have the keys enrolled, e.g. `41-edk2-ovmf-2m-raw-x64-sb.json`.

In other words there's little reason to forego feature-based firmware autoselection for the use case of booting unsigned operating systems.

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,510 @@
...
+
+Additionally, the firmware descriptor `50-edk2-ovmf-x64-nosb.json` declares
+`amd-sev` and `amd-sev-es` features. Requesting `<feature enabled='no'
+name='secure-boot'/>` could still match this SEV-capable descriptor when a
+non-SEV firmware is intended, introducing ambiguity.
```
andreabolognani: The firmware descriptors for the "regular" builds of edk2 are intentionally sorted before those intended for the CVM use case. The fact that that specific firmware descriptor advertises those SEV-related features might be a bug, but either way that descriptor is perfectly suitable for non-SEV VMs and there's no reason to avoid using it.

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,510 @@
...
+
+KubeVirt uses `OVMF_CODE.cc.fd` (the confidential-computing-specific OVMF
+build) for SEV/SEV-ES guests. This firmware is purpose-built for confidential
+computing with SEV-specific initialization and lacks SMM support.
+
+The CentOS Stream 9 `edk2-ovmf` package does **not** ship a firmware
+descriptor for `OVMF_CODE.cc.fd`. The only descriptor with `amd-sev` /
+`amd-sev-es` features is `50-edk2-ovmf-x64-nosb.json`, which maps to the
+generic `OVMF_CODE.fd` — a different firmware binary. Using auto-selection
+for SEV would silently select the wrong firmware.
```
andreabolognani, line 204: Hopefully this has been fixed in CentOS Stream 10, but if not it definitely needs to be addressed. The intention is to have feature-based firmware autoselection cover all use cases.

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,510 @@
...
+
+While firmware descriptors exist for these (`60-edk2-ovmf-x64-amdsev.json`,
+`60-edk2-ovmf-x64-inteltdx.json`), these VM types have additional device
+and launch security configuration requirements beyond firmware selection.
+Changing the firmware selection mechanism for confidential computing VMs
+carries higher risk with minimal benefit, since the firmware file names
+are stable and purpose-specific.
+
+These cases could be explored in a future follow-up VEP once the Secure Boot
+auto-selection is proven stable.
```
andreabolognani, line 220: No problem from my side to making the process of moving towards full adoption of feature-based firmware autoselection an iterative one. As mentioned above, the intention is for it to cover all use cases.

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,510 @@
...
+```xml
+<os firmware='efi'>
+  <type arch='x86_64' machine='pc-q35-rhel10.2.0'>hvm</type>
+  <firmware>
+    <feature enabled='yes' name='secure-boot'/>
+    <feature enabled='yes' name='enrolled-keys'/>
+  </firmware>
+  <nvram>/var/run/kubevirt-private/libvirt/qemu/nvram/my-vm_VARS.fd</nvram>
+</os>
+```
```
andreabolognani, line 290: I'm concerned about the upgrade scenario. If you have an existing VM with an existing NVRAM file in raw format, but the firmware descriptors are sorted in a way that makes qcow2 format the preferred one, then you will effectively lose the contents of NVRAM or worse end up with a corrupted one, potentially leading to an unbootable VM.

One way to avoid that scenario would be to explicitly ask libvirt to use raw format while providing the path:

```xml
<os firmware='efi'>
  <nvram format='raw'>/path/to/nram.fd</nvram>
</os>
```

That should ensure that things work consistently.

andreabolognani (commented):
@lyarwood thanks for working on this! It's been a long time coming. I've left a few comments.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Thanks for the clarification @andreabolognani — that's really helpful. I wasn't aware that `<feature enabled='no' name='enrolled-keys'/>` would correctly match both the no-SB build and the SB-capable-but-unenrolled build. That does remove the main concern about the non-Secure-Boot EFI case.

I'll keep the alpha scope to Secure Boot only as planned, but this gives us a clear path to expand to non-Secure-Boot EFI in beta without the fallback logic concern I described. I'll update the VEP to reflect this.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Good to know the sorting is intentional — that resolves the ambiguity concern. I'll update the VEP to remove that as a blocker for the non-Secure-Boot case.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Agreed — if CS10 ships a descriptor for the CC-specific build that would unblock auto-selection for SEV/SEV-ES as well. I'll note this in the Future Work section.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Thanks, appreciate the confirmation.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Great catch — I hadn't considered the raw vs qcow2 NVRAM format mismatch on upgrade. I'll add `format='raw'` to the `<nvram>` element in the VEP and implementation to ensure existing raw NVRAM files aren't silently corrupted if the firmware descriptors prefer qcow2.

veps/sig-compute/firmware-auto-selection/vep.md
andreabolognani: Yeah, unfortunately features are not named very intuitively but when someone talks about "enable / disable Secure Boot" they usually mean "block / allow an unsigned OS to boot" respectively, and effectively that maps to the enrolled-keys feature.

enrolled-keys=yes implies secure-boot=yes, but enrolled-keys=no can match any value of secure-boot, so effectively you only need to set the value for enrolled-key; if you explicitly set secure-boot=no, that won't match a firmware descriptor with secure-boot=yes and enrolled-keys=no, which would have satisfied your use case of "allow an unsigned OS to boot".

For reference, virt-install has implemented its convenience option for Secure Boot configuration by setting the enrolled-keys feature only: https://github.com/virt-manager/virt-manager/pull/1038

lyarwood: Thanks @andreabolognani, I've simplified the XML snippets to only request `enrolled-keys` as you suggested, dropping the redundant `secure-boot` feature. This also aligns with how virt-install handles it.

dasionov:
Thank you @lyarwood recent changes looks good to me.
/lgtm

veps/sig-compute/firmware-auto-selection/vep.md
iholder101: @lyarwood Do we want to encode this plan in the document somewhere? We can also do that in a follow-up

veps/sig-compute/firmware-auto-selection/vep.md
iholder101: Do we need an update here as well?

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,514 @@
...
+
+### Scope of Change
+
+| VM Configuration | Current | Proposed |
+|---|---|---|
+| EFI + Secure Boot (standard) | Explicit paths | **Firmware auto-selection** |
+| EFI without Secure Boot | Explicit paths | Explicit paths (unchanged) |
+| EFI + SEV/SEV-ES | Explicit paths | Explicit paths (unchanged) |
+| EFI + SEV-SNP | Explicit paths (stateless) | Explicit paths (unchanged) |
+| EFI + TDX | Explicit paths (stateless) | Explicit paths (unchanged) |
```
iholder101: Is this true specifically for alpha?
Would we able to use firmware auto-selection for all of these cases by the time this GAs? Looking at @andreabolognani's comments I guess this would be possible, perhaps depending on a few bug fixes from the libvirt side?

veps/sig-compute/firmware-auto-selection/vep.md
iholder101: Makes total sense for currently running VMs.

What about new VMs though? IIUC qcow2 has the benefit of supporting more optimization and is now the preferable approach. Would we support that for "new" VMs?

veps/sig-compute/firmware-auto-selection/vep.md
```diff
@@ -0,0 +1,514 @@
...
+
+Adopt firmware auto-selection for all EFI configurations (including
+non-Secure-Boot and confidential computing). This would be a larger change
+with more risk — see [Why Only Secure Boot?](#why-only-secure-boot) for
+the specific technical blockers for each excluded case.
+
+**Deferred:** Could be pursued as a follow-up once the Secure Boot case is
+proven stable and missing firmware descriptors (e.g. for `OVMF_CODE.cc.fd`)
+are added to `edk2-ovmf`. See also [Future Work](#future-work) below for
+how CentOS Stream 10 edk2 updates could expand the scope.
```
iholder101, line 417: Okay I think this answers my previous question.
We can re-think if it's worth supporting as part of this VEP in the next cycle if we think it's not too complicated.

iholder101 (commented):
Thanks @lyarwood! Looks great to me!

Left a few questions, but I don't think they are blocking.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: I've updated the Non-Secure-Boot EFI section to reflect @andreabolognani's clarifications — the previous technical blocker rationale (fallback logic, descriptor ambiguity) has been replaced with a clear statement that there's no fundamental blocker. The plan to expand via `enrolled-keys=no` is now documented in the section and added to the Beta graduation checklist.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Yes — I've rewritten the section to reflect @andreabolognani's clarification that descriptor sorting is intentional and `enrolled-keys=no` handles this case correctly. The old code snippet and ambiguity concern are removed.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Good point — I've added Alpha and Target (GA) columns to the table. Non-Secure-Boot EFI is targeted for firmware auto-selection by GA, and is now a Beta graduation checklist item. SEV/SEV-ES remains blocked on missing upstream firmware descriptors. SEV-SNP and TDX stay on explicit paths since their firmware names are stable and purpose-specific.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Good question. I've added a paragraph addressing this — for alpha we use `format='raw'` uniformly to keep the rollback path simple (a VM created with the gate enabled must still work if rolled back to explicit pflash paths). Supporting qcow2 NVRAM for new VMs could be explored in beta/GA, but ideally the `uefi-vars` transition (Future Work) will eliminate pflash NVRAM entirely, making this moot.

veps/sig-compute/firmware-auto-selection/vep.md
lyarwood: Agreed, thanks!

veps/sig-compute/firmware-auto-selection/vep.md
andreabolognani: @lyarwood can you please remind me again, why would SEV-SNP and TDX not use firmware autoselection too? Is there a specific concern blocking that? Those use cases are definitely in scope as far as libvirt is concerned.

lyarwood: See https://github.com/kubevirt/enhancements/pull/242/changes/BASE..30e0292023a1d03fe074ee57cdbbdae0a59c7fae#diff-536a51a0eba09c2158a752da6f05d050a27d530553e35bdda5074ec56fa5c661R208 I just punted out of the initial Alpha scope for now, definitely something we can address ahead of or with the move to beta for this feature.

andreabolognani: That link goes nowhere for me.

andreabolognani: Anyway I'm okay with it being out of scope for Alpha, but unless there's a very compelling argument against it, I would like to see *all* paths go through firmware autoselection for GA. The issues mentioned with the descriptors for SEV/SEV-ES should just be addressed in edk2 before then.

vladikr: > Anyway I'm okay with it being out of scope for Alpha, but unless there's a very compelling argument against it, I would like to see _all_ paths go through firmware autoselection for GA. The issues mentioned with the descriptors for SEV/SEV-ES should just be addressed in edk2 before then.

ok, sounds good. 
@alancaldelas fyi

Looks good to me for alpha.
Thanks @lyarwood!
/approve

dasionov (approved):

Issue #17, opened 2025-03-10
VEP 17: instancetype.kubevirt.io/v1


lyarwood:
**Primary contact (assignee)**:

/assign lyarwood

**Current Feature Stage**:

Beta

**Feature Gate**:

N/A

**Responsible SIGs**:

/sig compute

**Enhancement link**:

[Original flavor and preferences design proposal](https://github.com/kubevirt/community/blob/main/design-proposals/instancetype.kubevirt.io/flavors-and-preferences.md)

**Timeline**:

* [x] v1alpha2 [v0.56.0](https://github.com/kubevirt/kubevirt/releases/tag/v0.56.0)
* [x] v1beta1 [v1.0.0](https://github.com/kubevirt/kubevirt/releases/tag/v1.0.0)
* [x] v1alpha{1,2} deprecation v1.1.0
* [x] v1alpha{1,2} removal v1.7.0
  * [x] https://github.com/kubevirt/kubevirt/pull/15400
* [ ] GA v1.8.0
  * [ ] https://github.com/kubevirt/kubevirt/pull/16109
  * [ ] https://github.com/kubevirt/kubevirt/pull/16110
  * [ ] https://github.com/kubevirt/kubevirt/pull/16130
* [ ] https://github.com/kubevirt/community/pull/318 - v1.9.0 (?)
* [ ] v1beta1 deprecation v1.9.0
* [ ] v1beta1 removal v1.10.0

**Additional context**:

lyarwood:
@vladikr are you okay with this as a tracker for something without an active enhancement associated with it?

vladikr:
> [@vladikr](https://github.com/vladikr) are you okay with this as a tracker for something without an active enhancement associated with it?

Sorry @lyarwood I thought I replied here. 
Yes, absolutely. It's great to have this to track the maturity of this feature.

lyarwood:
I've moved this to deferred as there is still a desire to address https://github.com/kubevirt/community/pull/318 (will transition into a VEP for v1.7.0) before we graduate the API.

lyarwood:
@0xFelix / @jcanocan - Are there any plans to implement v1beta2 during v1.7.0?

lyarwood:
> [@0xFelix](https://github.com/0xFelix) / [@jcanocan](https://github.com/jcanocan) - Are there any plans to implement v1beta2 during v1.7.0?

[...]

PR #244, opened 2026-03-26
VEP 17: Instancetype API v1 graduation and custom sizing

0xFelix:
## Summary

- Introduces VEP for graduating `instancetype.kubevirt.io` from `v1beta1` to `v1` through an intermediate `v1beta2` version
- **Custom sizing**: instancetypes become non-conflicting defaults - VM-specified values take precedence, aligning with `VirtualMachinePreference` behavior
- **Optional CPU/Memory**: enables device-only instancetypes (e.g., GPU passthrough without mandating compute resources)
- **5-release migration cycle** (v1.9.0 - v1.13.0) with proper storage version management and `kube-storage-version-migrator` support

Tracked by https://github.com/kubevirt/enhancements/issues/17

### DIFF
--- a/veps/sig-compute/17-instancetype-v1/instancetype-v1.md
+++ b/veps/sig-compute/17-instancetype-v1/instancetype-v1.md
@@ -1,0 +1,806 @@
+# VEP #17: Instancetype API v1 Graduation and Custom Sizing
+
+## VEP Status Metadata
+
+### Target releases
+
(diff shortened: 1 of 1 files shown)


0xFelix:
Closing this PR. After careful discussion, the consensus is to stick with standard sizing rather than introducing the complexity of custom sizing. If customizeability is required `VirtualMachineTemplates` can be used.

Issue #246, opened 2026-03-30
VEP 246: Live Migration Compression


michalskrivanek:
**Primary contact (assignee)**:
/assign @michalskrivanek

**Current Feature Stage**:
Alpha

**Feature Gate**:
AdvancedLiveMigration

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/247

**Timeline**:

* Alpha release target: v1.9
* Beta release target:
* Stable release target:
* Alpha
  * VEP PRs: https://github.com/kubevirt/enhancements/pull/247
  * Code PRs: https://github.com/kubevirt/kubevirt/pull/17509
  * Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@michalskrivanek hey, is there any work you are planinig to do this cycle?

michalskrivanek:
not in this cycle. need someone to actually use it before we can move forward with beta

PR #247, opened 2026-03-30
VEP 246: Live Migration Compression

michalskrivanek:
### VEP Metadata

**Tracking issue**: #246
**SIG label**: /sig compute

### What this PR does
Proposes adding a new spec.experimental section to MigrationPolicy as means to add advanced migration options, and adds "compression" option to enable migration stream zstd compression. This reduces network bandwidth consumption for memory-intensive workloads with minimal CPU overhead, especially when combined with the already-enabled multifd parallel migration.

### DIFF
--- a/veps/sig-compute/246-migration-compression/migration-compression.md
+++ b/veps/sig-compute/246-migration-compression/migration-compression.md
@@ -1,0 +1,230 @@
+# VEP #246: Live Migration Compression
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements](https://github.com/kubevirt/enhancements/issues/246)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Enable optional compression for live migration data streams via
+`MigrationPolicy`, reducing bandwidth consumption and improving migration
+times. In Alpha the user selects the algorithm explicitly (`zstd`). The
+long-term goal is automatic enablement, at which point the algorithm
+selection moves to an advanced override or is removed entirely.
+
+## Motivation
+
+Large-memory VMs or bandwidth-constrained environments often hit migration
+timeouts because the dirty rate exceeds the available bandwidth. QEMU/libvirt
+support zstd compression on multifd channels (since libvirt 9.4.0), but
+KubeVirt does not expose this.
+
+Zstd on the existing multifd channels provides ~3× compression with modest
+CPU overhead. Testing on a low-end 1 Gbps / Xeon Silver 4210R setup with
+50 GiB VMs showed:
+
+- **3.6× less data on the wire** (3.3 GiB vs 12 GiB)
+- **~0.4 CPU cores** overhead per VM (8 multifd streams)
+- Migrations that consistently failed without compression always succeeded
+
+Compression trades CPU for bandwidth. This is valuable when bandwidth is the
+bottleneck but not universally beneficial, hence explicit opt-in.
+
+## Goals
+
+- Expose compression in `MigrationPolicy` under `spec.experimental.compression`
+- Keep the default disabled
+- Gated behind the experimental migration options feature gate defined by
+  [VEP 293](https://github.com/kubevirt/enhancements/pull/295)
+- Path toward automatic enablement in future versions
+
+## Non Goals
+
+- Fine-grained tuning (compression level, thread count) in Alpha
+- User-facing documentation of tunables in Alpha
+- Automatic heuristic-based decisions in Alpha (dirty rates during migration
+  are 2–3× higher than pre-migration baseline, making predictions unreliable)
+- Block (disk) migration compression
+
+## Definition of Users
+
+- **Cluster Administrators**: Environments with constrained migration bandwidth
+  or large-memory VMs that frequently time out.
+- **Platform Engineers**: Building migration policies for multi-tenant clusters.
+
+## User Stories
+
+- As a platform engineer, I want to enable compression for memory-heavy VMs
+  via `MigrationPolicy` so they converge within the timeout window without
+  affecting other workloads.
+- As a cluster admin, I want migrations to eventually compress automatically
+  when bandwidth is the bottleneck.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### API and feature gate
+
+Compression is exposed as a field under `spec.experimental` in the
+`MigrationPolicy` CRD. This experimental section,
+its feature gate, and the propagation mechanics are defined by
+[VEP 293: Experimental Migration Options](https://github.com/kubevirt/enhancements/pull/295).
+This VEP only specifies the compression-specific field and its mapping to
+the hypervisor.
+
+The compression field is a string enum: `"none"` (disabled) or `"zstd"`.
+When omitted (`nil`) or set to `"none"`, compression is disabled.
+
+```go
+// +kubebuilder:validation:Enum=none;zstd
+type MigrationCompression string
+
+const (
+    MigrationCompressionNone MigrationCompression = "none"
+    MigrationCompressionZstd MigrationCompression = "zstd"
+)
+```
+
+### Implementation
+
+`virt-launcher` maps the API enum to the libvirt method name via an
+explicit mapping table and sets `VIR_MIGRATE_COMPRESSED` +
+`VIR_MIGRATE_PARAM_COMPRESSION`.
+
+### How it maps to libvirt/QEMU
+
+| KubeVirt API | libvirt |
+|---|---|
+| `experimental.compression: "zstd"` | `VIR_MIGRATE_COMPRESSED` flag + `VIR_MIGRATE_PARAM_COMPRESSION = "zstd"` |
+
+With multifd active (default when no CPU limit is set), zstd runs per
+multifd channel. Without multifd, QEMU falls back to single-threaded zstd.
+
+### Interaction with existing features
+
+| Feature | Interaction |
+|---|---|
+| Multifd (parallel) | Compatible — zstd runs per channel |
+| Auto-converge | Compatible — orthogonal |
+| Post-copy | Compatible — applies to pre-copy phase |
+| TLS | Compatible — compression before encryption |
+| Bandwidth limit | Compatible — compressed data counts toward limit |
+
+## API Examples
+
+```yaml
+apiVersion: migrations.kubevirt.io/v1alpha1
+kind: MigrationPolicy
+metadata:
+  name: compress-heavy-vms
+spec:
+  selectors:
+    namespaceSelector:
+      workload-type: memory-intensive
+  experimental:
+    compression: zstd
+```
+
+## Design Rationale
+
+### Why MigrationPolicy only
+
+Compression is a bandwidth-for-CPU tradeoff that depends on workload
+characteristics. `MigrationPolicy` provides the right granularity — admins
+target specific namespaces or VM classes without risking cluster-wide
+regressions.
+
+### Why `spec.experimental`
+
+The explicit algorithm selection is intended as a transitional API. The
+long-term goal is automatic enablement based on dirty rate, bandwidth, and
+convergence monitoring. If that proves reliable, compression becomes an
+implementation detail and the explicit knob either moves to an
+advanced-only section or is removed. The `spec.experimental` section and
+its lifecycle are defined by
+[VEP 293](https://github.com/kubevirt/enhancements/pull/295).
+
+### Consideration for future improvements
+
+Concerns about high CPU usage can be reduced if QEMU can support dynamically changing
+compression setting during a migration. New algorithms like QATzip (already present in QEMU)
+or lz4 can reduce the CPU usage further, altering the cost equation between the effective
+transfer speed vs additional CPU usage.
+
+## Alternatives
+
+1. **Boolean toggle**: Simpler, but no room for additional algorithms.
+2. **Cluster-wide default in KubeVirt CR**: Risks regressions for
+   CPU-constrained workloads. Can be reconsidered later.
+3. **Full configuration** (level, threads): Too much API surface for Alpha.
+4. **Always enable**: CPU overhead makes this unsuitable as a default.
+5. **Automatic-only**: Dirty-rate predictions not reliable enough yet.
+6. **xbzrle instead of zstd**: xbzrle is cache-based; zstd gives better
+   general-purpose ratio and works naturally with multifd.
+
+## Scalability
+
+No concerns. Compression runs in virt-launcher on the source node. CPU
+overhead is modest at zstd level 1 and bounded by migration bandwidth.
+
+## Update/Rollback Compatibility
+
+- **Update**: Field defaults to `nil` (disabled). Existing migrations unaffected.
+- **Rollback**: Feature gate removal causes the field to be ignored.
+  No persistent state changes.
+- **Mixed-version**: Source virt-launcher controls compression. Target
+  decompresses transparently.
+
+## Functional Testing Approach
+
+1. **Unit**: `generateMigrationFlags()` includes `MIGRATE_COMPRESSED` and
+   params include `Compression: "zstd"` when enabled.
+2. **E2E**: Migration succeeds with compression; verify non-zero compression
+   bytes in job stats.
+3. **Negative**: `spec.experimental` ignored when gate is disabled.
+
+## Implementation History
+
+<!-- Updated as implementation progresses -->
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Experimental migration options framework from
+      [VEP 293](https://github.com/kubevirt/enhancements/pull/295)
+      (feature gate, `spec.experimental` section, propagation path)
+- [ ] `MigrationCompression` enum and `spec.experimental.compression` field
+      added to `ExperimentalMigrationConfiguration`
+- [ ] `virt-launcher` maps API enum to libvirt compression method and
+      sets hypervisor flags
+- [ ] E2E test
+
+### Beta
+
+- [ ] Soak testing confirming acceptable CPU overhead
+- [ ] Investigate automatic enablement (dirty-rate / bandwidth / convergence)
+- [ ] If viable: implement automatic, move explicit algorithm to
+      advanced override or remove
+- [ ] Compression metrics in Prometheus
+
+### GA
+
+- [ ] Automatic enablement proven or explicit opt-in confirmed as long-term API
+- [ ] Stable across multiple releases


michalskrivanek:
> Was there any work done to explore the impact of the CPU and memory overhead that surely exists, and from my knowledge, is far from negligible? Why are we rushing to add API before we have any data and prove that this is useful? 

it's a tradeoff between CPU overhead and available bandwidth. There always is a bandwidth limitation, with VMs generally getting bigger over time it's more pronounced, if you're stuck with low end networking (let's say 1Gb these days)

As for the overhead, with the suggested default settings with multifd that's 1.5MB/stream on both sending and receiving side. That is pretty negligent. 
As for CPU *roughly* - on my dated Xeon Silver 4210R CPU @ 2.40GHz it's ~ 500MB/s, this is per stream as well. (btw on Apple M4 it's 6GB/s). 
Combination with autoConverge - considering we slow down the guest CPU pretty aggressively (down to 1% in 10 iterations) the additional CPU overhead is not really taken away from the guest in this case.


> It is very easy to use KubeVirt's external provider to deploy it on an existing Kubernetes based cluster, with modified code to simulate different scenarios.

The tradeoffs are pretty clear, the biggest unknown is about the actual real world workloads, and whether one wants to sacrifice CPU in order to save bandwidth.

> I am not against a new API, but I don't understand why we are adding an API to collect data instead of collecting data first and then designing a feature around it.

the compression feature itself is ~14 years old, its API is stable and proven. 
Again, this is about tradeoff, there's no data to show it's absolutely the best thing ever and it should be used as a default from now on. Absolutely not. 
But it's pretty trivial to show that with specific workload the throughput goes down by X% at the expense of Y% CPU.
I'll dig something out, though I'm not sure there were many tests with the newer zstd
(there is a dated comparison of the legacy multithread compression in https://lists.gnu.org/archive/html/qemu-devel/2020-11/msg02207.html , I would use that as a rationale for picking zstd)

Aseeef:
I also want to add in support of migration compression: We have a large-scale real world analogy to this proposal in the form of zram. Ram compression has been pretty successful achieving compression ratios of 2-4:1 (according to data in https://notes.xeome.dev/notes/Zram) with zstd.

Finding data on CPU overheads was harder, but I would imagine if zstd works well enough on real systems for real-time ram compression, there is a good chance it will work well here in almost all cases...?

[This also brings the subject that if the underlying VM is already using ram compression, then we'd just be wasting CPU cycles re-compressing that data. But I think even so, you can expect enabling compression to help most end-users because typically it's the less frequently accessed parts of ram that get compressed which in a typical migration would probably only get transferred once anyway...]

Further improvements as I suggested in my comments could be made if support for changing compression of an inflight migration is added to QEMU (very well-scoped and technically feasible addition according to AI, though I haven't personally looked at it). This could allow us to switch between compress + don't compress and observe if compression helps balancing CPU and bandwidth in the future.

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,256 @@
...
+**`MigrationConfiguration`** (in `staging/src/kubevirt.io/api/core/v1/types.go`):
+
+```go
+type MigrationConfiguration struct {
+    // ... existing fields ...
+
+    // MigrationCompression enables zstd compression of the live migration
+    // data stream, reducing network bandwidth at the cost of additional CPU.
+    // Defaults to false
+    MigrationCompression *bool `json:"migrationCompression,omitempty"`
```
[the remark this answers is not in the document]
michalskrivanek: whatever is acceptable for minimal API. never/always would work for the starts, we can add auto or even the different methods later on when we have them.

iholder101:
In terms of API, I think there are two concerns here:

Firstly, Tuning migrations is not a trivial thing to do. Even KubeVirt developers are sometimes having a hard time to capture all of the different options, which means it might be extremely hard for some of the users. Since live-migration is a core feature that is so important, the direction that was discussed multiple times in the past is moving towards a more declarative API in which the user would express what he wants to achieve, or what's the nature of his workloads, and KubeVirt would be the one deciding which lower level tunings to use to make it happen.

Some work was done to set the ground for that. We now have dirty-rate tracking for example, and the idea was to automatically set paused migration / post copy according to the migration bandwidth and dirty-rate. In this case of compression, I would try to think "given a certain dirty-rate, bandwidth and available resources on the node - when is it the right thing to just enable compression behind the scenes?". Adding more and more tunables is not scalable from the aspect of the ability to reason about everything at ones by users that are not expert on the matter.

Secondly, we're trying to decouple ourselves from Libvirt and QEMU. Our long term vision is to not depend on these at all and allow more hypervisors (e.g. Cloud-Hypervisor) to be supported as first-class citizens. KubeVirt's API is already too coupled to libvirt and we're trying to move away from this approach.

michalskrivanek:
> In terms of API, I think there are two concerns here:
> 
> Firstly, Tuning migrations is not a trivial thing to do. Even KubeVirt developers are sometimes having a hard time to capture all of the different options, which means it might be extremely hard for some of the users. Since live-migration is a core feature that is so important, the direction that was discussed multiple times in the past is moving towards a more declarative API in which the user would express what he wants to achieve, or what's the nature of his workloads, and KubeVirt would be the one deciding which lower level tunings to use to make it happen.

fair, what is your suggestion for such declarative API in this case? It cannot be decided without any user input really, it's a tradeoff one needs to be willing to make between "higher chance of migration success due to higher effective bandwidth" vs "additional cpu load on the node".

> Some work was done to set the ground for that. We now have dirty-rate tracking for example, and the idea was to automatically set paused migration / post copy according to the migration bandwidth and dirty-rate. In this case of compression, I would try to think "given a certain dirty-rate, bandwidth and available resources on the node - when is it the right thing to just enable compression behind the scenes?". Adding more and more tunables is not scalable from the aspect of the ability to reason about everything at ones by users that are not expert on the matter.

This is not necessarily meant for end-users, I can imagine an automated component to decide when to use it - but it does need _some_ API in kubevirt to use.

> Secondly, we're trying to decouple ourselves from Libvirt and QEMU. Our long term vision is to not depend on these at all and allow more hypervisors (e.g. Cloud-Hypervisor) to be supported as first-class citizens. KubeVirt's API is already too coupled to libvirt and we're trying to move away from this approach.

that's fair as well, but that's again a separate issue - there is at least one such hypervisor offering this kind of tradeoff (compression), so what is the right API that's not closly coupled to libvirt? One way or the other the actual implementation of such feature will always be hypervisor dependent.

iholder101:
> It cannot be decided without any user input really, it's a tradeoff one needs to be willing to make between "higher chance of migration success due to higher effective bandwidth" vs "additional cpu load on the node".

> This is not necessarily meant for end-users, I can imagine an automated component to decide when to use it - but it does need _some_ API in kubevirt to use.

I agree there's a tradeoff, but I'm not sure I agree that the user must take this decision.

Let's say that we do have an API for it. Now the question becomes - how should this API be used? Which logic do we have to comply to in order to take use of this in an optimized way that makes sense?

If we're able to intelligently decide, based on some real-time factors, when it makes sense to use compression and how - let's just apply it on the user's behalf. Essentially - this is what an "automated component" would do exactly, so we can achieve the same with less APIs. And if we can't intelligently decide that yet, then my point is that the user (or a third level component) would as well not have the answer.

michalskrivanek:
> If we're able to intelligently decide, based on some real-time factors, when it makes sense to use compression and how - let's just apply it on the user's behalf. Essentially - this is what an "automated component" would do exactly, so we can achieve the same with less APIs. And if we can't intelligently decide that yet, then my point is that the user (or a third level component) would as well not have the answer.

TBH I do not see a tangible data to be able to automatically decide. Even that other component would need some way how to express "yeah for this VM it makes sense to compress, but not for the other", it just could be a bit more higher level and/or specific to certain deployments so it can make assumptions generic kubevirt cannot.
just a point in case - I tried to compare the ongoing dirty rate with the actual dirty rate during migration in my setup and the difference is 2-3 times....I'm not sure its really usable to make an automatic decision. Aseef's earlier idea of maybe a new feature to flip compression on _during_ the migration is interesting, but it doesn't exist yet. And you'd still probably want some place to express "yeah i really want it to switch in these conditions". Even if that owuld be a low level API.

iholder101:
> would need some way how to express "yeah for this VM it makes sense to compress, but not for the other"

My question is - as a user - how can I know whether it makes sense or not?

michalskrivanek:
> > would need some way how to express "yeah for this VM it makes sense to compress, but not for the other"
> 
> My question is - as a user - how can I know whether it makes sense or not?

Most naively - your migrations are not usually converging and you desperately want them to (or you want to speed them up), and you do not care about additional CPU load on nodes.
Whether that's the right setting or not depends on other factors that kubevirt cannot know, some would rather use post-copy, some would maybe prefer higher max_downtime....it depends on what they want to accept.
 I do not see a way how you can make a decision for the user without their input.
For compression specifically it's not hard to describe what it means and what are the tradeoffs in a hypervisor- or algorithm-agnostic way

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+4. **MigrationPolicy** (`staging/src/kubevirt.io/api/migrations/v1alpha1/types.go`):
+   In `GetMigrationConfByPolicy()`, propagate `MigrationCompression` override
+   the same way as `AllowAutoConverge`.
+
+### How it maps to libvirt/QEMU
+
+| KubeVirt API                      | libvirt flag / param                        |
+|-----------------------------------|---------------------------------------------|
+| `migrationCompression: true`      | `VIR_MIGRATE_COMPRESSED` flag               |
```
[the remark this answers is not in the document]
michalskrivanek: I think both is useful. I disagree with _not_ exposing knobs that can solve issues. You are *never* able to catch 100% of the usecases with a single magic implementation. There are always valid reasons outside of the "system" that the system can't know about and can't take into account. For that you need the knobs to exist.
That's not to say you shouldn't strive for best default. "auto" makes sense too, and we should try to make it work for as many usecases as possible. That would currently be zstd and indeed can change in 5y. But even in 10y there could be a very valid reason for someone to choose zstd

I think enum with "auto" and "zstd" and "none" makes sense - auto would be zstd and subject o change, default is still none, and zstd wouldbe explicit choice

@iholder101, thoughts? does it resolve your API concern?

michalskrivanek:
updated the knob to enum, add rationale and example for compression use, and for choosing zstd as the default algorithm
@vladikr @iholder101 @Aseeef PTAL

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,324 @@
...
+```go
+type MigrationConfiguration struct {
+    // ... existing fields ...
+
+    // MigrationCompression selects the compression strategy for the live
+    // migration data stream. "auto" uses the best available algorithm
+    // (currently zstd), "zstd" selects zstd explicitly, and "" (empty/default)
+    // disables compression.
+    //+optional
+    MigrationCompression *MigrationCompressionMode `json:"migrationCompression,omitempty"`
```
Aseeef: Im skeptical whether the inclusion of "MigrationCompressionMode" that lets users pick the compression algorithm to use is fully justifiable right now. I understand that the intention is that in the future, we might want to support other compression algorithms like QATzip, but perhaps it would be better to hold off on exposing this as a configurable field until then.

In fact, it might not even make sense even after QATzip is support is added to allow this to be a CR configurable field. This is because it is entirely possible that some nodes might support QATzip and others don't so to specify QATzip in the CR wouldn't make sense.

So I think maybe its better to repurpose this field to just be whether compression is enabled. Something like 
```go
type CompressionStatus string

const (
    MigrationCompressionDisabled CompressionStatus = "disabled"
    MigrationCompressionAuto CompressionStatus =  "auto"
    MigrationCompressionEnabled CompressionStatus = "enabled"
)
type MigrationConfiguration struct {
  ...
  MigrationCompression *CompressionStatus `json:"migrationCompression,omitempty"`
}
```

TLDR, I think we should defer the field that allows you to specify the compression algorithm for a future addition since right now it doesn't seem super useful. And instead repurpose this field only to specify whether compression is enabled.

michalskrivanek: That's why I like the enum more than a bool. Even if there would be just two values for now it allows extending the methods (or auto-enablement once it exists) easily. 
Oh I see I left the "mode" in the text, let me fix that, I meant to expose MigrationCompression only, as currently two values, "zstd" or "none" 

(as for qatzip in its current shape - I'm not suggesting to add it yet, however it does have zlib fallback so it would work. Poorly, with a lot more overhead than zstd)

michalskrivanek: done in 704f94d

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+- Migrations that consistently failed without compression always succeeded
+
+Compression trades CPU for bandwidth. This is valuable when bandwidth is the
+bottleneck but not universally beneficial, hence explicit opt-in.
+
+## Goals
+
+- Expose compression in `MigrationPolicy` under `spec.experimental.compression`
+- Keep the default disabled
+- Guard behind a single `AdvancedLiveMigration` Alpha feature gate
```
Aseeef: What do you think about something like `ExperimentalMigrationOptions` for the FG name?

Aseeef: Also if I understand correctly, this FG is intended to test current and future experimental options for live migration right? So will this perpetually be an Alpha FG?

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+  covering the entire `spec.experimental` section
+- Path toward automatic enablement in future versions
+
+## Non Goals
+
+- Cluster-wide default in KubeVirt CR in Alpha
+- Fine-grained tuning (compression level, thread count) in Alpha
+- Automatic heuristic-based decisions in Alpha (dirty rates during migration
```
Aseeef, line 58: > dirty rates during migration
  are 2–3× higher than pre-migration baseline.

Out of curiosity: Could you talk more about this (not in the VEP, I mean in the comments). Why does the act of live migration increase dirty rates by 2-3x?

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+`MigrationPolicySpec` and `MigrationConfiguration`, propagating through
+the existing policy resolution path.
+
+### Feature gate
+
+Rather than gating each experimental knob individually, gate the entire
+`spec.experimental` section behind a single `AdvancedLiveMigration`
+feature gate at Alpha. This covers compression and any future experimental
+fields without proliferating gates. The `experimental` block is ignored
+when the gate is disabled.
```
Aseeef: OK good, this approach is similar to k8s: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md
(see section under New field in existing API version)

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+- [ ] Feature gate `AdvancedLiveMigration` guards `spec.experimental`
+- [ ] `MigrationCompression` enum and `spec.experimental.compression` field
+      added to `MigrationPolicySpec` via shared `ExperimentalMigrationConfiguration`
+- [ ] `GetMigrationConfByPolicy()` propagates the override
+- [ ] `virt-handler` gate-checks and passes setting to `virt-launcher`
+      via `MigrationOptions.Experimental`
+- [ ] `virt-launcher` maps API enum to libvirt compression method and
+      sets hypervisor flags
+- [ ] E2E test
+- [ ] User guide docs
```
Aseeef: Especially given the experimental nature of API do we want to commit to a user doc this early in alpha?

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+)
+
+type ExperimentalMigrationConfiguration struct {
+    //+optional
+    Compression *MigrationCompression `json:"compression,omitempty"`
+}
+```
+
+The same `ExperimentalMigrationConfiguration` struct is reused in both
+`MigrationPolicySpec` and `MigrationConfiguration`, propagating through
```
Aseeef: "Cluster-wide default in KubeVirt CR in Alpha" is listed as a non-goal above. But this implies `ExperimentalMigrationConfiguration` will be a part of the cluster wide config.

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+- [ ] User guide docs
+
+### Beta
+
+- [ ] Soak testing confirming acceptable CPU overhead
+- [ ] Investigate automatic enablement (dirty-rate / bandwidth / convergence)
+- [ ] If viable: implement automatic, move explicit algorithm to
+      advanced override or remove
+- [ ] Compression metrics in Prometheus
+
```
Aseeef, line 226: Not sure if this is worth mentioning in the VEP under the beta section but as previously mentioned, a lot of the concerns with CPU usage can be reduced if QEMU can support dynamically changing compression setting during a migration. Or perhaps if support for lz4 compression is added (faster alternative to zstd)... Some changes QEMU side can significantly alter the cost equation on the side of having some sort of compression by default eventually in KubeVirt.

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,241 @@
...
+```go
+// +kubebuilder:validation:Enum=none;zstd
+type MigrationCompression string
+
+const (
+    MigrationCompressionNone MigrationCompression = "none"
+    MigrationCompressionZstd MigrationCompression = "zstd"
+)
+
+type ExperimentalMigrationConfiguration struct {
```
Aseeef: In #249 you call this `ExperimentalMigrationPolicy`. But I think you want to use ths same struct (right?)

veps/sig-compute/246-migration-compression/migration-compression.md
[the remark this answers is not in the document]
michalskrivanek: we'll go with `.spec.experimental` as discussed in other comments

veps/sig-compute/246-migration-compression/migration-compression.md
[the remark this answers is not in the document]
michalskrivanek: changed to a string enum for more flexibility

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: I haven't looked at more details TBH, just observed in my runs. It could be that the sampling is just not really hitting the right spots (it's a 1s sample of 0.5% of pages with uniform distribution), or that when you actually go and read every single page the cache eviction has such high negative effect on the application/workload.

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: I would want to see `.spec.experimental` and `.spec.advanced` coexist.
So I named it `AdvancedLiveMigration` as I see it as a way to customize and fine tune even after the individual features are stable. But maybe we can leave that for the new hypothetical VEP, there's nothing to put to `.spec.advanced` right now when these newly proposed things are at alpha.

As for the `.spec.experimental` alone, from the API POV woulnd't it be sufficient to keep the API as perpetual alpha forever, not requiring any gate?

veps/sig-compute/246-migration-compression/migration-compression.md
Aseeef: I see. That is just very bizarre to me that the act of live migration could increase dirty rates by that much, it doesn't make sense.

veps/sig-compute/246-migration-compression/migration-compression.md
Aseeef: Here is the PR for that hypothetical VEP: https://github.com/kubevirt/enhancements/pull/295

I also looked at how k8s does there experimental fields, and they do protect it with feature gates too. So based on that, I think it makes sense for us to do that too. Also, I don't think its worth having two different `.spec.experimental `and `.spec.advanced `. Can you check this VEP out and see if it is inline with your vision?

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: yes. I hope once we gather more accurate metrics we'll be able to identify some pattern.
It's not new, there's been workloads where live migration caused huge increase in activity (e.g., RHV with JBoss workloads ~10y ago) but there never has been monitoring of ongoing dirty rates, so maybe now we'll have a chance...

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: I would remove it from the goals then, the policy structs are included in both kubevirt and vmi so it is indeed present there too

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: added "Consideration for future improvements" section

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: right, thanks for noticing. i'll fix it there

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: Thank you for opening it, it probably makes sense to clarify indeed, even if it just serves as a record of how we got to the final state.

My vision was:
`.spec.experimental `for new things we do not really know how they'll behave or we're not sure about optimal defaults
`.spec.advanced` for permanent low level tunables, functionality is mature, but there are justifiable cases where users may want to change
`spec.<top_level>` for permanent, user-facing, well documented options

I personally do not really care how the feature gates are structured, I do not think they serve much of a purpose here, but I do not mind either way.

Aseeef:
Since the state of VEP 293 (experimental API) is up in the air, and VEP freeze days away, one way around it could be "we will worry about API later in Beta" and make it so simply enabling the feature gate unconditionally enables compression. Just a thought. I do hope we can reach an agreement on 293 though.

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,230 @@
...
+- As a cluster admin, I want migrations to eventually compress automatically
+  when bandwidth is the bottleneck.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### API and feature gate
```
Aseeef, line 82: I don't know, I have been bothered by the `spec.experimental.migrationCompression` but I couldn't formulate why... I think I see the problem now. This is dead code/dead API.

This is because you mention:
> In Alpha the user selects the algorithm explicitly (`zstd`). The
> long-term goal is automatic enablement, at which point the algorithm
> selection moves to an advanced override or is removed entirely`

So especially given that `spec.experimental.migrationCompression`  is already intended to likely to removed by GA and currently only has two options: `zstd` and `none`, there is no point in having this API and I think it should be completely removed for Alpha. It does nothing except control whether compression is enabled or disabled. For alpha, I think it is better to just use the feature gate to enable/disable compression.

Then you can mention in the VEP *possible* APIs for what a GA API might have (i.e. being able to change algorithms, policies for even potentially dynamically deciding whether to use compression or not, or as I previously mentioned even turning compression on/off mid-migration if QEMU adds support for it). You can mention something along the lines of how we do not have enough information to commit to an API at the moment in the VEP.

veps/sig-compute/246-migration-compression/migration-compression.md
```diff
@@ -0,0 +1,230 @@
...
+
+Compression is exposed as a field under `spec.experimental` in
+`MigrationPolicy` and `MigrationConfiguration`. The experimental section,
```
iholder101: Should we focus on https://github.com/kubevirt/enhancements/issues/294 then, and come back to evaluate this VEP with an API that won't expose direct libvirt knobs?

michalskrivanek: @iholder101 I don't understand the "libvirt knobs" exposure here. The algorithm choice is limited by what QEMU and libvirt supports, yes, but I do not see how you can avoid setting a libvirt parameter when that is the interface you have to currently use.

IOW it's not a "compression libvirt knob" that's exposed, it's a "compression algorithm".

veps/sig-compute/246-migration-compression/migration-compression.md
michalskrivanek: my intention was to move it to `spec.advanced` if it eventually gets created. 
I thought we settled on AdvancedExperimentalMigrationOptions FG (in #295) and with that the `spec.experimental.migrationCompression` makes sense to me

You cannot do it via another FG only, this option needs to be applied via MigrationPolicies to a subset of VMs, not unconditionally to all.

veps/sig-compute/246-migration-compression/migration-compression.md
iholder101: I'm not sure the users even need to decide or be aware of the algorithm to be used. But let's not drift over to this discussion.

In any case - I think we can fold this one to https://github.com/kubevirt/enhancements/issues/294. There you can introduce the fields which is anyways what this VEP proposes. Then, in the next cycle, we can re-think (alongside some data that you'd able to collect) how to integrate it within our APIs.

Does that make sense?

michalskrivanek: sure, I did refer to it in an update rather than repeating
Unfortunately I won't be able to make text updates until monday, but it sounds good to me

veps/sig-compute/246-migration-compression/migration-compression.md
Aseeef: > You cannot do it via another FG only, this option needs to be applied via MigrationPolicies to a subset of VMs, not unconditionally to all.

Good point... I forgot about this.

iholder101:
Thanks @michalskrivanek,

Let's get this in.
I emphasize that the libvirt knobs exposed here should be temporary and would not end up in the API in the final state of this VEP.
By the next cycle we will go to the next alpha phase and revisit the API to remove them.

By that time, I expect you would have some data that will allow us to plan how we move forward, alongside any relevant data that will represent how compression will affect CPU and memory for both the Burstable and Guaranteed QoS cases.

/lgtm

vladikr:
> Thanks @michalskrivanek,
> 
> Let's get this in. I emphasize that the libvirt knobs exposed here should be temporary and would not end up in the API in the final state of this VEP. By the next cycle we will go to the next alpha phase and revisit the API to remove them.
> 
> By that time, I expect you would have some data that will allow us to plan how we move forward, alongside any relevant data that will represent how compression will affect CPU and memory for both the Burstable and Guaranteed QoS cases.
> 
> /lgtm

Extremely important from my perspective as well! 
/approve
/hold
@Aseeef, please remove the hold when you are satisfied.

veps/sig-compute/246-migration-compression/migration-compression.md
iholder101: @Aseeef brought up a good point that elsewhere we agreed to add this to MigrationPolicy only - not to the Kubevirt CR.

To emphasize - this VEP will need to be updated and this is a blocker for the implementation PR.
FYI @michalskrivanek

Aseeef: @iholder101, this has been addressed. PR looks fine to me now.

vladikr: > @Aseeef brought up a good point that elsewhere we agreed to add this to MigrationPolicy only - not to the Kubevirt CR.
> 
> To emphasize - this VEP will need to be updated and this is a blocker for the implementation PR. FYI @michalskrivanek

@Aseeef has pushed an updated version that addresses these issues, so I'm reinstating your lgmt @iholder101 
/lgtm

PR #249, opened 2026-04-01
VEP 291: Dynamic Downtime Tuning and Iteration-Based Convergence

michalskrivanek:
### VEP Metadata
**Tracking issue**: #291
**SIG label**: /sig compute

### What this PR does
Proposes an iteration-aware algorithm for gradually increasing QEMU's
`max_downtime` during live migration to improve convergence. The approach
is built on .spec.experimental added by #247, fits entirely within the existing
virt-launcher monitoring loop, and is completely optional.
Instead of the current all-or-nothing behavior (wait for timeout → pause /
post-copy / abort), the algorithm tracks `MemIteration` from libvirt job
stats and linearly increases `max_downtime` up to a configurable ceiling
with sane defaults. This lets QEMU converge naturally with sub-second
switchover downtime rather than requiring drastic intervention.

### Related VEPs
- **VEP 246: Live Migration Compression** (#246 / PR #247) — Adds zstd compression; shares the `spec.experimental` section and `ExperimentalMigrationPolicy` feature gate.
- **VEP 248: Live Migration Stall Detection** (#248 / PR #252) — Complementary stall detection that could replace the fixed `startAfterIteration` with automatic detection.

### DIFF
--- a/veps/sig-compute/291-dynamic-downtime-tuning/vep.md
+++ b/veps/sig-compute/291-dynamic-downtime-tuning/vep.md
@@ -1,0 +1,336 @@
+# VEP 291: Live Migration — Dynamic Downtime Tuning and Iteration-Based Convergence
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Improve live migration convergence by gradually increasing QEMU's
+`max_downtime` as migration iterations progress, with a progressive ramp up of a `target_downtime` up to the configured `maxDowntimeMs`, replacing the current all-or-nothing approach (run until timeout, then pause / post-copy / abort).
+
+The feature is activated by the presence of `spec.experimental.downtimeTuning`
+in a `MigrationPolicy`. It ramps `max_downtime` up to the `maxDowntimeMs`
+ceiling (see [VEP 252](https://github.com/kubevirt/enhancements/pull/252)).
+
+## Motivation
+
+### Current behavior
+
+KubeVirt's migration monitor makes convergence decisions based purely
+on wall-clock time:
+
+1. **`completionTimeoutPerGiB × vmSize`** — if exceeded, trigger
+   post-copy / pause / abort.
+2. **`progressTimeout`** — if `DataRemaining` hasn't decreased for N
+   seconds, abort.
+
+This has several problems:
+
+- **No awareness of iteration progress.** QEMU reports `MemIteration`
+  but KubeVirt ignores it. A VM that has completed 15 iterations and
+  is close to converging gets the same treatment as one that hasn't
+  finished even the initial bulk copy.
+
+- **`max_downtime` is never tuned.** QEMU's default 300ms is too low
+  for workloads with moderate dirty rates — estimated remaining time
+  can hover just above it, never converging.
+
+- **Time-based decisions penalize large VMs.** Large VMs are granted
+  excessive timeouts even though their active memory (working set)
+  does not necessarily scale with VM size.
+
+### Proposed behavior
+
+QEMU switches over when `estimated_remaining_time ≤ max_downtime`.
+By gradually raising `max_downtime` as iterations progress (inspired
+by oVirt's migration policy engine), we give QEMU progressively more
+room to converge *before* existing timeouts fire, without overriding
+existing safety nets. If the ceiling is reached without convergence,
+the existing time-based logic still applies.
+
+## Goals
+
+- Activate via presence of `spec.experimental.downtimeTuning` in
+  `MigrationPolicy`, gated behind the experimental migration options
+  feature gate defined by
+  [VEP 293](https://github.com/kubevirt/enhancements/pull/295)
+- Use `maxDowntimeMs` (top-level field per
+  [VEP 252](https://github.com/kubevirt/enhancements/pull/252)) as the
+  ceiling for the ramp
+- Keep disabled by default
+- Preserve all existing timeout, progress, and abort semantics as fallback
+
+## Non Goals
+
+- Cluster-wide default in KubeVirt CR in Alpha
+- Replacing post-copy / pause / auto-converge mechanisms
+- Bandwidth or CPU throttling
+- Changing migration behavior for any user who does not opt in
+
+## Definition of Users
+
+- **Cluster Administrators**: Workloads with moderate-to-high dirty rates
+  that frequently time out or require post-copy/pause.
+- **Platform Engineers**: Wanting finer convergence control without
+  enabling auto-converge (which throttles guest CPU).
+
+## User Stories
+
+- As a cluster admin, I want large-memory VMs to converge with a bounded,
+  sub-second switchover pause instead of hitting a timeout and being
+  forcibly paused, falling back to post-copy, or being aborted.
+- As a platform engineer, I want to enable downtime tuning for a subset
+  of VMs via `MigrationPolicy` without affecting the rest of the cluster.
+
+## Repos
+
+kubevirt/kubevirt.
+
+## Design
+
+### API and feature gate
+
+Downtime tuning is activated by the presence of
+`spec.experimental.downtimeTuning` in a `MigrationPolicy`, gated
+behind the experimental migration options feature gate
+([VEP 293](https://github.com/kubevirt/enhancements/pull/295)).
+
+The algorithm ramps `max_downtime` up to the `maxDowntimeMs` ceiling —
+a field in `MigrationPolicySpec` / `MigrationConfiguration`.
+This VEP reuses that field; see VEP 252 for its definition and
+semantics.
+
+Parameters under `spec.experimental.downtimeTuning`:
+
+| Parameter             | Default | Description                                          |
+|-----------------------|---------|------------------------------------------------------|
+| `initialMs`           | 150     | Starting `target_downtime` (lower than QEMU's 300ms) |
+| `steps`               | 7       | Equal steps from initial to `maxDowntimeMs` ceiling  |
+| `startAfterIteration` | 3       | Start cooldown after this iteration                  |
+| `cooldownSeconds`     | 10      | Minimum seconds between increases                    |
+
+An empty `downtimeTuning: {}` enables tuning with all defaults.
+
+### Implementation and algorithm
+
+`virt-launcher` runs the tuning algorithm in the existing monitoring loop.
+
+On each monitoring pass, `processInflightMigration` reads
+`MemIteration` and elapsed time. The tuning logic:
+
+1. If `experimental.downtimeTuning` is absent → no-op.
+2. On first pass set the initial `target_downtime` → `dom.MigrateSetMaxDowntime(initialMs)`.
+3. If `MemIteration < startAfterIteration` → wait.
+4. When threshold reached → start cooldown timer.
+5. After `cooldownSeconds` elapse → bump by one step, capped at
+   `maxDowntimeMs`, call `dom.MigrateSetMaxDowntime()`, reset cooldown.
+6. Log each adjustment for observability.
+
+All existing timeout / abort / post-copy / pause logic is unmodified.
+
+With defaults (`maxDowntimeMs=1050`, `steps=7`, cooldown 10s):
+
+```
+time:         0s    ~75s        ~85s   ~95s   ~105s   ~115s   ~125s   ~135s
+iter:         1     3           ~10    ~20    ~30     ~40     ~50     ~60
+max_downtime: 150   150(start)  300    450    600     750     900     1050
+```
+
+The time-based cooldown prevents rapid escalation on fast-iterating
+1 Gbps links where iterations advance every ~400ms. At 10s cooldown,
+each step spans ~10 GiB of transferred data at 1 Gbps — enough for
+multiple full iterations, giving QEMU a fair chance to converge at
+each downtime level.
+
+### Interaction with existing features
+
+| Feature               | Interaction                                                           |
+|-----------------------|-----------------------------------------------------------------------|
+| completionTimeoutPerGiB | Unmodified. Tuning may help converge before timeout fires.          |
+| progressTimeout       | Unmodified. Abort-on-stuck still works.                               |
+| AllowPostCopy         | Unmodified. Post-copy triggers after timeout if tuning didn't help.   |
+| AllowAutoConverge     | Compatible. Auto-converge throttles CPU; this increases downtime tolerance. They complement each other. |
+| VFIO switchover       | Unmodified. Separate code path on timeout.                            |
+| MigrationCompression   | Compatible. Compression reduces data; tuning helps convergence timing.|
+
+The stall-detection work proposed in
+[VEP 252](https://github.com/kubevirt/enhancements/pull/252) attempts a smarter and more comprehensive approach, at the price of higher complexity.
+
+### Example scenario
+
+50 GiB VM, dirty rate ~700 Mbps, migration bandwidth ~900 Mbps,
+`maxDowntimeMs = 1050`, cooldown 10s.
+Because the dirty rate is close to the available bandwidth, QEMU
+makes progress but DataRemaining oscillates around a plateau —
+estimated downtime hovers around 500–800ms, always above QEMU's
+default `max_downtime`.
+
+| Iteration | Time   | max_downtime | Action                                             |
+|-----------|--------|--------------|----------------------------------------------------|
+| 1–3       | 0–70s  | 150ms        | Bulk copy; below `startAfterIteration`, no cooldown |
+| 4         | ~70s   | 150ms        | Threshold reached → cooldown timer starts           |
+| ~30       | ~80s   | 300ms        | Iter advanced AND 10s elapsed → first bump          |
+| ~55       | ~90s   | 450ms        | Iter advanced AND 10s elapsed → second bump         |
+| ~80       | ~100s  | 600ms        | est. < max → QEMU switches over                    |
+
+**Without tuning**: estimated downtime stays around 500–800ms, always
+above QEMU's default 300ms — migration never converges and eventually
+hits the completion timeout.
+
+**With tuning**: `max_downtime` catches up to `estimated_downtime`
+at ~600ms — sub-second switchover, bounded by the ceiling.
+
+### Benchmark results
+
+Tested on 1 Gbps network, 8 GiB OCP HCP worker node VM (idle hosted
+cluster workload, avg dirty rate ~650 Mbps). Defaults: start after
+iteration 3, initial 150ms, 7 steps with 10s cooldown, up to 1050ms.
+10 migrations with tuning vs 10 without:
+
+| Metric               | With Tuning | Without Tuning | Delta          |
+|----------------------|-------------|----------------|----------------|
+| Avg Total Time       | 92.2s       | 112.4s         | −18% (20s)     |
+| Avg Convergence Time | 24.7s       | 45.0s          | −45%           |
+| Median Convergence   | 24.3s       | 35.0s          | −31%           |
+| Avg Downtime         | 611ms       | 451ms          | +160ms (+35%)  |
+| Avg Iterations       | 28          | 69             | −59%           |
+| Avg Data Transferred | 9,694 MiB   | 11,862 MiB     | −18%           |
+| Worst Total Time     | 98s         | 161s           | −39%           |
+| Convergence Range    | 19–30s (11s)| 24–93s (69s)   | 6× tighter     |
+
+**Under load** (constant ~900 Mbps dirty rate near bandwidth limit):
+converges with tuning at ~1s downtime; fails to converge without.
+
+**Idle VM** (CentOS Stream 10, no workload): tuning adds no overhead
+and actually achieves lower downtime (129ms vs 244ms) thanks to the
+lower initial `target_downtime` of 150ms vs QEMU's default 300ms.
+
+## API Examples
+
+Enable downtime tuning with all defaults (ramps up to `maxDowntimeMs`):
+
+```yaml
+apiVersion: migrations.kubevirt.io/v1alpha1
+kind: MigrationPolicy
+metadata:
+  name: tune-downtime
+spec:
+  selectors:
+    namespaceSelector:
+      workload-type: memory-intensive
+  maxDowntimeMs: 1050
+  experimental:
+    downtimeTuning: {}
+```
+
+With custom low-level overrides:
+
+```yaml
+apiVersion: migrations.kubevirt.io/v1alpha1
+kind: MigrationPolicy
+metadata:
+  name: tune-downtime-advanced
+spec:
+  selectors:
+    namespaceSelector:
+      workload-type: memory-intensive
+  maxDowntimeMs: 1500
+  experimental:
+    downtimeTuning:
+      initialMs: 200
+      steps: 5
+      startAfterIteration: 2
+      cooldownSeconds: 15
+```
+
+## Alternatives
+
+1. **Always enable with defaults**: Desirable long-term, but a
+   universally-correct `max_downtime` ceiling doesn't exist. Opt-in
+   first; enable by default once benchmarks confirm the defaults.
+2. **Rely solely on auto-converge**: Throttles guest CPU — effective for
+   CPU-bound dirty workloads but impacts performance. Downtime tuning
+   affects only the switchover pause. The two complement each other.
+3. **Exponential ramp**: Overshoots quickly (150→225→338→506→759→1139ms
+   in five steps). Linear gives predictable, gradual increases.
+4. **Pure iteration-based ramp**: Rapid escalation on fast links where
+   iterations advance in milliseconds. Time-based cooldown ensures each
+   level persists long enough for QEMU to attempt convergence.
+5. **Cluster-wide default in KubeVirt CR**: Risks regressions. Can be
+   considered for Beta/GA once defaults are proven.
+
+## Scalability
+
+No scalability impact. Runs within the existing per-migration
+monitoring goroutine. Adds one `MigrateSetMaxDowntime` libvirt call
+per step. No new API resources or watches.
+
+## Update/Rollback Compatibility
+
+- **Update**: Field defaults to `nil` (disabled). Existing migrations
+  unaffected.
+- **Rollback**: Feature gate removal causes the field to be ignored.
+  No persistent state changes.
+- **Mixed-version**: Source virt-launcher controls tuning. Target node
+  requires no changes.
+
+## Functional Testing Approach
+
+1. **Unit**: Verify `MigrateSetMaxDowntime` called with expected values
+   at correct iterations. No-op when `experimental.downtimeTuning` is
+   absent. Ceiling respected.
+2. **E2E**: Migration with tuning enabled + moderately dirty workload.
+   Verify successful completion and log lines showing downtime steps. In constrained environment it won't be possible to reliably reach a fixed downtime estimate, excessive non-default values would need to be used.
+3. **Negative**: `spec.experimental` ignored when gate is disabled.
+
+## Implementation History
+
+<!--
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] `maxDowntimeMs` top-level field available as the ceiling (per
+      [VEP 252](https://github.com/kubevirt/enhancements/pull/252))
+- [ ] Experimental options framework from [VEP 293](https://github.com/kubevirt/enhancements/pull/295)
+      gates `spec.experimental`
+- [ ] `DowntimeTuningPolicy` struct under `spec.experimental.downtimeTuning`;
+      its presence activates the tuning algorithm
+- [ ] `virt-launcher` runs tuning algorithm in monitoring loop
+- [ ] E2E test
+- [ ] User guide docs
+
+### Beta
+
+- [ ] Low-level tunables remain in `spec.experimental` or are dropped
+- [ ] Soak testing across workload profiles (idle, moderate, heavy)
+- [ ] Prometheus metric for downtime adjustments
+- [ ] Feature gate enabled by default
+
+### GA
+
+- [ ] Feature gate removed
+- [ ] Consider enabling by default with sensible `maxDowntimeMs`
+- [ ] Stable across multiple releases
+- [ ] Evaluate overlap with [VEP 252](https://github.com/kubevirt/enhancements/pull/252)
+      stall-detection; if that approach subsumes the fixed ramp, this
+      feature may be folded in or replaced


veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+
+50 GiB VM, 200 Mbps dirty rate, 800 Mbps migration bandwidth:
+
+| Iteration | DataRemaining | max_downtime | Action                        |
+|-----------|---------------|--------------|-------------------------------|
+| 1         | 12,400 MiB    | 300ms        | QEMU default                  |
+| 2         | 4,200 MiB     | 300ms        | —                             |
+| 3         | 1,800 MiB     | 300ms        | `iterationsBeforeStart` reached |
+| 5         | 850 MiB       | 450ms        | +150ms                        |
+| 7         | 420 MiB       | 600ms        | +150ms                        |
```
Aseeef: Im confused at this example.

a) At the rate we were progressing, migration would have converged without any tuning anyway. This example doesn't line up with this statement:
> Without tuning, this VM continues iterating until the 150\*50s timeout,
then gets forcibly paused. With tuning, it converges at iteration ~10
with 750ms switchover downtime — sub-second, far better than a
multi-second pause, post-copy, or outright abort.

Please let me know if I misunderstood something.

b) With 800mbps, and 600ms max downtime, the switchover should have happened at iteration 7 (420mib/8000mbps=525ms)?

Aseeef (commented):
__General comments:__

As you know I'm working on resolving a similar problem. 

But it looks like the focus of your work is essentially to allow slightly higher downtimes, while still preferring lower downtimes. The key idea (from my perspective) of your VEP is to aid the use case: "OK, I can tolerate downtimes of up to 1200ms, but I still prefer if the downtime was smaller". This use case is applicable even when "allowWorkloadDisruptions", "allowPostcopy", or "allowAutoConverge" are false.

On the other hand, I'm working on trying to detect as soon as possible whether migration has stalled and in the case workload disruptions are not allowed to cancel the migration, or when they are allowed, find the right time to trigger post copy or pause-and-copy to make the switchover.

There is room for some synergy between what I'm working on and this. Particularly, my work can help remove the "iterationsBeforeStart" to an automatic select at what point we stop making progress on remainingData so that we should start tuning downtime.


__Next, I have a suggestion/something to think about:__

Suppose your migration is stalled and `estimated_downtime` is 1000ms. Currently, the `max_downtime` defaults to 300ms. You are proposing (assuming I understood the proposal correctly) that after x amount of iterations, we bump max_downtime by +150ms each iteration. However in this case, those extra iterations you spend going from 300ms -> 450ms -> 600ms -> 750ms -> 900ms -> 1050ms are wasted. This is because if we stopped making progress on remainingData, it was better to just jump to 1050ms from the start.

There is also another layer here: in reality when migration stalls, `estimated_downtime` won't be "1000ms" but rather be fluctuating between (for example) 800ms-1200ms. My work tries to take advantage of this by triggering downtime at local minima.

I still need a few more days to finish my proposal, but we should definitely sync afterward as this still addresses certain things I was ignoring in my own proposal.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+  dirty rates that frequently time out or require post-copy/pause.
+* Platform engineers wanting finer convergence control without
+  enabling auto-converge (which throttles guest CPU).
+
+## User Stories
+
+* As a cluster admin, I want my large-memory VMs to converge without
+  being paused or falling back to post-copy, by letting QEMU accept
+  a slightly longer switchover downtime after sufficient iteration
+  progress.
```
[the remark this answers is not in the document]
fabiand: Yes, it is pretrty much automating what we can do manually today.

Today during migration, we can pause a VM leading to a pause which is lasting as long as the migration the takes to converge.
With this feature, by configuring the max downtime we effectively set an upper cap to this.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+### Configuration via annotations
+
+The feature is enabled per-VMI via annotation on the VMI:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  annotations:
+    kubevirt.io/migration-downtime-tuning: |
```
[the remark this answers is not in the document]
michalskrivanek: me as well:) I've been told so

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+      {
+        "stepMs": 150,
+        "maxDowntimeMs": 1500,
```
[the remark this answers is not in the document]
michalskrivanek: yeah, I think the max downtime is the most important one deserving to be configurable. I can imagine the others are not required (similarly as autoConverge parameters are not exposed) and we can put in decent defaults. Even for the maximum I think we can use a good enough value that's better than today, out of the box.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+### Always enable with defaults
+
+Risky — changing default `max_downtime` behavior could surprise users.
+Opt-in via annotation is safer for alpha.
+
+### Rely solely on auto-converge
+
+Auto-converge throttles guest CPU, directly impacting workload
+performance. Downtime tuning only affects switchover pause duration,
+which is far less disruptive.
```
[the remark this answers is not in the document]
michalskrivanek: it's less intrusive to the workload. - timers still fire in autoConverge, but the time it takes to e.g. service those handlers differ a lot, as well as pure cpu performance. IOW it's a weird kind of slowdown that may have very different effect depending on workload. Whereas with max_downtime the only variable is teh length of the pause. Indeed it's not always better, it's just less unpredictable from the guest perspective.
In reality I would suggest to combine the two configs

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,311 @@
...
+
+Less precise. Iterations directly reflect QEMU's convergence
+progress. Time-based ramps increase too aggressively for slow
+migrations or too timidly for fast ones.
+
+### Formal API fields in MigrationConfiguration / MigrationPolicy
+
+More adoption friction, CRD/webhook changes, larger review surface.
+Annotation lets us iterate quickly in alpha; promote to API in beta
+if proven.
```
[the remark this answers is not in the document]
michalskrivanek: or the default might be good enough and we may not need them, sure.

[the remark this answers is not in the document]
michalskrivanek: I'll update with a paragraph about the suggested default implementation and update the examples. I would want to start with 150ms, and step up by 1/7th of max in 7 steps, starting with iteration 3 (1 is the initial copy, 2 is supposedly still large singe it's accumulated over the duration of the initial large copy, 3 let's give it a chance for one hopefully small-enough iteration to make it within the minimum downtime), and the default maximum of 1050ms, so you get:
150 150 150 300 450 600 750 900 1050 1050 1050 ...

and potentially with larger configured max_downtime of e.g. 4900ms would step it up more steeply by 700ms, with the first 2 iterations still hardcoded to 150
150 150 700 1400 2100 2800 3500 4200 4900

michalskrivanek:
clarified a bit, updated examples and proposed api.
@Aseeef it's not supposed to replace your proposal, rather offer a more simple and less intrusive way until that stall detection is more mature. The main purpose here is not to detect and act on stalls, it's just about improving migration convergence without disruption.

michalskrivanek:
> The key idea (from my perspective) of your VEP is to aid the use case: "OK, I can tolerate downtimes of up to 1200ms, but I still prefer if the downtime was smaller". This use case is applicable even when "allowWorkloadDisruptions", "allowPostcopy", or "allowAutoConverge" are false.

indeed. it's complementary/different, and IMO desperately missing in kubevirt today

> There is room for some synergy between what I'm working on and this. Particularly, my work can help remove the "iterationsBeforeStart" to an automatic select at what point we stop making progress on remainingData so that we should start tuning downtime.

yes

> Suppose your migration is stalled and `estimated_downtime` is 1000ms. Currently, the `max_downtime` defaults to 300ms. You are proposing (assuming I understood the proposal correctly) that after x amount of iterations, we bump max_downtime by +150ms each iteration. However in this case, those extra iterations you spend going from 300ms -> 450ms -> 600ms -> 750ms -> 900ms -> 1050ms are wasted. This is because if we stopped making progress on remainingData, it was better to just jump to 1050ms from the start.

the reason why not to start with max is that if after the first (or first few) iterations qemu sees it can make it within the 1050ms it would do it - and ideally you would prefer a smaller downtime if possible, say, if one or two more iterations that take just couple of seconds would bring down the effective downtime down to 500ms making migration more seamless. A universal heuristic is hard, I think, hence this simplified proposal that should work good enough most of the times - the steps are just a few, potential waste is not so large.

> There is also another layer here: in reality when migration stalls, `estimated_downtime` won't be "1000ms" but rather be fluctuating between (for example) 800ms-1200ms. My work tries to take advantage of this by triggering downtime at local minima.

yes, this downtime tuning is not to address cases when it fluctuates.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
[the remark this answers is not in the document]
fabiand: Please note that the downtime is impacting an application, and the quesiton is if the application is tolerating a certain downtime, and application oweners are usualy != cluster admins

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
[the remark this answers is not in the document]
fabiand: Check this and others: https://github.com/kubevirt/kubevirt/blob/51814453617a23f0e57b0db556b2e38c690086b3/staging/src/kubevirt.io/api/core/v1/types.go#L1348

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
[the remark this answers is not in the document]
michalskrivanek: AFAIU vMotion is using 1s max stun time for "normal" circumstances, I think that's what we should target as a sensible default as well

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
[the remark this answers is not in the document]
iholder101: A lot of mistakes were made in the current API :)

In general, I tend to think that annotations ARE de-facto APIs that are hidden and are less structured.
TBH, I think there needs to be a good reason to avoid APIs in order to use annotations for configurations, but generally it's an anti-pattern.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,376 @@
...
+section above.
+
+No new CRD fields, webhooks, or RBAC changes are required.
+
+## Alternatives
+
+### Always enable with defaults
+
+Risky — changing default `max_downtime` behavior could surprise users.
+Opt-in via annotation is safer for alpha.
```
iholder101: My feedback is very similar to the feedback for the other VEP here: https://github.com/kubevirt/enhancements/pull/247#issuecomment-4214518348.

Aside from the fact that this couples us to libvirt, I'm not sure how users will have the knowledge to configure this. I tend to think we should be doing this for them behind the scenes. AFAICT most users want migrations to "just work", and are not expert about all of these low-level knobs that are complicated to optimize correctly.

iholder101 (commented):
Thank you for this VEP.

I think this direction is important, and you bring some very good points here.

That being said, I think more work needs to be done. What I have in mind is:
1. Getting some data about which configurations are the best to use + some evidence about how it helped different kinds of workloads to converge faster.
2. Thinking about a behind-the-scenes truly no-API algorithm that we can use in KubeVirt without any user intervention.

With these in place, we'd be able to tell our users to just enable the feature gate and see if migrations improve or not. They key difference here is that we put the responsibility to tune this correctly on **KubeVirt**, not on its users.

I'd be absolutely happy to see you push this forward!

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
michalskrivanek: I do not see how this is coupling anything to libvirt - except for leveraging the API (the _only_ API available) to set the downtime. I do not see any other way how to do that unless you mean to throw away libvirt completely already and replace that with direct interaction with qemu (i think that would be a crazy idea at this point) or not doing it at all - considering that the default of 300ms is just too low I do not see how we can live with it when using qemu/kvm.

As for the "just work" I think that's unrealistic not to have API - there's no "just work" value that covers all scenarios. The default shall be the best possible values that cover most of them, but it will never cover them all. So for the remaining part what is your suggestion? Ignore those? Or just keep piling up heuristics? I can imagine soft and hard limit for downtime combined with e.g. AllowWorkloadDisruption to infer what the user would prefer to do, but it's always going to be heuristics and sooner or later someone won't like it. And then we're in trouble that the behavior cannot be changed when it doesn't have an API.
Whether that's annotations or proper API doesn't really matter

michalskrivanek:
few notes from testing on 1Gbps network
- as expected default 1s doesn't help when bandwidth << dirty rate. 
- the iterations are often too fast and the increases are fairly quick. It probably makes sense to have a cooldown period (time based or iteration based) to give it some chance to converge with the new downtime setting. Also simply just to to call the API too often.  Probably time-based makes more sense here - with 10s for 1Gbps bandwidth that's up to 1GB worth of data(so a good chance to iterate couple times) and still with default number of steps means ~60s of tuning to get to the max

10 migrations with and without tuning - start in 3rd iteration, initial 150ms, 7 steps with 10s cooldown, up to 1050ms max, 1Gbps link, idle OCP guest workload - avg dirty rate ~650Mbps but it varies a lot since it's supposedly idle there are some quiet times.

Metric | With Tuning | Without Tuning | Delta
-- | -- | -- | --
Avg Iter1 Time | 67.5s | 67.4s | ~0 (identical)
Avg Convergence | 24.7s | 45.0s | -45% (20s faster)
Median Convergence | 24.3s | 35.0s | -31%
Avg Total Time | 92.2s | 112.4s | -18% (20s faster)
Avg Downtime | 611ms | 451ms | +160ms (+35%)
Avg Iterations | 28 | 69 | -59%
Avg Data Transferred | 9,694 MiB | 11,862 MiB | -18%
Worst Total Time | 98s | 161s | -39%
Best Total Time | 86s | 91s | -5s
Convergence Range | 19-30s (11s) | 24-93s (69s) | 6x tighter
P90 Convergence | 29.8s | 67.6s | -56%

Light load:
When adding light constant load, dirty rate gets closer to bandwidth (~900Mbps) but there are no idle periods, as expected it converges with ~1s downtime vs never converges.

Completely idle VM - Centos 10 idle VM sitting in login prompt after boot (many constant pages, no churn) - testing effect of proposed lower initial downtime:

Metric | With Tuning | Without Tuning
-- | -- | -- 
Avg Total Time | 9.5s | 9.5s
Avg Downtime | 129ms | 244ms
Avg Iterations | 3.0 | 3.5
Data Transferred | 763 MiB | 763 MiB

Additionally to test synergy with also enabling #246 the results are pretty good when one wants to sacrifice CPU, results for the the first "idle worker node" case:
Metric | With zstd | Without compression | Improvement
-- | -- | -- | --
Total time | 21.3s | 87.5s | 4.1x faster
Data transferred | 2145 MiB | 9210 MiB | 4.3x less
Iterations | 5.6 | 13 | 2.3x fewer
Downtime | 124ms | ~450ms (1050ms max) | 3.6x lower

michalskrivanek:
added [draft implementation](https://github.com/kubevirt/kubevirt/pull/17480) (for now using annotations). It's quite self-contained.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
michalskrivanek: @iholder101 what I think makes sense is to expose the truly user facing API as regular option (that would only be the max tolerable downtime here), and keep the low level knobs as annotations. 

Kubevirt's problem is not too many knobs, it's the lack of good defaults and differentiation between high level knob with appropriate documentation/guidance and lower level knobs for fine tuning and experimentation (on user's side)

iholder101: I think that it is realistic to come up with a solution that "just works" for most use-cases. That being said, I don't necessarily think that we don't need to expose any APIs, but if we do these APIs need to be declarative and extremely easy to understand from a user perspective - not libvirt/QEMU low-level knobs that most users won't know how to configure and that couple us to libvirt/QEMU which we want to be able to replace in the future.

As I see it, these APIs should express things in the spirit of "I'm OK with these VMs having a higher downtime" or "I'm okay with using more CPU for these VMs", etc. From a user perspective this is something understandable, unlike setting `max_downtime` to 300/600 ms or choosing if the compression algorithm should be `zstd` or `zip`.

That being said - a good argument to **temporarily** expose these tunables is if it is the only way to test this. This can be considered, but if we go this route I think the VEP has to clearly state that the next **alpha** phase has to remove these APIs.

I suggest we discuss this in today's VEP discussion meeting.

michalskrivanek: downtime/stun time is relatively well understood term. I would be cautions with making it too simplistic. E.g. the AlowWorkloadDisruption is an example of a well meant simplification which turned into a misleading and confusing option.

veps/sig-compute/248-dynamic-downtime-tuning/vep.md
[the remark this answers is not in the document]
fabiand: @iholder101 then please provide a suggestion. IIUIC then we consider VM level API adiditions problematic, and we are also considering annotations to be harmful? Then I wonder: Are we saying we want to put KubeVirt into maintenance mode? ;)

My question is: We want to explore to add new stuff, what is the path to opt-into these features that we want to try out.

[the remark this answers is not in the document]
iholder101: Hey @fabiand!
Please see my comment here. I think it answers this exactly: https://github.com/kubevirt/enhancements/pull/249#discussion_r3077754764.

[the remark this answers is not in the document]
vladikr: > @iholder101 then please provide a suggestion. IIUIC then we consider VM level API adiditions problematic, and we are also considering annotations to be harmful? Then I wonder: Are we saying we want to put KubeVirt into maintenance mode? ;)
> 
> My question is: We want to explore to add new stuff, what is the path to opt-into these features that we want to try out.

I don't think this API belongs in the VMI object. VM users shouldn't care about migration as it needs to be transparent for them and shouldn't be able to affect it. It's a cluster admin responsibility. We have several APIs to affect the migration: The VMIMigration object, A cluster wide `spec.configuration.migration` and migration policies.

However, I'm not against an experimental API, but this should be explicitly stated in the VEP that this API is only experimental and will not go forward in the current form. 
Moreover, I there are other ways to experiment with APIs, is to deploy a modified kubevirt on a cluster using KubeVirt's external provider.

Generally, I don't think an API is necessary in this case. I don't see how admins would know what option to set.
From my perspective, we should develop a heuristic to allow KubeVirt to be smart enough to adjust this parameter.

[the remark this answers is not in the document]
michalskrivanek: it's a bit mixed responsibility - there's cluster admin's POV with infrastructure details knowledge and desired metrics/performance for migrations, drains. But there's also workload/application aspect of tolerance to downtime.
Overall I agree a single "max downtime" API is very likely sufficient, but it makes sense to me to have that for both personas, ideally.
The others are indeed very low level and doesn't need to be exposed to users after the feature is mature.
However I disagree with the heuristic-only (no knobs) option for the start, it's impractical in the field where recompiling or replacing stable kubevirt release is just not an option. I think an experimental API/annotations/configmap/whatever is a necessity initially.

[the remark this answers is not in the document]
vladikr: > it's a bit mixed responsibility - there's cluster admin's POV with infrastructure details knowledge and desired metrics/performance for migrations, drains. But there's also workload/application aspect of tolerance to downtime.

That's what the migration polices are for. It's up to the admin to craft policies that target a specific class of workloads.
My strong preference is to move this experimental tunable away from the VMI API to a migration policy CRD and a cluster wide migration config as a fallback.
It's literally designed for this exact use-case ( a per VM group migration config via label selectors)

Again, I understand the "mixed responsibility", but this is exactly the philosophy... Admins craft the policy for a specific group/class/individual VM, and the VM owner labels the VM accordingly.
The application cares about the resulting downtime, not about the exact max downtime value or iteration steps.

I think if we are adding an experimental API it needs to be a single high level value. 

I get the need for an experimental API, but what I'm trying to prevent is that the convenience for experimenting would turn into a de facto API. 

I also agree that it's impossible to arrive at a smart heuristic that "just works" in most cases right now. 
Therefore, I'm fine with either an eperimental api or a simple mechanism + data gathering and improving it from there - but that vision of where we are going with this should be reflected in the VEP.

I hope it makes sense.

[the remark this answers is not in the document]
michalskrivanek: it does. 
( though I would say that, generally, if you find people using the experimental knobs as the de-facto API it's a pretty good indication that the "let's design a perfect API without field experimentation" practice is not working well:-P )

Migration policies with cluster default fallback make sense, for the max downtime single high level value.

I genuinely wonder what is the issue with annotations for experimental knobs. It can't be any clearer that it is distinct from the CRD API. Is that common that people misunderstands that? Is it causing issues? AFAICT this is pretty common to allow for "advanced configuration" in some way, esxi has ~40 low level migration-related options that you can play with, yet it is considered to be the golden standard of good defaults and simplicity to use.

michalskrivanek:
from further feedback on API - let's try to push the low level knobs to a clearly marked experimental section in migration policies. Most of the defaults do not need cluster default configurable, perhaps except the max_downtime itself.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+
+### Current behavior
+
+KubeVirt's migration monitor makes convergence decisions based purely
+on wall-clock time:
+
+1. **`completionTimeoutPerGiB × vmSize`** — if exceeded, trigger
+   post-copy / pause / abort.
+2. **`progressTimeout`** — if `DataRemaining` hasn't decreased for N
+   seconds, abort.
```
germanovm, line 39: This is what I expected progressTimeout to do given its name, but its not what it does. It checks if any data was transferred at all in the period and aborts if not. Its not a simple stall detector. See CNV-71164.

michalskrivanek: yes, in the stall detection proposal it's repurposed, here https://github.com/kubevirt/enhancements/pull/252/changes#diff-4df3450838ded564e5715a456a09794f9f801ed99f176a8b32a51af04d39b1bcR145-R146

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+  `max_downtime` of 300ms and will only switch over when estimated
+  remaining transfer time drops below it. For workloads with moderate
+  dirty rates, remaining time hovers slightly above 300ms — never
+  quite converging. KubeVirt's only response is the nuclear option:
+  pause the guest, trigger post-copy, or abort after timeout.
+
+- **Time-based decisions penalize large VMs.** A 256 GiB VM with a
+  healthy dirty rate at iteration 20 gets the same treatment as a
+  4 GiB VM stuck at iteration 1.
+
```
Aseeef, line 55: Suggestion:

1. This section can be made shorter in a few parts.

Particularly, this can be shortened a lot.
> QEMU reports `MemIteration`
  — how many full passes over dirty memory have completed. A migration
  at iteration 15 is very close to converging and just needs a small
  downtime increase, but KubeVirt doesn't distinguish it from one
  stuck at iteration 1.

Some statements here also don't sound confusing to me. For example when you say "A migration
  at iteration 15 is very close to converging" is this an example?

2. There is an overloading of how `max_downtime` is used in this VEP. Consider using `target_downtime` instead.

3. I don't follow this:

> **Time-based decisions penalize large VMs.** A 256 GiB VM with a
  healthy dirty rate at iteration 20 gets the same treatment as a
  4 GiB VM stuck at iteration 1.

What do you mean by "stuck at iteration 1". Based of our prior discussions, I think you mean to say something along the lines of "large VMs are granted an excessive amount of timeouts even though the working set (active memory) does not necessarily scale with VM size".

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+### Proposed behavior
+
+A new algorithm, inspired by QEMU's own auto-converge CPU throttle
+and highly effective migration policies in oVirt project, is inserted
+into the existing monitoring loop:
+
+1. **Set initial `max_downtime` to 150ms** — lower than QEMU's default
```
Aseeef:
```suggestion
1. **Set initial `target_downtime` to 150ms** — lower than QEMU's default
```

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+   300ms, giving low-dirty-rate VMs a chance at shorter switchover.
+2. **Track iteration count** via `MemIteration` from libvirt job stats.
```
Aseeef: I would suggest not querying job stats here. Instead, the event that fires at the start of each iteration already contains information about which iteration count we are on. (Allows you to save a few milliseconds, and act closer to iteration boundaries)

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+3. **After iteration 3** (allowing the initial bulk copy and first
```
Aseeef:
```suggestion
3. **After iteration `startAfterIteration` (default 3)** (allowing the initial bulk copy and first
```

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+   re-dirty pass to complete), start a cooldown timer.
+4. **Ramp linearly with time-based cooldown** — increase `max_downtime`
```
Aseeef:
```suggestion
4. **Ramp linearly with time-based cooldown** — increase `target_downtime`
```

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+section or be dropped.
+
+### Feature gate
+
+Same `ExperimentalMigrationPolicy` gate as compression — covers the
+entire `spec.experimental` section. The `downtimeTuning` block is
+ignored when the gate is disabled.
+
+### How the ramp works
+
```
Aseeef, line 134: This section has too much overlap with the section on "Proposed behavior". Its probably best to put this all under implementation/algorithm section

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+## Design Rationale
+
+### Why MigrationPolicy only
+
+The tolerable switchover pause depends on the application.
+`MigrationPolicy` lets platform engineers target specific workloads
+without risking cluster-wide regressions.
+
+### Why `spec.experimental`
```
Aseeef: I wonder if its better to have a seperate VEP for spec.experimental. Especially since I'm going to be using it in my VEP to as per @vladikr's suggestion. So instead of having 3 different VEPs re-describe the same thing, and having to update 3 different VEPs anytime we want to change something, just have a dedicated VEP for this part and each of our 3 VEPs can reference that?

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+  No persistent state changes.
+- **Mixed-version**: Source virt-launcher controls tuning. Target node
+  requires no changes.
+
+## Functional Testing Approach
+
+1. **Unit**: Verify `MigrateSetMaxDowntime` called with expected values
+   at correct iterations. No-op when `downtimeTuning` is nil. Ceiling
+   respected.
+2. **E2E**: Migration with tuning enabled + moderately dirty workload.
```
Aseeef, line 296: E2E testing for this is non-trivial because ideally E2E tests should be agnostic to how much bandwidth capability the particular testing machines have. So we also need to limit bandwidth for migrations. This is code to do this already but due to https://gitlab.com/qemu-project/qemu/-/work_items/3364, limiting bandwidth on migrations using multifd doesn't work.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,428 @@
...
+For example:
+01-02-1921: Implemented mechanism for doing great stuff. PR: <LINK>.
+03-04-1922: Added support for doing even greater stuff. PR: <LINK>.
+-->
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `ExperimentalMigrationPolicy` guards `spec.experimental`
```
Aseeef: I'm not fully certain how feature gating should be handled across the three VEPs. I'm thinking the three VEPs should each have their own feature gates so their graduation can be tracked independently. And then we can let `ExperimentalMigrationPolicy` just expose additional tuning options for those VEPs. Thoughts?

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: yeah, let's shorten. 
2. - I think it makes sense to strictly distinguish the max_dowtime (the user defined maximum) from the target_downtime(internal algorithm target, setting the qemu's max_downtime).
3. "stuck" - yes, just saying it the other way around - if you try to use sensible values they get applied to large VMs too early. same outcome, they're unusable.

michalskrivanek: I'll also drop the detailed API change to shorten things a bit, it's redundant with the PR and just annoying to keep in sync.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: I think for the purpose of the description it doesn't matter how it is determined. Indeed event should be the primary way. 
I think it makes sense to cherrypick your https://github.com/kubevirt/kubevirt/pull/17544/changes/5536d7c05b0af5a440130feae0545444d5dcc067 (minus the stall guard) and adapt the poll interval (400ms->1s?) as an immediate improvement.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: we can try to force multifd off. It's unfortunate it's no longer configurable (removed by https://github.com/kubevirt/kubevirt/pull/12705), perhaps we can bring it back as an advanced option now.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: I want to avoid overly granular gates. I think `.spec.experimental` provides a sufficient choice to turn things on and off without implying guarantees. I think the actual graduation of migration features would then be expressed by moving the options/tunables away from experimental to top level

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: my only concern is when the project struggles with high number of VEPs this would be a fourth one about live migration:/
Maybe it's fine to introduce it now and later on build a stronger story for migration options. I would anyway want to see .spec.advanced in addition to experimental for permanent low level tunables, and shift the legacy lowlevel tunables (like diabling tls and others) to advanced. But it feels it would have to be rushed too much right now before 1.9

michalskrivanek:
removed some of the slop, rephrased and shorten in several places to align with your suggestions
@Aseeef thanks for you review!

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
Aseeef: >  I think the actual graduation of migration features would then be expressed by moving the options/tunables away from experimental to top level

Yes but then graduation will be binary. There would be no intermediate like "beta".

Here is what I'm imagining:
* We would have this `ExperimentalMigrationPolicies` feature (or `AdvancedMigrationPolicies` or whatever we want to call it). This feature gate can be perpetually alpha containing tunables that are either only intended for use for debugging and not something we ever intend to expose or something we eventually intend to sift up to top level. API under this gate have no guarantees version to version.
* `MigrationCompression`, `IterativeMigrationTuning`, `MigrationStallDetection` are all independent FGs with some fields that live inside `ExperimentalMigrationPolicies`. Even when `ExperimentalMigrationPolicies` is not enabled, sane defaults name them work as intended. In fact, we can even graduate some of these features while keeping certain tunables locked behind `ExperimentalMigrationPolicies`.

michalskrivanek: > * We would have this `ExperimentalMigrationPolicies` feature (or `AdvancedMigrationPolicies` or whatever we want to call it). This feature gate can be perpetually alpha containing tunables that are either only intended for use for debugging and not something we ever intend to expose or something we eventually intend to sift up to top level. API under this gate have no guarantees version to version.

hm, but that defeats the purpose of having them readily available. The point of `experimental` section is to deter people from using them or consider them stable, but at the same time allow setting them on demand for troubleshooting/experimentation. So ok...it could be a forever-alpha gate...you mean one would have to turn it on before the debugging. Is it necessary though? I do not see a reason to ever turn the gate off - you can simply not use any of those options.
If it is necessary I don't see a particular problem, I would just try to turn it on by default/always.


> * `MigrationCompression`, `IterativeMigrationTuning`, `MigrationStallDetection` are all independent FGs with some fields that live inside `ExperimentalMigrationPolicies`. Even when `ExperimentalMigrationPolicies` is not enabled, sane defaults name them work as intended. In fact, we can even graduate some of these features while keeping certain tunables locked behind `ExperimentalMigrationPolicies`.
I understand, but I guess I'm jut not a big fan of so many gates for no real gain. When it's clearly marked as experimental why bother with the significant overhead around deployments, especially downstream ?

Aseeef: Considering the options exposed here are implementation specific, advanced/experimental options with no guarantees regarding API stability I do think this would be a justified design decision.

Aseeef: @vladikr thoughts?

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
```diff
@@ -0,0 +1,315 @@
...
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Improve live migration convergence by gradually increasing QEMU's
+`max_downtime` as migration iterations progress, with a progressive ramp up of a `target_downtime` up to the configured `maxDowntimeMs`, replacing the current all-or-nothing approach (run until timeout, then pause / post-copy / abort).
+
+Configuration is exposed via `MigrationPolicy` under `spec.experimental`.
+All tunables start in the experimental section; `maxDowntimeMs` is
```
Aseeef: In my VEP (https://github.com/kubevirt/enhancements/pull/252) I'm exposing `maxDowntimeMs` directly now (i.e. no experimental). I think for `maxDowntimeMs` this makes perfect sense (instead of having to migrate this field later out of experimental). This simplifies future use case too, and I think you should do the same.

Aseeef (commented):
Overall, looks good to me. Choices are reasonable. Particularly, it is interesting how compared to this VEP, my VEP (https://github.com/kubevirt/enhancements/pull/252) which uses time-based steps, you use iteration based steps.

Thinking about it, this probably makes sense for this VEP. For example, this **wouldn't** be a valid critique:
> If iterations are really long, then the downtime tuning won't start for a really long time!

This is because if iterations are long, that means downtime tuning is probably not going to help you either because iteration duration and expected downtime at that iteration are very closely related.

---

Let's converge on (a) location for `maxDowntimeMs` (b) define the functional testing approach (i.e. setting multifd, or waiting on https://gitlab.com/qemu-project/qemu/-/work_items/3364 to get fixed). Otherwise, this is an `lgtm` from me.

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: ack. I think this is one of the most important setting indeed.

michalskrivanek:
> Let's converge on (a) location for `maxDowntimeMs` (b) define the functional testing approach (i.e. setting multifd, or waiting on https://gitlab.com/qemu-project/qemu/-/work_items/3364 to get fixed). Otherwise, this is an `lgtm` from me.

top level is fine, for me that's not even a case of `advanced` setting, this is a fundamental live migration parameter

I think bringing back multifd configuration through `experimental` is the practical choice since waiting on qemu side may take some time (I can see how it's not the most important bug for QEMU), and besides tests there are use cases to adjust in deployments as well (very fast networks where 8 may not be enough, or the opposite - setting it to 1 to actually limit the throughput "naturally" without any limiter where LM happens over shared network and full utilization is undesirable)

veps/sig-compute/291-dynamic-downtime-tuning/vep.md
michalskrivanek: I'll make changes to VEP 291 to expect/reuse this, and activate the feature via presence of `.experimental.downtimeTuning`

Aseeef:
Great :)
/lgtm

iholder101:
Important note: As said elsewhere, in the final state of this VEP should not consist of libvirt-level tunables. KubeVirt should be hypervisor-agnostic and smart enough by default that the vast majority of workloads just migrate well without any special knobs. We will revisit it in the next alpha phase of this VEP.

Thanks @michalskrivanek !

vladikr:
> Important note: As said elsewhere, in the final state of this VEP should not consist of libvirt-level tunables. KubeVirt should be hypervisor-agnostic and smart enough by default that the vast majority of workloads just migrate well without any special knobs. We will revisit it in the next alpha phase of this VEP.
> 
> Thanks @michalskrivanek !

+100! 
/approve
Thank you!

Issue #165, opened 2025-12-19
VEP 161: ContainerPath Volumes


mhenriks:
**Primary contact (assignee)**:

/assign @mhenriks 

**Current Feature Stage**:

New

**Feature Gate**: 
`ContainerPathVolumes`

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/164

**Timeline**:

* Alpha release target (v1.8):
* Beta release target (v1.9):
* Stable release target (v1.10):

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/164
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

mhenriks:
> Hi [@mhenriks](https://github.com/mhenriks), can you please adjust the title to VEP 165? Thanks.

oops, all set @orelmisan

orelmisan:
Thanks :)

vladikr:
@mhenriks do you think we could get any feedback from any users on this?

vladikr:
@mhenriks is this going to remain alpha ?

mhenriks:
Thanks @vladikr I created #437 to take this to beta

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

mhenriks:
> Hey!
> 
> Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22). Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

Thanks @iholder101 added @awels to the spreadsheet

PR #251, opened 2026-04-03
VEP 250: Add serviceAccountName to VirtualMachineInstance Spec

mhenriks:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/250
**SIG label**: /sig/compute

### What this PR does

- Decouples pod identity ("run as this service account") from token exposure ("mount the SA token into the VM"), which are currently conflated in the serviceAccount volume source
- Motivated by ContainerPath volumes (VEP #165) and cloud workload identity patterns (AWS IRSA, Azure WorkloadIdentity) where users need the pod to run as a specific SA for webhook injection but don't want the Kubernetes SA token exposed to the VM
- Mirrors the Kubernetes Pod API convention of spec.serviceAccountName
- Fully backward compatible: existing serviceAccount volumes continue to work unchanged

### DIFF
--- a/veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
+++ b/veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
@@ -1,0 +1,420 @@
+# VEP #250: Add serviceAccountName to VirtualMachineInstance Spec
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets version: v1.9
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+
+## Overview
+
+This proposal adds a `serviceAccountName` field to `VirtualMachineInstanceSpec`, decoupling the concern of "which service account the virt-launcher pod runs as" from "whether the service account token is exposed to the VM as a disk/filesystem." Today these two concerns are conflated in the `serviceAccount` volume source, forcing users to expose a service account token volume to the VM even when they only need the pod to run under a specific identity.
+
+## Motivation
+
+KubeVirt's current mechanism for specifying which service account a VM runs as is the `serviceAccount` volume source. When a user specifies this volume, KubeVirt does two things:
+
+1. Sets `pod.spec.serviceAccountName` on the virt-launcher pod
+2. Mounts the service account token as an ISO disk or virtiofs filesystem exposed to the VM guest
+
+This coupling made sense when the primary use case was "I want the VM to access the Kubernetes API using a specific service account." However, with the growing adoption of cloud provider workload identity systems (AWS IRSA, Azure Workload Identity), a new pattern has emerged: users need the virt-launcher pod to run as a specific service account so that external mutating webhooks inject the correct credentials, but they **do not want** the Kubernetes service account token exposed to the VM. This pattern pairs naturally with mechanisms like ContainerPath volumes (VEP #165) that can expose webhook-injected credentials to the VM, though this proposal does not depend on ContainerPath.
+
+For example, with AWS IRSA:
+
+1. User creates a ServiceAccount annotated with `eks.amazonaws.com/role-arn`
+2. User wants the virt-launcher pod to run as this ServiceAccount so the EKS Pod Identity Webhook injects AWS credentials
+3. User wants to expose the injected AWS credentials to the VM
+4. User does **not** want the Kubernetes service account token exposed as a disk to the VM
+
+Currently, the user is forced to configure:
+
+```yaml
+volumes:
+- name: aws-irsa-sa
+  serviceAccount:
+    serviceAccountName: aws-irsa-sa  # Required just to set pod identity
+- name: aws-token
+  containerPath:
+    path: /var/run/secrets/eks.amazonaws.com/serviceaccount
+    readOnly: true
+```
+
+The `serviceAccount` volume creates an unnecessary disk/filesystem in the VM that the user must configure but does not want. With `spec.serviceAccountName`, the user can simply write:
+
+```yaml
+serviceAccountName: aws-irsa-sa
+volumes:
+- name: aws-token
+  containerPath:
+    path: /var/run/secrets/eks.amazonaws.com/serviceaccount
+    readOnly: true
+```
+
+Cloud provider credential injection is the most prominent example, but there are other reasons a workload needs a specific service account identity without wanting the token exposed to the VM:
+
+- **Image pull secrets**: ServiceAccounts can have `imagePullSecrets` attached, allowing the virt-launcher pod to pull from private registries without the token being relevant to the VM workload
+- **Service mesh authorization**: Istio AuthorizationPolicy and similar mesh configurations use service account identity to control traffic between workloads
+- **Admission and policy enforcement**: OPA/Gatekeeper, Kyverno, or Pod Security Admission policies that scope permissions by service account
+- **External secret injection**: Vault Agent Injector and similar tools use service account identity to determine which secrets to inject into the pod
+- **Audit attribution**: Service account identity appears in Kubernetes audit logs for traceability
+
+In all of these cases, the service account functions as a **workload identity**. It determines how the workload authenticates to external systems and how infrastructure components authorize it. This is a workload-level concern, not a pod implementation detail. KubeVirt already recognizes this by exposing service account control through the `serviceAccount` volume. Adding `serviceAccountName` as a standalone field doesn't introduce a new abstraction leak; it corrects an existing one by separating the identity concern from the credential concern.
+
+## Goals
+
+- Allow users to control the VM's service account identity without requiring a service account volume
+- Maintain backward compatibility with existing `serviceAccount` volume usage
+- Simplify the user experience for specifying workload identity
+
+## Non Goals
+
+- **Removing the `serviceAccount` volume type**: The existing volume type remains useful when users explicitly want the token exposed to the VM
+- **Changing the behavior of existing `serviceAccount` volumes**: Existing VMs using `serviceAccount` volumes continue to work identically
+- **Adding `automountServiceAccountToken` to VMI spec**: This proposal focuses solely on `serviceAccountName`. Whether to expose the token to the VM is handled by the presence or absence of a `serviceAccount` volume
+- **Token projection configuration**: Advanced token projection settings (audiences, expiration) are out of scope for this proposal
+
+## Definition of Users
+
+- **Platform Engineers**: Teams deploying KubeVirt in cloud environments that use workload identity systems (AWS IRSA, Azure Workload Identity)
+- **Namespace Admins**: Administrators who define and manage VMs that need cloud provider credentials but don't need direct Kubernetes API access from within the VM
+- **Cluster Admins**: Teams that use service account identity for policy enforcement, service mesh authorization, image pull secrets, or audit attribution
+- **Security-Conscious Operators**: Administrators who want to follow the principle of least privilege by not exposing Kubernetes service account tokens to VMs unnecessarily
+
+## User Stories
+
+- As a platform engineer, I want my VirtualMachine to run as a specific ServiceAccount for cloud provider credential injection, without exposing the Kubernetes service account token to the VM guest
+- As a namespace admin, I want to specify which ServiceAccount my VM runs as without having to configure an unwanted service account token volume
+- As a security-conscious operator, I want to set a custom service account for my VM without exposing unnecessary Kubernetes service account tokens, reducing the attack surface inside the VM
+
+## Repos
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### New API Field
+
+A new `serviceAccountName` field is added to `VirtualMachineInstanceSpec`:
+
+```go
+type VirtualMachineInstanceSpec struct {
+    // ...existing fields...
+
+    // ServiceAccountName is the name of the ServiceAccount to use to run the
+    // virt-launcher pod. This sets pod.spec.serviceAccountName but does NOT
+    // automatically expose the service account token to the VM guest.
+    // To expose the token to the VM, use a serviceAccount volume.
+    // +optional
+    ServiceAccountName string `json:"serviceAccountName,omitempty"`
+}
+```
+
+### Interaction with ServiceAccount Volumes
+
+When both `spec.serviceAccountName` and a `serviceAccount` volume are specified, they **must reference the same service account**. If they differ, the VM will be rejected with a validation error. This prevents confusing configurations where the pod identity and the exposed token come from different service accounts.
+
+| `spec.serviceAccountName` | `serviceAccount` volume | Pod serviceAccountName | Token exposed to VM |
+|---|---|---|---|
+| Set to "my-sa" | Not present | "my-sa" | No |
+| Not set | Present with "my-sa" | "my-sa" | Yes |
+| Set to "my-sa" | Present with "my-sa" | "my-sa" | Yes |
+| Set to "my-sa" | Present with "other-sa" | **Validation error** | N/A |
+| Not set | Not present | "" (default) | No |
+
+### Pod Template Rendering
+
+The virt-controller's pod template rendering logic is updated:
+
+```go
+func serviceAccountName(vmi *v1.VirtualMachineInstance) string {
+    // spec.serviceAccountName takes precedence as the source of truth
+    if vmi.Spec.ServiceAccountName != "" {
+        return vmi.Spec.ServiceAccountName
+    }
+    // Fall back to serviceAccount volume for backward compatibility
+    for _, volume := range vmi.Spec.Volumes {
+        if volume.ServiceAccount != nil {
+            return volume.ServiceAccount.ServiceAccountName
+        }
+    }
+    return ""
+}
+```
+
+The pod's `automountServiceAccountToken` is set to `true` only when a `serviceAccount` volume is present. When only `spec.serviceAccountName` is set (no `serviceAccount` volume), `automountServiceAccountToken` remains `false`. Cloud provider webhooks (IRSA, Azure Workload Identity) inject their own projected volumes independently of this setting, so they work regardless.
+
+### Immutability
+
+Changing `serviceAccountName` requires recreating the VMI. For VMIs managed by a VirtualMachine object, this means a VM restart.
+
+### Validation
+
+1. If both `spec.serviceAccountName` and a `serviceAccount` volume are present, the service account names must match
+2. `spec.serviceAccountName`, if set, must be a valid Kubernetes service account name (DNS subdomain)
+3. At most one `serviceAccount` volume is allowed (existing validation, unchanged)
+
+### Migration and Backward Compatibility
+
+- Existing VMs using `serviceAccount` volumes continue to work without modification
+- No migration is needed; both mechanisms coexist
+- The `serviceAccount` volume is **not** deprecated by this proposal. It remains the correct choice when you want the token exposed to the VM
+
+## API Examples
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,411 @@
...
+3. User wants to expose the injected AWS credentials to the VM via ContainerPath
+4. User does **not** want the Kubernetes service account token exposed as a disk to the VM
+
+Currently, the user is forced to configure:
+
+```yaml
+volumes:
+- name: aws-irsa-sa
+  serviceAccount:
+    serviceAccountName: aws-irsa-sa  # Required just to set pod identity
```
0xFelix, line 42: Can only have a single volume of this type?

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,411 @@
...
+- Both source and target virt-controllers must support the field for it to take effect
+
+## Functional Testing Approach
+
+### Unit Tests
+- Validation: `spec.serviceAccountName` alone sets pod service account correctly
+- Validation: `spec.serviceAccountName` matching `serviceAccount` volume is accepted
+- Validation: `spec.serviceAccountName` conflicting with `serviceAccount` volume is rejected
+- Validation: `spec.serviceAccountName` must be a valid DNS subdomain name
+- Pod rendering: `automountServiceAccountToken` is true when `serviceAccountName` is set
```
0xFelix: Should be false, no? Only true when a `serviceAccount` volume is present?

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Implementation History
+
+_To be filled in as implementation progresses_
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `VMIServiceAccountName` guards the new `serviceAccountName` field
```
0xFelix: I wonder if this really requires a feature gate

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: Yeah this is currently enforced

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: Good question! This is not really any new functionality just a new way to enable existing functionality

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: Good catch!

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: @vladikr I've removed the featuregate since this is not really a new feature. Since there is no featuregate does this really need to be a VEP at all?

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Motivation
+
+KubeVirt's current mechanism for specifying which service account a VM runs as is the `serviceAccount` volume source. When a user specifies this volume, KubeVirt does two things:
+
+1. Sets `pod.spec.serviceAccountName` on the virt-launcher pod
+2. Mounts the service account token as an ISO disk or virtiofs filesystem exposed to the VM guest
+
+This coupling made sense when the primary use case was "I want the VM to access the Kubernetes API using a specific service account." However, with the introduction of ContainerPath volumes (VEP #165) and the growing adoption of cloud provider workload identity systems (AWS IRSA, Azure Workload Identity), a new pattern has emerged: users need the virt-launcher pod to run as a specific service account so that external mutating webhooks inject the correct credentials, but they **do not want** the Kubernetes service account token exposed to the VM.
```
EdDev: It seems odd to ref an Alpha feature on a VEP that attempts to go directly to GA.
The ref to that feature repeats through this VEP, consider keeping it as a soft note to decouple this proposal from it.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- **External secret injection**: Vault Agent Injector and similar tools use service account identity to determine which secrets to inject into the pod
+- **Audit attribution**: Service account identity appears in Kubernetes audit logs for traceability
+
+In all of these cases, the service account is used as a **pod-level identity** rather than a credential to expose inside the VM. The current API forces users to conflate these concerns.
+
+This mirrors Kubernetes Pods, which have always had `spec.serviceAccountName` as a first-class field separate from volume configuration.
+
+## Goals
+
+- Add `serviceAccountName` to `VirtualMachineInstanceSpec` to set the virt-launcher pod's service account identity without requiring a volume
```
EdDev: This seems as the solution, not the goal.
If I understand this correctly, the goal is to allow users* to control the pod service account without requiring a volume.

* I am unsure who this "user" is. The one who creates individual VMs is in this case a "user"?

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- Maintain backward compatibility with existing `serviceAccount` volume usage
+- Align KubeVirt's API with Kubernetes Pod API conventions
```
EdDev: This enters the implementation domain IMO. There are many Pod related fields which we do not expose through the VMI directly. The exposure is now is breaking in a sense the abstraction/encapsulation over pods.
I would drop it, as it has no functional implications IMO. See later other ways to do this without requiring VMI spec changes, making this an implementation detail.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- Simplify the user experience for workload identity patterns (IRSA, Azure Workload Identity)
+
+## Non Goals
+
+- **Removing the `serviceAccount` volume type**: The existing volume type remains useful when users explicitly want the token exposed to the VM
+- **Changing the behavior of existing `serviceAccount` volumes**: Existing VMs using `serviceAccount` volumes continue to work identically
+- **Adding `automountServiceAccountToken` to VMI spec**: This proposal focuses solely on `serviceAccountName`. Whether to expose the token to the VM is handled by the presence or absence of a `serviceAccount` volume
```
EdDev, line 80: Please try not to mix implementation details in this section. You could just rephrase to something like:
> Expose the SA token to the VM.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- **Token projection configuration**: Advanced token projection settings (audiences, expiration) are out of scope for this proposal
+
+## Definition of Users
+
+- **Platform Engineers**: Teams deploying KubeVirt in cloud environments that use workload identity systems (AWS IRSA, Azure Workload Identity)
+- **Application Developers**: Developers running VMs that need cloud provider credentials but don't need direct Kubernetes API access from within the VM
```
EdDev: Application developers sound to me like the ones operating inside the guests.
The ones that define and start VMs are usually namespace admins.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- **Kubernetes Administrators**: Teams that use service account identity for policy enforcement, service mesh authorization, image pull secrets, or audit attribution
```
EdDev: Are these "Cluster Admins"?

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- **Security-Conscious Operators**: Administrators who want to follow the principle of least privilege by not exposing Kubernetes service account tokens to VMs unnecessarily
+
+## User Stories
+
+- As a platform engineer on EKS, I want my VirtualMachine's virt-launcher pod to run as a specific ServiceAccount for IRSA credential injection, without exposing the Kubernetes service account token to the VM guest
+- As an application developer, I want to specify which ServiceAccount my VM runs as without having to configure an unwanted service account token volume
+- As a security-conscious operator, I want to avoid exposing Kubernetes service account tokens to VMs that only need cloud provider credentials, reducing the attack surface inside the VM
+- As a KubeVirt user migrating from Pods, I expect `serviceAccountName` to work similarly to how it works on Pods, as a simple field on the spec rather than requiring a volume
```
EdDev: Can you rephrase it to avoid the implementation detail? Just to mention that the SA needs to be defined with a custom one.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+    }
+    return ""
+}
+```
+
+The pod's `automountServiceAccountToken` is set to `true` only when a `serviceAccount` volume is present. When only `spec.serviceAccountName` is set (no `serviceAccount` volume), `automountServiceAccountToken` remains `false`. Cloud provider webhooks (IRSA, Azure Workload Identity) inject their own projected volumes independently of this setting, so they work regardless.
+
+### Immutability
+
+`spec.serviceAccountName` is immutable after VMI creation, consistent with Kubernetes Pod's `spec.serviceAccountName` and the general immutability of VMI spec fields.
```
EdDev: I suggest you focus on it requiring VM restart. VMI spec is immutable to regular users, but it is mutable to the VM controller in order to support hotplug operations on some fields.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+  template:
+    spec:
+      serviceAccountName: my-sa
+      domain:
+        resources:
+          requests:
+            memory: 1Gi
+```
+
+## Alternatives
```
EdDev, line 247: I think that alternatives which are not answering the requirements/goals, do not need to be listed.

I would also suggest to consider two points:
- Managing this per a group of VM/VMI. It is not clear if there is a scenario where individual VMs need specific SA. Some grouping may be more appropriate and save low level API fields.
- Consider [MutatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/) instead, as an addon deployed by the user. If KubeVirt is to assist, then we could expose it with a small abstraction through a VM/VMI subresource (virt-api).

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+
+**Rollback:**
+- Rolling back to a version without `serviceAccountName` will cause VMs that use only `spec.serviceAccountName` (without a `serviceAccount` volume) to lose their pod identity setting
+- VMs will default to the namespace's default service account
+- Users should add `serviceAccount` volumes before rollback if pod identity is required
+- VMs using `serviceAccount` volumes (with or without `spec.serviceAccountName`) are unaffected by rollback
+
+**VM Migration:**
+- `serviceAccountName` is part of the VMI spec, not runtime state, so migration is unaffected
+- Both source and target virt-controllers must support the field for it to take effect
```
EdDev: There is only one active virt-controller component AFAIK. It includes the VM, VMI and VMIM controllers.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- Simulated IRSA pattern: VM with `serviceAccountName` + ContainerPath accesses injected credentials
+- VM migration with `spec.serviceAccountName` set
+
+## Implementation History
+
+_To be filled in as implementation progresses_
+
+## Graduation Requirements
+
+This feature does not require a feature gate because it introduces no new functionality — it provides a new, more ergonomic way to configure existing behavior (setting `pod.spec.serviceAccountName` on the virt-launcher pod). The `serviceAccount` volume already enables this, and `spec.serviceAccountName` simply decouples pod identity from token exposure.
```
EdDev, line 410: The addition of a new field seems to quality to an API change and the need to evaluate its relevance. There are alternatives IMO which could also fit here, and the logic does exist (e.g. to co-exist with an existing functionality, the volume one).

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- Validation: `spec.serviceAccountName` must be a valid DNS subdomain name
+- Pod rendering: `automountServiceAccountToken` is false when only `serviceAccountName` is set (no `serviceAccount` volume)
+- Pod rendering: `automountServiceAccountToken` is true when a `serviceAccount` volume is present
+- Pod rendering: no service account token disk is created when only `serviceAccountName` is set (no `serviceAccount` volume)
+- Backward compatibility: existing `serviceAccount` volume behavior unchanged
+
+### Functional Tests
+- VM with only `spec.serviceAccountName` boots successfully with correct pod identity
+- VM with `spec.serviceAccountName` and matching `serviceAccount` volume works correctly
+- VM with mismatched `spec.serviceAccountName` and `serviceAccount` volume is rejected
```
EdDev: These seem to fit unit tests to me.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,402 @@
...
+- VM with only `serviceAccount` volume (no `spec.serviceAccountName`) works as before
```
EdDev: I would expect existing tests to cover it, no? Maybe just mention it already exists, and that it appears here for completness.

EdDev (changes requested):

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
EdDev: I think the VEP is required because of the API change.
The FG thing is questionable, I am not sure it is that straightforward to assume it is correct to expose this new field without trying more things first (or at least discussing them).

Ref: https://github.com/kubevirt/enhancements/pull/251#discussion_r3148513730

mhenriks: Updated the alternatives section

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: deemphasized ContainerPath

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: clarified

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
mhenriks: I get the concern about Pod mapping and removed that but I'd argue that service account identity is a workload-level concern, not a pod implementation detail. Also, VMI spec already exposes this concept through the serviceAccount volume

mhenriks:
Thanks for the thorough review @EdDev. Made considerable updates

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
vladikr: Yes, I agree that the service account ID belongs on the VMI object. Overall, it decouples the pod id from mounting the SA token into the VM.
So the new serviceAccountName (id) would be set without the need of mounting the token volume into the VM.
To me, the new field could be about identity management/federation and authentication of that workload, not just the underlying pod.   
@mhenriks I'd add a short section on why we need to expose this field to directly answer @EdDev question, explaining why service account identity crosses the line into workload API, unlike other pod specific fields.

I'm +1 for not requiring FG, as this isn't a new functionality, to me, this is an improvement that helps with decoupling and keeps the existing functionality backward compatible

mhenriks: Thanks @vladikr! I've added a paragraph at the end of the Motivation section making the case that service account identity is a workload-level concern, not a pod implementation detail, and that the existing serviceAccount volume already established that this belongs on the VMI spec. Let me know if that covers what you had in mind

vladikr:
> Thanks @vladikr! I've added a paragraph at the end of the Motivation section making the case that service account identity is a workload-level concern, not a pod implementation detail, and that the existing serviceAccount volume already established that this belongs on the VMI spec. Let me know if that covers what you had in mind

Perfect, looks good to me! Thank you.

0xFelix (commented):
/lgtm
/hold

Looks good to me, putting hold to give others a chance to chime in

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,420 @@
...
+- Doesn't solve the problem of not wanting the token at all
+- Confusing interaction between spec-level and volume-level service account names
+
+**Conclusion:** Rejected because it doesn't address the core issue of unwanted token exposure
+
+### Alternative 5: Group-Level Service Account Management
+
+Manage service account assignment at a group level (e.g., per namespace or label selector) rather than per-VM.
+
+**Pros:**
```
EdDev, line 336: There is another pro: Only the cluster-admin controls the SA assignment, which may provide some security protection (e.g. choosing wrongly the SA for a specific VM).

I guess that could also be considered as a con, because it requires admin intervention.

veps/sig-compute/250-vmi-serviceaccountname/vmi-serviceaccountname.md
```diff
@@ -0,0 +1,420 @@
...
+
+Deploy a [MutatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/) that mutates the virt-launcher pod's service account. KubeVirt could optionally assist by exposing this through a VM/VMI subresource.
+
+**Pros:**
+- No VMI API changes needed
+- Leverages standard Kubernetes extension mechanisms
+
+**Cons:**
+- MutatingAdmissionPolicy is a cluster-scoped resource, so only cluster admins can create them. Namespace admins who want to set a service account on their VMs would need to request a cluster admin to create or modify an admission policy, creating an unnecessary operational dependency
+- Users don't create virt-launcher pods directly — virt-controller does. A policy would need to match pods created by virt-controller and correlate them back to the originating VMI, which is fragile and indirect
```
EdDev, line 356: I do not think it is fragile. We link firmly between the two and depend on it today with other integrations.

EdDev (approved):
Thank you for the changes, it looks convincing now :)

vladikr:
Thank you!
/approve

PR #252, opened 2026-04-04
VEP 248: Live Migration Convergence Improvements

Aseeef:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/248
**SIG label**:  
/sig compute

### What this PR does
Proposes significant enhancements to live migration by proposing a migration stall detection feature that also simultaneous tries to optimize the downtime even while stalled.


veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,544 @@
...
+-->
+
+## Repos
+
+<!--
+List of repositories this design impacts
+-->
+- kubevirt/kubevirt
+
+## Design
```
Aseeef, line 101: TODO: design neglects interaction with autoConverge

veps/sig-compute/migration-convergence-enhancements.md
```diff
@@ -0,0 +1,544 @@
...
+
+## Motivation
+
+<!--
+Why this enhancement is important
+-->
+
+1. On high dirty-rate VMs where dirty-rates are more than the network bandwidth, migration runs for a long time wasting energy, compute and bandwidth even if the migration had already gotten as close to convergence as it could much earlier. There are two main reasons for this:
+    (a) The defaults on `completionTimeoutPerGiB` are conservative with a value of 150s. For example, a moderately sized VM with 16GB memory can attempt to continue migrating for over 40 minutes before timing out. This is not necessarily a bad thing because, while uncommon, there are indeed times when migrations benefit from larger completion times.
```
michalskrivanek: I would go a bit further than "uncommon", and still say 99% of the time it's a bad thing. You're killing the network for 40 minutes for....nothing. I highly doubt there is a case where admin wants to both wait for 40 minutes _and_ overloading connectivity _and_ have 300ms downtime. And this example is just for 16GB VM that I would consider small by current standards.
For 256GB VM this number is just outright wrong.

veps/sig-compute/migration-convergence-enhancements.md
```diff
@@ -0,0 +1,544 @@
...
+    (b) The existing stall detection mechanism in KubeVirt is weak. Currently, it works by monitoring progress on remaining bytes during a migration. The timeout on the migration is controlled by the `progressTimeout` field and resets whenever remaining bytes decrease. However, this ignores that during a typical stall, remaining bytes on the migration typically fluctuate up and down due to phase variations in the underlying workload that cause dirty rates to vary.
```
michalskrivanek: yeah it shouldn't reset. that's just wrong re-implementation of the original ovirt behavior where the timer/timeout should reset on the absolute minimum since the start.

veps/sig-compute/migration-convergence-enhancements.md
```diff
@@ -0,0 +1,544 @@
...
+
+> __Bandwidth Note (`bandwidthBpms`)__
+Libvirt exposes a `bandwidth` estimate via `DomainJobInfo` that is effectively a ~100 ms snapshot, so the raw value jitters between iterations. We smooth it with an exponentially weighted moving average (EWMA): `ewmaEstimate ← α·sample + (1−α)·ewmaEstimate` (seeded with the first sample). We use **α = 0.6**, matching the implementation—a reasonable balance between reacting to real shifts and damping short-lived spikes.
+
+Even in cases where post-copy is permitted, we prefer using pre-copy only as long as this implied downtime stays under the `maxDowntime` threshold. The idea is to avoid post-copy where possible as post-copy by design is riskier than pre-copy and sensitive to failures.
```
michalskrivanek: the feedback I'm getting from the field is that the fear is exaggerated, so I wonder if we shouldn't rather shoot for minimizing the overall duration by switching to post-copy sooner. Not too soon, but also not wasting pre-copy time and bandwidth unnecessarily. I mean...if you've successfuly moved 64GB of data, and you anyway want to switch to post-copy for the last chunk, then the risk difference between switching at e.g. remaining 3GB vs 2GB is negligent.

Aseeef: Hmm... you are right. With typical values of maxDowntime and typical bandwidths we would probably be talking megabytes of difference.

Good catch. I'll modify this.

veps/sig-compute/migration-convergence-enhancements.md
```diff
@@ -0,0 +1,544 @@
...
+
+[TODO] This check could be triggered by a sudden drop in the network (i.e. bandwidth drops close to near zero causing downtime predictions to shoot up). Switching to post-copy in this scenario would be BAD.
+
+
+#### Choosing a Sane Default for Max Completion Time Per GiB
+
+In our opinion, the current default value of `completionTimeoutPerGib` (150s) is acceptable. The worst-case scenarios where such large completion times occur are when average dirty rates are just below average bandwidth. On a 1000 Mbps link, if dirty rates are between ~945 Mbps and 1000 Mbps, completion time per GiB would be at least 150s. That seemed a plausible enough scenario that we think it is reasonable not to change this default.
```
michalskrivanek: I think it's too much. on 1Gbps link you're supposedly able to move 1GB/s of pages (or 500MB/s if you consider the default of parallelOutboundMigrationsPerNode: 2) I honestly don't think there's much point in so many iterations if we are in the end switching to post-copy. Yes you may get lucky when you're close and succeed with pre-copy only in 103rd iteration, but considering you're fully utilizing the network...eh

Aseeef: So my goal here was to actually propose a lower value for completionTimeoutPerGib but I wanted to propose a value I could justify. So I created this spreadsheet to play around with some values: https://docs.google.com/spreadsheets/d/1V-FcX6gO4mNx5fBTgnBq-UnkXAn2y4pTjuYtgDZ1cCc/edit?gid=0#gid=0. However, I was struggling to justify a "sane default". Still though especially if your compression VEP gets approved, we can definitely make a case to reduce this value. But coming up with a good, defendable number is an open problem.

michalskrivanek: I think it's mostly due to the current default being no longer sensible, 300ms for 1Gbps is just too low for today's workloads/sizes. Planning for 100Mbps bandwidth is also too low, I would use 300Mbps at least...maybe that's a good way how to think about the autopilot warnings, @tiraboschi. Regardless the actual link speed, an arbitrary minimum bandwidth that we consider a standard today is sufficient.

Aseeef: They key metric is the effective transfer rate (bandwidth - dirty rate). Even if we assume typical rates of 300-1000mbps, workload dirty rates vary.

tiraboschi: Exposing more metrics + alerts is a valid fallback, but it shifts the problem to the cluster admin. I still believe we should aim for a default that is robust enough to handle the majority of workloads without manual tuning. We should try to solve the 80% case programmatically before expecting admins to react to alerts and tweak configurations.

michalskrivanek: yeah but there's a limit in what we can do - I would say we should define it as minimum supported bandwidth. Not like a guarantee to be able to migrate, but generally used by us to derive the timeouts.

veps/sig-compute/migration-convergence-enhancements.md
```diff
@@ -0,0 +1,488 @@
...
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
+For example, during the planning phase for version v1.123, do **not** target beta for v.124 in advance.
+-->
+
+- This VEP targets alpha for version: 1.9.0
+- This VEP targets beta for version: 1.10.0
+- This VEP targets GA for version: 1.11.0
```
iholder101: Please leave these empty. They should be filled on the planning phase of the relevant cycle (e.g. 1.10 / 1.11)

iholder101 (commented):
@Aseeef Thank you!
Please ping me when this is ready for review!

Aseeef:
@iholder101 apart for the section on "Choosing a Sane Default for Max Completion Time Per GiB", I think I'm done. I have no good answer for "Choosing a Sane Default for Max Completion Time Per GiB".

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,678 @@
...
+
+> __Target Downtime:__
+The "target downtime" refers to the downtime configured in QEMU. This is the threshold that when met, QEMU triggers its internal switch-over to stop-and-copy.
+
+We achieve this by proposing a new API field called `maxDowntime` which specifies the maximum acceptable downtime before we consider additional downtimes as a "workload disruption". Today migrations use QEMU’s fixed **300 ms** target and fail migration if pause would exceed it unless post-copy is permitted or workload disruptions are allowed (e.g. `allowWorkloadDisruption`).
+
+In pre-copy only mode, after stalling, we compare `maxDowntime` to an **implied downtime** derived from the switch-over target: `impliedDowntimeMs ≈ bestRemainingBytes / bandwidthBpms` (bytes per ms). Recall that `bestRemainingBytes` is collected during the "sample iterations" during the stall detection phase. **If** `impliedDowntimeMs` is less than or equal to `maxDowntime`, we set QEMU’s target downtime as this `impliedDowntimeMs` so pre-copy can eventually converge. **else if** `maxDowntime * 1.5 >= impliedDowntimeMs`, we set target downtime to `maxDowntime` and hope a later
```
Aseeef: So basically if the "implied downtime" is 1400ms, and maxDowntime is 1000ms, since the implied downtime is with in maxDowntime * 1.5, we consider it still to be in a plausible range. And since the only alternative is to abort the migration, this is the only case where our stall detection algorithm defers to the max competition timeout to act as the final line of defense.

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,676 @@
...
+To handle this case, we keep sorted, distinct **`remainingBytesHistory`** since **after** stall and **relax** by raising **`bestRemainingBytes`** to the **next smallest** if, after a patience window, the desirable value is not reached. The "since after stall" makes sense because if this failure to find a good point is due to a workload shift, we want to look at the more recent points for the next best value.
+
+1. Let `patience` = **`progressTimeout`** seconds.
+2. Wait up to `patience` seconds for `remainingBytes ≤ bestRemainingBytes × 1.04`.
+3. If still not converged, set `bestRemainingBytes` to the **next smallest** distinct value in `remainingBytesHistory`.
+4. Set `patience = max(patience / 2, 1)`.
+5. Repeat steps 2–4 until switch-over or timeout.
+
+
+### Choosing a Sane Default for Max Completion Time Per GiB
```
michalskrivanek: @Aseeef as we also discussed offline, I wonder if it's not worth another separate proposal to deprecate and remove. The initial Iteration 1 copy is a mandatory and "fixed", depending on bandwidth, memory size (and optionally compression efficiency), but you can't do anything about it, it just needs to happen. Only then it makes sense to actively watch progress/convergence, and it's no longer VM size dependent.

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,676 @@
...
+
+
+### Feature Gate
+
+A new **Alpha** feature gate `MigrationStallDetection` is introduced and registered in `pkg/virt-config/featuregate/active.go` with `State: Alpha`. When disabled, all behavior remains identical to the current implementation.
+
+
+### Plumbing: Event-Based Iteration Loop
+
+The migration monitor loop needs to react to iteration boundaries fired by QEMU. QEMU emits a `MIGRATION_PASS` QMP event each time it completes a dirty bitmap sync at the start of a new pre-copy iteration (i.e., the previous iteration's page transfer is complete and the dirty bitmap has been re-scanned). Libvirt translates this into a `VIR_DOMAIN_EVENT_ID_MIGRATION_ITERATION` domain event. In the Go API, hooking into this domain event only requires us to expose a `DomainEventMigrationIterationRegister(callback) (int, error)` method in the `Connection` interface in `pkg/virt-launcher/virtwrap/cli/libvirt.go`.
```
michalskrivanek, line 335: what's the benefit of adding the event handling when the monitor will still do 400ms polls? We may miss some iterations if relying solely on polls. 
But when that happens today It should mean that you're in the 300-400ms range so you see more than one iteration passed during one monitor loop. But is that really interesting if we're considering  ~1 max, or even today with 300s default if you miss an iteration here or there...

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: Hmm. Perhaps we can mention this at either sig compute or today's VEP discussion. But your proposal to:
1. Exclude the first iteration from max completion time timeout
2. Decrease max completion time timeout

I think you said this in response to me mentioning what if for large VMs memory size after a certain point increases the timeout by a lower amount. So I think that would be the third part of this proposal, correct?

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: I think you are referring to running the polling loop alongside the event loop (i.e. we keep the event loops, but polling loops run in parallel).

This is indeed a concern that I was thinking about. For now, I decided to ignore it until telemetry shows delays from the polling loop is significant. Based on the code however, Gemini said it expected on the order of low, single digit ms delays. The main performance concern here would be grabbing the current job stats (which by the way is also internally synchronized via locks so putting the polling loop in a different go routine probably won't help much).

The point of the polling loop is to (a) logging visibility especially between long iterations, (b) performing timeouts in case iteration takes too long, (c) updating the bandwidth EWMA estimator.

(c) is important because QEMU's reported bandwidth via QMP can be unstable because they are based on 100ms intervals. One alternative is to apply EWMA regularization to bandwidth at the start of each iteration on the event but my concern with that was that the duration between each event can be large. A second alternative was to skip the polling loop if based on remaining data, bandwidth, time elapsed since last iteration, we expect the current pre-copy round to finish soon (thus trigger the event). But that introduces complexity concerns and I'm not sure how accurate that would be...

In any case, the silver lining is this is not a correctness concern (stall detection still will not be wrong) but rather a quality concern (trigger switch-over at a non-optimal time).

veps/sig-compute/248-migration-convergence/migration-convergence.md
michalskrivanek: The problem is the "perGib" part only makes sense for Iteration 1. From that time on the time to converge is basically independent of memory size. We can apply the sliding window and tuning and whatnot, but ultimately it does make sense to have an overall time-based (or iteration-based) upper bound so we don't migrate for weeks.
I could imagine we an repurpose this to just apply to iteration 1, but then we still need some upper bound (not necessarily configurable) for convergence phase. It could perhaps be the same value, just multiplied _not_ by size. 

example: if we say reasonable minimum bandwidth is 400Mbps, so 50MB/s, take some buffer and use e.g. 33MB/s at least, so roughly "30s" per GB. Then apply that separately as 30s * 8 to 8GB VM for initial pass (240s) and then 30s * 6 (180s) for convergence. 6 is arbitrary, loosely meaning "6 times as long as you're assuming 1GB of dirty pages takes, or "for how many GB of dirty pages you want to keep trying")
- aborting if the bandwidth is significantly lower than expected minimum
- aborting (or AllowWorkloadDistuption) if VM doesn't converge in reasonable time

veps/sig-compute/248-migration-convergence/migration-convergence.md
michalskrivanek: not worried so much about the delay (though they may fire pretty quickly if we would set a low downtime...so maybe that concern is real in some circumstances, dunno), but more about the concurrency of the JobStats calls. It is guarded by lock but that's exactly the issue, if the current poll loop holds the lock it's going to block the event handling. Do we need to act exactly at the top of the iteration via event, or would it be perhaps sufficient to react every 400ms i.e. just counting iterations. Yes there is a possibility there would be 2 or more within the poll, but does it really matter? IIUC that event doesn't carry any data other than iteration #

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: Thinking about this some more, one concern for not including the first iteration in the timeout is that if the first iteration is starved for bandwidth (busy node?) that we could have unbounded iterations for that too right? Or maybe that's not a real concern?

I don't know, or maybe the per gb approach is fundamentally just wrong. Claude suggested something like:
> T_max ≈ (memory_size * 2) / bandwidth

Then there is also the aspect that many (but not all) applications actually don't scale their working set (i.e. the pages actively getting dirtied) as the amount of total memory increases. And this results in a scenario where in many cases, we need a larger timeout per Gb for smaller VMs but can get away with smaller timeouts per Gb for larger VMs.

Especially if compression is integrated that might change what we set timeouts to as well... I think this problem is non-trivial enough to warrant a dedicated proposal or if not a proposal I do think it is definitely out of scope for this VEP and I'm considering deleting this section all together.

Thoughts?

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: In the current design, there is no locking. Polling and events both run in a shared loop (see implementation pseudocode). Synchronization is achieved through channels.

> Do we need to act exactly at the top of the iteration via event, or would it be perhaps sufficient to react every 400ms i.e. just counting iterations... IIUC that event doesn't carry any data other than iteration

Acting at event boundaries is important because that's when the remaining bytes is most accurate (since we just synced the recently dirtied bytes and not much time has passed for additional dirtying to happen). So what's important about the event-based trigger is the timing to act at a local minima.

Aseeef: Just ran some tests profiling how long the polling loop takes to execute (. The test was run on a kubeadm cluster. Migration was between two machines both with Intel(R) Xeon(R) W-1290 CPU @ 3.20GHz. The machines weren't doing anything else so were relatively idle. Across two migrations the stats were (in ns).

Min    : 343,888.00
Max    : 1,411,081.00
Avg    : 694,688.50
Std    : 92,518.21
Median : 690,270.00

Tests were on top of https://github.com/kubevirt/kubevirt/pull/17301 though. 

Edit: Also, while I didn't capture this in a my tests, once I while I see in the final loop, it can take 100ms+ for GetJobStats to return. I think that is due to a benign race in #17301 for when migrationErr chan closes while in the middle of the monitoring loop. Not really super relevant for this PR since this only triggers after the migration already succeeded.

michalskrivanek: yeah the expensive part is the jobstats. Wouldn't it be best and pretty non-intrusive to just kick the monitor loop via channel from the iteration event? that pattern is already there for migrationErr, and you get as close to iterations boundaries as possible, with the only "drawback" being the monitor loop samples at irregular intervals, which is IMHO fine. It probably makes sense as a standalone small functional enhancement - we can poll less frequently (for the purpose of timeouts and long iterations), and we'll get more accurate numbers...

Aseeef: > kick the monitor loop via channel from the iteration event

This is exactly the plan :)

> we can poll less frequently

I'm not fully sure about this mainly because im relying on poling loop for bandwidth info. But it might be fine to reduce frequency.

michalskrivanek: There's nothing wrong with 400ms, but it does feel needlessly frequent for the first iteration and insufficient when you're closer to converge. With iteration event trigger it could be that 1s is optimal for the first phase. And we can then log every poll instead of every 10th right now as a small improvement to debugging...

veps/sig-compute/248-migration-convergence/migration-convergence.md
michalskrivanek: I do not see any issue with applying this to the first iteration only besides the compression aspect. And it's just about inaccuracy, not that it would cause any real problem. We can only overshoot the timeout with higher compression ratios, but worst case scenario is the same as no compression.

It does make sense to take it out, I would say it could even be a bug fix instead of yet another VEP. The current behavior is just...misleading and barely usable in practice.

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: > but it does feel needlessly frequent for the first iteration and insufficient when you're closer to converge.

I didn't fully follow what you were referring to here (i.e. insufficient in what), but at least in terms of logging, ATM I'm planning on:
* Logging every single iteration at least once
* If an iteration is taking too long, poll loop will still log (right now it logs every 4 sec)

michalskrivanek: Yes, exactly. With slightly longer polls (say 1s) you can log every time. As well as on each iterarion since it has now significant meaning for the algorithm. 
Sorry for not being clear

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,676 @@
...
+
+UX research often cites **~10s** before users treat a UI as hung; Kubernetes’ default probe period is **10s**; many app timeouts cluster near **30s** (SSH, browsers); Windows gives up a TCP handshake after **~21s** without reply (Linux is more lenient). We infer operators care most about keeping **final** downtime **well under ~15s**; beyond that, marginal gains from finer tuning usually do not justify much longer stall-detection delay.
+
+At **60s** we still expect $\geq 4$ sample iterations. Assuming each sample iteration draws remaining bytes from a stationary(ish) distribution, $P(\text{at least one} \leq \text{median in 4 tries})$ = $1 − 0.5^4 = 0.9375$ which are acceptable odds of acting near a favorable remaining-byte level.
+
+
+### The Stall Margin
+
+While this VEP proposes "4%" as the threshold to be to the Stall Margin, values between 3% to 10% are likely still reasonable. However, this VEP choses to be more on the conservative side.
```
Aseeef: Thought: maybe for post-copy this stall margin can be higher. This is because the outcome in stop-and-copy is service downtime. But in post-copy is service degradation. Perhaps it makes sense to permit slightly worst remaining bytes (by bumping up stall margin when in post-copy) since the outcome isn't that bad? Thoughts?

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,713 @@
...
+
+> __Target Downtime:__
+The "target downtime" refers to the downtime configured in QEMU. This is the threshold that when met, QEMU triggers its internal switch-over to stop-and-copy.
+
+We achieve this by proposing a new API field called `maxDowntime` which specifies the maximum acceptable downtime before we consider additional downtimes as a "workload disruption". Today migrations use QEMU’s fixed **300 ms** target and fail migration if pause would exceed it unless post-copy is permitted or workload disruptions are allowed (e.g. `allowWorkloadDisruption`).
+
+In pre-copy only mode, after stalling, we compare `maxDowntime` to an **implied downtime** derived from the switch-over target: `impliedDowntimeMs ≈ bestRemainingBytes / bandwidthBpms` (bytes per ms). Recall that `bestRemainingBytes` is collected during the "sample iterations" during the stall detection phase. **If** `impliedDowntimeMs` is less than or equal to `maxDowntime`, we set QEMU’s target downtime as this `impliedDowntimeMs` so pre-copy can eventually converge. **else if** `maxDowntime * PRE_COPY_POSSIBLE_FACTOR >= impliedDowntimeMs`, we set target downtime to `maxDowntime` and hope a later
```
Aseeef: Not a big fan of this policy in the else-if on second thought since its the only part relying on max competition timeout.

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,713 @@
...
+
+We go through each of these examples including the workload setup and how the presented stall detection algorithm might react to each of these scenerios in the below presentation:
+**Video Walkthrough**: https://youtu.be/T5_W7X7o70k
+**Presentation Slides**: https://docs.google.com/presentation/d/e/2PACX-1vQj2xFyHHLCKvjXjyFjH3DpIaLKsAGexadrPm1nmoLbtiidU-XJhSbDJzpvMdyAWISXYFAHfN6JMFhT/pub?start=false&loop=false&delayms=10000
+
+
+### Limitations & Risks
+
+In this section we acknowledge certain limitations of this approach and list scenerios where the current timeout approach wins:
+* Network drops can cause the stall detection algorithm to trigger (due to remaining bytes increasing). While we won't trigger switchover until we get remaining bytes to be at least as low as the best previous iteration, nevertheless, it is possible to the switchover decision to trigger while pre-copy is still making improvements after network recovers. Perhaps future work can better address this edge case.
```
Aseeef: Maybe a simple fail safe: never trigger post-copy if bandwidth is lower than 10(?) Mbps?

veps/sig-compute/248-migration-convergence/migration-convergence.md
michalskrivanek: I wonder if i wouldn't be better to keep the #249 separate and independent. It seems like a more straightforward logic, we wouldn't need to introduce impliedDowntime or do anything else here for pre-copy, just keep comparing the current "target downtime" with remaining bytes.

veps/sig-compute/248-migration-convergence/migration-convergence.md
michalskrivanek: I would still alarm on unusual slow speeds (outside of this VEP).
With 10Mbps or lower even the first iteration would take ages...and if we repurpose the perGib setting for the first iteration only it would effectively work like a minimum bandwidth guard

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: But if both are calling the set_max_downtime in QEMU, then the implementations are conflicting. We should discuss tomorrow.

michalskrivanek: yeah i meant you would NOT set downtime here, the downtime iterations would be separately increasing it up to user's limit, and you won't compensate anything here, just use the currently set value for the switch-over calculation

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: Right. The intention here is to ensure we never trigger post-copy while the network is down. Whether that's the stall detection or the timeout trigger it.

vladikr:
@Aseeef Thank you for this important work!
Personally, I'm truly happy to see such a thoughtful proposal.

On one hand, any improvement to the current approach would be better than just waiting until the acceptableCompletionTime expires, but I also recognize that we don't have good defaults now. 
I have a few thoughts/observations about the current status of the VEP.

1. I think we should move all the experimental tunables into the `MigrationPolicy.spec.experimental.migrationConvergence` (somewhere in that area)
```
migrationConvergence:
  enabled: true
  observeOnly: true   (I'll explain later :) )
  stallMargin: 0.05
  ewmaAlpha: 0.3
  preCopyPossibleFactor: 1.3
  maxDowntime: 500ms
  minBandwidthMbpsForPostCopy: 50
```
And it would be useful to explain the logic in some easy code, I thought something like:

```
// StallDetectionLoop  -  runs on every MIGRATION_PASS event + periodic poll
func detectStall(currentRemaining, bandwidthBps float64, history []float64) (shouldAct bool, impliedDowntime time.Duration) {
    // lowest remaining bytes seen in sampling window
    best := min(history)                    
    if currentRemaining > best * (1 + stallMargin) {
        return false, 0
    }

    ewmaBw := updateEWMA(bandwidthBps, ewmaAlpha)
    impliedMs := (best / ewmaBw) * 1000

    if impliedMs <= maxDowntime.Milliseconds() {
        // only in active mode
        setQemuDowntime(impliedMs)
        return true, time.Duration(impliedMs)*time.Millisecond
    } else if impliedMs <= maxDowntime.Milliseconds()*preCopyPossibleFactor {
        setQemuDowntime(maxDowntime.Milliseconds())
        return true, maxDowntime
    } else if ewmaBw < minBandwidthMbpsForPostCopy*1e6 {
        log.Warn("bandwidth too low for safe post-copy")
        return false, 0
    } else {
        // trigger post-copy or stop-and-copy only in active mode
        return true, impliedDowntime
    }
}

```
For me it was easier to follow.

2. Regarding the max downtime setting, that both the stall detector and the work in #249 are planning on calling. 
I would suggest that the stall detector would set the *desired* downtime and something else in the main loop will act on it

3. I think we need to consider running the stall detection as a shadow data collector.  This way, we could collect metrics for at least one cycle on real clusters.
We could add a tunable to enable `observeOnly: true` or  `dataCollectionOnly: true` setting to the experimental API to enable this.
We are already collecting a ot of migration metrics via Prometheus and we could simply extend this pattern.
One of the important values would be to calculate the impliedDowntime, best remaining bytes, EWMA bandwidth, etc...

we could add somethign like 

```
kubevirt_vmi_migration_ewma_bandwidth_bytes_per_second
kubevirt_vmi_migration_best_remaining_bytes
kubevirt_vmi_migration_stall_margin
```
And also report the decision the new logic would have made. 
I'm not very good with Promethous - maybe @machadovilaca and @sradco could help us with this.

I looked at how this is done in other places and it's a counter:
```
kubevirt_vmi_migration_stall_decision_total
    Counter
    Labels:
      - mode: "shadow" | "active"
      - decision: "continue_precopy" | "tighten_downtime" | "switch_postcopy" | "use_timeout_fallback" | "abort_low_bandwidth"
      - vmi, namespace, etc. 

kubevirt_vmi_migration_stall_implied_downtime_ms
    Gauge
    Labels:
      - mode: "shadow" | "active"
      - action_taken: "set_implied" | "set_maxdowntime" | "none"
```

What do you think?

@iholder101 @xpivarc @jean-edouard  fyi

michalskrivanek:
1 - I tried to add that as part of the compression in standalone commit https://github.com/kubevirt/kubevirt/pull/17509/commits/dc3b289b8a6aa3c3659b0f48b6e63d0adc3cfd96 , I would propose to hide this behind the same AdvancedMigration gate (just to limit the amount of gates really, and then `.spec.experimental.something.enabled` would drive this feature)
2 - w.r.t. PreCopy the more I think about it the more it makes sense to me to keep downtime settings in qemu separated from stall detection. The sample downtime tuning implementation is straightforward and non-intrusive, and it seems to me it would play well with the stall detection when the same way as autoconverge slows down the guest "on its own".
3 - there's already bunch of useful stats that are not gathered but only logged in virt-launcher. They're going to be relatively reliable with https://github.com/kubevirt/kubevirt/pull/17301 in, it should be easier to bubble them up. One separate metric IMHO makes sense to add to long-term per-VM stats - the experienced/effective Downtime for completed migrations.

Aseeef:
@Barakmor1 and @vladikr could you take a look when possible? I think I have covered all my bases at this stage.

Barakmor1:
Hey, this VEP is quite verbose and hard to review. It also seems related to VEPs 293, 295 and 291, so the context is pretty large. Unfortunately, I won’t be able to review it before the deadline.

vladikr:
I think that the direction here is solid, the safety nets are better, and the observability story is now excellent. 
This is exactly how we want experimental features to be introduced in KubeVirt.
/approve

vladikr:
Going forward, we must produce an adaptive convergence logic that doesn’t require manual stallMargin or ewmaAlpha.

Aseeef:
> Going forward, we must produce an adaptive convergence logic that doesn’t require manual stallMargin or ewmaAlpha.

Subject to findings during late alpha/beta after implementation, I see two directions:

1. ewmaAlpha, stallMargin, and others will become hard coded constants using values that  work for everyone.

2. Another direction is that if results are sensitive enough to changes in hyperparameters, we might offer different "configuration profiles" that cluster admins can specify for how they want to balance trade-offs. i.e. whats more important to them? Minimizing migration durations at the cost of downtime? Minimize downtime at the cost of total migration duration? Something in between?

vladikr:
yes, I think it can be adaptive based on some profile but the key is that in beta we won't to rely on these manual settings, but either rely on data that we will collect during alpha and/or based on a "profile" as you've suggested. I would put this as a graduation criterion.

Going forward I see MigrationPolicy with high-level intents like:
```
migrationProfile: balanced | aggressive | conservative | lowDowntime | postCopyPreferred
allowWorkloadDisruption: true/false
```
Maybe a few workload-class hints (workloadType: database | stateless | ai-training)

Aseeef:
I updated this statement as a Beta requirement:
```
Revisit hyper-parameter fine-tuning and how trade-offs should be exposed to users (since users should not be expected to manually configure knobs like "stallMargin").
```

 The migration profiles and the overall "declarative migration API" we want to go towards is a more general conversation that applies even beyond this VEP. Exactly how this API should be is probably its own VEP and conversation hence the generic comment. Is that fine?

vladikr:
> I updated this statement as a Beta requirement:
> 
> ```
> Revisit hyper-parameter fine-tuning and how trade-offs should be exposed to users (since users should not be expected to manually configure knobs like "stallMargin").
> ```
> 
> The migration profiles and the overall "declarative migration API" we want to go towards is a more general conversation that applies even beyond this VEP. Exactly how this API should be is probably its own VEP and conversation hence the generic comment. Is that fine?

Yes absolutely!

iholder101:
@michalskrivanek are you ready to LGTM?

Aseeef:
> @michalskrivanek are you ready to LGTM?

I believe he is on PTO today and tomorrow.

michalskrivanek:
Yep, looks good! Thanks!
/lgtm

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,693 @@
...
+Today, we already expose:
+* DataRemaining (Bytes Remaining)*
+* MemDirtyRate (Dirty Rate)
+* MemoryBps (Bandwidth)*
+* DataProcessed (Total Bytes Pushed)
+* DataTotal (Total Bytes Pushed + Bytes Remaining)
+* Migration Start Time*
+* Migration End Time*
+
+Currently, this information is not very useful because it is collected at a fixed interval of 5s, whereas our requirements for when this data should be collected are different. Therefore, we propose:
```
machadovilaca, line 251: not useful for the stall detection mechanism or for customer visibility?

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,693 @@
...
+* Emit `DataRemaining` and `MemDirtyRate` only at pre-copy boundaries.
+* `DataTotal` should only be pushed once at the start of a migration (as it does not change mid-migration).
+* `MemoryBps` should be pushed at the same frequency that we update our EWMA metrics (every 400ms, once per polling loop).
```
machadovilaca, line 254: pushed where? prometheus?

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,693 @@
...
+* At the end of each migration, push actual downtime / post-copy duration.
+We also attach application generated timestamps to each event. This is because Prometheus itself uses timestamps based on when it scrapes these metrics resulting in lost precision.
```
machadovilaca, line 256: what does "attach application generated timestamps" mean?

veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -0,0 +1,693 @@
...
+* `patienceWindowDecayFactor`
+* `stallMargin`
+* `ewmaAlpha`
+* `precopyPossibleFactor`
+
+We present an argument for reasonable defaults that we can start with for these hyper-parameters in [migration-hyperparameters.md](./migration-hyperparameters.md), at least for the alpha phase of this proposal.
+
+We hope to fine-tune these values throughout the alpha and beta phase using real-world data from live clusters to deduce optimal values for these hyper-parameters, or perhaps to even re-evaluate parts of this proposal entirely. In order to facilitate this, proper logging and visibility of the proposed stall detector will be required.
+
+###  Logging & Visibility
```
machadovilaca, line 234: i think we need to refactor this section a bit, the terminology you used raises some concern about how you are thinking about the implementation

you often mention "push", but there is not push mechanism, Prometheus polls metrics in 30 second intervals, so lets make sure we are on the same page using the correct names to avoid confusion

i think we need to clarify the collection intervals, what's really needed for user visibility, and how do you intend to control the data flow

machadovilaca, line 234:
```suggestion
###  Observability
```

veps/sig-compute/248-migration-convergence/migration-convergence.md
Aseeef: I mean that KubeVirt itself can attach the timestamp the event occurred because the Prometheus generated timestamps come from when the data gets scrapped (right?) and that can be at a variable frequency. But here, I need high resolution data for it to be useful. Does that make sense?

machadovilaca: you can pass a timestamp to operatormetrics.CollectorResult, we should already be doing that for migration metrics, because data is collected every 5sec, but only polled by prometheus every 30s

Issue #254, opened 2026-04-09
VEP 254: Guest GPU Metrics via VSOCK


machadovilaca:
**Primary contact (assignee)**:

/assign @machadovilaca

**Current Feature Stage**: New

**Feature Gate**: GPUMetrics

**Responsible SIGs**:

Primary SIG:
/sig observability

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (1.9):
* Beta release target (x.):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/255
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:

Collect GPU metrics (utilization, memory, temperature, power, ECC errors) from inside guest VMs via a virtio-serial channel and expose them as `kubevirt_vmi_gpu_*` Prometheus metrics from virt-handler. A lightweight guest agent (`gpu-metrics-agent`) uses NVML to collect metrics and communicates with the host over a newline-delimited JSON protocol. Supports both Linux and Windows guests. See the VEP for full design details.

> [!IMPORTANT]
> Please keep this description up to date.

enp0s3:
@machadovilaca Can you please update the issue title to align with the VEP title?

PR #255, opened 2026-04-10
VEP #254: Guest GPU Metrics via VSOCK

machadovilaca:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/254
**SIG label**: /sig observability /sig compute

### What this PR does

GPU workloads running inside KubeVirt virtual machines currently lack observability. Cluster administrators and users have no way to monitor GPU utilization, memory usage, temperature, power consumption, or error counts for GPUs passed through to VMs.

This VEP introduces a mechanism for collecting GPU metrics from inside the guest and exposing them as Prometheus metrics on the host.


veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+## Alternatives
+
+### VSOCK Instead of Virtio-Serial
+
+VSOCK (`AF_VSOCK`) provides socket-based communication between guest and host without virtio-serial.
+
+**Rejected because:**
+- VSOCK requires kernel support that is not universally available, especially on older guests and Windows.
```
dominikholler: Windows support is upcoming, and I would be surprised if Linux guests which are too old for vsock would utilize GPUs

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
```diff
@@ -0,0 +1,312 @@
...
+- Virtio-serial is already used by KubeVirt for qemu-guest-agent and downward metrics, making it a proven transport.
```
dominikholler: @kostyanf14 (and maybe @jcanocan ) Do you agree?

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
machadovilaca: @enp0s3 said the exactly same on kv vep discussion meeting, if the general opinion is that "Windows support is upcoming" is fine, then I would be okay with it

I actually tried VSOCK first, and was able to get a poc running, ditched it because of the windows support
and since we already used virtio-serial, seemed like a good approach

kostyanf14: Windows already supported

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
```diff
@@ -0,0 +1,312 @@
...
+- Virtio-serial channels appear as simple character devices in the guest, making the agent trivial to implement on both Linux and Windows.
```
kostyanf14: Reject for `Virtio-serial` because we see some problems with huge data transfer https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1462

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
kostyanf14: @YanVugenfirer

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
machadovilaca: here the data sent should be small, i shared the structure in the VEP document

machadovilaca:
@dominikholler @kostyanf14 added vsock as an alternative in the design section

machadovilaca:
/cc @michalskrivanek 

added guest-file-read as an alternative in the design section

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,316 @@
...
+GPU utilization, memory usage, temperature, power consumption, or error counts for GPUs passed through to VMs.
+
+This VEP introduces a mechanism for collecting GPU metrics from inside the guest and exposing them as Prometheus metrics on the host. A
+lightweight guest agent communicates with the host via a virtio-serial channel, and virt-handler scrapes the agent on each Prometheus
+collection cycle to produce `kubevirt_vmi_gpu_*` metrics.
+
+## Motivation
+
+GPU passthrough and vGPU workloads are increasingly common in KubeVirt for AI/ML training, inference, and media processing. Host-level GPU
+monitoring tools like NVIDIA DCGM exporter are not available in these configurations. The NVIDIA GPU Operator does not deploy this service
```
rthallisey, line 34: Should kubevirt be responsible for maintaining a gpu metric system?  This seems generally useful for virtualization.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,316 @@
...
+
+Linux has native support and Windows is supported through the virtio-win driver package.
+
+**Downsides:**
+- Known issues with large data transfers: Windows drivers fail WriteFile calls >2MB.
+- No flow control. The guest can't detect whether the host has connected or disconnected from the socket.
+
+#### VSOCK
+
+VSOCK (`AF_VSOCK`) is a socket address family for guest-host communication using the virtio-vsock transport. KubeVirt already has VSOCK
```
rthallisey: Using Vsock can bypass the launcher and go right to the handler or another node-loacl exporter.  I can see some advantages to that.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,316 @@
...
+```
+
+## Alternatives
+
+### Host-Side GPU Metrics (DCGM / Node Exporter)
+
+Collect GPU metrics from the host using NVIDIA DCGM or the GPU node exporter.
+
+**Rejected because:**
+- The NVIDIA GPU Operator does not deploy DCGM exporter on nodes where GPUs are configured for passthrough or vGPU, because the host no
```
rthallisey, line 260: This is changing.  The DCGM team is working on adding vsock support so we can gather metrics from inside guest and share it over vsock.  Once implemented, it would make sense to always have DCGM exporter on the host.

machadovilaca: do you have any pointers for that work? I tried proposing the support for vsock on dcgm exporter and they suggested pursuing other approaches

https://github.com/NVIDIA/dcgm-exporter/pull/649#issuecomment-4209670676

rthallisey: https://docs.nvidia.com/datacenter/dcgm/latest/release-notes/changelog.html#features

```
Host engine

        Added support for listening on the VSOCK protocol.

        Added support for the following fields:

                DCGM_FI_DEV_GET_GPU_RECOVERY_ACTION

                DCGM_FI_DEV_GPU_RECOVERY_ACTION

                DCGM_FI_DEV_MEMORY_UNREPAIRABLE_FLAG

                DCGM_FI_DEV_NVLINK_ECC_DATA_ERROR_COUNT_TOTAL

                DCGM_FI_DEV_NVLINK_PPCNT_IBPC_PORT_XMIT_WAIT


```

veps/sig-observability/gpu-metrics-via-vsock/vep.md
dominikholler: I do not understand the full meaning of this comment. I just want to highlight that it might be beneficial to avoid bypassing KubeVirt APIs, e.g. because of https://github.com/kubevirt/enhancements/pull/223

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
jcanocan: Virtio-serial is being used already by downward metrics and it works fine. However, I don't have any experience in Windows guests.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
rthallisey: The data path in the vsock approach is simpler, e.g. guest -> exporter.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
machadovilaca: updated to use DCGM with VSOCK

veps/sig-observability/gpu-metrics-via-virtio-serial/vep.md
machadovilaca: updated to use DCGM with VSOCK

veps/sig-observability/gpu-metrics-via-vsock/vep.md
machadovilaca: KubeVirt wouldn't really be maintaining a GPU metric system. The heavy work would be done entirely by DCGM inside the guest, and KubeVirt would handle the communication and exposition of the data.

With that said, NVIDIA DCGM exporter is the de facto way to expose GPU metrics to Prometheus, so it would be ideal and simpler for users if it stayed that way. What I think that would mean:

1- KubeVirt creates a VSOCK for the VMI if any GPU device is configured 
2- DCGM listening of VSOCK is running on the guest
3- NVIDIA GPU Operator deploys DCGM exporter on nodes configured for vGPU and GPU passthrough
4- DCGM exporter connects to VMI's via VSOCK to query DCGM data and expose metrics to Prometheus

I would then add some Prometheus recording rules to correlate DCGM exporter metrics with KubeVirt VMI's.

enp0s3: @machadovilaca I think that adding the VOSCK automatically will be a problematic approach, the correct way to configure VSOCK for a VM is via the KubeVirt API.

enp0s3 (commented):
@machadovilaca Hi. Actually I don't see any added value for the vGPU metrics to be maintained by the KubeVirt tree. It will couple DCGM metrics development with KV release cycle. I think that DCGM can manage with its own metrics exporter.

enp0s3:
@machadovilaca @rthallisey To be more specific, I don't understand the drawbacks of the following approach:
1. Aggregate VMI to vGPU metrics on a higher level using Prometheus PromQL
2. Using the regular VM pod network, attach kubernetes service the VM network and collect the metrics that way, instead of using VSOCK.

enp0s3:
@rthallisey Another drawback of using VSOCK is that we are willing to graduate it to be namespace confined, which might impose obstacles for current DCGM design, leading to the need of escalating the privileges of DCGM to be able to access every network namespace on the node.

machadovilaca:
> @machadovilaca @rthallisey To be more specific, I don't understand the drawbacks of the following approach:
> 
> 1. Aggregate VMI to vGPU metrics on a higher level using Prometheus PromQL
> 2. Using the regular VM pod network, attach kubernetes service the VM network and collect the metrics that way, instead of using VSOCK.

@enp0s3 

1. whether or not we decide to support this VSOCK DCGM approach, if the metrics Prometheus ingests are coming from DCGM directly, we would always need a way to correlate them to VMIs. And even if we end up not doing any action to simplify this metric collection, we can still provide recording rules for the correlations

2. it is just a question of simplifying the work for the end user, the approach you are suggesting should even work right now. But for each VMI the user now needs to: 1. configure network for the vmi, 2. install dcgm in the vmi, 3. create the k8s service, 4. create a service monitor, 5. correlate gpu and vmi metrics

imo we should look for ways to simplify this

rthallisey:
>Another drawback of using VSOCK is that we are willing to graduate it to be namespace confined
>
Fair point.  Perhaps though there's a way to solve `global` vsock with some security hardening.  I'd like to at least explore that option.

The two use cases I see are:
1. Advertise kubevirt specific gpu metrics that tie vmi to gpu
2. Advertise all gpu metrics from a passthrough GPU to the dcgm-exporter

Use case 1 is meant for kubevirt admins.  Use case 2 can tie into the existing GPU tooling ecosystem.

enp0s3:
> > @machadovilaca @rthallisey To be more specific, I don't understand the drawbacks of the following approach:
> > 
> > 1. Aggregate VMI to vGPU metrics on a higher level using Prometheus PromQL
> > 2. Using the regular VM pod network, attach kubernetes service the VM network and collect the metrics that way, instead of using VSOCK.
> 
> @enp0s3
> 
> 1. whether or not we decide to support this VSOCK DCGM approach, if the metrics Prometheus ingests are coming from DCGM directly, we would always need a way to correlate them to VMIs. And even if we end up not doing any action to simplify this metric collection, we can still provide recording rules for the correlations

+1 for recording rules, IMO its much more simpler then the current approach. 

> 2. it is just a question of simplifying the work for the end user, the approach you are suggesting should even work right now. But for each VMI the user now needs to: 1. configure network for the vmi, 2. install dcgm in the vmi, 3. create the k8s service, 4. create a service monitor, 5. correlate gpu and vmi metrics
> 
> imo we should look for ways to simplify this

this can be simplified using tools like Helm.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,292 @@
...
+NVIDIA DCGM 4.5.0 introduced native VSOCK support, allowing the DCGM daemon inside a guest VM to accept connections from the host over
+VSOCK. By leveraging this capability, KubeVirt can collect GPU metrics directly from DCGM without maintaining a custom guest agent, providing
+per-VM, per-GPU observability that is consistent with the existing `kubevirt_vmi_*` metrics namespace and enabling unified dashboards and
+alerting.
+
+## Goals
+
+- Expose per-VM, per-GPU utilization metrics as Prometheus metrics from virt-handler.
+- Support both GPU passthrough and vGPU devices.
+- Support Linux and Windows guests.
```
enp0s3: How are we going to support Windows? Do we have a way to test this?

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Error Detection
+As an operations engineer, I want to be alerted when a GPU inside a VM reports ECC errors so I can proactively migrate the workload before
+hardware failure.
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+
+## Design
```
enp0s3, line 84: We need to mention that VSOCK had to be requested in the VM spec

enp0s3 (commented):
I would prefer this VEP to be eventually converged with VEP 143, if we will go with the VSOCK approach I would like the code to live in the separate monitoring stack

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,292 @@
...
+  name: gpu-workload
+spec:
+  domain:
+    devices:
+      gpus:
+        - name: gpu1
+          deviceName: nvidia.com/A100
+```
+
+## Alternatives
```
enp0s3, line 235: @machadovilaca Can you please add in the alternative why inline communication method using regular kubernetes network was rejected? What are the drawbacks of deploying additional kubernetes resources in order to expose the metric service that is running inside the guest?

In addition what are the drawbacks of using recording rules or any other high level tools to tie VM to GPU? Why explicit instrumentation is needed to tie these resources?

veps/sig-observability/gpu-metrics-via-vsock/vep.md
machadovilaca: my idea was that if the feature gate is enabled and GPUs are present in the spec, virt-controller would automatically attach a VSOCK device to the domain XML

enp0s3: @machadovilaca But its only relevant for NVIDIA GPUs, isn't it?

veps/sig-observability/gpu-metrics-via-vsock/vep.md
machadovilaca: added DCGM via regular Kubernetes Networking alternative

recording rules are not an issue, even if we expose VSOCK for DCGM exporter to collect the DCGM metrics, we would need the recording rules

veps/sig-observability/gpu-metrics-via-vsock/vep.md
machadovilaca: initially my idea was to keep this generic enough to handle other gpus in the future
but with some of the work moving to the dgcm exporter supporting vsock, and nvidia operator deploying the exporter on all use cases, it is now more focused on nvidia, yes

enp0s3: IMO its suboptimal to always attach VSOCK upon GPU resource request.

machadovilaca: i think many times if a gpu is attached, dcgm will be installed in the guest alongside with the drivers, so the dcgm exporter would be able to collect metrics with no additional user operation over what he does today

but i can change that. what do you think is better, using the existing attach vsock field or creating a new one specific for gpu metrics, that if enabled, would create the vsock?

veps/sig-observability/gpu-metrics-via-vsock/vep.md
enp0s3: @machadovilaca ~~I would leave the responsibility to attach the VSOCK for the user. In case the VSOCK isn't attached via the VM spec we won't collect the metrics.~~ sorry confused with another thread.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
enp0s3: @machadovilaca I would leave the responsibility to attach the VSOCK for the user. In case the VSOCK isn't attached via the VM spec we won't collect the metrics.

machadovilaca: ok

then with the new dcgm exporter support for vsock
we should be able to just add some documentation for the users to enable vsock on the spec, install dcgm in the guest, and start the dcgm service listening on the vsock

now, nvidia needs to update nvidia gpu operator to deploy the the dcgm exporter on nodes configured for vgpu or gpu passthrough, which it currently doesn't

on kubevirt side, we would only need to add the recording rules, that correlated the dcgm exporter metrics, with the vmis

enp0s3 (commented):
/lgtm

@machadovilaca Thank you!

enp0s3:
@vladikr Hi, could you please have a look?

enp0s3:
@machadovilaca Three more topics we should converge:
* Windows support, can we commit to it?
* Combined consumption of VSOCK, both by the metric collector and by KubeVirt API. We should mention the limitations.
* The trigger to collect the metrics, there can be non-NVIDIA GPU with attached VSOCK, how would we differ that case?

vladikr:
@machadovilaca @enp0s3 To be honest, personally, I didn't have the capacity to review this proposal deep enough to understand whether it fits KubeVirt or not ...
I am not able to give my approval for this cycle. 
Let's defer it to the next one, which will give us enough time to have a proper discussion.

enp0s3:
@vladikr Makes sense. Sorry for the noise. It looks like we need more time.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,371 @@
...
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+GPU workloads running inside KubeVirt virtual machines currently lack observability. Cluster administrators and users have no way to monitor
+GPU utilization, memory usage, temperature, power consumption, or error counts for GPUs passed through to VMs.
+
+This VEP introduces a mechanism for collecting GPU metrics from inside the guest and exposing them as Prometheus metrics on the host. NVIDIA
+DCGM (Data Center GPU Manager) 4.5.0 added native support for listening on the VSOCK protocol, enabling direct guest-to-host communication
+without a custom guest agent. virt-launcher connects to DCGM inside the guest via VSOCK and exposes the GPU metrics data through the unified
```
rthallisey, line 27: DCGM-exporter running on the host will poll for metrics over the socket.  DCGM should be running in the guest and listening on the socket.  Having dcgm-exporter or dcgm connect to the launcher over vosck doesn't sound right to me.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,371 @@
...
+per-VM, per-GPU observability that is consistent with the existing `kubevirt_vmi_*` metrics namespace and enabling unified dashboards and
+alerting.
+
+## Goals
+
+- Expose per-VM, per-GPU utilization metrics as Prometheus metrics from the observability-controller.
+- Support both GPU passthrough and vGPU devices.
+- Support Linux guests.
+- Leverage DCGM's native VSOCK support to avoid maintaining a custom guest agent.
+- Integrate with the unified `GetVMStats` gRPC call rather than introducing a separate RPC.
```
EdDev, line 49: Implementations detail, less of a goal.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,371 @@
...
+spec:
+  domain:
+    devices:
+      autoattachVSOCK: true
+      gpus:
+        - name: gpu1
+          deviceName: nvidia.com/A100
+```
+
+## Alternatives
```
EdDev, line 235: You could define a goal that restricts customized guest-agents and explain it in the motivation. By that, several alternatives will be removed here.

veps/sig-observability/gpu-metrics-via-vsock/vep.md
```diff
@@ -0,0 +1,371 @@
...
+|  - listens on VSOCK      | vsock  |      TimeDefinedCache      |     |                     |     |                         |
+|                          |        |    scrapeGPUMetrics()      |     |                     |     |  queries virt-handler   |
++--------------------------+        |                            |     |  GetVMStats()       | <-- |  for all VMI data       |
+                                    |  cmd-server (gRPC)         |     |   includes gpuStats |     |                         |
+                                    |    GetVMStats() handler    | <-- |                     |     |  emits:                 |
+                                    |      includes gpuStats     |     |                     |     |    kubevirt_vmi_gpu_*   |
+                                    +----------------------------+     +---------------------+     +-------------------------+
+```
+
+### 1. Opting In: Annotation and VSOCK Enablement
```
EdDev, line 105: This is forcing the core project to handle GPU vendor specific metrics.
It would be easier to adopt and evaluate an add-on solution, out-of-tree.

For example, using an injected sidecar that can operate on the VSOCK socket, exposing the needed to a central collector.
Is there an option to keep this external as much as possible from the core project?

EdDev (commented):
Thank you.

Can you please provide some ref to other virtualization projects or vendors which collect and expose GPU statistics?
I am interested to understand if it makes sense to collect the information by the GPU vendors, independently, and only at the end to merge/cross-link the data of the VM and the v/GPUs.

Issue #256, opened 2026-04-10
VEP 256: OCI Artifact Export for VirtualMachines and VirtualMachineTemplates


0xFelix:
**Primary contact (assignee)**:

/assign 0xFelix

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig storage

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (1.9):
* Stable release target (x.y):

* [ ] Beta
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/257
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

PR #257, opened 2026-04-10
VEP #256: Add OCI artifact export proposal

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/pull/256
**SIG label**: /sig compute

### What this PR does

Add OCI artifact export proposal

### DIFF
--- a/veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
+++ b/veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
@@ -1,0 +1,595 @@
+# VEP #256: OCI Artifact Export for VirtualMachines and VirtualMachineTemplates
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir
+  in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+`VirtualMachineExport` does not support packaging exports into a single
+distributable artifact, and it does not support `VirtualMachineTemplate`
+(VEP #76) as a source kind.
+
+This proposal defines two OCI artifact formats -
+`application/vnd.kubevirt.virtualmachine.v1` for `VirtualMachines` and
+`application/vnd.kubevirt.virtualmachinetemplate.v1` for
+`VirtualMachineTemplates` - and extends `VirtualMachineExport` with an OCI
+layout streaming endpoint. No new CRDs or controllers outside of
+`kubevirt/kubevirt` are required.
+
+The artifacts can be downloaded as
+[OCI image layout](https://github.com/opencontainers/image-spec/releases/tag/v1.1.1)
+archives for local handling with standard tools (oras, skopeo, podman,
+crane), and then pushed to an OCI-compliant registry for distribution across
+clusters. Direct registry push from the export pod may be added in a future
+version.
+
+## Motivation
+
+By packaging VMs and VM templates as OCI artifacts, we get:
+
+* **Registry-native distribution** - push/pull using existing registry
+  infrastructure
+* **Content-addressable deduplication** - shared disk images are stored once
+  in the registry, even when referenced by multiple artifacts
+* **Standard tooling** - oras, skopeo, podman, crane
+* **Signing and verification** - artifacts in registries can be signed with
+  cosign or notation out of the box
+* **Lossless round-tripping** - the full CRD object is stored in the config
+  blob, preserving all fields including template parameters
+* **Native disk format** - raw disk images are stored directly as OCI blobs
+  without any conversion
+
+## Goals
+
+### Phase 1 - `VirtualMachine` export
+
+* Define an OCI artifact format for `VirtualMachine` that stores the
+  resource definition in the config blob and disk images as individual layers
+* Extend the export server with an `/export.oci.tar` endpoint that streams
+  OCI image layout archives
+* Use an OCI Image Index as the top-level object to support
+  multi-architecture VMs from the start
+* Extend `virtctl vmexport` with `--format=oci`
+
+### Phase 2 - `VirtualMachineTemplate` export
+
+* Extend the OCI artifact format to `VirtualMachineTemplate`
+* Extend `VirtualMachineExport` to accept `VirtualMachineTemplate` as a
+  source kind, with the export controller resolving template volume sources
+
+## Non Goals
+
+* Direct registry push from the export pod (deferred to a future version)
+* Import of OCI artifacts into the cluster (deferred to a follow-up design)
+* Support for multi-VM appliances (single VM or template per artifact)
+* Guest-level operations such as sysprep or sealing during export
+* Live export of running VMs
+* Cross-hypervisor portability (the OCI format is KubeVirt-native)
+
+## User Stories
+
+* As a VM owner, I want to export my `VirtualMachine` including all its
+  disks to an OCI artifact so I can migrate it to another cluster.
+* As a VM owner, I want to push my VM to an OCI registry so another cluster
+  can pull and import it.
+* As a VM template owner, I want to push a `VirtualMachineTemplate` to an
+  OCI registry so I can share it across clusters.
+* As a VM template owner, I want to download a template as a file so I can
+  transfer it to an air-gapped cluster.
+* As a VM template owner, I want template parameters preserved in the
+  artifact so the receiving cluster can re-import the template with its
+  parameterization intact.
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Feature Gate
+
+The OCI export functionality is gated behind a new `OCIExport` feature gate
+in the `KubeVirt` CR. When disabled, the export server returns 404 for
+`/export.oci.tar` requests and `VirtualMachineExport` rejects
+`VirtualMachineTemplate` as a source kind.
+
+### OCI Artifact Format Specification
+
+`VirtualMachines` and `VirtualMachineTemplates` are packaged as OCI
+artifacts using the standard OCI image manifest with custom media types. The
+top-level object is always an OCI Image Index, even for single-architecture
+exports, for forward compatibility.
+
+Two artifact types are defined:
+
+| Artifact Type                                        | Source Kind              |
+|------------------------------------------------------|--------------------------|
+| `application/vnd.kubevirt.virtualmachine.v1`         | `VirtualMachine`         |
+| `application/vnd.kubevirt.virtualmachinetemplate.v1` | `VirtualMachineTemplate` |
+
+The version suffix refers to the artifact format version, not the CRD API
+version. The actual CRD API version is recorded inside the config blob.
+`VirtualMachineSnapshot` exports use the `virtualmachine.v1` artifact type
+since a snapshot represents VM state.
+
+#### Image Index
+
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
+  "artifactType": "application/vnd.kubevirt.virtualmachine.v1",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "digest": "sha256:abc...",
+      "size": 1234,
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux"
+      }
+    }
+  ],
+  "annotations": {
+    "org.opencontainers.image.title": "my-vm",
+    "org.opencontainers.image.description": "Fedora 42 Server VM",
+    "org.opencontainers.image.created": "2026-04-09T12:00:00Z"
+  }
+}
+```
+
+The `platform.architecture` field maps to `spec.template.spec.architecture`,
+defaulting to `amd64` when unset. The `platform.os` field is always `linux`.
+
+#### Image Manifest
+
+Each manifest in the index describes one architecture variant:
+
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "artifactType": "application/vnd.kubevirt.virtualmachine.v1",
+  "config": {
+    "mediaType": "application/vnd.kubevirt.virtualmachine.config.v1+json",
+    "digest": "sha256:def...",
+    "size": 4096
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.kubevirt.disk.raw+zstd",
+      "digest": "sha256:111...",
+      "size": 2147483648,
+      "annotations": {
+        "io.kubevirt.disk.name": "rootdisk",
+        "io.kubevirt.pvc.capacity": "10Gi",
+        "io.kubevirt.pvc.volumeMode": "Block",
+        "io.kubevirt.pvc.accessModes": "ReadWriteOnce",
+        "org.opencontainers.image.title": "rootdisk.raw.zst"
+      }
+    },
+    {
+      "mediaType": "application/vnd.kubevirt.disk.raw+zstd",
+      "digest": "sha256:222...",
+      "size": 1073741824,
+      "annotations": {
+        "io.kubevirt.disk.name": "datadisk",
+        "io.kubevirt.pvc.capacity": "5Gi",
+        "io.kubevirt.pvc.volumeMode": "Block",
+        "io.kubevirt.pvc.accessModes": "ReadWriteOnce",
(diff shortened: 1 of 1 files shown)


veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+|------------------------------------------------------------------|------------------------------------------|
+| `application/vnd.kubevirt.virtualmachine.v1`                     | `artifactType`: `VirtualMachine`         |
+| `application/vnd.kubevirt.virtualmachine.config.v1+json`         | Config blob: VM definition               |
+| `application/vnd.kubevirt.virtualmachinetemplate.v1`             | `artifactType`: `VirtualMachineTemplate` |
+| `application/vnd.kubevirt.virtualmachinetemplate.config.v1+json` | Config blob: VM template definition      |
+| `application/vnd.kubevirt.disk.raw+gzip`                         | Layer: gzip-compressed raw disk          |
+| `application/vnd.kubevirt.disk.qcow2`                            | Layer: qcow2 disk image                  |
+
+Raw disks are always gzip-compressed during export. qcow2 images are stored
+as-is since qcow2 supports internal compression.
```
mhenriks: All disks populated by CDI will be raw format. I think the only place qcow is used is as a base layer for containerdisks and below it is stated that "ContainerDisk volumes keep their image reference as-is - the referenced image is not re-packaged into the artifact."

0xFelix: In this case we can drop qcow altogether. I don't see any other uses cases for it. Or would you like to see containerdisk export as well?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
+# VEP #256: OCI Artifact Export for VirtualMachines and VirtualMachineTemplates
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
+- This VEP targets beta for version: v1.9
```
awels: Are we skipping alpha? Or did I miss alphs

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+* Extend `VirtualMachineExport` to accept `VirtualMachineTemplate` as a
+  source kind, with the export controller resolving template volume sources
+
+## Non Goals
+
+* Import of OCI artifacts into the cluster (deferred to a follow-up design)
+* Support for multi-VM appliances (single VM or template per artifact)
+* Guest-level operations such as sysprep or sealing during export
+* Live export of running VMs
+* Cross-hypervisor portability (the OCI format is KubeVirt-native)
```
awels, line 81: Maybe point out that direct registry push is a non goal in this version?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+  "annotations": {
+    "org.opencontainers.image.title": "my-vm",
+    "org.opencontainers.image.description": "Fedora 42 Server VM",
+    "org.opencontainers.image.created": "2026-04-09T12:00:00Z"
+  }
+}
+```
+
+The `platform.architecture` field maps to `spec.template.spec.architecture`,
+defaulting to `amd64` when unset. The `platform.os` field is always `linux`.
```
awels, line 156: Maybe note that the platform.os does NOT represent the guest OS, but the OS that would process the image when importing (which is always some form of linux). How would this work with different hyper visors like hyper-v? Clusters with that hypervisor would likely be some form of windows, and in that case the OS would be different?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+    "size": 4096
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.kubevirt.disk.qcow2",
+      "digest": "sha256:111...",
+      "size": 1073741824,
+      "annotations": {
+        "io.kubevirt.disk.name": "rootdisk",
+        "org.opencontainers.image.title": "rootdisk.qcow2"
```
awels: If we are only referencing PVCs/Snapshots here, I don't think we have any qcow2 formats, everything would be RAW.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
awels: I agree with Michael here, I don't think we need to worry about qcow2 here since none of the sources will be qcow2.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+| Media Type                                                       | Usage                                    |
+|------------------------------------------------------------------|------------------------------------------|
+| `application/vnd.kubevirt.virtualmachine.v1`                     | `artifactType`: `VirtualMachine`         |
+| `application/vnd.kubevirt.virtualmachine.config.v1+json`         | Config blob: VM definition               |
+| `application/vnd.kubevirt.virtualmachinetemplate.v1`             | `artifactType`: `VirtualMachineTemplate` |
+| `application/vnd.kubevirt.virtualmachinetemplate.config.v1+json` | Config blob: VM template definition      |
+| `application/vnd.kubevirt.disk.raw+gzip`                         | Layer: gzip-compressed raw disk          |
+| `application/vnd.kubevirt.disk.qcow2`                            | Layer: qcow2 disk image                  |
+
+Raw disks are always gzip-compressed during export. qcow2 images are stored
```
awels: isn't gzip bad for streaming? Maybe something else like xz or zstd would work better? I guess it depends on what the library we are using supports.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,537 @@
...
+#### Streaming Strategy
+
+The OCI manifest must contain the SHA-256 digest of every blob before any
+blob is written to the TAR. This requires a two-pass approach:
+
+**Pass 1 - compute digests (runs on startup, export becomes ready after):**
+
+1. Generate the config blob (VM or template JSON), compute its digest/size
+2. Stat each disk file to determine its size
+3. Read each disk file sequentially, compressing raw disks through gzip,
```
awels: Random thought, if we move the hash of the disks to the end of the OCI-artifact, could we keep the hash calculation in memory without having to read and hash the entire disk twice? While we are streaming the data, we can calculate the hash on the fly. Then after we are done we can also send the calculated hashes. I really hate to have to read the entire disk twice.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: ACK, then I'll drop qcow2.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: Good question, would you put this behind another feature gate?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: You've got a point. Zstd should be better suited for streaming purposes, but since it is not available in the go stdlib, we need an external library (`github.com/klauspost/compress/zstd`) for it.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: Not necessarily true, you can have a HyperV host running Linux VMs. With the recently merged MSHV support, it will be possible to avoid the penalty of running virt in virt by passing through / leveraging the MSHV of the host directly. So IIUC `platform.OS` should always be `Linux`. (see also [this FOSDEM talk](https://fosdem.org/2026/schedule/event/BFQ8XA-introducing-mshv-accelerator-in-qemu))

@harshitgupta1337 Would appreciate if you could confirm my understanding?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: Great suggestion and it seems to be in conformance with the OCI spec. That would work for both the download as well as the push to registry case. However, there still seems to be one issue with the download case and that is the TAR header. It needs to come first and it needs to contain the sizes of all included files. We don't know the size of compressed disks upfront unfortunately. I don't think there's a way around the two pass?

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
vladikr: @0xFelix Yes, I would...
or we could delay and stay in alpha - then we can reuse the FG.

Aren't you worried that this is a non-trivial change and should have its own feature lifecycle?

mhenriks:
Looks pretty great!
/lgtm

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
harshitgupta1337: Hi @0xFelix, your understanding - about the K8s node running Linux while leveraging HyperV on the host - is correct.
This is the primary scenario for which the MSHV support was introduced.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,548 @@
...
+|------------------------------------------------------------------|------------------------------------------|
+| `application/vnd.kubevirt.virtualmachine.v1`                     | `artifactType`: `VirtualMachine`         |
+| `application/vnd.kubevirt.virtualmachine.config.v1+json`         | Config blob: VM definition               |
+| `application/vnd.kubevirt.virtualmachinetemplate.v1`             | `artifactType`: `VirtualMachineTemplate` |
+| `application/vnd.kubevirt.virtualmachinetemplate.config.v1+json` | Config blob: VM template definition      |
+| `application/vnd.kubevirt.disk.raw+zstd`                         | Layer: zstd-compressed raw disk          |
+
+Raw disks are always compressed with zstd during export. Zstd is provided
+by the external Go library
+[`github.com/klauspost/compress/zstd`](https://github.com/klauspost/compress).
```
dominikholler, line 243: https://github.com/klauspost/compress/discussions/496 and
```
dholler@dholler-thinkpadp1gen3:/tmp$ truncate -s 1T bigfile
dholler@dholler-thinkpadp1gen3:/tmp$ ls -lsh bigfile
45M -rw-r--r--. 1 dholler dholler 1.0T Apr 15 08:37 bigfile
dholler@dholler-thinkpadp1gen3:/tmp$ time zstd bigfile
bigfile              :  0.00%   ( 1.000 TiB =>   32.9 MiB, bigfile.zst)        

real	3m15.016s
user	10m42.872s
sys	1m35.398s
dholler@dholler-thinkpadp1gen3:/tmp$ ls -lsh bigfile.zst 
33M -rw-r--r--. 1 dholler dholler 33M Apr 15 08:37 bigfile.zst
dholler@dholler-thinkpadp1gen3:/tmp$ rm bigfile
dholler@dholler-thinkpadp1gen3:/tmp$ time unzstd bigfile.zst 
bigfile.zst          : 24.9 GiB...     zstd: error 70 : Write error : cannot write block : Disk quota exceeded 

real	0m6.359s
user	0m4.598s
sys	0m6.416s
dholler@dholler-thinkpadp1gen3:/tmp$ truncate -s 1T bigfile
dholler@dholler-thinkpadp1gen3:/tmp$ time unzstd bigfile.zst 
zstd: bigfile already exists; overwrite (y/n) ? y
bigfile.zst         : 1099511627776 bytes                                      

real	2m55.905s
user	4m10.446s
sys	0m13.208s
dholler@dholler-thinkpadp1gen3:/tmp$ ls -lsh bigfile
92M -rw-r--r--. 1 dholler dholler 1.0T Apr 15 08:37 bigfile

```

shows that the related sparse implementation seems to be possible, even it might not come out of the box.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: VMExportGate was graduated in 1.9, so we need another FG.

I see this as an extension of the existing VMExport API. Since the new fields should be purely additive, I don't think a new API version of VMExport is required. Also we cannot not hide them behind a FG.

But I can add a FG, so the OCI format would only be available when the FG was enabled.

0xFelix:
Added a feature gate (`OCIExport`) and updated the phases.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
0xFelix: @vladikr Added the `OCIExport` gate

vladikr: Thank you!

vladikr:
/approve
Thank you @0xFelix 
Looks great from my side.
/hold
Feel free to drop the hold once you're satisfied :)

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,573 @@
...
+  metadata is lost
+* While it would be theoretically possible to export each disk as a
+  separate ContainerDisk, this would require external orchestration to
+  track which images belong together and how to reassemble them into a
+  VM - multiple artifacts to reproduce what a single OCI artifact
+  achieves
+
+OCI artifacts store the complete resource definition alongside
+arbitrarily many disk layers in one self-contained artifact.
+
```
dominikholler, line 471: While it would be possible to extend the definition of a containerdisk in a way that it could contain the VM spec as a file (and even multiple disks) inside the image, I understood that that this would require pulling the whole file system blob. The justification to introduce a new format is, that it would be beneficial to pull only some small json blobs, avoid pulling the big file system blob. Ist this correct?

0xFelix: Yes, extending the containerdisk to embed the spec would inefficiently force us to pull a blob just to read a small configuration file. Extending the spec would essentially duplicate what OCI artifacts already provide out of the box.

veps/sig-compute/256-oci-export-proposal/oci-export-proposal.md
```diff
@@ -0,0 +1,588 @@
...
+### Export Controller: `VirtualMachineTemplate` Source Handler
+
+A new source handler implementing the `ExportSource` interface:
+
+1. Fetch the `VirtualMachineTemplate` and validate it has `Ready=True`
+2. Extract the embedded VM spec from `spec.virtualMachine`
+3. Resolve volume sources:
+    - DataVolumeTemplates with a direct PVC source -> resolve to PVCs,
+      export disk data as layers
+    - DataVolumeTemplates with `sourceRef` or other non-PVC sources -> kept
```
mhenriks, line 357: How are DataVolumeTemplates handled for VirtualMachine Source? Does the VM spec in the archive include dataVolumeTemplates at all? Assuming VM with populated DVTs but I guess it is possible to export a VM that was never started and may not have populated DVs

0xFelix: Good point. If an existing VM has already been started, we should strip the DVTs and replace the referring volumes with PVC volumes.

The edge case is VMs with DVTs that were never started. Should we keep the DVTs in that scenario? We could determine this by checking for the corresponding DataVolume: if it exists, we strip the DVT and if not, we keep it.

However, an alternative approach would be to simply reject the export if any DVs are missing. A VM with unstarted DVTs should probably just be considered a "template". If that's the case, `VirtualMachineTemplate` should be used as the export source instead.

awels: I believe the DVs will always get created (but could be in WaitingForConsumer phase), but if we have WaitForFirstConsumer storage and the VM is never started, then the corresponding PVCs are never populated (the PVCs will be created as well, but remain pending).

But I am not sure if the VMExport will even start if the PVCs are not bound, I would have to double check the code. If it does start, then I guess the VMExport pod will be the first consumer...

mhenriks: @awels has a good point, in WFFC case the PVCs owned by DVs _should_ get bound/populated when the export starts so I think it makes sense to strip the DVTs from the VM spec when including in the archive

I wouldn't reject anything in a webhook

0xFelix: Added a paragraph about stripping DVTs from VMs to the proposal

mhenriks:
/lgtm
Thanks @0xFelix!

0xFelix:
/hold cancel

Let's get it in :)

Issue #259, opened 2026-04-13
VEP 259: Raw I/O Support for LUN Disks


mhenriks:
**Primary contact (assignee)**:

/assign mhenriks

**Current Feature Stage**:

**Feature Gate**: `RawIO`

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (1.9):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/260
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
removing this from the 1.10 milestone as there is no straight froward solution.

PR #260, opened 2026-04-13
VEP #259: Raw I/O Support for LUN Disks

mhenriks:
### VEP Metadata

**Tracking issue**: 
**SIG label**: 

/sig storage
/sig compute

### What this PR does

API for rawio to LUNs

### DIFF
--- a/veps/sig-storage/259-rawio-lun-disks.md
+++ b/veps/sig-storage/259-rawio-lun-disks.md
@@ -1,0 +1,327 @@
+# VEP #259: Raw I/O Support for LUN Disks
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This proposal adds a `rawio` boolean field to the KubeVirt disk API for LUN-type disks. When enabled, the field sets `rawio='yes'` on the libvirt domain disk element and grants `CAP_SYS_RAWIO` to the virt-launcher pod. The `virt-controller` SecurityContextConstraint (SCC) is also updated to permit this capability.
+
+## Motivation
+
+Enterprise storage environments commonly use dedicated management VMs that communicate with external storage arrays over SCSI protocols. These VMs issue privileged SCSI commands — such as mode page configuration, diagnostic routines, defect management, and firmware updates via `WRITE BUFFER` — to manage the lifecycle of storage devices. This is a standard operational pattern in data centers where the storage management plane runs inside VMs rather than requiring physical access to the host or out-of-band management interfaces.
+
+KubeVirt currently supports exposing volumes as LUN devices (`disk.device: lun`), which allows VMs to issue SCSI commands via the SG_IO interface. However, the kernel's SCSI command filter divides commands into unrestricted and restricted groups. Restricted commands — including maintenance operations (`WRITE BUFFER`, `SEND DIAGNOSTIC`), vendor-specific commands, and service actions — are blocked unless the process has `CAP_SYS_RAWIO`. This means storage management VMs migrated to KubeVirt lose the ability to perform the privileged SCSI operations they were designed for.
+
+Today there is no way for a KubeVirt user to request raw I/O access for a LUN disk. Adding opt-in `rawio` support closes this gap while keeping the elevated privilege gated behind a feature gate and an explicit per-disk opt-in, so that the broader security posture of KubeVirt is not affected for workloads that do not need it.
+
+## Goals
+
+- Allow users to opt in to raw I/O for LUN-type disks via a new API field so that VMs can issue restricted SCSI commands to provisioned devices.
+- Ensure the feature does not grant capabilities beyond what the user's namespace PSA level and SCC bindings already permit.
+- Gate the feature behind a feature gate (`RawIO`) until GA.
+
+## Non Goals
+
+- Granting `CAP_SYS_RAWIO` globally to all virt-launcher pods. The capability is only added when explicitly requested.
+- Supporting `rawio` for non-LUN disk device types (e.g., `disk`, `cdrom`). Libvirt only supports `rawio` on `device='lun'`.
+- Modifying the `sgio` (SCSI Generic I/O) filtering attribute. The `sgio` setting is orthogonal and can be addressed separately.
+- Changes to PodSecurityPolicy (PSP). KubeVirt does not use PSP, and PSP has been removed from Kubernetes as of v1.25.
+- Implementing the validating webhook for non-OpenShift clusters. On upstream Kubernetes, there is no SCC layer and no per-user capability restrictions by default. PSA enforcement at the namespace level is sufficient.
+
+## Definition of Users
+
+- **Storage administrators** who operate storage array management VMs that require privileged SCSI access to manage external storage devices.
+- **Cluster administrators** who need to understand and control the security implications of granting raw I/O access.
+
+## User Stories
+
+- As a storage administrator, I want to run a storage array management VM on KubeVirt with LUN passthrough, so that the VM can issue privileged SCSI commands (mode page configuration, diagnostics, firmware updates) to manage the storage array — the same operations it performed on the previous bare-metal or traditional virtualization platform.
+- As a storage administrator, I want to update firmware on SCSI storage devices from within my VM using standard tools like `sg_write_buffer`, so that I can manage device lifecycle without physical or out-of-band access to the storage hardware.
+- As a storage administrator, I want to issue vendor-specific SCSI commands (e.g., `SEND DIAGNOSTIC`, `READ DEFECT DATA`, vendor-specific mode pages) to a passthrough LUN device from within my VM for ongoing storage health monitoring and management.
+- As a cluster administrator, I want raw I/O support to be gated behind a feature gate, so that I can control whether this capability is available on my cluster.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### API Changes
+
+A new optional boolean field `rawio` is added to the `LunTarget` struct (the per-disk configuration under `spec.domain.devices.disks[].lun`):
+
+```go
+type LunTarget struct {
+    // ...existing fields...
+
+    // Rawio indicates whether the disk should have raw I/O access enabled.
+    // When set to true, the libvirt domain will have rawio='yes' on the disk
+    // element, and the virt-launcher pod will be granted CAP_SYS_RAWIO.
+    // Requires the RawIO feature gate.
+    // +optional
+    Rawio *bool `json:"rawio,omitempty"`
+}
+```
+
+### Validation
+
+- A webhook validates that `rawio` is only set when `disk.device` is `"lun"`. Setting `rawio` on any other device type is rejected.
+- A webhook rejects `rawio: true` on create (and on update if not already set) when the `RawIO` feature gate is disabled. This prevents users from setting a field that would be silently ignored.
+- If a VM already has `rawio: true` set and the feature gate is subsequently disabled, the field is allowed to persist on updates (ratcheting) to avoid breaking existing manifests.
+
+### Libvirt Domain Generation
+
+When `rawio: true` is set on a LUN disk and the `RawIO` feature gate is enabled, the domain converter in virt-launcher sets `rawio='yes'` on the corresponding `<disk>` element in the libvirt domain XML. If the feature gate is disabled, the `rawio` field is ignored even if set on the VMI.
+
+```xml
+<disk type='block' device='lun' rawio='yes'>
+  ...
+</disk>
+```
+
+### virt-launcher Pod Security
+
+When any disk on a VMI has `rawio: true` and the `RawIO` feature gate is enabled, virt-controller adds `CAP_SYS_RAWIO` to the virt-launcher container's security context. If the feature gate is disabled, the capability is not added regardless of the field value.
+
+```yaml
+securityContext:
+  capabilities:
+    add:
+      - SYS_RAWIO
+```
+
+This is only added when needed — virt-launcher pods for VMIs without `rawio` disks are unaffected.
+
+### SecurityContextConstraint (SCC) Changes
+
+The `kubevirt-controller` SCC must be updated to include `SYS_RAWIO` in `allowedCapabilities` (not `defaultCapabilities`). This is a one-time deployment change that permits virt-controller to request this capability for virt-launcher pods when needed. Pods that do not set `rawio: true` on any disk are unaffected.
+
+### Validating Webhook for SCC Capability Enforcement (OpenShift)
+
+Adding `SYS_RAWIO` to the `kubevirt-controller` SCC means virt-controller can create virt-launcher pods with that capability. Since virt-controller creates pods directly (rather than through a Deployment or ReplicaSet), OpenShift's SCC admission evaluates virt-controller's service account, not the pod's. Without additional enforcement, any user who can create a VM could get `CAP_SYS_RAWIO` on their virt-launcher pod regardless of whether their service account permits it.
+
+To mitigate this, a **validating admission webhook** is added that fires on pod creation for virt-launcher pods. The webhook uses the OpenShift `PodSecurityPolicySubjectReview` API (`security.openshift.io/v1`) to check whether the **pod's service account** — not virt-controller's service account — is permitted by its SCC bindings to use the requested capabilities. If the pod's service account does not have an SCC allowing `SYS_RAWIO`, the pod creation is denied.
+
+```yaml
+apiVersion: security.openshift.io/v1
+kind: PodSecurityPolicySubjectReview
+spec:
+  template:
+    spec:
+      containers:
+      - name: compute
+        securityContext:
+          capabilities:
+            add: ["SYS_RAWIO"]
+  user: "system:serviceaccount:<namespace>:<pod-sa>"
+  groups:
+  - "system:serviceaccounts"
+  - "system:serviceaccounts:<namespace>"
+status:
+  allowedBy:
+    name: <scc-name>  # nil if no SCC permits the capability
+```
+
+This approach has two key properties:
+
+1. **Analogous to Deployment behavior.** When a ReplicaSet controller creates a pod in OpenShift, the SCC admission controller automatically checks the pod's service account, not the controller's. Since virt-controller creates pods directly, this automatic check does not apply — the webhook replicates it.
+
+2. **Robust against mutation.** Because this is a validating (not mutating) webhook, it runs after all mutating webhooks in the admission chain. It evaluates the final pod spec, so any capabilities injected by a mutating webhook are also caught.
+
+The webhook is **OpenShift-specific**. On upstream Kubernetes there is no SCC layer, and any user with pod creation RBAC can already request any capability. If a cluster uses Pod Security Admission (PSA), enforcement is at the namespace level — `SYS_RAWIO` is blocked for all pods in a `baseline` or `restricted` namespace regardless of who creates them.
+
+### Pod Security Standards (PSS) Impact
+
+`CAP_SYS_RAWIO` is not in the set of capabilities permitted by the `baseline` or `restricted` Pod Security Standards. When `rawio: true` is requested, the virt-launcher pod will require the `privileged` security level. This should be noted in documentation so that cluster administrators using Pod Security Admission (PSA) enforcement can configure their namespaces accordingly.
+
+### Feature Gate
+
+The feature is gated behind `RawIO`. The gate is checked in two places:
+
+1. **Webhook (API layer):** Rejects new `rawio: true` fields when the gate is disabled, preventing users from setting a field that would have no effect. Existing values are allowed to persist through updates (ratcheting).
+2. **Controller and domain converter:** Ignore `rawio: true` when the gate is disabled, ensuring that disabling the gate effectively revokes the capability for all VMs on their next pod creation, even if the field is still present in the spec.
+
+## API Examples
+
+### Enabling raw I/O on a LUN disk
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: storage-mgmt-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          disks:
+            - name: scsi-lun
+              lun:
+                bus: scsi
+                rawio: true
+      volumes:
+        - name: scsi-lun
+          persistentVolumeClaim:
+            claimName: my-scsi-device
+```
+
+### Without raw I/O (default behavior, unchanged)
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: standard-lun-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          disks:
+            - name: scsi-lun
+              lun:
+                bus: scsi
+      volumes:
+        - name: scsi-lun
+          persistentVolumeClaim:
+            claimName: my-scsi-device
+```
+
+## Security Considerations
+
+`CAP_SYS_RAWIO` is a broad Linux capability. Beyond enabling raw SCSI commands, it gates access to several privileged operations. This section analyzes each attack surface and explains why, in the context of a virt-launcher container, the capability is effectively scoped to SCSI command filtering on explicitly provisioned devices.
+
+### Attack surface analysis
+
+**`/dev/mem` (physical memory read/write):** In an unrestricted environment, this device allows reading and writing arbitrary physical memory. However, in a virt-launcher pod this is blocked by multiple independent layers: (1) the container's device cgroup uses a deny-all policy and does not whitelist `/dev/mem` — the kernel blocks `open()` at the cgroup level regardless of capabilities; (2) the device node does not exist in the container's tmpfs `/dev`; (3) `CONFIG_STRICT_DEVMEM` (enabled by default since Linux 4.16 and on RHEL/Fedora) restricts access to PCI/BIOS regions only; (4) `CONFIG_IO_STRICT_DEVMEM` further restricts to unclaimed I/O regions; (5) SELinux's `container_t` context blocks access on OpenShift.
+
+**`/dev/kmem` (kernel virtual memory):** Removed from the upstream kernel entirely in version 5.13 (2021). Does not exist on any supported platform.
+
+**`/proc/kcore` (kernel memory dump, read-only):** Masked by the OCI runtime — bind-mounted to `/dev/null` in all standard container runtimes (CRI-O, containerd). Reading it returns nothing.
+
+**`iopl()` / `ioperm()` (I/O port access):** These syscalls grant direct hardware port access, which could in theory be used to reprogram DMA engines via PCI configuration ports (0xCF8/0xCFC) and achieve arbitrary physical memory access. However, both syscalls are **blocked by the default seccomp profile** in CRI-O and containerd. The process receives EPERM before the capability check runs. KubeVirt also ships a custom seccomp profile for virt-launcher (since v0.59.0).
+
+**`/dev/cpu/*/msr` (model-specific registers):** Not accessible — blocked by the device cgroup, and the device nodes do not exist in the container.
+
+**`/proc/sys/vm/mmap_min_addr` (low-address memory mapping):** `/proc/sys` is mounted read-only in containers, preventing modification.
+
+**`FIBMAP` ioctl (file block mapping):** Read-only information disclosure about disk block layout. Low risk — the equivalent `FIEMAP` ioctl provides the same information without any capability requirement.
+
+**`/proc/bus/pci` (PCI configuration space):** Masked by the OCI runtime and blocked by the device cgroup. PCI device nodes are not present in the container.
+
+**`hpsa(4)` / `cciss(4)` devices (host RAID controllers):** Their device nodes are not present in the container (blocked by device cgroup deny-all policy), so the `CAP_SYS_RAWIO`-gated ioctls are unreachable.
+
+**Other device-specific operations:** `CAP_SYS_RAWIO` gates ioctls on several other device types. The device cgroup deny-all policy ensures those devices are not accessible in the container.
+
+**Raw SCSI commands (SG_IO):** This is the intended use case. The device cgroup restricts which block devices are visible to the container, so raw SCSI access is scoped to the specific LUN devices explicitly provisioned for the VM.
+
+### Effective scope in virt-launcher
+
+In summary, `CAP_SYS_RAWIO` in a virt-launcher pod is inert for all attack surfaces except SCSI command filtering, because the actual exploitation vectors require device access or syscall availability that containers block through independent mechanisms (device cgroup, seccomp, masked paths, strict devmem, read-only procfs). The capability is necessary but not sufficient for any of the dangerous operations it theoretically gates. Its practical effect is limited to allowing restricted SCSI commands on explicitly provisioned LUN devices.
+
+### Administrative controls
+
+1. **Opt-in only:** The capability is never added unless the user explicitly sets `rawio: true` on a LUN disk.
+2. **Feature gate:** The entire feature is gated, giving cluster administrators control over availability.
+3. **Validating webhook (OpenShift):** A validating webhook checks that the virt-launcher pod's service account has an SCC permitting the requested capabilities, preventing privilege escalation through virt-controller or mutating webhooks. See [Validating Webhook for SCC Capability Enforcement](#validating-webhook-for-scc-capability-enforcement-openshift).
+4. **Documentation:** Users are advised that enabling `rawio` elevates the privilege level of the virt-launcher pod and takes it out of `restricted` PSS compliance.
+
+## Alternatives
+
+### Use `unpriv_sgio` instead of `CAP_SYS_RAWIO`
+
+The kernel provides a per-device sysfs knob (`/sys/dev/block/<maj:min>/queue/unpriv_sgio`) that allows unprivileged SCSI Generic I/O without requiring `CAP_SYS_RAWIO`. This would avoid granting a broad capability.
+
+However, this approach has significant challenges in KubeVirt:
+- sysfs is mounted read-only inside containers, even privileged ones (see [kubevirt/kubevirt#6507](https://github.com/kubevirt/kubevirt/issues/6507)).
+- virt-handler could set this value on the host, but it adds complexity: resolving device major:minor, setting the knob before VM start, clearing it on stop and migration, and coordinating between virt-controller and virt-handler.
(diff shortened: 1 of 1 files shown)


vladikr:
Based on the discussion [here](https://github.com/kubevirt/enhancements/pull/191#issuecomment-4247479361), I agree with this approach now. 
I still think we need to explore the dynamic loading of the capability only when it's actually necessary.
The only concern is the change in the SCC. 
I wonder what @jean-edouard and @xpivarc think about this.

mhenriks:
> Based on the discussion [here](https://github.com/kubevirt/enhancements/pull/191#issuecomment-4247479361), I agree with this approach now. I still think we need to explore the dynamic loading of the capability only when it's actually necessary. The only concern is the change in the SCC. I wonder what @jean-edouard and @xpivarc think about this.

@vladikr you were right to push back when I said this VEP was simple and straightforward! Updating the SCC and granting cap rawio can lead to privilege escalation in certain cases, like Openshift out of the box with PSA not enforcing

I'm proposing we add a validating webhook on pod creation that fires for pods created by virt-controller. The webhook would use the OpenShift `PodSecurityPolicySubjectReview` API to check whether the pod's service account is permitted by its SCC bindings to use the capabilities requested in the pod's security context. If the SA doesn't have an SCC allowing SYS_RAWIO, the pod creation is denied

Suggesting webhook (rather than virt-controller check) because it will run after all mutating webhooks in the admission chain. This means it catches the final pod spec and if any mutating webhook were to inject `CAP_SYS_RAWIO` or other elevated capabilities into a virt-launcher pod, the validating webhook would still evaluate and deny it. This protects against privilege escalation from any source, not just virt-controller itself

This is analogous to how Deployments are validated in Openshift. When the ReplicaSet controller creates a pod, the SCC admission controller checks the pod's SA, not the controller's SA. Since virt-controller creates pods directly, we don't get that behavior automatically, so the webhook replicates it

This webhook would be OpenShift-specific. On upstream Kubernetes, there is no privilege escalation concern by default and any user with pod creation RBAC can already request any capability. If a cluster uses PSA, it enforces at the namespace level (not per-user), so `CAP_SYS_RAWIO` would be blocked for all pods in a baseline or restricted namespace regardless of who creates them

With this approach, the kubevirt-controller SCC change is safe: the SCC permits the capability at the virt-controller level, but the validating webhook ensures it's only actually used when the workload's SA is also authorized regardless of what mutations happened earlier in the admission chain

cc @fabiand

akalenyu (commented):
Pretty logical straightforward stuff. I think the main thing (and this has come up before with @alromeros work on velero) is in which kubevirt adjacenet project can we deploy openshift specific webhooks.

This is a much needed escape hatch that we are missing today which limits our flexibility (even outside this VEP, with shipping of backports to older releases).

akalenyu:
> Pretty logical straightforward stuff. I think the main thing (and this has come up before with @alromeros work on velero) is in which kubevirt adjacenet project can we deploy openshift specific webhooks.
> 
> This is a much needed escape hatch that we are missing today which limits our flexibility (even outside this VEP, with shipping of backports to older releases).

/cc @nunnatsa 
Maybe you can help with that angle of this VEP :pray:

nunnatsa:
> > Pretty logical straightforward stuff. I think the main thing (and this has come up before with @alromeros work on velero) is in which kubevirt adjacenet project can we deploy openshift specific webhooks.
> > This is a much needed escape hatch that we are missing today which limits our flexibility (even outside this VEP, with shipping of backports to older releases).
> 
> /cc @nunnatsa Maybe you can help with that angle of this VEP 🙏

Is it a new webhook, or a new check in an existing webhook? 

KubeVirt already check if it's running on an openshift cluster, so this validation can be wrapped by this condition, as done in other places.

mhenriks:
> > > Pretty logical straightforward stuff. I think the main thing (and this has come up before with @alromeros work on velero) is in which kubevirt adjacenet project can we deploy openshift specific webhooks.
> > > This is a much needed escape hatch that we are missing today which limits our flexibility (even outside this VEP, with shipping of backports to older releases).
> > 
> > 
> > /cc @nunnatsa Maybe you can help with that angle of this VEP 🙏
> 
> Is it a new webhook, or a new check in an existing webhook?
> 
> KubeVirt already check if it's running on an openshift cluster, so this validation can be wrapped by this condition, as done in other places.

Thanks to @akalenyu for bringing this up! My preference would be to keep the webhook implementation in kubevirt/kubevirt unless there is a very good reason not to. I know we already do some OpenShift specific configuration but there has also been a recent push to move non essential components out of KubeVirt core so I am definitely open to discussion here

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,308 @@
...
+- Update the `virt-controller` SCC to permit `CAP_SYS_RAWIO` for virt-launcher pods that require it.
+- Add an OpenShift-specific validating webhook to ensure virt-launcher pods are only granted `CAP_SYS_RAWIO` when the pod's service account has an SCC permitting it.
+- Gate the feature behind a feature gate (`RawIO`) until GA.
+
+## Non Goals
+
+- Granting `CAP_SYS_RAWIO` globally to all virt-launcher pods. The capability is only added when explicitly requested.
+- Supporting `rawio` for non-LUN disk device types (e.g., `disk`, `cdrom`). Libvirt only supports `rawio` on `device='lun'`.
+- Modifying the `sgio` (SCSI Generic I/O) filtering attribute. The `sgio` setting is orthogonal and can be addressed separately.
+- Changes to PodSecurityPolicy (PSP). KubeVirt does not use PSP, and PSP has been removed from Kubernetes as of v1.25.
```
Acedus: Seems like a bit of a redundant goal.

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,308 @@
...
+            claimName: my-scsi-device
+```
+
+## Security Considerations
+
+`CAP_SYS_RAWIO` is a broad Linux capability. Beyond enabling raw SCSI commands, it also grants:
+
+- Access to `/dev/mem`, `/dev/kmem`, and `/proc/kcore` (direct physical memory access)
+- `iopl()` and `ioperm()` system calls (I/O port access)
+- `FIBMAP` ioctl for disk block mapping
```
xpivarc: /hold
This to me is as no go, it uncomfortably reminds me of hostdisk in a way. Is there any alternative to the proposed solution also how likely is this use case?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,308 @@
...
+This proposal adds a `rawio` boolean field to the KubeVirt disk API for LUN-type disks. When enabled, the field sets `rawio='yes'` on the libvirt domain disk element and grants `CAP_SYS_RAWIO` to the virt-launcher pod. The `virt-controller` SecurityContextConstraint (SCC) is also updated to permit this capability.
+
+## Motivation
+
+KubeVirt currently supports exposing volumes as LUN devices (`disk.device: lun`), which allows VMs to issue SCSI commands via the SG_IO interface. However, certain privileged SCSI operations — such as the `WRITE BUFFER` command used for device firmware updates — are blocked by the kernel's SCSI command filter unless the process has `CAP_SYS_RAWIO`.
+
+Today there is no way for a KubeVirt user to request raw I/O access for a LUN disk. Users who need to perform firmware updates or other privileged SCSI operations from within a VM have no supported path to do so.
+
+Adding opt-in `rawio` support closes this gap while keeping the elevated privilege gated behind a feature gate and an explicit per-disk opt-in, so that the broader security posture of KubeVirt is not affected for workloads that do not need it.
+
```
xpivarc, line 31: Can you also include some use case for this?

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: The hostdisk concern is fair but I think the situations are different. hostdisk granted broad host filesystem access from inside the container. CAP_SYS_RAWIO looks broad on paper but in a virt-launcher pod all the dangerous surfaces — /dev/mem, iopl()/ioperm(), /proc/kcore, etc are independently blocked by the device cgroup, seccomp profile, and OCI masked paths. The capability's only practical effect is relaxing the kernel's SCSI command filter on the specific LUN devices provisioned for the VM. I've added a detailed attack surface analysis to the Security Considerations section that walks through each vector

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Sure, the primary driver is storage array management VMs which are dedicated VMs that manage external storage arrays over SCSI protocols (mode page configuration, diagnostics, defect management, firmware updates). Basically, there is a special LUN that is mounted by the management VM that exports an API over SCSI. This is a standard pattern in enterprise data centers. When these VMs migrate to KubeVirt, they lose the ability to issue restricted SCSI commands because the kernel's command filter blocks them without CAP_SYS_RAWIO

veps/sig-storage/259-rawio-lun-disks.md
akalenyu: Not sure this is as bad as HostDisk. On serious clusters with PSA, we should be protected
> If a cluster uses Pod Security Admission (PSA), enforcement is at the namespace level — SYS_RAWIO is blocked for all pods in a baseline or restricted namespace regardless of who creates them.

akalenyu:
> > > > Pretty logical straightforward stuff. I think the main thing (and this has come up before with @alromeros work on velero) is in which kubevirt adjacenet project can we deploy openshift specific webhooks.
> > > > This is a much needed escape hatch that we are missing today which limits our flexibility (even outside this VEP, with shipping of backports to older releases).
> > > 
> > > 
> > > /cc @nunnatsa Maybe you can help with that angle of this VEP 🙏
> > 
> > 
> > Is it a new webhook, or a new check in an existing webhook?
> > KubeVirt already check if it's running on an openshift cluster, so this validation can be wrapped by this condition, as done in other places.
> 
> Thanks to @akalenyu for bringing this up! My preference would be to keep the webhook implementation in kubevirt/kubevirt unless there is a very good reason not to. I know we already do some OpenShift specific configuration but there has also been a recent push to move non essential components out of KubeVirt core so I am definitely open to discussion here

I must have misunderstood you in the sig-storage meeting, I thought you were suggesting that we don't want it in kubevirt. I wasn't too concerned

veps/sig-storage/259-rawio-lun-disks.md
akalenyu: Fair question, it was my first thought about this too. To be completely blunt, do folks want this badly enough for us to go through with the development. My understanding is that this is a strong yes.

mhenriks: Yeah this is a must have requirement for users with high end enterprise storage

veps/sig-storage/259-rawio-lun-disks.md
akalenyu: To further address this, I think we could just introduce a toggle in addition to the feature gate. This is arguably a nicer way to go about security sensitive features in general. @aglitke raised this in one of the forums earlier.

xpivarc: @mhenriks Please write it down, so bystander will also know why you think this is safe

mhenriks: Have you looked at `### Attack surface analysis` yet?

mhenriks: > To further address this, I think we could just introduce a toggle in addition to the feature gate. This is arguably a nicer way to go about security sensitive features in general. @aglitke raised this in one of the forums earlier.

I personally think featuregate + PSA + SCC validating webhook is sufficient. Don't see a toggle giving any additional protection except with PSA in privileged namespace. And in that case user could just create a pod with the RAWIO cap

iholder101:
Was it considered to implement this as a [plugin](https://github.com/kubevirt/enhancements/issues/190)?

A webhook could be used to inject the added capability while a DomainHook could inject the XML pieces.
WDYT?

mhenriks:
> Was it considered to implement this as a [plugin](https://github.com/kubevirt/enhancements/issues/190)?
> 
> 
> 
> A webhook could be used to inject the added capability while a DomainHook could inject the XML pieces.
> 
> WDYT?

@iholder101 yeah see discussion here https://github.com/kubevirt/enhancements/pull/191#issuecomment-4247479361

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This proposal adds a `rawio` boolean field to the KubeVirt disk API for LUN-type disks. When enabled, the field sets `rawio='yes'` on the libvirt domain disk element and grants `CAP_SYS_RAWIO` to the virt-launcher pod. The `virt-controller` SecurityContextConstraint (SCC) is also updated to permit this capability.
```
xpivarc, line 22: The SCC should not be simply appended, we should use dedicated one if anything...

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Motivation
+
+Enterprise storage environments commonly use dedicated management VMs that communicate with external storage arrays over SCSI protocols. These VMs issue privileged SCSI commands — such as mode page configuration, diagnostic routines, defect management, and firmware updates via `WRITE BUFFER` — to manage the lifecycle of storage devices. This is a standard operational pattern in data centers where the storage management plane runs inside VMs rather than requiring physical access to the host or out-of-band management interfaces.
```
xpivarc, line 26: Why this software cannot run inside Pod instead of VM?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+KubeVirt currently supports exposing volumes as LUN devices (`disk.device: lun`), which allows VMs to issue SCSI commands via the SG_IO interface. However, the kernel's SCSI command filter divides commands into unrestricted and restricted groups. Restricted commands — including maintenance operations (`WRITE BUFFER`, `SEND DIAGNOSTIC`), vendor-specific commands, and service actions — are blocked unless the process has `CAP_SYS_RAWIO`. This means storage management VMs migrated to KubeVirt lose the ability to perform the privileged SCSI operations they were designed for.
+
+Today there is no way for a KubeVirt user to request raw I/O access for a LUN disk. Adding opt-in `rawio` support closes this gap while keeping the elevated privilege gated behind a feature gate and an explicit per-disk opt-in, so that the broader security posture of KubeVirt is not affected for workloads that do not need it.
+
+## Goals
+
+- Allow users to request raw I/O for LUN-type disks via a new API field.
+- Set `rawio='yes'` on the corresponding libvirt `<disk>` element when the field is enabled.
+- Grant `CAP_SYS_RAWIO` to the virt-launcher container when any disk on the VMI has `rawio: true`.
```
xpivarc: This can be simply expressed within one sentence, too much detail

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+- Update the `virt-controller` SCC to permit `CAP_SYS_RAWIO` for virt-launcher pods that require it.
+- Add an OpenShift-specific validating webhook to ensure virt-launcher pods are only granted `CAP_SYS_RAWIO` when the pod's service account has an SCC permitting it.
```
xpivarc: This is another no-go in my opinion, we need to stay vendor agnostic as much as we can. Wouldn't the SCC guarantee this anyway?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+- Gate the feature behind a feature gate (`RawIO`) until GA.
```
xpivarc, line 36: What about after GA?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+When `rawio: true` is set on a LUN disk and the `RawIO` feature gate is enabled, the domain converter in virt-launcher sets `rawio='yes'` on the corresponding `<disk>` element in the libvirt domain XML. If the feature gate is disabled, the `rawio` field is ignored even if set on the VMI.
+
+```xml
+<disk type='block' device='lun' rawio='yes'>
+  ...
+</disk>
+```
+
+### virt-launcher Pod Security
```
xpivarc, line 97: We worked so much on security of the virt-launcher Pod and this simply goes against every possible effort we had/have... 
Even if the capability gives the launcher one concrete usage withing the Pod, a successful escape would allow additional exploit IMHO

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+**`/dev/cpu/*/msr` (model-specific registers):** Not accessible — blocked by the device cgroup, and the device nodes do not exist in the container.
+
+**`/proc/sys/vm/mmap_min_addr` (low-address memory mapping):** `/proc/sys` is mounted read-only in containers, preventing modification.
+
+**`FIBMAP` ioctl (file block mapping):** Read-only information disclosure about disk block layout. Low risk — the equivalent `FIEMAP` ioctl provides the same information without any capability requirement.
+
+**Raw SCSI commands (SG_IO):** This is the intended use case. The device cgroup restricts which block devices are visible to the container, so raw SCSI access is scoped to the specific LUN devices explicitly provisioned for the VM.
+
```
xpivarc, line 233: What about:
`create memory mappings at addresses below the value specified by /proc/sys/vm/mmap_min_addr;`
and
`map files in /proc/bus/pci;`
`perform certain operations on hpsa(4) and cciss(4) devices`
`perform a range of device-specific operations on other devices`

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+4. **Documentation:** Users are advised that enabling `rawio` elevates the privilege level of the virt-launcher pod and takes it out of `restricted` PSS compliance.
+
+## Alternatives
+
+### Use `unpriv_sgio` instead of `CAP_SYS_RAWIO`
+
+The kernel provides a per-device sysfs knob (`/sys/dev/block/<maj:min>/queue/unpriv_sgio`) that allows unprivileged SCSI Generic I/O without requiring `CAP_SYS_RAWIO`. This would avoid granting a broad capability.
+
+However, this approach has significant challenges in KubeVirt:
+- sysfs is mounted read-only inside containers, even privileged ones (see [kubevirt/kubevirt#6507](https://github.com/kubevirt/kubevirt/issues/6507)).
```
xpivarc, line 252: Can't the handler writes it? @alicefr would you help here?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+- virt-handler could potentially set this value on the host, but it would need to know which devices require it and manage the lifecycle of the setting.
+- The `unpriv_sgio` knob does not cover all operations that `CAP_SYS_RAWIO` enables.
```
xpivarc: Which?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+This could be explored as a future optimization to reduce the privilege surface for specific use cases.
```
xpivarc: This needs to be explored as we are not compromising on security.

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+
+2. **Robust against mutation.** Because this is a validating (not mutating) webhook, it runs after all mutating webhooks in the admission chain. It evaluates the final pod spec, so any capabilities injected by a mutating webhook are also caught.
+
+The webhook is **OpenShift-specific**. On upstream Kubernetes there is no SCC layer, and any user with pod creation RBAC can already request any capability. If a cluster uses Pod Security Admission (PSA), enforcement is at the namespace level — `SYS_RAWIO` is blocked for all pods in a `baseline` or `restricted` namespace regardless of who creates them.
+
+### Pod Security Standards (PSS) Impact
+
+`CAP_SYS_RAWIO` is not in the set of capabilities permitted by the `restricted` Pod Security Standard. When `rawio: true` is requested, the virt-launcher pod will require at least the `baseline` security level. This should be noted in documentation so that cluster administrators using Pod Security Admission (PSA) enforcement can configure their namespaces accordingly.
```
jean-edouard: In the previous paragraph, you say `SYS_RAWIO is blocked for all pods in a baseline or restricted namespace`, and here `the virt-launcher pod will require at least the baseline security level`. Isn't that a contradiction?

veps/sig-storage/259-rawio-lun-disks.md
```diff
@@ -0,0 +1,326 @@
...
+    // Requires the RawIO feature gate.
+    // +optional
+    Rawio *bool `json:"rawio,omitempty"`
+}
+```
+
+### Validation
+
+- A webhook validates that `rawio` is only set when `disk.device` is `"lun"`. Setting `rawio` on any other device type is rejected.
+- A webhook rejects `rawio: true` on create (and on update if not already set) when the `RawIO` feature gate is disabled. This prevents users from setting a field that would be silently ignored.
```
jean-edouard, line 84: The feature gate is meant to be temporary and removed upon GA. At that point, regular users have access to a simple boolean that weakens virt-launcher security without a way for admins to block it.
How about a CR setting with an explicit scary name (AllowRawIOSysCapOnVirtLaunchers or something), that would be false by default and forever?

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: A dedicated SCC would need to be a full copy of the existing kubevirt-controller SCC with just SYS_RAWIO added. What's the benefit? Same service account creates the pods either way

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: KubeVirt's purpose is to run VM workloads. If a user wants to run SCSI management software in a VM (I believe the software is shipped as a vm image), they shouldn't need to justify that over a pod. This feature does not grant any privilege the user doesn't already have

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Can simplify this, sure

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: > This is another no-go in my opinion, we need to stay vendor agnostic as much as we can

As I've stated here https://github.com/kubevirt/enhancements/pull/260#issuecomment-4281874563, I am open to this living elsewhere, but we already enable openshift specific features in KubeVirt so there is precedent, for example creating SCCs, Routes, OLM integration, and recently added Template conversion. Given the relative simplicity of this webhook, I don't think it is unreasonable to add it as well to kubevirt/kubevirt but if there are strong objections it can live elsewhere

> Wouldn't the SCC guarantee this anyway?

No. When a Pod is created, the SCC of the user/service account that *created* the Pod is evaluated. And the kubevirt-controller SA will have the appropriate permission.

We should handle VMIs the same way that OpenShift handles Deployments/Replicasets. We should evaluate the the SCC of the service account *associated* to the Pod not the creator. There is a special admission controller for this. We will essentially do the same thing in a webhook for virt-launcher pods

In practice, on an out of the box OpenShift cluster this means that in order to run a VM with rawio, users will typically have to create a dedicated SA for the VMI and add that SA to an SCC that they likely created just for this purpose. This is a lot of hoops to jump through but keeps things secure.

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Feature gate is removed, rawio is a standard api and the capability is still only added when `rawio: true` is explicitly set, PSA still enforces at the namespace level, and the validating webhook still checks the pod's service account SCC bindings.

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: When the feature is GA, all the security checks will still be in place. The service account associated with the VM has to be allowed rawio via SCC on openshift or the VM must be running in a namespace with `privileged` PSA enforcement on vanilla kubernetes. In either case, the user could also create a Deployment with rawio. If the PSA is `restricted` the user can't do either

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: The goal of that work is "secure by default" not "prevent all capability usage." Kubernetes allows pods to request additional capabilities for a reason. Some workloads need them.

This feature is secure by default. There is enforcement by PSA at the namespace level and SCC is validated by the webhook. The attack surface analysis section shows, CAP_SYS_RAWIO in virt-launcher is inert for everything except scsi.

Nevertheless, we are not adding any new attack vectors. A user that can create a VM with rawio could also create a Deployment with rawio and exploit the same hypothetical escape

A user that cannot create a Deployment with rawio cannot create a VM with rawio

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Good catch, meant `privileged`

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Good catch. mmap_min_addr is covered (procfs is read-only so the value cannot be lowered), but I'll add the missing items:

- /proc/bus/pci (PCI config space): Masked by the OCI runtime and blocked by the device cgroup. PCI device nodes are not present in the container.
- hpsa(4) / cciss(4) devices: These are host RAID controller drivers. Their device nodes are not present in the container (blocked by device cgroup deny-all policy), so the ioctls are unreachable.
- Other device-specific operations: Same principle. CAP_SYS_RAWIO gates ioctls on specific device types, but the device cgroup ensures those devices are not accessible in the container.

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: virt-handler could probably do this, but it adds a lot of complexity. And crucially, having virt-handler toggle host sysfs knobs to grant unprivileged processes access to restricted SCSI commands is fundamentally privilege escalation. We should use the standard Kubernetes capability model instead

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Updated the VEP to drop this point. There are operations like `SG_SCSI_RESET` that are not supported but the stronger argument against unpriv_sgio is the complexity and that it is fundamentally privilege escalation outside the standard Kubernetes capability model

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: I don't think we are compromising on security. The unpriv_sgio via virt-handler alternative actually compromises security because it is privilege escalation outside the standard Kubernetes capability model. 

Can you point to a specific security concern that the analysis doesn't address?

veps/sig-storage/259-rawio-lun-disks.md
iholder101: > but we already enable openshift specific features in KubeVirt so there is precedent, for example creating SCCs, Routes, OLM integration, and recently added Template conversion. Given the relative simplicity of this webhook, I don't think it is unreasonable to add it as well to kubevirt/kubevirt but if there are strong objections it can live elsewhere

FWIW I find this reasoning a bit problematic. There's a precedent for a lot of bad stuff.
We should aim to reduce bad patterns, and not to amplify them just because there's a precedent.

veps/sig-storage/259-rawio-lun-disks.md
akalenyu: I guess the CSI drivers in question could handle this too. Obviously going to be hard to get them all on the same page but it might come up from their regular pod users at some point too
(storage class param for unpriv_sgio)

I wonder if csi-addons could help us with this alternative if we end up keeping this in our realm

akalenyu: Here's how it looks like in practice https://github.com/ceph/ceph-csi/blob/a5474f81497297aa9cd341e12529923b1349084d/internal/csi-addons/rbd/reclaimspace.go#L120-L123

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Obviously the precedent argument is weak but it is unclear to me (and I think others in this review) how bad it is compared to the alternatives? Given that this feature requires a SCC check in order to be secure, what do you propose?

veps/sig-storage/259-rawio-lun-disks.md
mhenriks: Sure, a CSI driver could theoretically expose rawio capable disks and do privilege escalation on our behalf. Still have to configure the libvirt domain though

Not sure csi-addons is a great fit. It currently handles storage infrastructure operations (reclaim space, replication, fencing) via custom resources, not user facing disk semantics

Could it be possible to have a privileged component that raw-ifies an arbitrary pvc? Probably. But I think synchronization would be tough. You'd likely want this component to do its work before the disk is used by the pod

akalenyu: > Could it be possible to have a privileged component that raw-ifies an arbitrary pvc? Probably.

Thinking through this, the unpriv_sgio alternative and avoiding the capability may be augmented to work for this one case but I'm afraid it won't scale. When the next request comes (CAP_*) we would still need to find a way to cleanly allow a certain capability for eligible entities
(We can't predict what other enterprise VM workloads will need)

So in that sense we may be building reusable security infrastructure

akalenyu: https://github.com/kubevirt/kubevirt/pull/17675

Acedus: I don't think unpriv_sgio makes sense because it simply never landed in upstream Linux so it wouldn't work on non-RHEL based nodes.

cc @bonzini

veps/sig-storage/259-rawio-lun-disks.md
Acedus: > The goal of that work is "secure by default" not "prevent all capability usage." Kubernetes allows pods to request additional capabilities for a reason. Some workloads need them.

+1

> Even if the capability gives the launcher one concrete usage withing the Pod, a successful escape would allow additional exploit IMHO

Is the attack surface analysis insufficient? Seems to me like we already have sufficient extra limitations in place (default seccomp already prevents most of the concerns that at least I had AFAIK). 

We also already conditionally add `CAP_SYS_NICE` for root VMs, how would opting-in to `CAP_SYS_RAWIO` be any different (if anything it'd be more restrictive)?

veps/sig-storage/259-rawio-lun-disks.md
Acedus: +1, I don't see how the attack surface changes whether it's the existing SCC or a different one.

veps/sig-storage/259-rawio-lun-disks.md
Acedus: > The feature gate is meant to be temporary and removed upon GA. At that point, regular users have access to a simple boolean that weakens virt-launcher security without a way for admins to block it.

I think that cluster admins who are cautious enough about cluster security may choose to deploy a VAP to block it entirely but @mhenriks's comment stands on its own.

Acedus (commented):
I'm inclined to cautiously lgtm this, some important notes though: 

1. The `KubevirtSeccompProfile` FG will be on by default after https://github.com/kubevirt/kubevirt/pull/17405, which means that virt-handler will install our custom seccomp profile (RuntimeDefault + userfaultfd for post-copy live migration) on all nodes in the cluster. Such profile is required to prevent `iopl(2)` and `ioperm(2)` use in a guest escape scenario. With that said, we'd need to render virt-launchers with disk rawio to use that seccomp profile since right now (even with the FG enabled) we don't set a default seccomp profile unless explicitly configured in the KV CR's `SeccompConfiguration.VirtualMachineInstanceProfile` field.
2. `mmap(2)` below `mmap_min_addr` is permitted by CAP_SYS_RAWIO, the kernel bypasses the minimum address check entirely in `cap_mmap_addr()`[1] when the capability is present in the init user namespace, which enables mapping the NULL page and lowers the bar for exploiting kernel NULL pointer dereference vulnerabilities. On Openhift where SELinux is enabled by default and container processes are granted `container_t`, this is mitigated since the capability check in `cap_mmap_addr()` passes but SELinux's independent check in `selinux_mmap_addr()` [2] denies it because `container_t` lacks `MEMPROTECT__MMAP_ZERO`. This is still worth noting since vanilla k8s doesn't ship with SELinux.

Having said all that, I think that provided we address my first point, this proposal is valid and doesn't risk the security posture of the project beyond what's necessary.

[1] https://github.com/torvalds/linux/blob/a293ec25d59dd96309058c70df5a4dd0f889a1e4/security/commoncap.c#L1474-L1478
[2] https://github.com/torvalds/linux/blob/a293ec25d59dd96309058c70df5a4dd0f889a1e4/security/selinux/hooks.c#L4027-L4038

mhenriks:
@vladikr @xpivarc mind sharing you current thoughts on this proposal with the community?

fabiand:
What are the next steps with this PR?

/remove-lifecycle stale

akalenyu:
@vladikr is hope this is a fair summary from today's VEP meeting :pray: 

So basically, since https://github.com/kubevirt/enhancements/pull/191#issuecomment-4247479361, we have in fact found a way to make
> Ambient cap: virt-launcher-monitor → virt-launcher

dynamic (Vladik had a successful PoC). So this kind of reopens the door for a discussion about going with a plugin, though, SCC/openshift specifics remain an issue and there was a suggestion to dip into this problem realm in this very VEP. I honestly think the whole openshift coupling issue deserves it's own VEP, but I understand we don't want to be accumulating more coupling.

I think we could find some middle ground to let this VEP progress, in that we will let an OpenShift-aware component (say, HCO) extend kubevirt's SCC/create a new one alongside the additional plumbing needed to make this secure on OpenShift. 

We should also weigh if this is fundamentally a core API or plugin material. I can see Vladik was also hinting towards it in that old comment
> plugin doesn't make sense and this IS a core API.

Acedus:
> @vladikr is hope this is a fair summary from today's VEP meeting 🙏
> 
> So basically, since [#191 (comment)](https://github.com/kubevirt/enhancements/pull/191#issuecomment-4247479361), we have in fact found a way to make
> 
> > Ambient cap: virt-launcher-monitor → virt-launcher
> 
> dynamic (Vladik had a successful PoC). So this kind of reopens the door for a discussion about going with a plugin, though, SCC/openshift specifics remain an issue and there was a suggestion to dip into this problem realm in this very VEP. I honestly think the whole openshift coupling issue deserves it's own VEP, but I understand we don't want to be accumulating more coupling.
> 
> I think we could find some middle ground to let this VEP progress, in that we will let an OpenShift-aware component (say, HCO) extend kubevirt's SCC/create a new one alongside the additional plumbing needed to make this secure on OpenShift.
> 
> We should also weigh if this is fundamentally a core API or plugin material. I can see Vladik was also hinting towards it in that old comment
> 
> > plugin doesn't make sense and this IS a core API.

bump @vladikr @xpivarc are there any remaining concerns?

Issue #141, opened 2025-11-27
VEP 141: Target-Side Pre-Migration Hooks for Domain XML Modification


Barakmor1:
/assign @Barakmor1 

This change is internal, with no API changes or upgrade risk. The final XML remains the same, so users see no behavior change. A feature gate would add no value. We believe this makes it safe to GA directly, but we’d love to hear your thoughts.

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

/sig storage
/sig network

**Enhancement link**:

[](https://github.com/kubevirt/enhancements/pull/142)

**Timeline**:

* Stable release target (1.8):

* [ ] GA
  * [ ] VEP PRs:
  https://github.com/kubevirt/enhancements/pull/142
  * [ ] Code PRs:
https://github.com/kubevirt/kubevirt/pull/16212
https://github.com/kubevirt/kubevirt/pull/16225

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
VEP PR is merged, moving this one to "Tracked".

iholder101:
Any plans for v1.9 @Barakmor1 ?

Barakmor1:
@iholder101 Yes, I’m planning to move it to beta by 1.9.

iholder101:
> [@iholder101](https://github.com/iholder101) Yes, I’m planning to move it to beta by 1.9.

Awesome!
Waiting for the beta graduation PR (see https://groups.google.com/g/kubevirt-dev/c/p-MF3X-NegI)!

iholder101:
@Barakmor1 https://github.com/kubevirt/enhancements/issues/111 seems to be dependent on this for GA.
Do you plan to GA this in v1.10?

CC @orelmisan

Barakmor1:
> [@Barakmor1](https://github.com/Barakmor1) [#111](https://github.com/kubevirt/enhancements/issues/111) seems to be dependent on this for GA. Do you plan to GA this in v1.10?
> 
> CC [@orelmisan](https://github.com/orelmisan)

@iholder101 @orelmisan Yes, thanks for the reminder, Here it is: 

https://github.com/kubevirt/enhancements/pull/368

xpivarc:
@Barakmor1 I marked this VEP as tracked from 1.10. Please untrack it if you don't intend to work on this in the release

Issue #111, opened 2025-10-28
VEP 111: Upgrade Path for VMs Using Ordinal Naming for Secondary Networks


orelmisan:
**Primary contact (assignee)**:

/assign @orelmisan 

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig network

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

orelmisan:
> [@orelmisan](https://github.com/orelmisan) don't we need a PR to update the VEP for this cycle?

@vladikr, Apologies for missing your message. Since #141 only graduates to Beta in v1.9, this feature cannot graduate to GA before its prerequisite. Consequently, we need to postpone the GA milestone to at least v1.10.

Please see https://github.com/kubevirt/enhancements/pull/275.

PR #262, opened 2026-04-14
VEP #141: Target Beta Graduation for v1.9

Barakmor1:
Remove the requirement to enable it by default, as it is now a global requirement.

### VEP Metadata

**Tracking issue**: 
https://github.com/kubevirt/enhancements/issues/141

**SIG label**: 
/sig sig-compute

### DIFF
--- a/veps/sig-compute/target-xml-mutation-hooks.md
+++ b/veps/sig-compute/target-xml-mutation-hooks.md
@@ -5,14 +5,14 @@
 ### Target releases
 
 - This VEP targets alpha for version: v1.8
-- This VEP targets beta for version:
+- This VEP targets beta for version: v1.9
 - This VEP targets GA for version:
 
 ### Release Signoff Checklist
 
 - [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
 - [x] (R) Alpha target version is explicitly mentioned and approved
-- [ ] (R) Beta target version is explicitly mentioned and approved
+- [x] (R) Beta target version is explicitly mentioned and approved
 - [ ] (R) GA target version is explicitly mentioned and approved
 
 ## Overview
@@ -163,8 +163,7 @@ It should be a scalable and extendable hook system that allows adding new hooks.
 - Add unit tests for hooks.
 
 ### Beta
-- Enable feature gate by default.
-- Migrate all existing XML modifications from source to target-side hooks (dedicated CPU pinning, etc.).
+- Migrate all target-based modifications from source to target-side hooks (dedicated CPU pinning, disk source path for cross-namespace or cross-cluster live migration).
 
 ### GA
 


veps/sig-compute/target-xml-mutation-hooks.md
```diff
@@ -163,7 +163,6 @@ It should be a scalable and extendable hook system that allows adding new hooks.
 - Add unit tests for hooks.
 
 ### Beta
-- Enable feature gate by default.
 - Migrate all existing XML modifications from source to target-side hooks (dedicated CPU pinning, etc.).
```
lyarwood: nit - do you want to list the remaining modifications here?

lyarwood (commented):
/lgtm

Simple nit to explicitly list the remaining modifications to migrate but otherwise I think this is good to graduate in 1.9.

veps/sig-compute/target-xml-mutation-hooks.md
xpivarc: Nit: Migrate all (target-based) existing XML modifications from source to target-side hooks ...

xpivarc (commented):
/approve
/hold

veps/sig-compute/target-xml-mutation-hooks.md
Barakmor1: Just for context, we have an XML modification that depends on information available only on the source node, specifically the disk size when the volume source is HostDisk.

This is an example of a source-side XML modification that should remain there, since the required information exists only in the source context:

https://github.com/kubevirt/kubevirt/blob/58725b1186435fcbed002e7beffd81fc867facdf/pkg/virt-launcher/virtwrap/live-migration-source.go#L976

I fixed the phrasing Thanks!

Barakmor1: Actually, after looking into it, the only remaining target-based XML modification is the disk source path, which can change during cross-namespace or cross-cluster live migration. I will add this information.

veps/sig-compute/target-xml-mutation-hooks.md
orelmisan: nit:
```suggestion
 [X] (R) Beta target version is explicitly mentioned and approved
```

orelmisan, line 112: Is there a plan to deprecate these (and/or other similar fields)?

veps/sig-compute/target-xml-mutation-hooks.md
```diff
@@ -163,8 +163,7 @@ It should be a scalable and extendable hook system that allows adding new hooks.
 - Add unit tests for hooks.
 
 ### Beta
-- Enable feature gate by default.
-- Migrate all existing XML modifications from source to target-side hooks (dedicated CPU pinning, etc.).
+- Migrate all target-based modifications from source to target-side hooks (dedicated CPU pinning, disk source path for cross-namespace or cross-cluster live migration).
```
orelmisan, line 166: Are there any potential issues when upgrading the cluster from a version which had the FG turned off?

orelmisan (commented):
Thank you @Barakmor1.
I support this graduation, as it is a prerequisite for enabling the functionality of VEP #111 by default.

veps/sig-compute/target-xml-mutation-hooks.md
Barakmor1: No, because the final XML is always identical, as the same modification log is applied during migration to the target.

veps/sig-compute/target-xml-mutation-hooks.md
Barakmor1: Yes it is listed in the [issue](https://github.com/kubevirt/enhancements/issues/141) as a Post- GA task

Barakmor1: Done

veps/sig-compute/target-xml-mutation-hooks.md
orelmisan: Assuming the logic won't drift between versions.

Barakmor1: If the logic changes, the final XML will be determined by the target hook.

orelmisan (approved):
Thank you @Barakmor1!

vladikr:
Looks good to me!
@xpivarc do we still need the hold?

Issue #144, opened 2025-11-28
VEP #144: Adjust virt-launcher memory lock requirements


bgartzi:
**Primary contact (assignee)**:

/assign @bgartzi

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig network

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/145

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):] v1,8

  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/145
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/16197
  * [ ] Docs PRs:
-->

> [!IMPORTANT]
> Please keep this description up to date.

bgartzi:
@vladikr thanks, the PR is linked in the description as "Code PR" (the one that is not closed, yet; https://github.com/kubevirt/kubevirt/pull/16384).
The implementation PR should be referencing this issue as you mentioned already.

I also wanted to edit the PR title to mention VEP144, but for some reason the "Edit" icon does not appear as in other repositories for me.

vladikr:
> [@vladikr](https://github.com/vladikr), I suggest we remove the SIG network label from this one, as it has moved to SIG compute.

@orelmisan Sure.

iholder101:
> [@iholder101](https://github.com/iholder101) thanks for the heads up. It should be up to date now :)

Thanks a lot! :)
One small note though - the links are under "stable" and not "alpha" :)

iholder101:
Hey @bgartzi!
Are there any plans to move this forward in v1.9?

iholder101:
Also, can you please add https://github.com/kubevirt/kubevirt/pull/16384 to the issue description under code PRs for v1.8?

bgartzi:
@iholder101 

> Are there any plans to move this forward in v1.9?

I thought it would be nice to move it to beta. However, I was told that to do that it should be proved that at least someone (person/feature) should be using it.

The only _user_ I'm aware of is vDPA network interfaces, which is not publicly available yet, although it will be published soon. Do we have to wait until that to start moving this?

The only TODO for beta agreed in the VEP doc is to notify publicly that in the future the other memory extension mechanism will be deprecated.

[...]

PR #263, opened 2026-04-14
VEP144: graduate reservedOverhead to beta

bgartzi:
### VEP Metadata

**SIG label**: /sig compute

### What this PR does
- Require functional testing as a requirement to graduate `ReservedOverheadMemlock` to beta.
- Set target version for `ReservedOverheadMemlock` beta graduation to v1.9

Might not be of good etiquette but moved documentation and functional testing requirements out from Alpha requirements in the VEP md, which were not met and moved them into other graduation targets.

### DIFF
--- a/veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
+++ b/veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
@@ -133,26 +133,26 @@ specify the memlock RLimit needs of a VM.
 
 The new field will be `spec.domain.memory.reservedOverhead`. It will
 contain two optional subfields:
-- `requiresLock` (`string`): An enum indicating whether or not to extend
-  the exiting memory lock RLimits of the `virt-launcher` pod and its
-  qemu process.
-- `value` ([`Quantity`][k8s-resource-quantity]): An optional field to
-  let users specify how much memory could be locked apart from the VM's
-  memory size. Note this could be larger than the guest's maximum memory
-  and even larger than the maximum memory allowed for the
+- `memLock` (`string`): An enum indicating whether or not to extend the
+  exiting memory lock RLimits of the `virt-launcher` pod and its qemu
+  process. It supports two values: `Required` and `NotRequired`.
+- `addedOverhead` ([`Quantity`][k8s-resource-quantity]): An optional
+  field to let users specify how much memory could be locked apart from
+  the VM's memory size. Note this could be larger than the guest's
+  maximum memory and even larger than the maximum memory allowed for the
   `virt-launcher`.
 
-If `requiresLock: "true"`, the logic that adjusts `virt-launcher` pod's
-memory lock RLimits will be triggered. By default (when `value` is empty
-or zero), the RLimits will be updated to match expected minimum value of
-it. This base will equal the sum of VM's base memory size and whatever
-other needs that KubeVirt might consider internally (see
+If `memLock: "Required"`, the logic that adjusts `virt-launcher` pod's
+memory lock RLimits will be triggered. By default (when `addedOverhead`
+is empty or zero), the RLimits will be updated to match expected minimum
+value of it. This base will equal the sum of VM's base memory size and
+whatever other needs that KubeVirt might consider internally (see
 [`GetMemoryOverhead`][kbvirt-getMemOverhead] logic).  Whatever value
-that `value` has will be added to that base.
+that `addedOverhead` has will be added to that base.
 
-Note that if the expected memory RLimit is $2 * Mem_{VM}$, `value` only
-needs to carry $Mem_{VM}$, as the base RLimit will already consider the
-memory size of the VM once.
+Note that if the expected memory RLimit is $2 * Mem_{VM}$,
+`addedOverhead` only needs to carry $Mem_{VM}$, as the base RLimit will
+already consider the memory size of the VM once.
 
 Users could set those fields by themselves. Mutating admission webhooks
 could also watch for creation of VMs and compute memlock RLimit
@@ -182,9 +182,10 @@ the calculation existing for VFIO, SEV and RT VMs.
 ## API Examples
 
 Users can specify that a `VirtualMachine` could require to lock memory
-through `spec.domain.memory.reservedOverhead.requiresLock`. If any other
+through `spec.domain.memory.reservedOverhead.memLock`. If any other
 region apart from the VM memory could be locked too, the user can
-specify it by setting `spec.domain.memory.reservedOverhead.value`.
+specify it by setting
+`spec.domain.memory.reservedOverhead.addedOverhead`.
 
 An example with both fields set:
 
@@ -198,9 +199,9 @@ spec:
     spec:
       domain:
         memory:
-          value:
-            value: 1G
-            requiresLock: "true"
+          reservedOverhead:
+            addedOverhead: 1G
+            memLock: "Required"
   ...
 ```
 
@@ -289,7 +290,8 @@ current stage.
 Apart from that, users could set these fields by themselves, but we
 expect this general mechanism to be consumed mainly by mutating
 admission webhooks not to require the user to run some of the
-mathematical operations needed to come up with proper `value` values.
+mathematical operations needed to come up with proper `addedOverhead`
+values.
 
 Note that mutating admission webhooks do introduce certain latency
 during pod admission. This means that in case the number of devices and
@@ -309,9 +311,9 @@ Unit tests should cover that `virt-handler` specifies the right memory
 lock RLimit value for different `spec.domain.memory.reservedOverhead`
 values:
 - The field does not exist or is empty.
-- It exists with `requiresLock: "true"` but an empty `value`.
-- It exists with `requiresLock: "true"` and a valid `value`.
-- It exists with `requiresLock: "false"` and a valid `value`.
+- It exists with `memLock: "Required"` but an empty `addedOverhead`.
+- It exists with `memLock: "Required"` and a valid `addedOverhead`.
+- It exists with `memLock: "NotRequired"` and a valid `addedOverhead`.
 - Other conditions such as SEV/RT/SR-IOV are met and there are other
   requirements present in `reservedOverhead`.
 
@@ -347,7 +349,7 @@ The following e2e scenarios should be considered:
   `spec.domain.memory.reservedOverhead` and apply the memory lock limits
   to `virt-launcher` and QEMU processes accordingly under a feature
   gate.
-- Add feature gate `MemLockRLimitConfiguration`.
+- Add feature gate `ReservedOverheadMemlock`.
 
 ### Beta
 - Implement functional tests.
@@ -356,5 +358,5 @@ The following e2e scenarios should be considered:
 - Documentation.
 
 ### GA
-- Remove the `MemLockRLimitConfiguration` feature gate.
+- Remove the `ReservedOverheadMemlock` feature gate.
 - Deprecate other memory extension mechanisms.
--- a/veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
+++ b/veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
@@ -5,7 +5,7 @@
 ### Target releases
 
 - This VEP targets alpha for version: v1.8
-- This VEP targets beta for version:
+- This VEP targets beta for version: v1.9
 - This VEP targets GA for version:
 
 ### Release Signoff Checklist
@@ -14,7 +14,7 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 - [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
 - [x] (R) Alpha target version is explicitly mentioned and approved
-- [ ] (R) Beta target version is explicitly mentioned and approved
+- [x] (R) Beta target version is explicitly mentioned and approved
 - [ ] (R) GA target version is explicitly mentioned and approved
 
 ## Overview
@@ -330,6 +330,12 @@ The following e2e scenarios should be considered:
 
 ## Implementation History
 
+- 2025/04/29: Plans to graduate the feature into beta in v1.9 are
+  accepted.
+- 2025/03/04: Feature lands Alpha stage in v1.8
+  https://github.com/kubevirt/kubevirt/pull/16956
+- 2025/02/27: Feature implementation lands the main branch
+  https://github.com/kubevirt/kubevirt/pull/16384
 - 2025/12/10: Working on the draft implementation for the new design.
 - 2025/11/21: Draft implementation of the first proposed VEP:
   https://github.com/kubevirt/kubevirt/pull/16197
@@ -342,12 +348,12 @@ The following e2e scenarios should be considered:
   to `virt-launcher` and QEMU processes accordingly under a feature
   gate.
 - Add feature gate `MemLockRLimitConfiguration`.
-- e2e testing implementation.
-- Documentation.
 
 ### Beta
+- Implement functional tests.
 - Notify that other memory extension mechanisms will be replaced by this
   one in the future.
+- Documentation.
 
 ### GA
 - Remove the `MemLockRLimitConfiguration` feature gate.


bgartzi:
Using this VEP graduation PR as the bait, I would like to take the chance to have a little discussion about the original API design for `ReservedOverheadMemlock` before we move on.

One of the use-cases that we talked about during the initial design phase was using this field as a target for mutating admission webhooks that would add overhead/memlock requirements to this field based on their needs.

We tested a mutating admission webhook implementation that would track and modify that field on VMI CREATE. In case `reservedOverhead `already has a value, it's assumed that the user added it so add "our value" to that. However, we're having some problems to figure out how it should behave in front of UPDATE actions. It could be a trivial action, such as the user hotplugging another interface, for example. We need to increase the limit. But, what if the user updates both the `reservedOverhead` value and other related fields at the same time? It's difficult to know from a webhook perspective whether the new VMI spec's `reservedOverhead` values were taken from the spec that was already modified by the mutating webhook, or from the original spec that was modified in the user's local environment and sent as an update request.

I'm flagging this here not to beg for hints (which I would appreciate in a separate discussion thread) but to flag it as a possible initial design flaw that I would like to discuss with you.

cc: @iholder101, @vladikr, @pprabhu3

vladikr:
> Using this VEP graduation PR as the bait, I would like to take the chance to have a little discussion about the original API design for `ReservedOverheadMemlock` before we move on.
> 
> One of the use-cases that we talked about during the initial design phase was using this field as a target for mutating admission webhooks that would add overhead/memlock requirements to this field based on their needs.
> 
> We tested a mutating admission webhook implementation that would track and modify that field on VMI CREATE. In case `reservedOverhead `already has a value, it's assumed that the user added it so add "our value" to that. However, we're having some problems to figure out how it should behave in front of UPDATE actions. It could be a trivial action, such as the user hotplugging another interface, for example. We need to increase the limit. But, what if the user updates both the `reservedOverhead` value and other related fields at the same time? It's difficult to know from a webhook perspective whether the new VMI spec's `reservedOverhead` values were taken from the spec that was already modified by the mutating webhook, or from the original spec that was modified in the user's local environment and sent as an update request.
> 
> I'm flagging this here not to beg for hints (which I would appreciate in a separate discussion thread) but to flag it as a possible initial design flaw that I would like to discuss with you.
> 
> cc: @iholder101, @vladikr, @pprabhu3

Thank you @bgartzi , but I'm curious, why is `UPDATE`  needed in this case? What is the use case?

bgartzi:
> Thank you @bgartzi , but I'm curious, why is UPDATE needed in this case? What is the use case?

iface hot-(un)plug. If I understood correctly currently is not supported for network binding plugins, but could be interesting for other cases I guess?

vladikr:
> > Thank you @bgartzi , but I'm curious, why is UPDATE needed in this case? What is the use case?
> 
> iface hot-(un)plug. If I understood correctly currently is not supported for network binding plugins, but could be interesting for other cases I guess?

@bgartzi I don't have a clear picture of how that would work. However, I think that a specific webhook or a Mutation Admission Policy would need to know that something it cares/responsible for has changed and would need to "add" the additional overhead.  The main point is that the webhook wouldn't react to an arbitrary change but only track a specific resource.

iholder101:
> > > Thank you @bgartzi , but I'm curious, why is UPDATE needed in this case? What is the use case?
> > 
> > 
> > iface hot-(un)plug. If I understood correctly currently is not supported for network binding plugins, but could be interesting for other cases I guess?
> 
> @bgartzi I don't have a clear picture of how that would work. However, I think that a specific webhook or a Mutation Admission Policy would need to know that something it cares/responsible for has changed and would need to "add" the additional overhead. The main point is that the webhook wouldn't react to an arbitrary change but only track a specific resource.

I tend to agree to this line of thinking.

Mutating policies/webhooks run serially, and I think in this case they shouldn't care about what the others do. It is possible to create a mutating policy/webhook for each managed device, each one will only query the device it's responsible for. If there's a case where it's important to know which webhook updated the value and by how much, this can be done via helper annotations (although I'm not sure I see the use-case for that).

As a side note - if more logic is needed (to modify the domain XML / perform node-level operations) - it the mutating webhook/policy can be wrapped into a [plugin](https://github.com/kubevirt/enhancements/issues/190).

@bgartzi what do you think?

enp0s3:
> > > > Thank you @bgartzi , but I'm curious, why is UPDATE needed in this case? What is the use case?
> > > 
> > > 
> > > iface hot-(un)plug. If I understood correctly currently is not supported for network binding plugins, but could be interesting for other cases I guess?
> > 
> > 
> > @bgartzi I don't have a clear picture of how that would work. However, I think that a specific webhook or a Mutation Admission Policy would need to know that something it cares/responsible for has changed and would need to "add" the additional overhead. The main point is that the webhook wouldn't react to an arbitrary change but only track a specific resource.
> 
> I tend to agree to this line of thinking.
> 
> Mutating policies/webhooks run serially, and I think in this case they shouldn't care about what the others do. It is possible to create a mutating policy/webhook for each managed device, each one will only query the device it's responsible for. If there's a case where it's important to know which webhook updated the value and by how much, this can be done via helper annotations (although I'm not sure I see the use-case for that).
> 
> As a side note - if more logic is needed (to modify the domain XML / perform node-level operations) - it the mutating webhook/policy can be wrapped into a [plugin](https://github.com/kubevirt/enhancements/issues/190).
> 
> @bgartzi what do you think?

@iholder101 Hi, I think that the webhook approach is too risky here. I prefer to go with a plugin-based design so that the final calculation could be orchestrated by the handler. I think webhooks are risky in case we have few of them operating on the same field. There is no way to enforce proper implementation of the webhook.

EdDev:
> But, what if the user updates both the `reservedOverhead` value and other related fields at the same time? It's difficult to know from a webhook perspective whether the new VMI spec's `reservedOverhead` values were taken from the spec that was already modified by the mutating webhook, or from the original spec that was modified in the user's local environment and sent as an update request.

The webhook or any hook point in general should observe the state and operate in an idempotent manner. It should not care about changes and imperative operations.
This is how we expect the domain xml hook point that the network binding plugin is using to behave, i.e. the vendor should not assume it is called once or twice, it should be assumed to be called multiple times for different reasons.

If you depend on the change sequence, there is something wrong with the architecture.

bgartzi:
Hi, I'm replying everything in a single message.

@vladikr 
> I think that a specific webhook or a Mutation Admission Policy would need to know that something it cares/responsible for has changed and would need to "add" the additional overhead. The main point is that the webhook wouldn't react to an arbitrary change but only track a specific resource.

Yes, when a new VMI is created is pretty simple. The webhook scans for things that it is interested in (e.g. devices) and adds to the additional overhead. 1 device? 1Gi. 2? 2Gi, whatever.

My concern is about later in the VMI lifecycle. What if the user removes one of those devices? We had 2, now we have 1 so we need 1Gi. But how do we know if the user used the original spec for `k apply -f` or if they did `k get vm withReservedOverhead` + `k apply -f`. In other words, how does the webhook know if in this case, an `UPDATE` action in terms of the mutating admission webhook, needs to still add or remove 1Gi from the incoming spec?

@iholder101 
> Mutating policies/webhooks run serially, and I think in this case they shouldn't care about what the others do. It is possible to create a mutating policy/webhook for each managed device, each one will only query the device it's responsible for.

Yes, when a new VMI is defined, there's no problem. Each mutating webhook assumes that there could be others and limits itself to creating the field if it wasn't there, or adding some value to it.

It is during a spec update (device hot unplug, for example) that we don't know if the value that the user sent was meant to be the result of the sum of all mutating admission webhooks, if it is some value they came up for any other use-case they needed, or if it is the original vmi spec in which the `additionalOverhead` value was possibly empty. 

> If there's a case where it's important to know which webhook updated the value and by how much, this can be done via helper annotations (although I'm not sure I see the use-case for that).

It would be helpful if all interactions to the field were "recorded", even user's, which makes it harder to use IMO as they would need to track "the same thing" twice.

@enp0s3 
> Hi, I think that the webhook approach is too risky here. I prefer to go with a plugin-based design so that the final calculation could be orchestrated by the handler. 

That can be a way to make each webhook/plugin agnostic of the rest, I believe. Instead of operating on an overall value, in which webhooks do not know whether they need to add/remove, you have an array of plugins that declare what they need, which virt-handler limits to sum and apply in the end. This also makes it "hidden" so users can't retrieve the result and feed it back in a spec update, making the amounts double.

Another thing I had in mind was making the field somewhat like a dictionary, in which different use-cases could add their 'definition': value key-value pairs, which the virt-handler would sum and apply. However, these values are explicit to the rest, which could lead to bad webhook implementations that remove the existing ones or something like that.

@EdDev 
> The webhook or any hook point in general should observe the state and operate in an idempotent manner. It should not care about changes and imperative operations.

It is pretty much what we can do on VMI creation, as we know that the mutating webhook will be called just once for the [default reinvocationPolicy value](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#mutatingwebhook-v1-admissionregistration-k8s-io).

If it wasn't for that, we wouldn't know if the incoming value was set by the user, another webhook, or the webhook in a previous call to it.

> If you depend on the change sequence, there is something wrong with the architecture.

Which it is pretty much what I was trying to call out in this discussion. Even if the use-case I had in mind for updates was interface hot-(un)plug, which is not supported for network binding plugins afaict, this seemed like something that needed to be pointed out "early" in alpha, rather than later, when this was available generally and thus harder to get rid of.

iholder101:
@bgartzi so if I understand your line of thinking correctly, the main concern here is that the webhook doesn't know who set and current value of the overhead and why, i.e. if it was the user (assessing by himself how much overhead is needed) or the previous webhook call when the VMI was created. Is that correct?

My general opinion here is that KubeVirt should not enforce how exactly people will be using this mechanism, but will provide the necessary facility in order to achieve it, in multiple possible ways. It's fine if some users will delegate the overhead calculation to the VM owner, while others will decide that only webhooks will use this value to update it automatically. Some may even choose to prevent (via VAP) users from setting this value explicitly altogether. It's even possible, as said earlier, to "document" which webhooks did what via annotations, although I tend to agree with @EdDev that this might imply that there's something wrong with the architecture. Either way, this is a matter of how every user will decide the architecture for their environment.

WDYT?

bgartzi:
@iholder101 
> so if I understand your line of thinking correctly, the main concern here is that the webhook doesn't know who set and current value of the overhead and why, i.e. if it was the user (assessing by himself how much overhead is needed) or the previous webhook call when the VMI was created. Is that correct?

Pretty much, because we'd need to know which part of the result is owned by "us" and ensure that total is indeed the sum of all pieces (so no component is left behind).

> WDYT?

As long as everyone is aware of the implications of this, I'm fine. AFAICT, in-place `reservedOverhead` updates are not supported at the moment, as memlock limits are set on creation and VM migration. If some sort of migration-based hot (un)plug mechanism would rely on `reservedOverhead` we would need to tell users to be careful with how they set up the `addedOverhead` calculation mechanism.

This is a good hint of the limitations of the current API proposal in my opinion, which is worth it to at least document I believe.

Does this concern anyone else more than that?

enp0s3 (commented):
@bgartzi Hi,
Would you be able to align the design and API examples sections of the VEP with the API that you've used in the code PR? It looks like the VEP is outdated with respect to the merged API:
```
type ReservedOverhead struct {
	// AddedOverhead determines the memory overhead that will be reserved
	// for the VM. It increases the virt-launcher pod memory limit.
	// +optional
	AddedOverhead *resource.Quantity `json:"addedOverhead,omitempty"`
	// RequiresLock determines whether the VM's and its overhead memory
	// need to be locked or not. It is a common practice to enable this
	// if vDPA, VFIO or any other specialized hardware that depends on
	// DMA is being used by the VM.
	// False - (Default) memory lock RLimits are not modified.
	// True - Memory lock RLimits will be updated to consider VM memory
	//        size and memory overhead
	// +optional
	// +kubebuilder:validation:Enum=NotRequired;Required
	MemLock *MemLockRequirement `json:"memLock,omitempty"`
}

// MemLockRequirement describes whether the VM memory and its overhead
// needs to be locked or not.
type MemLockRequirement string

const (
	// MemLockRequired means that the VM memory and its overhead
	// could be locked by a device or any other condition.
	MemLockRequired = "Required"
	// MemLockNotRequired means that the VM memory and its overhead
	// is not going to be locked.
	MemLockNotRequired = "NotRequired"
)
```

enp0s3 (commented):
@bgartzi IIUC in some circumstances `MemLockNotRequired` will be ignored, for example when the VMI has VFIO device and the user configured the following:
```
  template:
    spec:
      domain:
        memory:
          reservedOverhead:
            addedOverhead: 1G
            memLock: "NotRequired"
```
It will be a surprise that the ~~1G was locked anyway~~ rlimit was stretched with the 1G.

enp0s3:
> Another thing I had in mind was making the field somewhat like a dictionary, in which different use-cases could add their 'definition': value key-value pairs, which the virt-handler would sum and apply. However, these values are explicit to the rest, which could lead to bad webhook implementations that remove the existing ones or something like that.

@bgartzi Following the comments I am convinced now that simplicity will both keep the API clean and give the freedom for users to implement their mechanisms. 

>   This is a good hint of the limitations of the current API proposal in my opinion, which is worth it to at least document I believe.

+1

bgartzi:
@enp0s3 
> @bgartzi IIUC in some circumstances `MemLockNotRequired` will be ignored, for example when the VMI has VFIO device and the user configured the following:
> 
> ```
>   template:
>     spec:
>       domain:
>         memory:
>           reservedOverhead:
>             addedOverhead: 1G
>             memLock: "NotRequired"
> ```
> 
> It will be a surprise that the ~1G was locked anyway~ rlimit was stretched with the 1G.

True. Sounds like a bug. I opened https://github.com/kubevirt/kubevirt/issues/17619 to track it.

This brings another topic, closely related to the previous discussion we've had on this PR, which is: what if users need set `addedOverhead` for two different reasons, one of which requires memlock and the other does not?

I'd say KubeVirt is kind of conservative in terms of the amount of things that takes into account when it sets the maximum memory lock limits as they are higher than libvirt's as to accommodate it properly.

I'd propose to go easy on that and say that if 1 needs memLock, all need memLock, but this is another limitation of the current API that might be as well worth it documenting.

Should these type of comments go here into the VEP or somewhere in kubevirt/kubevirt?

bgartzi:
@enp0s3 
> @bgartzi Hi, Would you be able to align the design and API examples sections of the VEP with the API that you've used in the code PR? It looks like the VEP is outdated with respect to the merged API:
> 

Last pushed version adds a commit fixing the API spec example and references spread all over the document text. I can squash the commits if you prefer it that way.

vladikr:
In general, I'm +1, looks good from my side.
/approve

vladikr:
@bgartzi did you have a chance to follow up with libvirt about the need for https://gitlab.com/libvirt/libvirt/-/blob/master/src/qemu/qemu_domain.c#L8425 ?

veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
```diff
@@ -342,13 +342,13 @@ The following e2e scenarios should be considered:
...
 
 ### Beta
+- Implement functional tests.
 - Notify that other memory extension mechanisms will be replaced by this
   one in the future.
 
 ### GA
 - Remove the `MemLockRLimitConfiguration` feature gate.
 - Deprecate other memory extension mechanisms.
+- Documentation.
```
iholder101: I think we should move this to Beta. WDYT?
Would you be able to handle that in the upcoming cycle?

iholder101:
> I'd propose to go easy on that and say that if 1 needs memLock, all need memLock, but this is another limitation of the current API that might be as well worth it documenting.

Commented on the issue you've opened. Let's continue the discussion over there.

> Should these type of comments go here into the VEP or somewhere in kubevirt/kubevirt?

I think that you can concisely mention these in the VEP for completeness (perhaps in a follow-up to not block this PR).
I also think a user-guide entry should be introduced, preferably in the beta phase (see my comment above).

Other than that - looks great to me!

veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
EdDev, line 343: I think the Alpha implementation merits linking here: https://github.com/kubevirt/kubevirt/pull/16384

veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
EdDev: +1

EdDev (approved):
This looks solid to me, thank you.

I second @iholder101 on the need for documentation during beta, encouraging consumption & usage.

Said that, I do not want to hold on this as it seems to be an easy one.

Holding to allow the discussion to continue.
/hold

veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
bgartzi: Moved into beta.

veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md
bgartzi: I also added a slight mention to the backport PR and today (date we are accepting plans for beta).

bgartzi:
@vladikr I'm trying to double confirm the iommufd and memory lock limit requirements on libvirt, I'll be back in a bit

iholder101:
Thanks a lot @bgartzi!
/lgtm

EdDev (approved):
Thank you!

bgartzi:
@vladikr memlock limits are still predicted using the same logic as before in libvirt, no matter iommufd is set as the backend for vfio devices or not. Limits might be optimized and lowered, but that requires further research.

EdDev:
> Holding to allow the discussion to continue.
> /hold

I think we are good. :pray:  

/unhold

vladikr:
> @vladikr memlock limits are still predicted using the same logic as before in libvirt, no matter iommufd is set as the backend for vfio devices or not. Limits might be optimized and lowered, but that requires further research.

@bgartzi - yes. but I think it isn't right to multiply the whole memory by the number of devices - instead, it should be the_whole_vm_memory + number_of_devices * fudge_factor(1 GiB)

Issue #199, opened 2026-02-05
VEP 197: Enabling KubeVirt on NVIDIA Grace-Blackwell/Vera-Rubin System


fanzhangio:
**Primary contact (assignee)**:

@fanzhangio 
/assign @fanzhangio 

**Current Feature Stage**:
New

**Feature Gate**:
`NVGRACE`

**Responsible SIGs**:

Primary SIG:
/sig $SIG

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:
TBD

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@fanzhangio @alaypatel07 We should sync about our plans for this work. In general, as @lyarwood mentioned, we have advanced in the https://github.com/kubevirt/kubevirt-aie repo (See PR 10 in particular and https://github.com/vladikr/iommufd-device-plugin)

lyarwood:
@fanzhangio would you mind updating the description with links to the VEP and implementation PRs for v1.9.0?

As discussed I don't think this FG can mature to Beta until the CentOS Stream AIE SIG changes land back upstream in CentOS Stream and are consumed by kubevirt/kubevirt. AFAIK that should be happening soon but I don't think it will be in time for v1.10.0 unfortunately.

lyarwood:
@fanzhangio as discussed I don't think we have  any plans to  advance  this to beta  in v1.10.0 given all of the outstanding changes in the CentOS Stream 10 kernel, QEMU and libvirt. Shall we drop it from the release for now and pick this back up in v1.11.0?

vladikr:
@lyarwood do we have any additional alpha work that we need to do in 1.10?

PR #267, opened 2026-04-15
VEP 266: Host Devices Assignment with IOMMUFD

fossedihelm:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/pull/266
**SIG label**: /sig compute

### What this PR does
Add first proposal for VEP 186: Host Devices Assignment with IOMMUFD

### Special notes for your reviewer
This is a continuation of https://github.com/kubevirt/enhancements/pull/187

Issue #266, opened 2026-04-15
VEP 266: Host Devices Assignment with IOMMUFD


fossedihelm:
**Primary contact (assignee)**:

/assign @fossedihelm 

**Current Feature Stage**:
New

**Feature Gate**:
`IOMMUFD`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig network

**Enhancement link**:
#267 

**Timeline**:

* Alpha release target (x.y): 1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: #267 
  * [ ] Code PRs:
  * [ ] Docs PRs:

This is a continuation of https://github.com/kubevirt/enhancements/issues/186

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@fossedihelm do we need to do anything in 1.10? can't we move it to beta?

fossedihelm:
@vladikr I am not sure honestly. I think we have to move libvirt version at least to  a version >= 12.2( currently 11.10).
That is not compatible with the "enable by default" from the Beta graduation.
https://github.com/kubevirt/kubevirt/pull/17956#discussion_r3393644132
WDYT?

vladikr:
> [@vladikr](https://github.com/vladikr) I am not sure honestly. I think we have to move libvirt version at least to a version >= 12.2( currently 11.10). That is not compatible with the "enable by default" from the Beta graduation. [kubevirt/kubevirt#17956 (comment)](https://github.com/kubevirt/kubevirt/pull/17956#discussion_r3393644132) WDYT?

Hey @fossedihelm , why is it related to libvirt? If the logic in virt-launcher is behind a featuregate, then this deice plugin has not effect at all... 
My thinking was that it would be beneficial to get rid of this feature gate more quickly for this device plugin.

fossedihelm:
Because moving to Beta means on by default, so I think that at least the default libvirt version should support it, or am I missing something?
Actually, this will not be an issue on 1.37 provider (CS10 based), but it will be with older ones (still CS9).
Please let me know :) @vladikr

vladikr:
> Because moving to Beta means on by default, so I think that at least the default libvirt version should support it, or am I missing something? Actually, this will not be an issue on 1.37 provider (CS10 based), but it will be with older ones (still CS9). Please let me know :) [@vladikr](https://github.com/vladikr)

yea... I agree :(

PR #270, opened 2026-04-15
VEP-199 NVIDIA Grace-Blackwell Support in KubeVirt

fanzhangio:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/199
**SIG label**:   /sig compute

### What this PR does

This VEP enables production-ready GPU passthrough on NVIDIA Grace Hopper and Grace Blackwell (GB200/GB300) systems in KubeVirt.

PR of key commits of implementation: https://github.com/kubevirt/kubevirt-aie/pull/37

### DIFF
--- a/veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
+++ b/veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
@@ -1,0 +1,489 @@
+# VEP 199: NVIDIA Grace GPU Passthrough Baseline in KubeVirt
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements](https://github.com/kubevirt/enhancements/issues/199) (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP introduces a small, **inference-first** baseline that lets KubeVirt run NVIDIA Grace Hopper and Grace Blackwell (GB200/GB300) GPU passthrough VMs on ARM64.
+
+It is gated by a new `GraceIOVirtualization` feature gate and is intentionally limited to what has already been validated in the [kubevirt-aie](https://github.com/kubevirt/kubevirt-aie) fork.
+
+The Phase 1 contract is:
+
+- A new `GraceIOVirtualization` feature gate controlling all Grace-specific domain conversion behavior.
+- Explicit dependency on `PCINUMAAwareTopology` ([VEP 115](https://github.com/kubevirt/enhancements/issues/115)) for NUMA-aware PCI placement.
+- Explicit dependency on the generic IOMMUFD path defined by [VEP 266](https://github.com/kubevirt/enhancements/issues/266) ([design PR #267](https://github.com/kubevirt/enhancements/pull/267)) for host device binding and FD delivery.
+- SMMUv3 IOMMU device emission per PCIe bus that hosts a passed-through Grace GPU.
+- ACPI Generic Initiator (GI) guest NUMA topology derived from the host via ACPI/sysfs and remapped onto guest NUMA cell IDs.
+- Automatic, platform-aware large-BAR `pcihole64` sizing from device BARs.
+- Admission validation for the Grace-specific configuration constraints introduced by Phase 1.
+
+EGM, vCMDQ, PCIe link speed/width modeling, PCIe switch topology mirroring, mixed GPU+NIC topologies, and any per-GPU PXB isolation work that goes beyond the kubevirt-aie baseline are explicitly **out of scope** for this VEP and will be specified in follow-up VEPs or design documents.
+
+## Motivation
+
+NVIDIA Grace Hopper and Grace Blackwell systems are becoming a standard substrate for AI and HPC on ARM64. Running these workloads inside KubeVirt VMs gives operators multi-tenancy, security isolation, and a Kubernetes-native lifecycle, but Grace introduces Cache Coherent Interconnect via NVLink-C2C between CPU and GPU and ARM64-specific I/O virtualization requirements (SMMUv3, ACPI Generic Initiator NUMA cells, very large device BARs) that x86 passthrough does not have.
+
+Without first-class support, operators must hand-craft QEMU command lines covering SMMUv3, GI NUMA cells, and 64-bit PCI hole sizing per VM. This VEP builds on the Grace baseline validated in [kubevirt-aie](https://github.com/kubevirt/kubevirt-aie) fork and defines the remaining Phase 1 guest-visible contract needed for upstream KubeVirt.
+
+## Grace Platform Requirements
+
+Grace GPU passthrough requires KubeVirt to construct more than a conventional PCI host-device assignment. The guest platform must expose the IOMMU, NUMA, and MMIO resources expected by the NVIDIA guest driver and firmware.
+
+### Guest-visible SMMUv3
+
+Grace systems use ARM SMMUv3 for hardware-accelerated DMA translation. For the Grace GPU passthrough configurations targeted by Phase 1, the guest-visible SMMUv3 topology is part of the device contract rather than an opaque host detail. KubeVirt must create the required SMMUv3 devices for the PCI buses that contain passed-through Grace GPUs and attach the host devices to those IOMMU instances during domain generation.
+
+### ACPI Generic Initiator Topology
+
+Grace platforms expose GPU-associated NUMA topology through ACPI SRAT and sysfs. The guest needs corresponding ACPI Generic Initiator structures so the NVIDIA guest driver can associate passed-through GPUs with the NUMA nodes and memory resources expected by the platform. KubeVirt derives this topology from host sysfs and emits the guest NUMA and GI information automatically.
+
+### Large-BAR MMIO Aperture
+
+Grace GPUs expose large 64-bit prefetchable PCI BARs that can exceed the default QEMU 64-bit PCI MMIO aperture. If the aperture is too small, guest firmware can fail to assign the GPU BARs, which can prevent the VM from booting or leave the GPU unusable. KubeVirt therefore auto-sizes `pcihole64` from the assigned device BARs during domain generation. The computed value is implementation-owned and is not exposed as a VMI API in Phase 1.
+
+## Goals
+
+- Make Grace GPU passthrough work on Grace Hopper and Grace Blackwell using the inference-first model already validated in kubevirt-aie.
+- Confine all Grace-specific domain conversion behavior behind the `GraceIOVirtualization` feature gate.
+- Reuse existing KubeVirt building blocks (VEP 115 NUMA-aware PCI placement, VEP 266 IOMMUFD) instead of duplicating them.
+- Derive Grace-specific guest topology (GI NUMA cells, NUMA distances, MMIO aperture) from host ACPI/sysfs at domain conversion time.
+- Reject statically invalid Grace VMI configurations at admission time, and fail fast with actionable errors for node-local discovery or runtime failures.
+
+## Non Goals
+
+- Defining a user-visible API for Grace topology (GI cell counts, NUMA distances, MMIO sizing, per-GPU PCI placement). All of these are inferred from the host.
+- EGM (Extended GPU Memory) backing for guest memory.
+- vCMDQ hardware command queue virtualization.
+- PCIe link speed and width modeling on guest root ports.
+- PCIe switch topology mirroring (e.g. ConnectX/BlueField/NVMe behind a switch).
+- Mixed GPU + SR-IOV NIC topologies with independent isolation policies.
+- Per-GPU PXB isolation beyond what is already validated by the kubevirt-aie baseline.
+- Generic large-BAR `pcihole64` API. This VEP does not introduce an override option to escape auto-configuration, see [Large-BAR MMIO Aperture (auto-sized)](#large-bar-mmio-aperture-auto-sized).
+- vGPU or MIG orchestration. This VEP covers passthrough only.
+- x86 platforms or non-Grace platforms.
+- Upstreaming the NVIDIA QEMU/libvirt patches; those are a platform prerequisite, not a KubeVirt code change.
+- Packaging NVIDIA QEMU/Libvirt with specific support on Grace-system for KubeVirt.
+
+## Definition of Users
+
+Infrastructure administrators, cloud providers, and platform engineers deploying NVIDIA Grace Hopper or Grace Blackwell systems for AI and HPC workloads on ARM64 Kubernetes clusters.
+
+## User Stories
+
+- As a platform engineer, I want to assign one or more Grace GPUs to a VMI using the standard KubeVirt host device API and have KubeVirt configure SMMUv3, GI NUMA cells, and the 64-bit PCI hole automatically.
+- As an infrastructure administrator, I want KubeVirt to reject Grace VMIs that are statically incompatible with the Phase 1 baseline (for example missing feature gates or an unsupported architecture) at admission time rather than at VM boot.
+- As a cluster operator, I want Grace-specific behavior to be off by default and opt-in through a single feature gate so existing non-Grace workloads are unaffected.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt) upstream.
+- [kubevirt/kubevirt-aie](https://github.com/kubevirt/kubevirt-aie) for the validated Grace baseline; see [release-1.8-aie-nv](https://github.com/kubevirt/kubevirt-aie/tree/release-1.8-aie-nv) and [NVIDIA fork branch](https://github.com/kubevirt/kubevirt-aie/tree/release-1.7-aie-nv)
+- [kubevirt/iommufd-device-plugin](https://github.com/kubevirt/iommufd-device-plugin) for the IOMMUFD device plugin used by VEP 266.
+
+---
+
+## Design
+
+### Design Principles
+
+This VEP follows an **inference-first** model.
+
+Operators do not describe Grace topology in the VMI spec. They request Grace GPUs through the existing host-device flow, and KubeVirt `virt-launcher` derives Grace-specific guest topology from host state at domain conversion time.
+
+| Behavior | Source |
+| --- | --- |
+| Whether Grace handling applies | `GraceIOVirtualization` feature gate, a Grace-class GPU host-device request, and Grace host detection in `virt-launcher` |
+| SMMUv3 device emission | VEP 199 guest contract, VEP 115 PCI bus placement, and host IOMMU capability discovery |
+| GI NUMA cell count and grouping per GPU | Host ACPI/sysfs, with documented implementation fallback for incomplete platform data |
+| Guest NUMA distance vectors involving GI cells | Host ACPI SLIT / `nodeX/distance` when available, remapped to guest cell IDs |
+| Large-BAR `pcihole64` size | Assigned device BARs plus implementation-owned platform alignment/safety margins |
+| Host device binding via IOMMUFD | Generic VEP 266 IOMMUFD path |
+| NUMA-aware PCI placement of Grace GPUs | VEP 115 `PCINUMAAwareTopology`; VEP 199 consumes the resulting guest PCI buses |
+
+This VEP defines the **required guest-visible behavior** for the Phase 1 baseline. It does not prescribe one exact discovery algorithm. An implementation is acceptable if it produces the same guest contract, fails closed when required platform support is missing, and remains compatible with VEP 115 and VEP 266.
+
+### Feature Gates and Dependencies
+
+Phase 1 introduces a single feature gate:
+
+- **`GraceIOVirtualization`** (alpha, off by default) gates all Grace-specific domain conversion behavior described in this VEP. `GraceIOVirtualization` does not implicitly enable any other feature gate. Where Grace depends on existing KubeVirt functionality, those dependencies must be enabled explicitly:
+
+- **`PCINUMAAwareTopology`** ([VEP 115](https://github.com/kubevirt/enhancements/issues/115)) must be enabled. Grace reuses the generic NUMA-aware PCI placement behavior and does not introduce a separate placement planner in Phase 1.
+
+- **IOMMUFD** ([VEP 266](https://github.com/kubevirt/enhancements/issues/266)) must be available for Grace host-device binding. VEP 199 consumes the generic device plugin, FD delivery, and domain-level `<iommufd>` behavior from VEP 266; it does not redefine those mechanisms.
+
+Admission rejects statically invalid Grace VMIs, such as missing required feature gates or unsupported architecture. Node-local details, including actual GI ranges, distance vectors, BAR sizes, and IOMMUFD FD availability, are validated by `virt-launcher` during domain conversion and must fail fast with actionable errors when unavailable.
+
+Dependency status at the time of this VEP:
+
+- **VEP 115 / `PCINUMAAwareTopology`** is tracked by [issue #115](https://github.com/kubevirt/enhancements/issues/115). The tracker is open and lists the feature gate as `PCINUMAAwareTopology`.
+- **VEP 266 / IOMMUFD host-device assignment** is tracked by [issue #266](https://github.com/kubevirt/enhancements/issues/266), with the design currently discussed in [PR #267](https://github.com/kubevirt/enhancements/pull/267).
+
+### Activation Surface
+
+#### Grace Device Classification
+
+A **Grace-class GPU host device** is a PCI host device whose KubeVirt `permittedHostDevices` entry maps the requested `deviceName` to a known NVIDIA Grace GPU PCI vendor/device ID. Resource names such as `nvidia.com/GB100_*` are operator-facing handles, but the classification is based on the underlying PCI vendor/device IDs, not on the resource name string alone.
+
+Admission may use the KubeVirt host-device configuration to determine whether a requested `spec.domain.devices.hostDevices[].deviceName` is Grace-class. During domain conversion, `virt-launcher` must verify the assigned BDF through sysfs, for example `/sys/bus/pci/devices/<bdf>/vendor` and `/sys/bus/pci/devices/<bdf>/device`, before applying Grace-specific topology.
+
+Node labels are not the classification mechanism. They may help scheduling, but Grace activation depends on the feature gate, the requested host-device resource, and runtime verification of the assigned PCI device.
+
+There is **no Grace-specific user-facing VMI API** in Phase 1. Grace handling is activated when all of the following are true:
+
+1. `GraceIOVirtualization` is enabled and the required VEP 115 / VEP 266 prerequisites are configured.
+2. The VMI requests one or more Grace-class GPU host devices through the existing host-device API.
+3. `virt-launcher` detects a Grace-capable host and matching Grace GPU devices.
+
+No annotation is required to opt a VMI into Grace handling. The previously proposed `alpha.kubevirt.io/graceVirtualization` annotation is removed from this VEP. Sub-feature toggles such as EGM and vCMDQ are deferred to follow-up VEPs.
+
+### Phase 1 Guest Contract
+
+When a Grace VMI is admitted, scheduled to a compatible Grace node, and Grace handling is active, KubeVirt provides the following guest-visible behavior:
+
+- Each PCIe bus that carries a passed-through Grace GPU has an associated guest-visible SMMUv3 IOMMU device in the libvirt domain.
+- Each passed-through Grace GPU is associated with a guest GI NUMA cell range, and the GPU `<hostdev>` references that range using `<acpi nodeset='...'>`.
+- Guest NUMA distances are derived from host ACPI/sysfs data when available and remapped to guest cell IDs. Documented platform defaults may be used only as implementation fallback.
+- The `pcie-root` controller has an auto-sized `<pcihole64>` value large enough for the aggregate 64-bit prefetchable BAR footprint plus KubeVirt-owned platform margins.
+- Host-device binding uses the generic VEP 266 IOMMUFD path.
+
+---
+
+## SMMUv3 Emission
+
+When Grace handling is active, KubeVirt emits one libvirt `<iommu model='smmuv3'>` element for each guest PCIe bus that carries a passed-through Grace GPU. The `pciBus` value is derived from the guest PCI bus or controller allocated by the VEP 115 NUMA-aware placement logic; VEP 199 does
+not prescribe a numbering scheme.
+
+```xml
+<devices>
+  <iommu model='smmuv3'>
+    <driver pciBus='...' accel='on' ats='...' ril='...' ssidSize='...' oas='...'/>
+  </iommu>
+</devices>
+```
+
+`pciBus` is derived from the guest PCI bus/controller allocated by the VEP 115 placement logic. `accel='on'` is required for the Phase 1 Grace contract; VEP 199 does not define a non-accelerated SMMUv3 fallback.
+
+The remaining SMMUv3 attributes are implementation-owned domain XML details. Where possible, KubeVirt should infer them from host IOMMU/device capabilities and the libvirt/QEMU platform defaults. The values used by the current kubevirt-aie baseline, such as `ats='on'`, `ril='off'`, `ssidSize='20'`, and `oas='48'`, may be used as alpha defaults for qualified Grace systems, but they are not user-visible API and are not stable VEP-level constants. If QEMU/libvirt grow reliable auto defaults for these attributes, KubeVirt may omit explicit values and let the lower layers select them.
+
+Notes:
+
+- VEP 199 owns the Grace guest-visible SMMUv3 topology.
+- VEP 266 owns the generic IOMMUFD path: device-plugin allocation, FD delivery, per-`<hostdev>` IOMMUFD binding, and the domain-level `<iommufd>` association.
+- `cmdqv` is not set in Phase 1. vCMDQ is deferred to a follow-up VEP.
+- If the selected node cannot provide the required VEP 266 IOMMUFD transport, KubeVirt must fail closed with an actionable error. Phase 1 does not define a silent downgrade to non-accelerated SMMUv3 for Grace passthrough.
+
+---
+
+## ACPI Generic Initiator Guest Topology
+
+NVIDIA Grace GPU passthrough requires guest ACPI Generic Initiator (GI) NUMA topology for the assigned GPUs. Without the expected GI topology, the NVIDIA guest driver can fail to initialize the passed-through GPU or leave it unusable for CUDA.
+
+### Sysfs-First Discovery Contract
+
+Phase 1 uses a sysfs-first discovery model. KubeVirt derives the guest GI topology from the selected host devices and the host-visible NUMA topology instead of making fixed Grace constants part of the VEP contract.
+
+The `release-1.8-aie-nv` already validated the core sysfs-first discovery primitive: it reads the GPU NUMA node, enumerates online host NUMA nodes, and identifies CPU-less, memory-less nodes from sysfs. VEP 199 builds the upstream contract around that host-discovered topology.
+
+The guest GI cell IDs, GI grouping, and distance matrix are therefore derived from host state. Phase 1 does not define a fixed contract such as exactly 8 GI cells per GPU or a fixed 10/11/40/80/120 distance table, which was validated as an alternative in `release-1.7-aie-nv` branch.
+
+### Guest NUMA Cell Layout
+
+The resulting guest topology adds CPU-less, zero-memory GI cells after the CPU NUMA cells produced by `guestMappingPassthrough`, and references those cells from the corresponding GPU `<hostdev>`:
+
+```xml
+<cpu>
+  <numa>
+    <!-- CPU cells from guestMappingPassthrough (count varies) -->
+    <cell id='0' cpus='0-N' memory='...' unit='KiB'/>
+    <!-- ... -->
+
+    <!-- GI cells for each passed-through Grace GPU. Count and grouping
+         are derived from host ACPI/sysfs. -->
+    <cell id='K'   memory='0' unit='KiB'/>
+    <cell id='K+1' memory='0' unit='KiB'/>
+    <!-- ... -->
+  </numa>
+</cpu>
+
+<devices>
+  <hostdev mode='subsystem' type='pci' managed='no'>
+    <source>
+      <address domain='0x0000' bus='0xNN' slot='0x00' function='0x0'/>
+    </source>
+    <acpi nodeset='K-K+M'/>
+  </hostdev>
+</devices>
+```
+
+The number of GI cells per GPU and the exact `nodeset` ranges depend on host ACPI data; they are not part of the VEP contract.
+
+### NUMA Distance Matrix
+
+The guest distance matrix for the synthesized topology is built by remapping host node IDs to guest cell IDs and reading the corresponding host distance vectors from `/sys/devices/system/node/nodeX/distance`. Because guest cells are renumbered (and a partial-passthrough VM may include only a subset of the host's GPUs), distances are projected through the host-to-guest mapping rather than copied verbatim.
+
+If the host does not expose a distance vector for a synthesized cell (for example a GI cell that is not visible via sysfs), the implementation MAY fall back to the kubevirt-aie baseline distance pattern. This fallback is documented as an implementation detail; the VEP contract is "use host ACPI/sysfs distances, remapped to guest cells".
+
+The schema impact (optional `<distances>` child on `<numa><cell>`, optional `cpus` attribute for CPU-less GI cells) is shared with VEP 115 and is not re-specified here.
+
+---
+
+## Large-BAR MMIO Aperture (auto-sized)
+
+Grace GPUs expose large 64-bit prefetchable BARs that exceed default QEMU 64-bit PCI MMIO aperture. If the aperture is too small, guest firmware can fail to assign the GPU BARs, which can prevent the VM from booting or leave the GPU unusable in the guest.
+
+Phase 1 **auto-sizes** `pcihole64` from the assigned PCI host devices. The value is not exposed as a VMI API.
+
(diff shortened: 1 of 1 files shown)


vladikr:
Hi @fanzhangio,

First of all, thank you for driving this VEP. The support for Grace/Blackwell GPU passthrough is a huge milestone for KubeVirt on ARM64.

I fully support upstreaming Grace-Hopper and GB200 (and Blackwell) support to KubeVirt 1.9

This is an initial review; however, I have some concerns with the current design direction of the VEP. In our validation on the [kubevirt-aie](https://github.com/kubevirt/kubevirt-aie) fork, we successfully implemented full GPU passthrough by having virt-launcher infer as much as possible dynamically from userspace rather than pushing configuration up to the VMI spec / API / controller level.
Specifically:

1. PCI Holes & Fake NUMA nodes
We should not require the VMI spec to define the 8 fake NUMA nodes or calculate the 64-bit PCI hole. virt-launcher already reads the real device BARs from /sys/bus/pci/devices/ and the ACPI NodeSet and injects everything into the libvirt domain automatically (see the merged PR [#10](https://github.com/kubevirt/kubevirt-aie/pull/10) in kubevirt-aie).

2. IOMMUFD management
We built the dedicated [iommufd-device-plugin](https://github.com/kubevirt/iommufd-device-plugin) that opens /dev/iommu, sets RLIMIT_MODE, and passes the FD to virt-launcher via SCM_RIGHTS. This keeps virt-launcher completely unprivileged and avoids polluting core KubeVirt with host-level device management.

I strongly believe virt-launcher must infer as much as possible from userspace to keep the KubeVirt API clean and abstract.

What I'll suggest is, let’s align this VEP and the implementation closer to the inference-first model already validated and merged in kubevirt-aie#10. 
The common case should “just work” with zero extra config.
Moving fast on validated features would be the least path of resistance from my perspective. Also the VEP freeze is on April 30th.

Regarding the truly advanced features (EGM, vCMDQ, per-GPU PXB, etc.) that cannot be safely auto-detected and were not validated elsewhere, I would suggest creating a follow-up VEP that builds on top of this one, where we could safely find an appropriate solution for KubeVirt or create its own experimental API. However, I can't say at this point if these features are ready to be integrated back into KubeVirt main.
I think that we should split the implementation into a small inference baseline + optional advanced features. 

Hope we could discuss this further on the AIE WG call.

Looking forward to your thoughts!

fanzhangio:
Hi @vladikr ,

Thank you for the thoughtful review. I fully agree on aligning with the inference-first model, and thanks for driving kubevirt-aie#10. it's great foundational work.

Let me clarify a few points and then address the scoping question.

**1. On inference vs. annotation**

The VEP does not require users to define the 8 GI NUMA nodes or calculate PCI hole sizes in the VMI spec. These are all auto-generated by virt-launcher at domain conversion time from sysfs topology discovery — exactly the 
inference-first model you describe. The `graceVirtualization` annotation is purely a feature enablement toggle (boolean flags: `hostDevices`, `smmuv3`, `vcmdq`, `egm`), not a configuration surface for NUMA cell counts or MMIO apertures.

On `pciHole64Size` specifically: I agree the common case should not require manual tuning. I'll implement auto-computation from aggregate device BAR sizes and keep `kubevirt.io/pciHole64Size` only as an advanced override for mixed or future platform configurations where default sizing may be insufficient.

**2. On iommufd-device-plugin**

Fully agree. The VEP's `<driver iommufd='yes'/>` on hostdev elements is compatible with the `SCM_RIGHTS` FD-passing approach. The VEP describes the required domain XML behavior, the exact FD delivery mechanism is an 
implementation detail that should remain open. I do not think the VEP needs to mandate it. I'll add a section clarifying this boundary.

**3. On scope and splitting**

I think we actually agree on the baseline more than it appears. The features that kubevirt-aie#10 already validates — per-GPU PXB isolation, SMMUv3 scoping, ACPI GI nodes, pcihole64, iommufd binding — are exactly what this 
VEP documents as the foundation. These are interdependent and cannot be separated:

- **ACPI GI nodes cannot be inferred from host sysfs** because they create a guest NUMA topology that does not exist on the host. The host has 2-4 NUMA nodes; a 4-GPU guest needs 34+ (2 CPU cells + 8 GI cells per GPU). The distance matrix (10/11/40/80/120) reflects NVLink-C2C interconnect characteristics — platform knowledge, not sysfs data.
- **Per-GPU PXB is required for SMMUv3** because SMMUv3 instances are scoped per PCI bus (`pciBus` attribute). Without isolated PXBs, you cannot have per-device SMMUv3 instances.
- **SMMUv3 nested translation is the baseline** for Grace GPU passthrough — the NVIDIA guest driver requires hardware-accelerated address translation for CUDA workloads.  

These are presented at [KubeVirt Summit 2026: KubeVirt On GB200: Virtualizing a Rack-Scale Supercomputer](https://www.youtube.com/watch?v=jtnRFgu4tdI).

On phasing, I'm open to narrowing the initial alpha scope. Here's a split that I think respects the platform interdependencies while keeping the v1.9 surface area focused:
- **Phase 1 (v1.9 alpha)**: GraceIOVirtualization feature gate, SMMUv3 + iommufd wiring, per-GPU PXB isolation, ACPI GI nodes, auto-computed pcihole64 and optional override, NUMA distance matrix, admission validation
- **Phase 2**: vEGM memory backend, vCMDQ hardware acceleration, PCIe link speed modeling, mixed GPU+NIC topology with PCIe switch mirroring

I should note that the device-plugin infrastructure for Phase 2 is already in place, vEGM support was merged in [kubevirt-gpu-device-plugin#179](https://github.com/NVIDIA/kubevirt-gpu-device-plugin/pull/179), so Phase 2 can move quickly once Phase 1 lands.

Happy to discuss the exact boundary on the AIE WG call — if there are specific Phase 1 items you'd want to move, I'd like to understand the concern so we can find the right cut.

lyarwood:
> ACPI GI nodes cannot be inferred from host sysfs because they create a guest NUMA topology that does not exist on the host. The host has 2-4 NUMA nodes; a 4-GPU guest needs 34+ (2 CPU cells + 8 GI cells per GPU). The distance matrix (10/11/40/80/120) reflects NVLink-C2C interconnect characteristics — platform knowledge, not sysfs data.

I don't think this is accurate. Looking at real GH200 systems (e.g. the [NCSA DeltaAI documentation](https://docs.ncsa.illinois.edu/systems/deltaai/en/latest/user-guide/architecture.html)), `numactl` output from a 4-GPU node shows memory binding across NUMA domains 0, 1, 2, 3 **and** GPU NUMA IDs 4, 12, 20, 28 — spaced exactly 8 apart. That's the full GI node structure visible on the host:

- Nodes 0-3: CPU NUMA nodes (one per superchip)
- Nodes 4-11: GPU 0 GI nodes (8 nodes)
- Nodes 12-19: GPU 1 GI nodes (8 nodes)
- Nodes 20-27: GPU 2 GI nodes (8 nodes)
- Nodes 28-35: GPU 3 GI nodes (8 nodes)

That's 36 NUMA nodes on the host, not 2-4. The GI nodes are ACPI firmware structures (SRAT/SLIT tables), not something created by the NVIDIA driver, so they should remain visible in `/sys/devices/system/node/` regardless of whether CDMM or NUMA mode is active and regardless of whether the GPU is bound to vfio-pci.

The NUMA distance matrix should also be readable from host sysfs via `/sys/devices/system/node/nodeX/distance`, which would let virt-launcher infer the distance values rather than hardcoding platform-specific constants like `giNodesPerGPU = 8` and the distance table.

This aligns with @vladikr's suggestion to use an inference-first approach — the GI structure, node count per GPU, and distance matrix could all be discovered from sysfs at domain conversion time rather than being treated as opaque platform knowledge.

fanzhangio:
Good catch on the GI node visibility @lyarwood. You're right that on Grace systems the ACPI SRAT/SLIT tables expose the full GI node structure to the host kernel, so the 8-per-GPU GI nodes and the distance matrix are discoverable via /sys/devices/system/node/. I overstated the case by saying "2-4 NUMA nodes" that's the CPU-only count, not the total.

I've updated the VEP to describe sysfs-based discovery for GI node ranges rather than hardcoding constants like giNodesPerGPU = 8. The planner now reads each GPU's primary node from `/sys/bus/pci/devices/<BDF>/numa_node` and scans consecutive CPU-less, memory-less nodes to determine the GI range, using other GPUs' primary nodes as boundaries. This adapts automatically across Grace Hopper (1C1G) and Grace Blackwell (1C2G) configurations.

For the NUMA distance matrix, host sysfs provides the raw distance values, but the guest cell IDs are renumbered (CPU cells from vCPU pinning, GI cells from discovered ranges) so a direct copy of host nodeX/distance vectors isn't valid as a guest distance matrix. The VEP documents a remapping policy for Phase 1 based on the qualified Grace
topology; a future enhancement can formalize a general host-to-guest distance remapping algorithm.

This doesn't change the Phase 1 scope — GI nodes and the distance matrix are still emitted in the guest domain XML.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+- [kubevirt](https://github.com/kubevirt/kubevirt) upstream
+- [kubevirt-aie nvidia branch](https://github.com/kubevirt/kubevirt-aie/tree/release-1.7-aie-nv) dedicated fork
+
+---
+
+## Design
+
+### Feature Gate
+
+A new `GraceIOVirtualization` feature gate controls all Grace-specific behavior. When enabled, it implicitly activates the `PCINUMAAwareTopology` behavior for Grace VMs regardless of whether the generic gate is also set.
```
fossedihelm: Do we really want to implicitly enable the `PCINUMAAwareTopology`?
I am not sure this was done before

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+- **Extended GPU Memory (EGM)** as the guest memory backend
+- **ACPI Generic Initiator (GI)** with 8 dedicated NUMA nodes per GPU for MIG support
+- **Per-GPU isolated PXB hierarchies** with per-bus SMMUv3 scoping
+- **Configurable 64-bit MMIO aperture** (pcihole64) for large-BAR GPUs
+- **PCIe link speed/width modeling** derived from host sysfs topology
+- **Mixed GPU+NIC PCI topology** with independent isolation policies
+- **VFIO overhead and memlock sizing** for large-BAR devices
+- **Admission validation** for Grace-specific configuration constraints
+
+This VEP depends on [VEP 115: PCIe NUMA Topology Awareness](https://github.com/kubevirt/enhancements/issues/115), which provides the generic NUMA-aware PXB planner architecture that this VEP extends for Grace-specific requirements. VEP 115 supplies the stateful PCI planner with resource reservation, collision prevention, and per-device-group PXB isolation. This VEP layers Grace-specific behaviors on top of that planner: per-GPU PXB isolation for SMMUv3 scoping, ACPI Generic Initiator NUMA cells, Grace-specific NUMA distances, and platform-specific subsystem configuration.
```
fossedihelm: Is it worth mentioning https://github.com/kubevirt/enhancements/pull/267 as well?

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+
+## Assumptions, Constraints, and Dependencies
+
+- **VEP 115**: This VEP requires the stateful NUMA-aware PXB planner architecture from VEP 115, including per-device-group PXB isolation and collision-free resource allocation.
+- **Grace Platform Software**: Requires NVIDIA-patched QEMU (`10.1.0+nvidia5`) and libvirt (`11.9.0+nvidia4-1`) for SMMUv3 nested
+  translation, vCMDQ, vEGM, and ACPI GI support.
+- **Grace Firmware**: NVIDIA GB200 firmware version 1.3 or later for ACPI IORT CANWBS memory access flag support.
+- **Machine Type**: ARM64 `virt` machine type.
+- **EGM Constraint**: EGM requires all GPUs on a compute tray to be passed through. Partial GPU configurations are not supported with EGM.
+- **Host Kernel**: Linux kernel with SMMUv3 nested translation support and iommufd enabled, e.g. 6.17-HWE (linux-nvidia Ubuntu Kernel).
```
fossedihelm: Can we mention https://github.com/kubevirt/enhancements/pull/267?

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+```
+
+| Field | Description |
+|---|---|
+| `hostDevices` | Enable Grace-specific host device topology (per-GPU PXB, GI nodes) |
+| `smmuv3` | Enable ARM SMMUv3 with nested translation per GPU bus |
+| `vcmdq` | Enable hardware command queue virtualization (requires `smmuv3`) |
+| `egm` | Use EGM character devices as guest memory backend |
+
+This annotation is a transitional alpha API. Once the design stabilizes, the intent is to promote these fields to a first-class VMI spec type.
```
fossedihelm: As per @vladikr and @lyarwood  comments, we should highlight that this is a fallback and we will focus on inference-first approach

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+
+Each attribute:
+
+| Attribute | Value | Description |
+|---|---|---|
+| `pciBus` | PXB bus number | Scopes this SMMUv3 to one PCI bus |
+| `accel` | `on` | Enable nested translation (host stage-2 + guest stage-1) |
+| `ats` | `on` | Address Translation Services for PCIe ATS capability |
+| `ril` | `off` | Range Invalidation Length (disabled for Grace) |
+| `pasid` | `on` | Process Address Space ID support |
```
fossedihelm: This was replaced by `ssidSize`

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+type IOMMUDriver struct {
+    PCIBus string `xml:"pciBus,attr,omitempty"`
+    Accel  string `xml:"accel,attr,omitempty"`
+    ATS    string `xml:"ats,attr,omitempty"`
+    RIL    string `xml:"ril,attr,omitempty"`
+    PASID  string `xml:"pasid,attr,omitempty"`
+    OAS    string `xml:"oas,attr,omitempty"`
+    CMDQV  string `xml:"cmdqv,attr,omitempty"`
+}
+```
```
fossedihelm: These implementation details can be omitted from VEP

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+(`accel=off`) to maintain basic functionality without nested translation.
+
+### Hostdev iommufd Binding
+
+Host devices on Grace use iommufd binding instead of legacy VFIO groups:
+
+```go
+type HostDeviceDriver struct {
+    IOMMUFD string `xml:"iommufd,attr,omitempty"`
+}
```
fossedihelm: https://github.com/kubevirt/enhancements/issues/266 will cover this

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1658 @@
...
+```
+
+```xml
+<hostdev mode='subsystem' type='pci' managed='no'>
+  <driver iommufd='yes'/>
```
fossedihelm: this was replaced by `<iommufd enabled='yes' fdgroup='iommu'/>` field under domain

fossedihelm (commented):
Thank you @fanzhangio.
Looks good overall. I agree with Vladik's comment about limiting the alpha to only the inferable capabilities and deferring the others to other cycles.
I also think we don't need too many implementation details in the VEP itself.
Last things... comments below :)

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Thanks @fossedihelm . Updated to ssidSize in the VEP.

To clarify:  
Our internal development branch (1.7.0-nvidia-aie) uses NVIDIA-patched QEMU/libvirt 
QEMU (10.1.0+nvidia5) and libvirt (11.9.0+nvidia4-1) where the attribute is still `pasid`. 
I noticed that upstream QEMU uses `ssidSize` (setting it to > 0 advertises the PASID capability). The VEP should target the upstream attribute name, so this is the right change.

I'm also aware that upstream may adopt `auto` defaults for `ats`, `ril`, and `oas` to match host SMMU properties, which would simplify the SMMUv3 element significantly. I've added a note in the VEP about
this (under the SMMUv3 attribute table).

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Done

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Done

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Good catch! Both must be explicitly enabled.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: yes, we align with inference-first as primary, annotation is just optional override.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Done

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Setting a global IOMMUFD is the right default (especially on Grace-system). We just wanted to keep the door open for supporting a per-<driver> override on individual hostdevs when there’s a need to deviate from the global setting.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+
+### GI Node Discovery from Host sysfs
+Host firmware exposes auxiliary GI nodes for each GPU via ACPI SRAT tables, visible in Linux sysfs at `/sys/devices/system/node/`. Typically, on Grace-Blackwell system, 8 NUMA nodes are associated with each GPU: the first serves as the GPU's device NUMA node, and the remaining 7 support MIG partitions.
+
+### Phase 1 Guest GI Model
+For Phase 1, KubeVirt synthesizes a fixed guest GI range for each selected Grace-Blackwell GPU rather than discovering host GI ranges from sysfs.
+This is intentional. Although qualified Grace hosts may expose GPU-related NUMA/GI information in `/sys/devices/system/node/`, host-visible GI node ranges are not a sufficiently robust source of truth for guest construction in the general case. In particular, partial GPU passthrough creates ambiguity when trying to infer GI group boundaries from only the selected devices. To keep guest behavior deterministic, KubeVirt constructs the guest GI topology directly.
```
lyarwood: This contradicts several other parts of the VEP:

- The Overview (line 26) says "ACPI Generic Initiator (GI) with dedicated NUMA nodes per GPU (discovered from host sysfs)"
- The Design Principles (line 122) say "This VEP follows an inference-first model"
- The Phase 1 Contract table (line 1573) says "ACPI GI nodes (sysfs-discovered) | NVIDIA guest driver fails without GI structures; nodes and count discovered from host sysfs"

Yet here the text says "synthesizes a fixed guest GI range ... rather than discovering host GI ranges from sysfs" and further argues that "host-visible GI node ranges are not a sufficiently robust source of truth."

This was discussed in the PR comments where both @vladikr and I pointed out that GI nodes are visible in host sysfs via ACPI SRAT tables, and you agreed to update the VEP to describe sysfs-based discovery. The VEP needs to be internally consistent - either Phase 1 discovers from sysfs (matching the Overview, Design Principles, and Phase 1 Contract table) or it uses fixed synthesis (and those other sections need updating).

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+| GI cell | Other GI cell for same GPU | 11 | Same GPU GI group |
+| GI cell | GI cell for different GPU | 40 | Cross-GPU / non-local GI group |
+
+This policy gives KubeVirt a deterministic guest contract independent of host node numbering and avoids overstating what can be inferred from generic sysfs alone.
+
+For a 4-GPU, 2-socket system with 34 NUMA cells (2 CPU + 32 GI), each cell carries a full 34-entry distance vector. Example for CPU cell 0 (socket 0):
+
+```xml
+<cell id='0' cpus='0-59' memory='468713472' unit='KiB'>
+  <distances>
```
lyarwood: This hardcoded Grace-specific distance policy contradicts the Phase 1 Contract table (line 1575) which says "NUMA distance matrix (sysfs-inferred) | Required for correct guest NUMA topology; distances read from host `/sys/devices/system/node/nodeX/distance`".

The body text here argues that host sysfs distances cannot be directly copied because guest NUMA cell IDs differ from host node IDs. That is a fair point - a remapping step is needed. But the Phase 1 Contract table claims the distances are sysfs-inferred, while this section defines a static lookup table with hardcoded values (10/11/40/80/120).

The VEP should settle on one approach and describe it consistently. If the distances are derived from host sysfs with remapping, describe the remapping. If they are hardcoded per-platform constants, update the Phase 1 table.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+### Auto-Computation from Device BARs
+
+By default, the planner auto-computes the required `pcihole64` size by reading the sysfs `resource` file for each passthrough device, summing total MMIO BAR sizes across all devices, and applying a **2x safety multiplier** to account for QEMU firmware alignment overhead and address space fragmentation. The result is clamped to a maximum of **16 TiB** (`17179869184` KiB), which is theoritical sufficient.
+
+### Optional Annotation Override
+
+For mixed or future platform configurations where the auto-computed value is insufficient, operators can override the pcihole64 size via annotation. The annotation value is in **KiB**:
+
+```
+kubevirt.io/pciHole64Size: "4294967296"  # 4 TiB in KiB
```
lyarwood: This annotation uses the `kubevirt.io/` prefix (not `alpha.kubevirt.io/`), implying it is a stable, non-Grace-specific annotation. But the auto-computation and usage is only described here in the Grace context. Is this intended to be a generic KubeVirt feature for any large-BAR device passthrough? If so, should it belong to VEP 115 or a separate VEP? If it is Grace-specific, should it use the `alpha.kubevirt.io/` prefix like the graceVirtualization annotation?

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+
+## Per-GPU Isolated PXB Hierarchy
+
+Unlike the original VEP 115 which uses one pxb-pcie per NUMA node with multiple root-ports, Grace requires **one pxb-pcie per GPU**. Each GPU gets its own isolated PCI bus so that SMMUv3 can be scoped per bus.
+
+NICs may share a PXB bus or receive isolated PXBs depending on topology and whether SMMUv3 is enabled for that bus.
+
+**PXB NUMA node vs. hostdev ACPI nodeset**: The PXB controller's `<node>` targets the *CPU NUMA node* (socket affinity for memory access), while the hostdev `<acpi nodeset>` targets the *GI NUMA cell range* (GPU device affinity for NVIDIA driver initialization). These serve different purposes and reference different guest NUMA cell IDs.
+
+The NUMA PCI planner groups devices by their host NUMA node and device type, then allocates PXB controllers accordingly:
```
lyarwood: "The planner" is referenced throughout this VEP (here, and in the MMIO, PCIe link, EGM, and mixed topology sections) but never formally introduced. Is this the NUMA PCI planner from VEP 115? Where does it run - in virt-controller, virt-handler, or the virt-launcher domain converter? A brief introduction at first mention would help readers understand the component boundaries.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+
+---
+
+## Configurable 64-bit MMIO Aperture
+
+Grace GPUs have BARs exceeding 256 GB. The default PCI 64-bit hole is insufficient.
+
+### Auto-Computation from Device BARs
+
+By default, the planner auto-computes the required `pcihole64` size by reading the sysfs `resource` file for each passthrough device, summing total MMIO BAR sizes across all devices, and applying a **2x safety multiplier** to account for QEMU firmware alignment overhead and address space fragmentation. The result is clamped to a maximum of **16 TiB** (`17179869184` KiB), which is theoritical sufficient.
```
lyarwood: Typo: "theoritical" -> "theoretical"

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+<iommufd enabled='yes' fdgroup='iommu'/>
+```
+
+This domain-level element activates iommufd for all hostdev elements. The iommufd device plugin (part of VEP 266) handles FD delivery via `SCM_RIGHTS`, keeping virt-launcher unprivileged.
+
+---
+
+## ACPI Generic Initiator
+
+Each Grace GPU requires dedicated Generic Initiator (GI) NUMA cells in the guest topology. The NVIDIA guest driver expects these ACPI GI structures to exist for GPU passthrough on Grace platforms. On current qualified Grace systems (e.g. Grace-Blackwell), each GPU requires a group of 8 GI cells in the guest: one device-affinity cell plus 8 auxiliary cells used by the NVIDIA guest driver to "onlines" GPU memory to the guest kernel. These GI cells are CPU-less and zero-memory. This is prerequisite for CUDA framework running on Grace-system with cache coherent GPUs.
```
lyarwood: Minor grammar: "This is prerequisite for CUDA framework running on Grace-system with cache coherent GPUs" -> "This is a prerequisite for the CUDA framework running on Grace systems with cache-coherent GPUs."

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,1612 @@
...
+  - name: rootdisk
+    persistentVolumeClaim:
+      claimName: pvc-rootdisk
+```
+
+---
+
+## Domain XML Example (Grace Blackwell, 120 vCPUs, 4 GPUs with EGM)
+
+The following is a condensed version of the actual domain XML generated by KubeVirt on a Grace Blackwell system (GB200) with 120 vCPUs (2 sockets x 60 cores), 4 GPU passthrough, EGM, SMMUv3, and vCMDQ. GPUs 0-1 are on socket 0, GPUs 2-3 are on socket 1. Distance entries and GI cells 4-33 are truncated for brevity but follow the same pattern.
```
lyarwood: The VEP template includes an "Alternatives" section that is missing from this VEP. Given the active design discussion around inference-first vs. annotation-driven approaches, and sysfs discovery vs. fixed synthesis for GI nodes and distance matrices, an Alternatives section would be valuable to document what was considered and why the chosen approach was selected.

Also, at 1612 lines this VEP is significantly longer than typical VEPs. The template says the Design section should be "brief and concise." Sections like the full domain XML example (lines 1143-1498), four separate VMI API examples, and detailed PCIe switch topology mirroring could be condensed substantially. A VEP should capture design decisions and contracts rather than serving as an implementation guide - the implementation details belong in design docs or PR descriptions.

lyarwood:
Thanks for all the work on this VEP @fanzhangio!

Talking to @vladikr, @fossedihelm and others, we think we'd make faster progress toward landing this if we simplified the VEP significantly. Right now at 1600+ lines it reads more like an implementation guide than a design proposal, and that's making it harder to review and reach consensus on the core design decisions.

What I'd suggest is trimming the VEP down to focus on:

- The high-level architecture and goals
- The inference-first design contract (what KubeVirt guarantees to the guest, not how it discovers or computes each value)
- The Phase 1 scope boundary and what the annotation controls
- The dependency relationship with VEP 115 and VEP 266

Sections like the full domain XML examples, PCIe switch topology mirroring details, EGM size partitioning arithmetic, and multiple VMI API variants could be removed or reduced to a single representative example each. Those details are valuable but belong in follow-up design docs or PR descriptions rather than the VEP itself.

A shorter VEP would also help resolve the internal consistency issues flagged in my inline comments (the GI node discovery and NUMA distance sections that still describe fixed synthesis despite the inference-first principle) since there would be fewer places where the approach needs to be stated consistently.

Happy to discuss further on the AIE WG call if that would help.

lyarwood (commented):
Thanks for the continued iteration on this @fanzhangio. The VEP reads much better at 468 lines with the inference-first approach stated consistently throughout. The scope narrowing, clean VEP 115/266 factoring, and removal of the `graceVirtualization` annotation are all good improvements.

I have some design and process comments below — mostly around areas where the VEP leaves key decisions undefined or where dependency risks should be acknowledged.

lyarwood (commented):
Thanks for the continued iteration on this @fanzhangio. The VEP reads much better at 468 lines with the inference-first approach stated consistently throughout. The scope narrowing, clean VEP 115/266 factoring, and removal of the `graceVirtualization` annotation are all good improvements.

I have some design and process comments below — mostly around areas where the VEP leaves key decisions undefined or where dependency risks should be acknowledged.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,468 @@
...
+
+This VEP introduces a small, **inference-first** baseline that lets KubeVirt run NVIDIA Grace Hopper and Grace Blackwell (GB200/GB300) GPU passthrough VMs on ARM64.
+
+It is gated by a new `GraceIOVirtualization` feature gate and is intentionally limited to what has already been validated in the [kubevirt-aie](https://github.com/kubevirt/kubevirt-aie) fork.
+
+The Phase 1 contract is:
+
+- A new `GraceIOVirtualization` feature gate controlling all Grace-specific domain conversion behavior.
+- Explicit dependency on `PCINUMAAwareTopology` ([VEP 115](https://github.com/kubevirt/enhancements/issues/115)) for NUMA-aware PCI placement.
+- Explicit dependency on the generic IOMMUFD path defined by [VEP 266](https://github.com/kubevirt/enhancements/pull/267) for host device binding and FD delivery.
```
lyarwood: Nit: This says "VEP 266" but links to PR #267. Consider linking to the tracking issue https://github.com/kubevirt/enhancements/issues/266 instead, or clarifying the numbering, to avoid confusion.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,468 @@
...
+## Non Goals
+
+- Defining a user-visible API for Grace topology (GI cell counts, NUMA distances, MMIO sizing, per-GPU PCI placement). All of these are inferred from the host.
+- EGM (Extended GPU Memory) backing for guest memory.
+- vCMDQ hardware command queue virtualization.
+- PCIe link speed and width modeling on guest root ports.
+- PCIe switch topology mirroring (e.g. ConnectX/BlueField/NVMe behind a switch).
+- Mixed GPU + SR-IOV NIC topologies with independent isolation policies.
+- Per-GPU PXB isolation beyond what is already validated by the kubevirt-aie baseline.
+- Generic large-BAR `pcihole64` API. This VEP does not introduce an override option to escape auto-configuration, see [Large-BAR MMIO Aperture (auto-sized)] (#large-bar-mmio-aperture-auto-sized).
```
lyarwood: Nit: The space between `]` and `(` breaks this markdown link. Should be:
```
see [Large-BAR MMIO Aperture (auto-sized)](#large-bar-mmio-aperture-auto-sized)
```

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,468 @@
...
+- **IOMMUFD** ([VEP 266](https://github.com/kubevirt/enhancements/pull/267)) must be available for Grace host-device binding. VEP 199 consumes the generic device plugin, FD delivery, and domain-level `<iommufd>` behavior from VEP 266; it does not redefine those mechanisms.
+
+Admission rejects statically invalid Grace VMIs, such as missing required feature gates or unsupported architecture. Node-local details, including actual GI ranges, distance vectors, BAR sizes, and IOMMUFD FD availability, are validated by `virt-launcher` during domain conversion and must fail fast with actionable errors when unavailable.
+
+### Activation Surface
+
+There is **no Grace-specific user-facing VMI API** in Phase 1. Grace handling is activated when all of the following are true:
+
+1. `GraceIOVirtualization` is enabled and the required VEP 115 / VEP 266 prerequisites are configured.
+2. The VMI requests one or more Grace-class GPU host devices through the existing host-device API.
```
lyarwood, line 151: The VEP repeatedly references "Grace-class GPU host devices" (here, in the Design Principles table, and in the admission table at line 265) but never defines the classification mechanism.

Is this by PCI vendor/device ID? By device plugin resource name pattern (e.g. `nvidia.com/GB100_*`)? By a node label? Even if the exact ID list is an implementation detail, the mechanism for classification should be specified - it determines how the feature gate actually activates in practice.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,468 @@
...
+## SMMUv3 Emission
+
+When Grace handling is active, KubeVirt emits one libvirt `<iommu model='smmuv3'>` element for each guest PCIe bus that carries a passed-through Grace GPU. The `pciBus` value is derived from the guest PCI bus or controller allocated by the VEP 115 NUMA-aware placement logic; VEP 199 does
+not prescribe a numbering scheme.
+
+```xml
+<devices>
+  <iommu model='smmuv3'>
+    <driver pciBus='...' accel='on' ats='on' ril='off'
+            ssidSize='20' oas='48'/>
```
lyarwood: These SMMUv3 attribute values are presented without justification. Are they Grace platform constants or should some be inferred from host IOMMU capabilities (following the inference-first principle)?

In earlier discussion you mentioned that upstream QEMU may adopt `auto` defaults for `ats`, `ril`, and `oas`. The VEP should clarify whether these are temporary fixed values for Phase 1 alpha that will move to auto-detection, or whether they are intentionally part of the Grace guest contract.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,468 @@
...
+
+### Feature Gates and Dependencies
+
+Phase 1 introduces a single feature gate:
+
+- **`GraceIOVirtualization`** (alpha, off by default) gates all Grace-specific domain conversion behavior described in this VEP. `GraceIOVirtualization` does not implicitly enable any other feature gate. Where Grace depends on existing KubeVirt functionality, those dependencies must be enabled explicitly:
+
+- **`PCINUMAAwareTopology`** ([VEP 115](https://github.com/kubevirt/enhancements/issues/115)) must be enabled. Grace reuses the generic NUMA-aware PCI placement behavior and does not introduce a separate placement planner in Phase 1.
+
+- **IOMMUFD** ([VEP 266](https://github.com/kubevirt/enhancements/pull/267)) must be available for Grace host-device binding. VEP 199 consumes the generic device plugin, FD delivery, and domain-level `<iommufd>` behavior from VEP 266; it does not redefine those mechanisms.
```
lyarwood: Both VEP 115 and VEP 266 are listed as hard dependencies. Could you add a brief note on their current status?

lyarwood (commented):
Thanks for the continued iteration on this @fanzhangio! The VEP reads much better at 468 lines with the inference-first approach stated consistently throughout. The scope narrowing, clean VEP 115/266 factoring, and removal of the graceVirtualization annotation are all good improvements.

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+The Phase 1 implementation may use platform-aware safety margins or temporary Grace-specific floors while the generic sizing algorithm is refined. These details are intentionally implementation-owned and are not exposed to users as VMI fields or annotations.
+
+If alpha users find a supported Grace topology where the inferred value is insufficient, the preferred fix is to improve KubeVirt's inference logic. A generic user override, if still required after alpha feedback, should be proposed in a follow-up large-BAR VEP rather than added to VEP 199.
+
+---
+
+## Admission Validation
+
+When `GraceIOVirtualization` is enabled, the admission webhook applies the following Phase 1 checks to VMIs that include Grace-class GPU host devices:
```
fossedihelm: Will there be a list of "device names" to achieve this?

fossedihelm: If yes, these names are strictly related to nvidia-device-plugin, no?
Should we mention this dependency somewhere?

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
```diff
@@ -0,0 +1,483 @@
...
+
+| Check | Reason |
+| --- | --- |
+| `PCINUMAAwareTopology` feature gate is enabled | Grace placement reuses the VEP 115 planner; without it, NUMA-aware PCI topology is unavailable. |
+| IOMMUFD prerequisites are configured at the cluster level | Grace passthrough requires the VEP 266 IOMMUFD path. |
+| `spec.architecture` is `arm64` and `spec.domain.machine.type` is `virt` | Grace is ARM64-only and uses the `virt` machine type. |
+| `dedicatedCpuPlacement` is `true` and `numa.guestMappingPassthrough` is set | Required so the planner can build deterministic guest NUMA cells. |
```
fossedihelm: do we need to mention `cpu.model: host-passthrough` as well?

fossedihelm (commented):
Thank you! Small comments below, otherwise it's quite fine for me.
Generic question: It's of course implicit that migration is not possible.
Do we need to explicitly say it? I leave it up to you. Thanks

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Good point. I think `cpu.model: host-passthrough` would not be as a hard Grace admission requirement unless a validated implementation failure without it.
Removed it

veps/sig-compute/199-nvidia-grace-virtualization-support/199-nvidia-grace-virtualization-support.md
fanzhangio: Good point. I will clarify this in the VEP.

VEP-199 does not define a fixed allow-list of `deviceName` strings, and it does not classify Grace devices from NVIDIA resource-name patterns for example `nvidia.com/GB100_*`.

`deviceName` remains the existing KubeVirt host-device resource handle. The classification source of truth is the PCI vendor/device ID associated with that resource, for example through KubeVirt `permittedHostDevices.pciHostDevices[].pciVendorSelector` / `resourceName`. During domain conversion, `virt-launcher` also verifies the assigned BDF through sysfs before applying Grace-specific topology.

The resource may be exposed by KubeVirt's generic PCI device plugin or by an external provider such as NVIDIA's KubeVirt GPU device plugin, but VEP-199 does not depend on NVIDIA-specific resource-name strings. Node labels are only for scheduling or operator policy; they are not the Grace classification mechanism.

fanzhangio:
Thanks @fossedihelm , I agree it is better to state this explicitly. I will keep live migration out of the Phase 1 contract and add a short note that Grace VMIs are not migratable in this VEP.

The reason is that Phase 1 guest construction depends on node-local host state: assigned PCI BDFs, SMMUv3/IOMMUFD setup, GI node discovery, NUMA distances, and large-BAR sizing. VEP 199 does not define compatibility checks, state transfer, or destination reconstruction for those resources. I've stated `Live migration` in out-of-scope. I just expanded the statement.

fossedihelm:
Thank you @fanzhangio!
It looks good from my side
/lgtm
@vladikr  can you take a look?

vladikr:
> Thank you @fanzhangio! It looks good from my side /lgtm @vladikr can you take a look?

Yes, thank you @fanzhangio Looks good from my side as well. This is now in very good shape, thank you for that!
The VEP has moved strongly toward the inference-first model.
/approve

Issue #140, opened 2025-11-24
VEP #140: Your short, descriptive title


frenzyfriday:
**Primary contact (assignee)**:

/assign @frenzyfriday 

**Current Feature Stage**:
New

**Feature Gate**:
LiveUpdateNadRef

**Responsible SIGs**:

Primary SIG:
/sig network

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/138 

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

orelmisan:
Hi @frenzyfriday, Please update the issue with all the relevant PRs so we could close it.
Including https://github.com/kubevirt/user-guide/pull/992.

orelmisan:
+1 for closing this as done for v1.9.
@EdDev do you agree?

PR #166, opened 2025-12-22
Updates to VEP 140 and GA NAD hotplug feature

frenzyfriday:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/140
**SIG label**: /SIG network

### What this PR does
- We dont want to restrict the NAD swap just to bridge bindings, but mention
that only scenarios with bridge binding will be tested.
- Explain why Alpha phase is skipped
- Amend the section about annotation patching
- Choosing Option 1 as the preferred implementation
- Proposes GA for the feature in v1.9

### Special notes for your reviewer

Addresses: https://github.com/kubevirt/enhancements/pull/138#pullrequestreview-3601612221

PR #271, opened 2026-04-16
VEP 140: GA NAD live update

frenzyfriday:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/140
**SIG label**: /sig NETWORK

### What this PR does
GA NAD live update

We have `considered other bindings` as a part of the GAing process. The implementation of this NAD hotplug itself if binding agnostic. 
`masquerade` and `passt` bindings are supported for the primary network interface and is hence irrelevant for this feature. Hotplugging NAD refs of SRIOV interfaces should be possible provided the resource is present on the node the VM migrates to. Hotplugging NADs of  Ifaces with binding plugins like macvtap will be possible as well as they do not require special resouces.
The e2e test right now covers only the bridge binding.  We can add other binding/plugins tests if there are more user requests.

### DIFF
--- a/veps/sig-network/hotpluggable-nad-ref.md
+++ b/veps/sig-network/hotpluggable-nad-ref.md
@@ -6,7 +6,7 @@
 
 - ~This VEP targets alpha for version:~
 - This VEP targets beta for version: v1.8
-- This VEP targets GA for version:
+- This VEP targets GA for version: v1.9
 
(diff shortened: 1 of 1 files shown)


frenzyfriday:
Merging this with https://github.com/kubevirt/enhancements/pull/166

PR #272, opened 2026-04-16
VEP 17: Instancetype API v1 graduation

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/17
**SIG label**: /sig compute

### What this PR does

Graduates VEP 17 to a straightforward v1beta1 -> v1 promotion based on
production stability across multiple releases. Custom sizing, optional
CPU/Memory, and resource visibility have been dropped from the scope. No
behavioral or schema changes are introduced - only the version promotion and
removal of previously deprecated fields.

The migration follows a 4-release cycle:

- **v1.9.0**: Introduce `v1`, deprecate `v1beta1`
- **v1.10.0**: Change storage and preferred version to `v1`
- **v1.11.0**: Deprecation window (`v1beta1` continues to be served)
- **v1.12.0**: Remove `v1beta1`

This follows Kubernetes Deprecation Policy Rules #4a and #4b and the
API Changes Guide (one release between storage version change and old
version removal for `kube-storage-version-migrator`).

### DIFF
--- a/veps/sig-compute/17-instancetype-v1/instancetype-v1.md
+++ b/veps/sig-compute/17-instancetype-v1/instancetype-v1.md
@@ -1,0 +1,227 @@
+# VEP #17: Instancetype API v1 Graduation
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v0.56.0
+- This VEP targets beta for version: v1.0.0
+- This VEP targets GA for version: v1.9.0
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone /
+release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in
+  [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [x] (R) Beta target version is explicitly mentioned and approved
+- [x] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP graduates the `instancetype.kubevirt.io` API group from `v1beta1` to
+`v1`. The instancetype API has been beta since v1.0.0, has proven stable in
+production across multiple releases, and is ready for GA. No behavioral or
+schema changes are introduced - this is a straightforward API version promotion
+with removal of previously deprecated fields.
+
+The migration follows a 4-release cycle:
+
+- **v1.9.0**: Introduce `v1`, deprecate `v1beta1`
+- **v1.10.0**: Change storage and preferred version to `v1`
+- **v1.11.0**: `v1beta1` continues to be served (deprecation window)
+- **v1.12.0**: Remove `v1beta1`
+
+This graduation plan follows
+[Rule #4a](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api)
+(beta versions served for 3 releases after deprecation) and
+[Rule #4b](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api)
+(preferred/storage must not advance until both versions are served) of the
+[Kubernetes Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/),
+as well as the
+[Kubernetes API Changes Guide](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas)
+(at least one release between storage version change and old version removal
+for [`kube-storage-version-migrator`](https://github.com/kubernetes-sigs/kube-storage-version-migrator)).
+
+## Motivation
+
+Graduating to `v1` signals API stability and allows the project to remove the
+beta version over time.
+
+## Goals
+
+- Graduate `instancetype.kubevirt.io` from `v1beta1` to `v1`
+- Remove previously deprecated fields from the `v1` API surface
+- Maintain full backward compatibility throughout the migration
+
+## Non Goals
+
+- Changing instancetype application semantics (conflict behavior stays as-is)
+- Making `CPU` or `Memory` optional in the instancetype spec
+- Surfacing applied resources in VM status
+- Merging instancetypes and preferences into a single resource type
+
+## Definition of Users
+
+- **VM User**: Creates VirtualMachines and references instancetypes.
+- **Cluster Admin**: Defines instancetypes as organizational standards or cloud
+  offerings.
+
+## User Stories
+
+- As a VM user, I want the instancetype API to be stable (`v1`), so I can rely
+  on it without worrying about breaking changes in future releases.
+
+- As a VM user, I want to gradually migrate from `v1beta1` to `v1` with clear
+  deprecation signals and no surprises in behavior.
+
+- As a cluster admin, I want a clear migration timeline so I can plan upgrades
+  and run storage migration tooling at the right time.
+
+## Repos
+
+- `kubevirt/kubevirt`
+- `kubevirt/common-instancetypes`
+
+## Design
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,277 @@
...
+
+**Version serving strategy:**
+
+- `v1beta1`: **removed**
+- `v1`: served, **storage version** (only version)
+
+### ControllerRevision Handling
+
+ControllerRevisions storing instancetype snapshots are upgraded to `v1` on use
+starting with v1.11.0 (when `v1beta1` is removed). Before that, existing
```
lyarwood: Shouldn't this happen in v1.10.0 while we still have support for the API structs in-tree?

0xFelix: Yep, updated the VEP because we actually need a 4-release cycle to satisfy the K8S requirements.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,289 @@
...
+  downgrade to pre-v1.9.0), deprecated
+- `v1`: served
+
+Conversion webhooks handle `v1beta1` <-> `v1` translation:
+
+- `v1beta1` -> `v1`: Deprecated `PreferredCPUTopology` values mapped to short
+  forms. Deprecated firmware fields mapped to `PreferredEfi`.
+- `v1` -> `v1beta1`: No mapping needed for `PreferredCPUTopology` - the short
+  forms (`cores`, `sockets`, `threads`, `spread`, `any`) are already valid
+  `v1beta1` values alongside the deprecated ones.
```
lyarwood: The v1beta1 -> v1 conversion direction describes mapping for both `PreferredCPUTopology` and the deprecated firmware fields, but the v1 -> v1beta1 direction only addresses `PreferredCPUTopology`. What happens when a v1 object with `PreferredEfi` is converted back to v1beta1? Should the deprecated `DeprecatedPreferredUseEfi` / `DeprecatedPreferredUseSecureBoot` fields be populated for backward compatibility with older clients reading via v1beta1?

If no mapping is needed because the v1beta1 schema also has `PreferredEfi` alongside the deprecated fields, it would be worth stating that explicitly here.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,289 @@
...
+- **v1.10.x -> v1.11.0**: No changes. `v1beta1` continues to be served
+  (deprecation window).
+- **v1.11.x -> v1.12.0**: `v1beta1` removed (migration must be complete).
+
+### Rollback considerations
+
+Rollback to any prior release is safe since each release can read the storage
+version of its successor. The only caveat is that rollback from v1.12.0
+requires that `kube-storage-version-migrator` completed the v1beta1 -> v1
+migration before the upgrade, since v1.12.0 removes `v1beta1`.
```
lyarwood: "Rollback to any prior release" is overstated - this is only safe for rollback to the immediately preceding release. For example, rolling back from v1.10.0 (storage: v1) to v1.8.x (which does not serve v1) would break since v1.8.x cannot read v1-stored objects.

Consider saying "Rollback to the immediately preceding release is safe" and briefly noting which rollback paths are valid at each phase.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,289 @@
...
+- [ ] Unit and functional tests for conversion and backward compatibility
+- [ ] Storage version: `v1beta1`, preferred version: `v1beta1`
+
+### v1.10.0
+
+- [ ] Storage version: `v1`, preferred version: `v1`
+- [ ] ControllerRevisions re-encoded to `v1` on use
+- [ ] Documentation recommends deploying `kube-storage-version-migrator`
+
+### v1.12.0
```
lyarwood, line 224: The graduation requirements jump from v1.10.0 to v1.12.0. Since v1.11.0 is an explicit phase in the 4-release cycle (the deprecation window per Rule #4a), it should have its own entry here for consistency - even if minimal (e.g., confirming v1beta1 is still served, verifying storage migration progress).

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,289 @@
...
+
+- `v1beta1`: **removed**
+- `v1`: served, **storage version** (only version)
+
+### ControllerRevision Handling
+
+ControllerRevisions storing instancetype snapshots are not managed by
+`kube-storage-version-migrator` since they contain serialized objects as opaque
+data. Starting with v1.10.0 (when storage moves to `v1`), ControllerRevisions
+are re-encoded from `v1beta1` to `v1` on use. This gives two full releases
```
lyarwood: "On use" is vague here. Looking at the existing implementation in `pkg/instancetype/upgrade/handler.go`, re-encoding is triggered during VM reconciliation in virt-controller (before VMI creation). Consider saying "during VM reconciliation" instead of "on use" so implementors and reviewers have a precise understanding of the trigger.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Graduation Requirements
+
+### v1.9.0
+
+- [ ] `v1` API types defined (same schema as `v1beta1`, deprecated fields
+  removed)
+- [ ] Conversion webhooks between `v1beta1` <-> `v1`
+- [ ] `v1beta1` marked as deprecated in API documentation
+- [ ] `common-instancetypes` updated to use `v1`
```
lyarwood: In v1.9.0, v1beta1 is still the preferred and storage version. Is this the intended timing for updating common-instancetypes to v1, or should it move to the v1.10.0 requirements when v1 becomes preferred?

lyarwood (commented):
> The migration follows a 3-release cycle: introduce v1 (v1.9.0), change storage (v1.10.0), remove v1beta1 (v1.11.0).

This still describes the old 3-release plan. The VEP content now has a 4-release cycle (v1.9.0, v1.10.0, v1.11.0 deprecation window, v1.12.0 removal).

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
0xFelix: Stated that `PreferredEfi` will not be converted back to the deprecated fields.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
0xFelix: Replaced with table describing this more precisely.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
0xFelix: I left it out deliberately, since there are no real action items for v1.11.0. There will be no changes in that release.

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
0xFelix: Updated to say "during VM reconciliation"

veps/sig-compute/17-instancetype-v1/instancetype-v1.md
0xFelix: Since v1 is already served in v1.9.0 IMO we are good to move common-instancetypes to v1 too

Issue #285, opened 2026-04-21
VEP 285: Preset launch security defaults via preference


jschintag:
**Primary contact (assignee)**:

/assign @jschintag 

**Current Feature Stage**:

New

**Feature Gate**:

TBD

**Responsible SIGs**:

Primary SIG:
/sig api

Additional SIGs (optional):
N/A

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
      * [ ] https://github.com/kubevirt/enhancements/pull/286
  * [ ] Code PRs:
      * [ ] https://github.com/kubevirt/kubevirt/pull/17551
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

fossedihelm:
@jschintag Can this be considered completed at all?
I see that you set Alpha in this tracker, but there is no FG.
Thank you

jschintag:
@fossedihelm From my perspective the alpha target has been met since all implementation goals for the phase have been merged and the feature is stable.
A separate Feature Gate for this was not planned, since the API itself is not yet GA and for launchSecurity only s390x is going GA with 1.9, so an additional FG here would serve little purpose.

I'm currently going through my VEPs and i will post an update PR for the document soon.

fossedihelm:
@jschintag So this is a sort of sub-VEP of something else?

jschintag:
> [@jschintag](https://github.com/jschintag) So this is a sort of sub-VEP of something else?

Yes kind of.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

p.s. even if the VEP is currently marked as "Tracked", a "VEP Reviewer" is a requirement. Without it, the VEP will be not remain tracked for v1.10.

jschintag:
@iholder101 Is the reviewer still required even though the code changes for 1.10 already merged since they where for the common-instancetypes?

PR #286, opened 2026-04-21
VEP 285: Preset launch security defaults via preference

jschintag:
This VEP proposes to enable easier management of confidential computing VMs by using common preferences.

### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/285
**SIG label**: /sig api

### What this PR does

Propose adding launchSecurity field to VirtualMachinePreference for configuring VMs.

### DIFF
--- a/veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
+++ b/veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
@@ -1,0 +1,224 @@
+# VEP #285: Preset launch security defaults via preference
+
+## VEP Status Metadata
+
+### Target releases
+
+<!--
+A PR must update this section during the planning phase of a given release in order to track it.
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
+For example, during the planning phase for version v1.123, do **not** target beta for v.124 in advance.
+-->
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add the `launchSecurity` parameter to `VirtualMachineClusterPreference` and
+`VirtualMachinePreference` to allow users to specify launch security settings for multiple VMs
+while still allowing overrides on a per-VM basis.
+Additionally remove `launchSecurity` from `VirtualMachineInstanceSpec`, as it does not make sense there.
+
+## Motivation
+
+Confidential Computing workloads are typically prepared from specific images created by cluster
+administrators with additional security measures.
+Since VMs started from these images will all run with launch security enabled, this allows cluster
+administrators to provide the image templates as a `DataSource` and specify common `launchSecurity`
+settings via preferences.
+Users can then fine-tune these settings for specific VM instances as needed.
+
+## Goals
+
+- Enable launch security settings via preferences
+- Improve the workflow of creating Confidential Computing VMs from common instance types
+- Prepare common preference types for IBM Secure Execution
+- Deprecate launch security setting from the `VirtualMachineInstanceSpec`
+
+## Non Goals
+
+- Block users from customizing the launch security settings per VM
+- Setting global defaults for launch security
+- Remove launch security setting from the `VirtualMachineInstanceSpec`, as it would require a new api version
+
+## Definition of Users
+
+**Cluster Administrators:**
+- Install the Kubernetes cluster and KubeVirt
+- Enable required features
+- Prepare common VM image sources
+- Perform other tasks related to cluster management, e.g., creating namespaces, providing storage,
+  access management, etc.
+
+**Users:**
+- Create VMs from prepared VM images
+- Manage the lifecycle of VMs
+- (Optional) Own the workload running inside the VM
+
+## User Stories
+
+**As a cluster administrator, I want to:**
+- Easily enable Confidential Computing in my cluster
+- Not spend additional effort leveraging Confidential Computing
+- Easily enable users to start Confidential Computing VMs
+
+**As a user, I want to:**
+- Easily create Confidential Computing VMs
+- Easily run Confidential Computing workloads
+- Spend no additional effort running Confidential Computing VMs compared to normal VMs
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+- https://github.com/kubevirt/api
+- https://github.com/kubevirt/common-instancetypes
+
+## Design
+
+For implementation details, see the [PoC PR](https://github.com/kubevirt/kubevirt/pull/17551).
+
+Add the `launchSecurity` field as `preferredLaunchSecurity` to `VirtualMachinePreferenceSpec` the same way as is already done
+for `VirtualMachineInstancetypeSpec`:
+
+```go
+type VirtualMachinePreferenceSpec struct {
+  ...
+
+	// Optionally defines the preferred LaunchSecurity
+	//
+	// +optional
+	PreferredLaunchSecurity *v1.LaunchSecurity `json:"preferredLaunchSecurity,omitempty"`
+}
+```
+
+When applying preferences to a new VMI, follow the same pattern as for other settings:
+
+1. If the `launchSecurity` field in the VMI is set, do nothing
+2. If the `launchSecurity` field in the preference is set, set the `launchSecurity` field in the
+   VMI to the value from the preference and return
+3. If neither the VMI nor the preference have a `launchSecurity` field, do nothing
+4. Settings defined in the choosen instancetype override everything else
+
+The following table summarizes the outcomes:
+
+| VMI   | Preference | Instancetype | Outcome      |
+| ----- | ---------- | ------------ | ------------ |
+| empty | empty      | empty        | empty        |
+| set   | empty      | empty        | VMI          |
+| set   | set        | empty        | VMI          |
+| empty | set        | empty        | Preference   |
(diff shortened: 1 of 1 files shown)


veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,212 @@
...
+Refer to https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#releases for more details
+-->
+
+### Alpha
+- [ ] API changes are implemented and available
+- [ ] New preference field is applied when rendering VMI
+- [ ] Unit tests are implemented and passing
+
+### Beta
+- [ ] Common preference types are added to common-templates for IBM Secure Execution
```
0xFelix: You mean `common-instancetypes`.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,212 @@
...
+When applying preferences to a new VMI, follow the same pattern as for other settings:
+
+1. If the `launchSecurity` field in the VMI is set, do nothing
+2. If the `launchSecurity` field in the preference is set, set the `launchSecurity` field in the
+   VMI to the value from the preference and return
+3. If neither the VMI nor the preference have a `launchSecurity` field, do nothing
+
+The following table summarizes the outcomes:
+
+| VMI     | Preference | Outcome    |
```
0xFelix: How does it go together with instancetypes?

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,212 @@
...
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add the `launchSecurity` parameter to `VirtualMachineClusterPreference` and
```
0xFelix, line 31: IIRC the intention for providing `launcheSecurity` with instance types is that nodes need to have the necessary hardware i.e. the resources to support the chosen setting. It's not simply a runtime preference of the guest.

0xFelix (commented):
At least we've got a conflict with #17.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: The hardware support for nodes is determined via nodeselectors (at least for AMD SEV and IBM Secure Execution, no idea about Interl/Arm).
The idea here is not related to the Hardware in the cluster, but rather to specific VM images.
For example IBM Secure Execution requires specific changes to the VM image, which would mean that image can only run in SE mode. If that is then used as a golden image in the cluster, linking it to a preference which means `launchSecurity` would be enabled automatically. No need for users to choose an instancetype with launchSecurity or enable it manually.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: Good catch

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: As far as i understand instancetypes override settings in the VMI, so it would override it here as well.
Let me add that here as well.

jschintag:
Thank you for the quick review @0xFelix.

I have updated the table to include instancetypes and fixed the mention of common-templates.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,217 @@
...
+for `VirtualMachineInstancetypeSpec`:
+
+```go
+type VirtualMachinePreferenceSpec struct {
+  ...
+
+	// Optionally defines the preferred LaunchSecurity
+	//
+	// +optional
+	LaunchSecurity *v1.LaunchSecurity `json:"launchSecurity,omitempty"`
```
lyarwood: I'm not a fan of it but everything is prefixed with `preferred`, I've wanted to drop it forever but the backwards compat required would be a nightmare.

@0xFelix if we do delay v1 maybe yet another thing to throw in the backlog?

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,217 @@
...
+
+| VMI   | Preference | Instancetype | Outcome      |
+| ----- | ---------- | ------------ | ------------ |
+| empty | empty      | empty        | empty        |
+| set   | empty      | empty        | VMI          |
+| set   | set        | empty        | VMI          |
+| empty | set        | empty        | Preference   |
+| empty | empty      | set          | Instancetype |
+| set   | empty      | set          | Instancetype |
+| set   | set        | set          | Instancetype |
```
lyarwood: nit - Both of these should conflict

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
lyarwood: The more I think about this the more I think we need to deprecate LaunchSecurity out of the instance type spec entirely and just provide it through preferences.

@0xFelix this will bork the v1 VEP for a cycle, would you be against this?

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: Good catch

0xFelix:
> @0xFelix this will bork the v1 VEP for a cycle, would you be against this?

I am fine with that, let's refine the instanctype API first and not rush v1 in this cycle.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
0xFelix: It's hard to draw a line on deciding what belongs into an instancetype or a preference. But I agree that we should move launchSecurity to preferences.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
0xFelix: @lyarwood Do you want this to keep using the `preferred` prefix? But I agree we could drop it in the future (even this cycle already?).

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
lyarwood: okay, @jschintag can you deprecate the instance type field as part of this work in v1beta1?

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
lyarwood: I'm pretty overloaded this cycle but I can write it up as a backlog VEP and we can try to address it either this cycle or next.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: Sure @lyarwood. I have added the removal to the beta target and added a deprecation notice to the alpha target.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
lyarwood: @jschintag for now can you prefix this with `preferred`.

lyarwood (commented):
quick nit otherwise this lgtm for alpha in v1.9.0 and we can address the prefix issue in a later release.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
jschintag: @lyarwood done

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,223 @@
...
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add the `launchSecurity` parameter to `VirtualMachineClusterPreference` and
+`VirtualMachinePreference` to allow users to specify launch security settings for multiple VMs
+while still allowing overrides on a per-VM basis.
+Additionally remove `launchSecurity` from `VirtualMachineInstanceSpec`, as it does not make sense there.
```
0xFelix, line 34: Remove or deprecate? Down below you only mention deprecating it.

Keep in mind that if we want to remove a field from the CRD, we need a new apiVersion. I suggest deprecating but also keeping it for now.

veps/sig-compute/285-launchsecurity-preference/launchsecurity-preference.md
```diff
@@ -0,0 +1,223 @@
...
+  preferredArchitecture: "s390x"
+  requirements:
+    cpu:
+      guest: 1
+    memory:
+      guest: 2Gi
+  preferredLaunchSecurity: {}
+```
+
+The removal of `launchSecurity` from `VirtualMachineInstanceSpec` follows in stages, starting with
```
0xFelix: That means you want to introduce a new apiVersion? Also, will warnings be emitted when the deprecated field is used?

jschintag: I think the removal would happen with v1, assuming that is not done in this release?
Otherwise it would only be deprecated for now, since it does not make sense to create a new API Version for it.

jschintag: I have updated the VEP to only include deprecating it via a warning, while leaving the functionality intact to ensure backwards compatibility.

0xFelix (approved):
/lgtm

Thanks!

jschintag:
@vladikr or @xpivarc could you PTAL?

vladikr:
/approve
Looks good to me for alpha.

Issue #287, opened 2026-04-21
VEP 287: TLS Group Preferences


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood

**Current Feature Stage**:

New

**Feature Gate**:

TLSGroupPreferences

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

* VEP: [veps/sig-compute/287-tls-group-preferences](https://github.com/kubevirt/enhancements/tree/main/veps/sig-compute/287-tls-group-preferences)
* VEP PR: https://github.com/kubevirt/enhancements/pull/288

**Timeline**:

* Alpha release target (v1.9.0):
* Beta release target:
* Stable release target:

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/288
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/17553
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@lyarwood so what is going to happen in 1.10? :)

lyarwood:
@Barakmor1 is taking this effort over moving forward

vladikr:
@Barakmor1 @lyarwood can one of you please update the status here?

lyarwood:
Status update — retargeting this to **v1.10.0 alpha**.

@Barakmor1 is now driving this effort however is out for a while ahead of VEP freeze so I'm going to try to move this forward. The VEP has been updated to:

- **Retarget alpha to v1.10.0** (was v1.9.0). This aligns with the CNV 5.1 / OpenShift TLS-curves mandate (VIRTSTRAT-618), which requires every TLS endpoint in every container to support configurable groups. HCO needs a released KubeVirt API (even alpha) to reconcile from, so landing the alpha in v1.10 is the priority.
- **Switch the API to a typed `Groups []TLSGroup` enum** (was `Groups []string`) to mirror OpenShift's TLS security profile API that HCO reconciles from.

Retarget VEP PR to follow.

> [!NOTE]
> The typed-enum switch reverses the design choice in the merged VEP and diverges from KubeVirt's existing `Ciphers []string` precedent — flagging for SIG / @Barakmor1 agreement before it's locked in. It also means the alpha code PR (kubevirt/kubevirt#17553), which currently implements `[]string`, needs rework.

/remove-lifecycle stale

lyarwood:
[...]

PR #288, opened 2026-04-21
VEP #287: TLS Group Preferences

lyarwood:
## Summary

- Adds a `Groups` field to KubeVirt's `TLSConfiguration` API for configuring TLS supported groups (elliptic curves) across all virt pod endpoints
- Enables Post-Quantum Cryptography (PQC) readiness via hybrid key exchange groups such as `X25519MLKEM768`
- Gated behind `TLSGroupPreferences` feature gate, targeting Alpha for v1.9.0

Tracking issue: https://github.com/kubevirt/enhancements/issues/287

/sig compute
/sig control-plane

### DIFF
--- a/veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
+++ b/veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
@@ -1,0 +1,445 @@
+# VEP #287: TLS Group Preferences
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue [#287](https://github.com/kubevirt/enhancements/issues/287) created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [X] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP adds a `Groups` field to KubeVirt's `TLSConfiguration` API, allowing
+cluster administrators to configure TLS supported groups (elliptic curves)
+negotiated during TLS handshakes across all virt pod endpoints. This enables
+Post-Quantum Cryptography (PQC) readiness by supporting hybrid key exchange
+groups such as `X25519MLKEM768`.
+
+## Motivation
+
+The transition to Post-Quantum Cryptography (PQC) is underway. NIST has
+finalised its first set of PQC standards and IETF has standardised hybrid key
+exchange groups such as `X25519MLKEM768` that combine classical and
+post-quantum algorithms. Go 1.24 (which KubeVirt already uses) includes native
+support for `X25519MLKEM768` as a `tls.CurveID`.
+
+KubeVirt's existing `TLSConfiguration` supports `MinTLSVersion` and `Ciphers`
+but has no mechanism to control which key exchange groups (elliptic curves) are
+offered during TLS negotiation. Without this, KubeVirt components use Go's
+default curve preferences, which means administrators cannot:
+
+- Enforce specific PQC-ready groups across all virt pod endpoints
+- Restrict curves to a known-safe set for compliance or FIPS requirements
+- Align KubeVirt's TLS behaviour with platform-wide or organisational TLS
+  policies
+
+### Why Not Automatically Enable PQC Groups?
+
+Go 1.24 already includes `X25519MLKEM768` in its default curve preferences, so
+KubeVirt components negotiate PQC key exchange automatically when both sides
+support it. However, an explicit API field is still needed because:
+
+- **Compliance restrictions**: PQC groups such as `X25519MLKEM768` are not yet
+  FIPS-approved. Deployments with FIPS requirements must be able to explicitly
+  exclude them.
+- **Performance trade-offs**: PQC key exchanges have larger key sizes and
+  higher computational cost. Administrators may need to control which groups
+  are offered based on their workload profile.
+- **Platform alignment**: Kubernetes platforms are beginning to expose TLS
+  group configuration in their cluster-wide TLS profiles. KubeVirt needs a
+  corresponding API so that meta-operators can propagate these settings into
+  the KubeVirt CR, ensuring all virt pod endpoints participate in cluster-wide
+  TLS policy rollouts.
+
+## Goals
+
+- Extend `TLSConfiguration` with a `Groups` field to configure TLS supported
+  groups
+- Apply the configured groups as `CurvePreferences` on all TLS server endpoints
+  in virt-api, virt-controller, virt-handler, virt-operator,
+  virt-exportproxy, virt-exportserver, virt-synchronization-controller,
+  virt-template-apiserver, and virt-template-controller
+- Validate group names in the KubeVirt CR admission webhook
+- Maintain backward compatibility — omitting `Groups` preserves current Go
+  default behaviour
+
+## Non Goals
+
+- Watching platform-specific TLS profile resources directly from virt-operator
+  — meta-operators (e.g. HCO) handle this translation
+- Configuring TLS groups on client-side connections — this VEP covers
+  server-side TLS endpoints only. Server-side configuration is sufficient to
+  control which group is actually negotiated: the server selects the group
+  from the client's offered set based on its `CurvePreferences`. Since all
+  KubeVirt components are Go programs, the client side already offers Go's
+  full default set (including PQC groups in Go 1.24+), so the server can
+  restrict or prioritise groups without client-side changes. Client-side
+  configuration would only be needed to control what appears in the
+  ClientHello itself (e.g. to avoid advertising PQC groups for strict
+  compliance, or to eliminate HelloRetryRequest round trips when the server
+  does not support a group the client sent a key share for).
+- Configuring certificate key curves (e.g. changing the hardcoded P-256 in
+  `pkg/certificates/triple/cert/cert.go`) — the certificate signing key is
+  independent of the key exchange groups negotiated during the TLS handshake.
+  This would require a separate API field and should be a separate VEP.
+- TLS adherence policy integration — some platforms define a TLS adherence
+  policy mechanism that reports whether components comply with the configured
+  TLS profile. Integrating with such mechanisms is out of scope for this VEP.
+
+## Definition of Users
+
+- **Cluster administrators**: Configure TLS groups via the KubeVirt CR
+  (`spec.configuration.tlsConfiguration.groups`) to enforce curve preferences
+  across all virt pod endpoints. On managed platforms, a meta-operator
+  typically populates this from the platform's cluster-wide TLS profile.
+- **Meta-operator developers**: Read the platform's TLS profile resource and
+  translate it into KubeVirt's `tlsConfiguration.groups`.
+
+## User Stories
+
+### Story 1: Enable PQC Hybrid Key Exchange
+
+As a cluster administrator, I want to configure TLS groups on my KubeVirt
+deployment to include `X25519MLKEM768` so that all virt pod endpoints negotiate
+PQC-ready key exchanges when clients support them.
+
+### Story 2: Restrict Curves for Compliance
+
+As a security officer, I want to restrict TLS groups to a known-safe subset
+(e.g. only `X25519` and `secp256r1`) so that non-approved curves are never
+negotiated, meeting our compliance requirements.
+
+### Story 3: Cluster-Wide TLS Profile Alignment
+
+As a platform administrator, I configure TLS groups in the cluster-wide TLS
+security profile. A meta-operator propagates these to the KubeVirt CR, and all
+virt pod endpoints honour the same group preferences as the rest of the
+cluster.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Feature Gate
+
+All changes are gated behind the `TLSGroupPreferences` feature gate (disabled
+by default). When the feature gate is disabled, the `Groups` field on
+`TLSConfiguration` is ignored and Go's default curve preferences apply.
+
+The gate allows controlled rollout and easy rollback: if a misconfiguration
+causes TLS issues, disabling the gate immediately restores Go defaults without
+needing to clear the `groups` field from the CR. It also allows meta-operators
+to enable the feature in lockstep with their platform's TLS group support.
+
+### API Change
+
+Add a `Groups` field to `TLSConfiguration` in
+`staging/src/kubevirt.io/api/core/v1/types.go`:
+
+```go
+type TLSConfiguration struct {
+    // ...existing fields...
+    // +optional
+    // +listType=set
+    Groups []string `json:"groups,omitempty"`
+}
+```
+
+The `Groups` field uses `[]string` (matching the existing `Ciphers` pattern)
+rather than a typed enum. Group names use IANA TLS Supported Groups registry
+names (e.g. `X25519`, `secp256r1`, `X25519MLKEM768`), validated at webhook
+time against Go's `crypto/tls` supported curves via a `CurvePreferenceNameMap`.
+
+This approach avoids requiring an API change when Go adds support for new
+groups (e.g. `SecP256r1MLKEM768` in a future Go version) and is consistent
+with how `Ciphers` already works in KubeVirt. It also accommodates builds
+that link alternative `crypto/tls` implementations.
+
+> **Note:** Go does not currently provide a public API for translating
+> curve/group names to `tls.CurveID` values (unlike cipher suites which have
+> `tls.CipherSuites()`). This requires maintaining a manual
+> `CurvePreferenceNameMap` in KubeVirt until Go provides an equivalent. There
+> is an upstream Go proposal to address this:
+> [golang/go#77712](https://github.com/golang/go/issues/77712). Once that
+> lands, KubeVirt can replace the manual map with Go's own API, eliminating the
+> maintenance burden. This is tracked as a Beta graduation requirement.
+
+### TLS Setup Changes
+
+All TLS server setup functions in `pkg/util/tls/tls.go` already read
+`getTLSConfiguration(kv)` and set `CipherSuites` and `MinVersion` on
+`tls.Config`. Each function needs one additional line to set
+`CurvePreferences`:
+
+```go
+kv := clusterConfig.GetConfigFromKubeVirtCR()
+tlsConfig := getTLSConfiguration(kv)
+ciphers := CipherSuiteIds(tlsConfig.Ciphers)
+minTLSVersion := TLSVersion(tlsConfig.MinTLSVersion)
+curvePreferences := CurvePreferenceIds(tlsConfig.Groups)
+config := &tls.Config{
+    CipherSuites:     ciphers,
+    MinVersion:       minTLSVersion,
+    CurvePreferences: curvePreferences,
+    // ...
+}
+```
+
+The affected functions are:
+
+| Function | Components | File Location |
+|----------|-----------|---------------|
+| `SetupPromTLS` | virt-controller, virt-handler, virt-operator | `pkg/util/tls/tls.go:29` |
+| `SetupExportProxyTLS` | virt-exportproxy | `pkg/util/tls/tls.go:64` |
+| `SetupTLSWithCertManager` | virt-api, virt-operator | `pkg/util/tls/tls.go:97` |
+| `SetupTLSForServer` | virt-handler, virt-synchronization-controller | `pkg/util/tls/tls.go:173` |
+| `buildServer` | virt-exportserver | `pkg/storage/export/virt-exportserver/exportserver.go:261` |
+
+When `Groups` is empty, `CurvePreferenceIds` returns `nil` and
+`CurvePreferences` is left unset, preserving Go's default behaviour (currently
+`X25519`, `secp256r1`, `secp384r1`, `secp521r1`, `X25519MLKEM768` in Go 1.24).
+
+#### virt-exportserver
+
+The virt-exportserver binary is a special case. Unlike the other components, it
+does not have access to the KubeVirt CR at runtime. Instead,
+`VMExportController.appendTLSEnvVars()`
+(`pkg/storage/export/export/export.go:1724`) reads `TLSConfiguration` from
+the KubeVirt CR and injects `TLS_MIN_VERSION` and `TLS_CIPHER_SUITES`
+environment variables into the export server pod at creation time.
+
+The `buildServer()` method (`exportserver.go:261`) then constructs a static
+`tls.Config` from these values. This path requires:
+
+1. A new `TLSCurvePreferences []tls.CurveID` field on `ExportServerConfig`
+   (`exportserver.go:102`)
+2. `buildServer()` sets `CurvePreferences` on the `tls.Config`
+3. A new `TLS_CURVE_PREFERENCES` environment variable read by
+   `cmd/virt-exportserver/virt-exportserver.go`
+4. `appendTLSEnvVars()` injects `TLS_CURVE_PREFERENCES` from the KubeVirt CR's
+   `tlsConfiguration.groups` field, using `CurvePreferenceIds` to convert
+   group names to `tls.CurveID` values
+
+### Curve Name to ID Mapping
+
+Add `CurvePreferenceIds` and `CurvePreferenceNameMap` functions in
+`pkg/util/tls/tls.go`, analogous to the existing `CipherSuiteIds` and
+`CipherSuiteNameMap`. `CurvePreferenceNameMap` maps IANA group names to
+`tls.CurveID` values and is used by both the TLS setup functions and the
+admission webhook for validation.
+
+### Deployment Injection
+
+Extend `InjectTLSConfigIntoDeployment` (`pkg/util/tls/tls.go:264`) to inject a
+`--tls-groups` flag for virt-template components, matching the existing pattern
+for `--tls-cipher-suites` and `--tls-min-version`. The virt-template
+components currently do not accept a `--tls-groups` flag, so the flag must
+be added to their startup code as part of this work.
+
+### Validation
+
+Extend `validateTLSConfiguration` in
+`pkg/virt-operator/webhooks/kubevirt-update-admitter.go` to:
+
+1. **Validate group names** against the `CurvePreferenceNameMap` (dynamically
+   built from Go's `crypto/tls` supported curves), rejecting unknown names.
+2. **Validate TLS version and group compatibility** — PQC groups such as
+   `X25519MLKEM768` are TLS 1.3-only (Go filters them out for TLS 1.2
+   connections). If `MinTLSVersion` is below TLS 1.3 and `Groups` contains
+   only PQC groups, TLS 1.2 handshakes would fail with an empty curve list.
+   The webhook must reject this, requiring at least one classical group
+   (`X25519`, `secp256r1`, `secp384r1`, or `secp521r1`) when `MinTLSVersion`
+   is below `VersionTLS13`.
+
+### Meta-Operator Integration
+
+KubeVirt does **not** watch platform-specific TLS profile resources directly.
+The integration model (unchanged from existing `ciphers`/`minTLSVersion`
+handling) is:
+
+1. Platform administrator configures the cluster-wide TLS profile
+2. A meta-operator watches the platform TLS profile resource and translates it
+   into `kubevirt.spec.configuration.tlsConfiguration.groups`
+3. KubeVirt components dynamically pick up the new groups via their existing
+   `GetConfigForClient` callbacks
+
+Note that virt-template components receive TLS configuration via command-line
+flags injected by virt-operator at deployment time. Changing the `groups`
+field triggers a deployment update and pod restart for these components, unlike
+the other components which pick up changes dynamically.
+
+## API Examples
+
+### KubeVirt CR with TLS Groups
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - TLSGroupPreferences
+    tlsConfiguration:
+      minTLSVersion: VersionTLS12
+      ciphers:
+        - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+      groups:
+        - X25519
+        - secp256r1
+        - secp384r1
+        - X25519MLKEM768
+```
+
+When `groups` is omitted, `CurvePreferences` is left unset on `tls.Config`
+and Go uses its built-in defaults.
+
+## Alternatives
+
+### Direct Platform TLS Profile Watching
+
+Have virt-operator watch a platform-specific TLS profile resource directly
+and apply the TLS configuration.
+
+**Rejected because:**
+
+- Introduces a direct dependency on platform-specific APIs in upstream KubeVirt
+- Duplicates the meta-operator's role — meta-operators already translate
+  platform configuration into KubeVirt CR fields
+- Breaks deployments on platforms that do not provide such a resource
+
+### Enum-Typed Groups
+
+Use a `TLSGroup` enum type with kubebuilder validation instead of `[]string`.
+
+**Rejected because:**
+
+- Requires an API change every time Go adds new group support
+- Doesn't accommodate builds linking alternative `crypto/tls` implementations
+- Inconsistent with how `Ciphers []string` already works in KubeVirt
+- Webhook validation against Go's runtime `crypto/tls` provides the same
+  safety without the maintenance burden
+
+### No Feature Gate
+
+Add the `Groups` field without a feature gate since it's optional and
+backward-compatible.
+
+**Rejected because:**
+
+- Allows controlled rollout and easy rollback
+- Allows meta-operators to enable the feature in lockstep with their
+  platform's TLS group support
+- Follows KubeVirt's feature lifecycle conventions
+
+## Scalability
+
+No scalability impact. The group configuration is read once per TLS handshake
+via the existing `GetConfigForClient` callback, which already reads cipher and
+version configuration. The mapping from group names to `tls.CurveID` values
+is O(n) over a list of at most 5 entries.
+
+## Update/Rollback Compatibility
+
+**Upgrade:**
+
+- Existing deployments with no `groups` field continue to work unchanged —
+  Go defaults apply
+- Adding groups to an existing `tlsConfiguration` takes effect dynamically
+  via `GetConfigForClient` callbacks for most components. virt-template
+  components receive groups via command-line flags and require a deployment
+  update (handled automatically by virt-operator).
+
+**Downgrade:**
+
+- If the `TLSGroupPreferences` feature gate is disabled, the `groups` field is
+  ignored and Go defaults apply
+- No impact on running connections; new handshakes use defaults
+
+**Version Skew:**
+
+- During upgrade, components at different versions may have different group
+  support. Components that do not yet understand the `groups` field will use
+  Go defaults, which include all the same groups — so there is no security
+  regression during rolling updates.
+- virt-template components receive TLS configuration via `--tls-groups` flags
+  injected by virt-operator. During a rolling upgrade, a new virt-operator may
+  inject the `--tls-groups` flag into a deployment running old virt-template
+  pods that do not recognise the flag. virt-operator already handles this by
+  triggering a rolling deployment update, so old pods are replaced with new
+  ones that understand the flag.
+- virt-exportserver pods are ephemeral and created on-demand by
+  virt-controller with TLS configuration injected as environment variables at
+  pod creation time. During a rolling upgrade, a new virt-controller may
+  inject the `TLS_CURVE_PREFERENCES` env var into a pod running an old
+  virt-exportserver binary — the old binary simply ignores the unknown env
+  var and uses Go defaults. Conversely, a new virt-exportserver binary
+  created by an old virt-controller will not receive the env var and also
+  falls back to Go defaults.
+
+## Functional Testing Approach
+
+### Unit Tests
+
+`pkg/util/tls/tls_test.go`:
+
+- `CurvePreferenceIds` returns correct `tls.CurveID` values for each group
+- `CurvePreferenceIds` returns nil for empty input
+- `CurvePreferenceIds` skips unknown group names
+- Validation rejects invalid group names
+- Validation accepts all valid group names
+
+### Functional Tests
+
+Extend `tests/infrastructure/tls-configuration.go`:
+
+- Configure `groups` on the KubeVirt CR
+- Verify TLS connections to virt-api, virt-handler, virt-exportproxy,
+  virt-template-apiserver, and virt-template-controller negotiate using the
+  configured groups
+- Verify that a client offering only a non-configured curve (where the
+  configured set excludes it) is rejected or falls back appropriately
+- Verify that omitting `groups` preserves default behaviour
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `TLSGroupPreferences` guards all code changes (disabled by
+  default)
+- [ ] `Groups []string` field added to `TLSConfiguration`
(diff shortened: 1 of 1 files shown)


veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,539 @@
...
+
+### API Change
+
+Add a `TLSGroup` enum type and a `Groups` field to `TLSConfiguration` in
+`staging/src/kubevirt.io/api/core/v1/types.go`:
+
+```go
+// TLSGroup defines a TLS supported group (key exchange curve) using IANA
+// names from the TLS Supported Groups registry.
+// +kubebuilder:validation:Enum=X25519;secp256r1;secp384r1;secp521r1;X25519MLKEM768
```
nunnatsa: Do we really want to limit? Maybe use the current golang library to validate, getting new groups when tbey are added to new go version.

Also, some buildd may link different implementation than the standard golang crypto package.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,539 @@
...
+cluster-wide TLS profiles. KubeVirt needs a corresponding API so that
+meta-operators (such as HCO on OpenShift) can propagate these settings into
+the KubeVirt CR, ensuring all virt pod endpoints participate in cluster-wide
+PQC rollouts.
+
+## Goals
+
+- Extend `TLSConfiguration` with a `Groups` field to configure TLS supported
+  groups
+- Apply the configured groups as `CurvePreferences` on all TLS server endpoints
```
0xFelix, line 68: Apply to `virt-template-apiserver` and `virt-template-controller` as well? They are being deployed by virt-operator.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,539 @@
...
+security profile. HCO propagates these to the KubeVirt CR, and all virt pod
+endpoints honour the same group preferences as the rest of the cluster.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Feature Gate
```
nunnatsa, line 135: Does a fg really needed? Setting the groups field is not mandatory, so if the user chooses to set the new field, isn't it enough? Also, tbe implementation is pretty simple and  uses the golsng standard lib, with no special logic, so what's the risk?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Good point. Switching to `[]string` (matching the existing `Ciphers` pattern) and validating dynamically against Go's `crypto/tls` supported curves at webhook time makes more sense here. This avoids needing an API change every time Go adds new groups (e.g. `SecP256r1MLKEM768` in a future Go version), and is consistent with how `Ciphers` already works in KubeVirt.

I'll update the VEP and implementation to use `[]string` with webhook validation via a `CurvePreferenceNameMap()` built from the Go standard library.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Good catch — virt-template-apiserver and virt-template-controller are already covered by the implementation via `InjectTLSConfigIntoDeployment` (which was extended with `--tls-groups`), but they're missing from the Goals list. I'll add them.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Good question. I think the feature gate is still valuable here for a few reasons:

1. **Alignment with meta-operators/platforms** — the OpenShift API gates this behind `TLSGroupPreferences`. Having a matching gate in KubeVirt allows HCO to enable both in lockstep rather than having KubeVirt unconditionally apply groups before the platform is ready.
2. **VEP process** — the feature lifecycle process expects a feature gate for Alpha features that can be promoted through Beta to GA.
3. **Rollback safety** — if a misconfiguration causes TLS issues, disabling the gate immediately restores Go defaults without needing to clear the `groups` field from the CR.

The implementation overhead is minimal (a single `if` check in each setup function), so I'd prefer to keep it.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- Align KubeVirt's TLS behaviour with platform-wide or organisational TLS
+  policies
+
+Kubernetes platforms are beginning to expose TLS group configuration in their
+cluster-wide TLS profiles. KubeVirt needs a corresponding API so that
+meta-operators (such as HCO on OpenShift) can propagate these settings into
+the KubeVirt CR, ensuring all virt pod endpoints participate in cluster-wide
+PQC rollouts.
+
+## Goals
```
enp0s3: Do we want the TLS configuration changes to take effect immediately by rolling out the virt-components once the configuration is changed?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- This VEP targets alpha for version: v1.9.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue [#287](https://github.com/kubevirt/enhancements/issues/287) created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
Barakmor1: I'm not sure, should mark this one as well?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+groups such as `X25519MLKEM768`.
+
+## Motivation
+
+The transition to Post-Quantum Cryptography (PQC) is underway. NIST has
+finalised its first set of PQC standards and IETF has standardised hybrid key
+exchange groups such as `X25519MLKEM768` that combine classical and
+post-quantum algorithms. Go 1.24 (which KubeVirt already uses) includes native
+support for `X25519MLKEM768` as a `tls.CurveID`.
+
```
Barakmor1, line 35: Consider adding a section on why a new API field is required here. Why not just append the new PQC groups to the default supported groups list automatically? Explicitly mentioning the performance and compatibility trade-offs would be helpful context.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+the KubeVirt CR, ensuring all virt pod endpoints participate in cluster-wide
+PQC rollouts.
+
+## Goals
+
+- Extend `TLSConfiguration` with a `Groups` field to configure TLS supported
+  groups
+- Apply the configured groups as `CurvePreferences` on all TLS server endpoints
+  in virt-api, virt-controller, virt-handler, virt-operator,
+  virt-exportproxy, virt-template-apiserver, and virt-template-controller
```
Barakmor1: Also `virt-exportserver` and `sync-controller`.

Barakmor1 (commented):
Thanks, i left some comments.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
Barakmor1, line 64: Could you clarify the expected behavior during a cross-cluster migration if there is a mismatch in group support (i.e., one side does not support a configured group)? Additionally, if a specific node within the cluster lacks support for a configured group, will there be a clear and actionable error message surfaced to the user?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+| X25519MLKEM768 | No (filtered out) | Yes |
+
+If `MinTLSVersion` is below TLS 1.3 and `Groups` contains **only** PQC groups
+(e.g. only `X25519MLKEM768`), the filtered curve list for TLS 1.2 connections
+would be empty, causing handshake failures with
+`"tls: no supported elliptic curves for ECDHE"`. Go does not validate this at
+config time — it fails silently at handshake time.
+
+The admission webhook must reject this misconfiguration, analogous to the
+existing rule that rejects custom cipher suites with TLS 1.3:
```
fossedihelm: Is this an implicit maintenance contract?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- Validate group names in the KubeVirt CR admission webhook
+- Maintain backward compatibility — omitting `Groups` preserves current Go
+  default behaviour
+
+## Non Goals
+
+- Watching platform-specific TLS profile resources directly from virt-operator
+  — meta-operators (e.g. HCO) handle this translation
+- Configuring TLS groups on client-side connections (server-side is the
+  priority; client-side alignment is a later initiative)
```
Barakmor1: I’m not sure I fully understand. Does this mean there won’t be support for cross-cluster or cross-namespace migration?
If that’s correct, could you explain why this would be difficult to achieve?
As far as I remember, cross-cluster live migration already supports TLS configuration.

FYI @awels

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
fossedihelm: I mean, if we make `Groups` open (no ENUM), and we enforce this validation, we will have to maintain this.
IOW, this is in contrast, IMO, with:
```
This approach avoids requiring an API change when Go adds support for new
groups (e.g. `SecP256r1MLKEM768` in a future Go version) and is consistent
with how `Ciphers` already works in KubeVirt. It also accommodates builds
that link alternative `crypto/tls` implementations.
```
Not sure about the solution, completely defer this to go or take full ownership.
Right now, it sounds like a middle way.
WDYT?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- [ ] Admission webhook validation for group names
+- [ ] Unit tests for mapping, validation, and TLS setup
+- [ ] Functional tests verifying group enforcement on virt pod endpoints
+
+### Beta
+
+- [ ] Certificate key curve configuration (replacing hardcoded P-256 in
+  `pkg/certificates/triple/cert/cert.go`)
+- [ ] Additional PQC hybrid groups supported automatically as Go adds them
+  (e.g. `SecP256r1MLKEM768`, `SecP384r1MLKEM1024`) — no API change needed
```
fossedihelm: this is what is concerning me

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+  default behaviour
+
+## Non Goals
+
+- Watching platform-specific TLS profile resources directly from virt-operator
+  — meta-operators (e.g. HCO) handle this translation
+- Configuring TLS groups on client-side connections (server-side is the
+  priority; client-side alignment is a later initiative)
+- Configuring certificate key curves (e.g. changing the hardcoded P-256 in
+  certificate generation) — deferred to Beta
```
Barakmor1: Can you elaborate?
why do we have this in the None Goals?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- TLS adherence policy integration (`tlsAdherence` field) — separate OpenShift
```
Barakmor1: Please elaborate

enp0s3 (approved):
@lyarwood Thank you! Enhancement seems clear to me

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
enp0s3: @Barakmor1 Hi. From what I understood PQC isn't FIPS-compliant, so you would need to explicitly opt-out from this option.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+deployment to include `X25519MLKEM768` so that all virt pod endpoints negotiate
+PQC-ready key exchanges when clients support them.
+
+### Story 2: Restrict Curves for Compliance
+
+As a security officer, I want to restrict TLS groups to a known-safe subset
+(e.g. only `X25519` and `secp256r1`) so that non-approved curves are never
+negotiated, meeting our compliance requirements.
+
+### Story 3: Cluster-Wide TLS Profile Alignment (OpenShift)
```
0xFelix: We're a bit on the OpenShift heavy side with this VEP 😛

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+1. A new `TLSCurvePreferences []tls.CurveID` field on `ExportServerConfig`
+   (`exportserver.go:102`)
+2. `buildServer()` sets `CurvePreferences` on the `tls.Config`
+3. A new `TLS_CURVE_PREFERENCES` environment variable read by
+   `cmd/virt-exportserver/virt-exportserver.go`
+4. `appendTLSEnvVars()` injects `TLS_CURVE_PREFERENCES` from the KubeVirt CR's
+   `tlsConfiguration.groups` field, using `CurvePreferenceIds` to convert
+   group names to `tls.CurveID` values
+
+### Curve Name to ID Mapping
```
0xFelix, line 235: We're a little implementation detail heavy here. It's nice to have but we don't need the full impl in the VEP yet.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+rather than a typed enum. Group names use IANA TLS Supported Groups registry
+names (e.g. `X25519`, `secp256r1`, `X25519MLKEM768`), validated at webhook
+time against Go's `crypto/tls` supported curves via a `CurvePreferenceNameMap`.
+
+This approach avoids requiring an API change when Go adds support for new
+groups (e.g. `SecP256r1MLKEM768` in a future Go version) and is consistent
+with how `Ciphers` already works in KubeVirt. It also accommodates builds
+that link alternative `crypto/tls` implementations.
+
+> **Note:** Go does not provide a public API for translating curve/group names
```
0xFelix: Will it be a requirement for beta/GA to adopt the new feature once it lands?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+```
+
+A corresponding `CurvePreferenceNameMap` function maps IANA group names to
+`tls.CurveID` values and is used by both the TLS setup functions and the
+admission webhook for validation.
+
+### Deployment Injection
+
+Extend `InjectTLSConfigIntoDeployment` (`pkg/util/tls/tls.go:264`) to inject a
+`--tls-groups` flag for virt-template components, matching the existing pattern
```
0xFelix, line 246: `virt-template` components need to implement this flag first. Mention that somewhere?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+for `--tls-cipher-suites` and `--tls-min-version`.
+
+### Validation
+
+Extend `validateTLSConfiguration` in
+`pkg/virt-operator/webhooks/kubevirt-update-admitter.go:249` to validate group
+names against the `CurvePreferenceNameMap` (dynamically built from Go's
+`crypto/tls` supported curves):
+
+```go
```
0xFelix: Same

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+
+### HCO Integration
+
+KubeVirt does **not** watch the OpenShift APIServer CR. The integration model
+(unchanged from existing `ciphers`/`minTLSVersion` handling) is:
+
+1. OpenShift admin configures `apiserver.spec.tlsSecurityProfile` (which
+   includes the new `groups` field once
+   [openshift/api#2583](https://github.com/openshift/api/pull/2583) merges)
+2. HCO watches the APIServer CR and translates the profile into
```
0xFelix: Add that HCO enables the FG unconditionally/if required?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+   `kubevirt.spec.configuration.tlsConfiguration.groups`
+3. KubeVirt components dynamically pick up the new groups via their existing
+   `GetConfigForClient` callbacks — no component restart required
+
+### FIPS Considerations
+
+`X25519MLKEM768` is documented as not FIPS-approved in the OpenShift API.
+KubeVirt itself applies whatever groups are configured — it is the
+responsibility of HCO (or the cluster administrator) to omit non-FIPS groups
+on FIPS-enabled clusters.
```
0xFelix: Do we need to mention it?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    tlsConfiguration:
+      minTLSVersion: VersionTLS12
+      ciphers:
+        - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+```
```
0xFelix: Do these examples add any benefit?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+version configuration. The mapping from group names to `tls.CurveID` values
+is O(n) over a list of at most 5 entries.
+
+## Update/Rollback Compatibility
+
+**Upgrade:**
+
+- Existing deployments with no `groups` field continue to work unchanged —
+  Go defaults apply
+- Adding groups to an existing `tlsConfiguration` takes effect immediately
```
0xFelix: It would restart virt-template components, since these take the groups as flags

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+  via `GetConfigForClient` callbacks (no component restart needed)
+
+**Downgrade:**
+
+- If the `TLSGroupPreferences` feature gate is disabled, the `groups` field is
+  ignored and Go defaults apply
+- No impact on running connections; new handshakes use defaults
+
+**Version Skew:**
```
0xFelix, line 375: Cannot run new operator with old virt-template Pods, since they don't have the flags.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,533 @@
...
+- `CurvePreferenceIds` skips unknown group names
+- Validation rejects invalid group names
+- Validation accepts all valid group names
+
+### Functional Tests
+
+Extend `tests/infrastructure/tls-configuration.go`:
+
+- Configure `groups` on the KubeVirt CR
+- Verify TLS connections to virt-api, virt-handler, and virt-exportproxy
```
0xFelix: Also virt-template components, there are existing tests for them already

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
enp0s3: @fossedihelm +1 thanks for bringing that up. I was confused by the statement `consistent
with how `Ciphers` already works in KubeVirt`. I see now that we are performing [input validation](https://github.com/kubevirt/kubevirt/blob/29371d76ac669c5137b3fae96bdeee2fc73a1c3b/pkg/virt-operator/webhooks/kubevirt-update-admitter.go#L267) for ciphers.
@lyarwood ~~I think we need to make this at least a requirement for further graduation.~~ I see that you already put this as a criteria for Alpha ` Admission webhook validation for group names`
I think that the intention to use `[]string` was to align in the future to the validation logic that we already have with ciphers - as soon as `tls/crypto` will provide us with the supported groups function we will be able to opt-in for validation w/o changing the API.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
enp0s3: @Barakmor1 I think it was left unchecked because we are not yet sure if we will make it to target version 1.9

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
enp0s3: @Barakmor1 +1 Good question. 
@awels I was looking at the [docs PR](https://github.com/kubevirt/user-guide/pull/914) and didn't find a reference for syncing cluster configurations, is this expected to be done manually by the cluster administrator?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Addressed in the updated VEP. Using `[]string` with webhook validation against `CurvePreferenceNameMap`. I've also clarified that the manual map is a temporary measure until golang/go#77712 lands, which is now tracked as a Beta graduation requirement.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Done - added virt-template-apiserver, virt-template-controller, virt-exportserver, and virt-synchronization-controller to the Goals list.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Kept the feature gate. I've expanded the rationale inline in the VEP: rollback safety (disabling the gate restores Go defaults without clearing the CR field) and allowing meta-operators to enable it in lockstep with their platform's TLS group support.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Added a "Why Not Automatically Enable PQC Groups?" subsection to Motivation covering FIPS compliance restrictions, performance trade-offs, and platform alignment.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Done - added virt-exportserver and virt-synchronization-controller to the Goals list.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: This VEP only configures server-side TLS endpoints. In TLS, the server selects the key exchange group from the client's offered set based on its `CurvePreferences`. Since both sides are Go programs, the client always offers Go's full default set, so the server controls what gets negotiated regardless of any mismatch in configuration.

For cross-cluster migration specifically, the migration tunnel TLS is a client-side connection initiated by virt-handler. This VEP does not configure client-side connections (listed as a Non Goal). The client would use Go defaults, and the remote server's `CurvePreferences` would determine the negotiated group. A mismatch in configured groups between clusters would not cause failures - it would just mean the negotiated group is determined by the server's preferences rather than both sides agreeing on a restricted set.

I've elaborated the client-side Non Goal to explain this.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Good catch. The manual `CurvePreferenceNameMap` is a temporary measure until golang/go#77712 lands, at which point we replace it with Go's own API. I've clarified this in the VEP and added it as a Beta graduation requirement. So the maintenance contract is time-bounded, not permanent.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: See my reply above on the cross-cluster migration thread (L52). Server-side configuration is sufficient - the server picks the group, and the client (using Go defaults) will always support it. A mismatch between clusters doesn't cause failures. Client-side configuration would only matter for strict compliance requirements about what appears in the ClientHello.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Addressed - the Beta graduation requirements now explicitly include adopting the upstream Go API (golang/go#77712) to replace the manual map.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Elaborated in the updated VEP. Server-side `CurvePreferences` controls which group is negotiated. Client-side would only be needed to control what appears in the ClientHello (strict compliance or avoiding HelloRetryRequest round trips).

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Elaborated - some platforms define a TLS adherence policy mechanism that reports whether components comply with the configured TLS profile. Integrating with such mechanisms is out of scope for this VEP.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Agreed - replaced OpenShift/HCO references with generic meta-operator terminology throughout. Only one HCO mention remains (as an example in Non Goals).

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Trimmed - removed the CurvePreferenceIds code block, the full validation code block, and the TLS version compatibility table. Replaced with prose descriptions.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Yes - added as a Beta graduation requirement.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Done - noted that virt-template components don't currently have a `--tls-groups` flag and it must be added as part of this work. Also added to the Alpha graduation checklist.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Trimmed along with the rest of the implementation detail.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Replaced the HCO Integration section with a generic Meta-Operator Integration section. The feature gate rationale now mentions that it allows meta-operators to enable in lockstep.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Removed the standalone FIPS section. FIPS compliance restrictions are now covered in the new "Why Not Automatically Enable PQC Groups?" motivation subsection.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Trimmed to a single representative example plus the default behaviour note.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Fixed - the VEP now notes that virt-template components receive groups via command-line flags and require a deployment update/pod restart, unlike the other components which pick up changes dynamically.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Fixed - added a version skew paragraph explaining that virt-operator triggers a rolling deployment update, replacing old pods with new ones that understand the `--tls-groups` flag.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Done - added virt-template-apiserver and virt-template-controller to the functional tests list and the Alpha graduation checklist.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,447 @@
...
+### Story 3: Cluster-Wide TLS Profile Alignment
+
+As a platform administrator, I configure TLS groups in the cluster-wide TLS
+security profile. A meta-operator propagates these to the KubeVirt CR, and all
+virt pod endpoints honour the same group preferences as the rest of the
+cluster.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
```
Barakmor1, line 131: It might be odd to have different APIs across virt components when it comes to TLS configuration.

lyarwood: Are you suggesting we list virt-template, AAQ etc here as well given they will need to make changes to consume this?

Barakmor1: Yes, maybe also HCO and all other components if they include a similar API for TLS configuration.

Barakmor1: FYI @nunnatsa

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
awels: > @Barakmor1 +1 Good question. @awels I was looking at the [docs PR](https://github.com/kubevirt/user-guide/pull/914) and didn't find a reference for syncing cluster configurations, is this expected to be done manually by the cluster administrator?

Whatever 'orchestrator' we are using is expected to keep the CAs in sync between clusters.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: @nunnatsa correct me if I'm wrong here but I think the HCO implementation sits outside the scope of this VEP and should be seperate right?

@Barakmor1 I'll add the additional component repos however that need to be modified to handle this work within kubevirt/kubevirt.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,447 @@
...
+
+Add a `Groups` field to `TLSConfiguration` in
+`staging/src/kubevirt.io/api/core/v1/types.go`:
+
+```go
+type TLSConfiguration struct {
+    // ...existing fields...
+    // +optional
+    // +listType=set
+    Groups []string `json:"groups,omitempty"`
```
Barakmor1, line 156: How about name Curves?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,447 @@
...
+that link alternative `crypto/tls` implementations.
+
+> **Note:** Go does not currently provide a public API for translating
+> curve/group names to `tls.CurveID` values (unlike cipher suites which have
+> `tls.CipherSuites()`). This requires maintaining a manual
+> `CurvePreferenceNameMap` in KubeVirt until Go provides an equivalent. There
+> is an upstream Go proposal to address this:
+> [golang/go#77712](https://github.com/golang/go/issues/77712). Once that
+> lands, KubeVirt can replace the manual map with Go's own API, eliminating the
+> maintenance burden. This is tracked as a Beta graduation requirement.
```
Barakmor1, line 177: I guess we can revisit this and add it as a requirement for Beta/GA.

Edit: just noticed we already have it on the Beta requirements please ignore

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,447 @@
...
+- [ ] Functional tests verifying group enforcement on virt pod endpoints
+  (including virt-template components)
+
+### Beta
+
+- [ ] Adopt upstream Go API for curve/group name resolution
+  ([golang/go#77712](https://github.com/golang/go/issues/77712)) if available,
+  replacing the manual `CurvePreferenceNameMap`
+- [ ] Certificate key curve configuration (replacing hardcoded P-256 in
+  `pkg/certificates/triple/cert/cert.go`)
```
Barakmor1: IMO, it would be better to drop this from the beta requirements at this point. This can be a future VEP if needed. It's better to keep this one focused, as IIUC this requirement may require an additional API.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
Barakmor1: I think we can at least include AAQ in this VEP. @akalenyu , what do you think about CDI does it fit as well?

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
nunnatsa: Please don't. This is going to be done in many places in many projects. Please keep the common name of "Groups"

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
nunnatsa: Yes, HCO implementation can be done elsewhere or in separate VEP. It will also add complicity to this VEP, as there are additional HCO use-cases, and they are more complex.

lyarwood: AAQ has its own operator, its own CR, and its own TLS setup - it doesn't use virt-operator or read TLSConfiguration from the KubeVirt CR. Same for CDI. So neither is in-scope here.

This VEP covers all components deployed by virt-operator that consume `TLSConfiguration` from the KubeVirt CR. AAQ and CDI would need to adopt group preferences independently if they want to align.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Agreed with @nunnatsa - `Groups` is the IANA TLS Supported Groups registry name and the term used in the TLS 1.3 spec (RFC 8446). Go's `CurvePreferences` field name is a legacy from when these were only elliptic curves, but the spec renamed them to "supported groups" when non-curve key exchange mechanisms (like X25519 and now MLKEM hybrids) were added. Keeping `Groups` is consistent with the standard terminology.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
lyarwood: Agreed - certificate key curve configuration is a distinct concern from key exchange group preferences and would likely need its own API field. I'll remove it from Beta requirements to keep this VEP focused.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
Acedus: > This VEP covers all components deployed by virt-operator that consume TLSConfiguration from the KubeVirt CR. AAQ and CDI would need to adopt group preferences independently if they want to align.

While I do concur with this statement, I do think all projects under the KubeVirt umbrella would benefit from a shared package for dynamically managing TLS configuration in a generic/agnostic way.

lyarwood: Yup I don't disagree but that's something we could look into once the initial implementation is in place and we can extract any shared logic somewhere for reuse.

akalenyu: Might be worth evaluating whether some of this will fit in controller-runtime directly. I think tons of openshift operators are going to do similar stuff and those are largely based on that SDK

lyarwood: I took a look at controller-runtime's current TLS support. It already provides a generic `TLSOpts []func(*tls.Config)` escape hatch on both the webhook server and metrics server, so operators built on controller-runtime can set `CurvePreferences` today without any upstream changes:

```go
webhook.NewServer(webhook.Options{
    TLSOpts: []func(*tls.Config){
        func(c *tls.Config) {
            c.CurvePreferences = []tls.CurveID{tls.X25519, tls.CurveP256}
        },
    },
})
```

There's no dedicated field for it though, and upstream explicitly moved *away* from individual TLS knobs — they added a `TLSMinVersion` field in 2021, then [deprecated and removed it](https://github.com/kubernetes-sigs/controller-runtime/pull/2041) in favour of `TLSOpts` in 2022. The rationale was that `TLSOpts` covers all `tls.Config` customization generically. There are no open issues or PRs requesting dedicated `CurvePreferences` support or any PQC/post-quantum discussion upstream.

So for operators based on controller-runtime (CDI, AAQ, etc.), the TLS plumbing is already there — they'd just need to wire their own CR field through to a `TLSOpts` function. What *could* be shared is the validation and name→ID mapping logic (the `CurvePreferenceNameMap` + webhook validation from this VEP), but that's more of a shared utility library than a controller-runtime feature.

KubeVirt itself doesn't use controller-runtime for its virt-* components (it constructs `tls.Config` directly in `pkg/util/tls/tls.go`), so this doesn't change the VEP-287 implementation approach.

akalenyu: Sorry I should have said I was explicitly referring to the plumbing for setting up a separate ctrl runtime "Runnable" that knows where to get the desired TLS config. This is largely similar between the various operators (get some CR, calculate something out of it's spec)

0xFelix: That would mean controllers would also have to watch the KubeVirt CR. Acceptable for controllers in-tree, but controllers like `virt-template-controller` are developed outside of the main repository and are unaware of the KubeVirt CR.

Barakmor1: I’m not sure about the other operators, but AAQ and CDI have similar TLS configurations in their CRs. The intention here is to agree that AAQ CR, CDI CR, and others that share similar TLS configuration fields will extend them in the same way. At least from an API perspective, this makes sense to me.

Sorry if that wasn’t clear, I didn’t mean for other operators to read the KubeVirt CR to set their TLS configuration.

enp0s3:
@xpivarc @vladikr Hey folks, would you be able to take a look? 
I spoke to @Acedus offline and he's fine with the approach.

veps/sig-compute/287-tls-group-preferences/tls-group-preferences.md
```diff
@@ -0,0 +1,445 @@
...
+## Motivation
+
+The transition to Post-Quantum Cryptography (PQC) is underway. NIST has
+finalised its first set of PQC standards and IETF has standardised hybrid key
+exchange groups such as `X25519MLKEM768` that combine classical and
+post-quantum algorithms. Go 1.24 (which KubeVirt already uses) includes native
+support for `X25519MLKEM768` as a `tls.CurveID`.
+
+KubeVirt's existing `TLSConfiguration` supports `MinTLSVersion` and `Ciphers`
+but has no mechanism to control which key exchange groups (elliptic curves) are
```
xpivarc, line 37: not sure about the correct usage of term "key exchange groups"

xpivarc (commented):
/approve
Alpha

PR #292, opened 2026-04-23
VEP-10: Graduate GPUsWithDRA feature gate to beta

alaypatel07:
Assisted-by: Cursor <noreply@cursor.com>

### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/10
**SIG label**:  /sig compute

### What this PR does

Graduate GPUsWithDRA feature gate to beta

### DIFF
--- a/veps/sig-compute/10-dra-devices/vep.md
+++ b/veps/sig-compute/10-dra-devices/vep.md
@@ -75,7 +75,7 @@ level of control when using Virtual Machines as they have with Containers.
 ### Supported Usecases
 
 1. GPU devices where the DRA driver publishes required attributes in device metadata files:
-   - `resources.kubernetes.io/pciBusID` for passthrough GPU 
+   - `resource.kubernetes.io/pciBusID` for passthrough GPU 
    - `mDevUUID` for mediated devices/vGPU (currently hardcoded in KubeVirt, will be standardized when [kubernetes/kubernetes#135552](https://github.com/kubernetes/kubernetes/issues/135552) is resolved)
 2. DRA driver must write metadata files via the DRA Attributes Downward API (see [KEP-5304](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5304-dra-attributes-downward-api))
 
@@ -87,7 +87,7 @@ level of control when using Virtual Machines as they have with Containers.
 
 ### Unsupported Usecases (in this VEP)
 
-1. GPUs where the DRA driver does not set the required attributes (e.g., `resources.kubernetes.io/pciBusID`, `mDevUUID`) will not
+1. GPUs where the DRA driver does not set the required attributes (e.g., `resource.kubernetes.io/pciBusID`, `mDevUUID`) will not
    be supported. Drivers must publish the necessary attributes for KubeVirt to generate the libvirt domain XML.
 2. HostDevices via DRA - will be addressed in a separate VEP (HostDevicesWithDRA)
 3. NetworkDevices via DRA - will be addressed in a separate VEP (NetworkDevicesWithDRA)
@@ -128,7 +128,13 @@ type VirtualMachineInstanceSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	// +optional
-	ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
+	ResourceClaims []VirtualMachineInstanceResourceClaim `json:"resourceClaims,omitempty"`
+}
+
+// VirtualMachineInstanceResourceClaim wraps the k8sv1.PodResourceClaim to allow
+// VMI-specific extensions while preserving the upstream wire format.
+type VirtualMachineInstanceResourceClaim struct {
+	k8sv1.PodResourceClaim `json:",inline"`
 }
 
 type GPU struct {
@@ -150,16 +156,59 @@ type GPU struct {
 // This type is shared infrastructure that will be reused by future device VEPs
 // (HostDevicesWithDRA, NetworkDevicesWithDRA, CPUsWithDRA).
 type ClaimRequest struct {
-	// ClaimName needs to be provided from the list vmi.spec.resourceClaims[].name where this
-	// device is allocated
-	// +optional
-	ClaimName *string `json:"claimName,omitempty"`
-	// RequestName needs to be provided from resourceClaim.spec.devices.requests[].name where this
-	// device is requested
-	// +optional
-	RequestName *string `json:"requestName,omitempty"`
+	// ClaimName references the name of an entry in the
+	// VMI's spec.resourceClaims[] array. The referenced
+	// entry may use either resourceClaimName or
+	// resourceClaimTemplateName.
+	// +kubebuilder:validation:MinLength=1
+	ClaimName string `json:"claimName,omitempty"`
+	// RequestName specifies which request from the
+	// ResourceClaim/ResourceClaimTemplate spec.devices.requests array this
+	// claim request corresponds to.
+	// +kubebuilder:validation:MinLength=1
+	RequestName string `json:"requestName,omitempty"`
 }
 ```
+
+### ClaimRequest String Change Rationale
+
+Changing `ClaimRequest.claimName` and `requestName` from `*string` to `string`
+is a **Go type change, not a wire-format break**. The JSON/YAML field names, the
+valid wire format, and existing persisted objects are all unchanged i.e.
+`claimName: "foo"` looks identical on the wire whether the Go field is
+`*string` or `string`, hence end users authoring VMIs will see no difference.
+
+The change is intentional and reflects how the API evolved from alpha to alpha2:
+
+1. In early alpha, using pointers for optional scalar fields was a reasonable
+   starting point and aligned with common Kubernetes API conventions:
+   <https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md>.
+2. In alpha2, DRA-specific validation was added and the DRA path now requires
+   both identifiers to be present and non-empty.
+3. For non-DRA/device-plugin entries, `ClaimRequest` remains absent (nil), so
+   `claimName` and `requestName` are naturally omitted.
+
+Given this behavior, plain strings are the cleaner model for the DRA path:
+they reduce pointer-handling complexity and make the "required when DRA is
+used" contract explicit. The feature is still alpha-gated, so the Go-side
+refactor can land cleanly with downstream consumers recompiling against the
+new types.
+
+### `vmi.spec.resourceClaims[]` Wrapper Type Rationale
+
+Starting in beta (v1.9), the element type of `vmi.spec.resourceClaims[]` is
+an in-tree type (`VirtualMachineInstanceResourceClaim`) that inlines the upstream
+`k8sv1.PodResourceClaim`. The wrapper exists because KubeVirt cannot add
+fields such as `managedClaim` (see [Appendix C](#c-managed-resource-claims))
+to `k8sv1.PodResourceClaim` itself; without it, additional policy fields
+would have to live in a parallel array, making the API awkward.
+
+Like the `*string` -> `string` change above, introducing the wrapper is a
+**Go type change, not a wire-format break**, hence no user manifests will
+be affected. Inlining `k8sv1.PodResourceClaim` keeps existing fields
+(`name`, `resourceClaimName`, `resourceClaimTemplateName`) at the same
+JSON/YAML level, so existing VMIs round-trip identically; only Go consumers
+compiling against the generated client-go types are affected.
 
 Note: VMI status does not include device attributes. Virt-launcher reads device metadata directly from files mounted
 by the DRA driver via the DRA Attributes Downward API (see [KEP-5304](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5304-dra-attributes-downward-api)).
@@ -192,20 +241,20 @@ KubeVirt is a pure consumer of this mechanism: virt-launcher only reads the meta
 that the framework and the driver produce. For how drivers publish metadata and how pods
 consume it in general, refer to the upstream Kubernetes documentation:
 
-- Concept: [DRA → Device metadata](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#device-metadata)
-- Workload-side usage: [Allocate devices to workloads with DRA → Request devices in workloads](https://kubernetes.io/docs/tasks/configure-pod-container/assign-resources/allocate-devices-dra/#request-devices-workloads)
+- Concept: [DRA -> Device metadata](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#device-metadata)
+- Workload-side usage: [Allocate devices to workloads with DRA -> Request devices in workloads](https://kubernetes.io/docs/tasks/configure-pod-container/assign-resources/allocate-devices-dra/#request-devices-workloads)
 
 #### Container-Side File Layout
 
-Virt-launcher reads metadata files from this base directory inside the virt-launcher
+Virt-launcher reads metadata files from this base directory inside the compute
 container:
 
 ```
 /var/run/kubernetes.io/dra-device-attributes/
-├── resourceclaims/
-│   └── {claimName}/{requestName}/{driverName}-metadata.json     # pre-existing claims
-└── resourceclaimtemplates/
-    └── {podClaimName}/{requestName}/{driverName}-metadata.json  # template-generated claims
+|-- resourceclaims/
+|   `-- {claimName}/{requestName}/{driverName}-metadata.json     # pre-existing claims
+`-- resourceclaimtemplates/
+    `-- {podClaimName}/{requestName}/{driverName}-metadata.json  # template-generated claims
 ```
 
 The two top-level subdirectories (`resourceclaims/` vs `resourceclaimtemplates/`) let
@@ -227,7 +276,7 @@ DRA GPU drivers that want to support KubeVirt must include the following attribu
 
 | Device Type | Required Attribute | JSON Path | Example |
 |-------------|-------------------|-----------|---------|
-| Passthrough GPU | `resources.kubernetes.io/pciBusID` | `requests[].devices[].attributes["resources.kubernetes.io/pciBusID"].stringValue` | `0000:65:00.0` |
+| Passthrough GPU | `resource.kubernetes.io/pciBusID` | `requests[].devices[].attributes["resource.kubernetes.io/pciBusID"].stringValue` | `0000:65:00.0` |
 | Mediated Device (vGPU) | `mDevUUID` | `requests[].devices[].attributes["mdevUUID"].stringValue` | `aa618089-8b16-4d01-a136-25a0f3c73123` |
 
 > **Note:** `mDevUUID` is not yet standardized under `resources.kubernetes.io`. See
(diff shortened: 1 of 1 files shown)


veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -178,207 +186,40 @@ DRA Attributes Downward API (defined by KEP-5304) to get device attributes and g
...
-| Field | Type | Source | Description |
-|-------|------|--------|-------------|
-| `metadata.name` | string | Automatic | Claim name |
-| `metadata.namespace` | string | Automatic | Claim namespace |
-| `requests[].name` | string | Automatic | Request name from ResourceClaim |
-| `requests[].devices[].driver` | string | Automatic | Driver name |
-| `requests[].devices[].pool` | string | Automatic | Pool name from allocation |
-| `requests[].devices[].device` | string | Automatic | Device name within the pool |
-| `requests[].devices[].attributes` | map | Automatic | Device attributes from ResourceSlice |
+In **beta**, kubevirt will drop the local copy and consume the metadata via the upstream
```
lyarwood: nit - beta (v1.9)

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -393,34 +234,72 @@ DRA GPU drivers that want to support KubeVirt must include the following attribu
...
-   attribute exposure via framework coordination (KEP-5304).
-3. Virt-controller does NOT need to watch ResourceClaims or ResourceSlices. Device attributes are made available to
-   virt-launcher via CDI-mounted metadata files written by the DRA driver.
+2. Starting in alpha2 (v1.8), virt-controller does NOT need to configure any attribute requests in the pod spec -
+   the DRA driver handles attribute exposure via framework coordination (KEP-5304).
+3. Starting in alpha2 (v1.8), virt-controller does NOT watch ResourceClaims or ResourceSlices. Device attributes
+   are made available to virt-launcher via CDI-mounted metadata files written by the DRA driver. The
+   `DRAStatusController` and the `ResourceClaim` / `ResourceSlice` informers that existed in alpha (v1.6) were
+   removed as part of this transition.
+4. Starting in beta (v1.8) virt-controller will skip the `permittedHostDevices` validation for DRA-managed devices
```
lyarwood: nit - v1.9

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -393,34 +234,72 @@ DRA GPU drivers that want to support KubeVirt must include the following attribu
...
-6. Use the extracted attribute to generate the libvirt domain XML
-
-### Webhook changes
-
-1. Allow the VMI only if vmi spec is correct, i.e. resource claims for requested devices are specified
+#### Virt-Launcher Parsing
 
-Note: The feature gate verification when VMI requesting DRA device will either be done in webhook or controller, TBD
+For each GPU or HostDevice entry in `vmi.spec.domain.devices.{gpus,hostDevices}[]` whose
+`ClaimRequest` is set, virt-launcher performs the following step in alpha2 (KubeVirt)
```
lyarwood: nit - "step" should be "steps" (6 steps listed below) and missing a colon at the end of the line.

veps/sig-compute/10-dra-devices/vep.md
lyarwood: Can you add a brief rationale for why `permittedHostDevices` validation is being skipped for DRA-managed devices? Is it because `permittedHostDevices` is a device-plugin concept that doesn't apply to DRA-managed devices? Also missing a comma after the closing parenthesis for consistency with items 2 and 3 above.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -853,6 +732,9 @@ writing an e2e test with a real GPU driver is outside the scope of alpha.
 - KEP-5304 merged and available in Kubernetes
 - Alpha2 design validated
 - Design must be proven stable with no major changes required during one full release cycle
+  - This is a general rule, however an exception can be made
+if the API can be used
+    for CPU and Network devices without any major changes
```
lyarwood: The indentation here is broken - line 736 starts at column 0 which will break the nested bullet when rendered.

Also, the wording "without any major changes" is quite vague. Could this be more precise? e.g.:

```suggestion
  - This is a general rule; however, an exception can be made if the API
    can be consumed by CPU and Network device types without breaking changes
    to the existing GPU interface.
```

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: updated

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: done

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: Added the reasoning, PTAL and resolve this if it looks good to you.

alaypatel07:
@vladikr @iholder101 @rthallisey @xpivarc @EdDev @0xFelix  PTAL, I have added a second commit here in this PR to address all the open question discussed during the sync on April 27

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -870,7 +755,7 @@ writing an e2e test with a real GPU driver is outside the scope of alpha.
 - Consider additional GPU use cases (vGPU, MIG)
 
 ### GA
-- Upgrade/downgrade testing
+- Upgrade/downgrade testing.
```
0xFelix: nit: drop change

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -393,34 +234,75 @@ DRA GPU drivers that want to support KubeVirt must include the following attribu
...
+     `podClaimName`, not the generated `ResourceClaim` name.
+3. Glob `*-metadata.json` in that directory. KubeVirt expects exactly one match (one driver
+   per request); zero or more-than-one is an error.
+4. Decode the JSON stream and pick the first object whose `apiVersion` is supported.
+5. Locate the single element of `requests[]` whose `name` equals the `requestName`. KubeVirt
+   expects exactly one device in `devices[]` (count > 1 not supported).
+6. Extract `mdevUUID` first, then fall back to `resource.kubernetes.io/pciBusID`, and build
+   the appropriate libvirt domain element.
+
+In **beta**, steps 2–4 will be replaced by a single call to the upstream
```
0xFelix:
```suggestion
In **beta (v1.9)**, steps 2–4 will be replaced by a single call to the upstream
```

0xFelix: nit

0xFelix (commented):
Contents look good to me, just a few nits. I'd prefer regular ASCII characters insttead of Unicode, but that's up to you.

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -178,207 +186,40 @@ DRA Attributes Downward API (defined by KEP-5304) to get device attributes and g
 
 ### Device Metadata via DRA Attributes Downward API
 
-> **Note:** This section describes the expected mechanism based on
-> [KEP-5304: DRA Device Attributes Downward API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5304-dra-attributes-downward-api).
+This design depends on
+[KEP-5304: DRA Device Attributes Downward API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5304-dra-attributes-downward-api).
```
iholder101, line 239: Do you think we should wait for the KEP to become beta as well?

I'm not sure we should, as it is almost certain that if the KEP will change it will effect the under-the-hood implementation details, not APIs.

WDYT?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -178,207 +186,40 @@ DRA Attributes Downward API (defined by KEP-5304) to get device attributes and g
...
-4. Workloads (virt-launcher) discover and read files via globbing or by claim name
+- Concept: [DRA → Device metadata](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#device-metadata)
+- Workload-side usage: [Allocate devices to workloads with DRA → Request devices in workloads](https://kubernetes.io/docs/tasks/configure-pod-container/assign-resources/allocate-devices-dra/#request-devices-workloads)
 
-#### File Location
+#### Container-Side File Layout
 
-Metadata files are written at a well-known path:
+Virt-launcher reads metadata files from this base directory inside the virt-launcher
+container:
```
iholder101:
```suggestion
Virt-launcher reads metadata files from this base directory inside the compute
container:
```

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -393,34 +234,75 @@ DRA GPU drivers that want to support KubeVirt must include the following attribu
...
+   - If `rc.resourceClaimTemplateName` is set (template-generated claim), use
+     `resourceclaimtemplates/<rc.name>/<requestName>/` — the path segment is the pod-local
+     `podClaimName`, not the generated `ResourceClaim` name.
+3. Glob `*-metadata.json` in that directory. KubeVirt expects exactly one match (one driver
+   per request); zero or more-than-one is an error.
+4. Decode the JSON stream and pick the first object whose `apiVersion` is supported.
+5. Locate the single element of `requests[]` whose `name` equals the `requestName`. KubeVirt
+   expects exactly one device in `devices[]` (count > 1 not supported).
+6. Extract `mdevUUID` first, then fall back to `resource.kubernetes.io/pciBusID`, and build
+   the appropriate libvirt domain element.
```
iholder101: Can a device expose both? What happens in such a scenario?

iholder101: `mmdevUUID` is not standardized yet (tracked by https://github.com/kubernetes/kubernetes/issues/135552).

Should we worry about that? What happens when it is standardized with a different key (e.g. `resource.kubernetes.io/mdevUUID`)?

I guess this will break our logic. Do we need to think about a migration plan?

veps/sig-compute/10-dra-devices/vep.md
iholder101, line 797: IIUC today we don't really have any functional tests. Do you have a test plan in mind?

iholder101 (commented):
Thanks @alaypatel07!

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: > Do you think we should wait for the KEP to become beta as well?

This is a valid question.

> I'm not sure we should, as it is almost certain that if the KEP will change it will effect the under-the-hood implementation details, not APIs.

I agree with this, the APIs presented in this VEP is what is going to beta, the metadata delivery mechanism is internal details.  Even if those details change, the external API should be able to remain stable and absorb those changes.

KEP-5304 to will be attempted to be in beta in 1.37, if risks are found and flagged there will still be enough time between KubeVirt vep-freeze and KubeVirt code-freeze to block beta graduation (as of now, chances of that happening is rare)

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: Yes, it is expressed a few lines above:
```
If both attributes are present on the same device, KubeVirt treats the device as a mediated
device (UUID takes precedence), because a parent mediated device may also expose a PCI address.
```

https://github.com/kubevirt/enhancements/pull/292/changes/BASE..a621eeff1898a28921711511cdd442707d074312#diff-ca56d7cf7c1e00e1a9f18d95a66ad56535d734121d47fcddb0831dca511cc506R237

alaypatel07: > I guess this will break our logic. Do we need to think about a migration plan?

This migration plan has to be supported by the driver. The driver should express both attributes, `mdevUUID` and `resource.kubernetes.io/mdevUUID` for couple releases and drop `mdevUUID` when it is safe. cc @varunrsekar for thoughts

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: Tests will be added as part of 1.9, [here](https://github.com/kubevirt/kubevirt/issues/16481) is the tracking issue

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -762,6 +765,294 @@ This VEP establishes the core DRA infrastructure for KubeVirt. The following VEP
...
+metadata:
+  name: vmi-with-grouped-cpu
+spec:
+  resourceClaims:
+  - name: cpu-claim
+    resourceClaimName: claim-cpu-capacity-10
+  domain:
+    cpu:
+      cores: 10
+      dra:
```
rthallisey, line 932: We were talking about this a little offline.  It's not clear to me if we need to have the embedded `dra` field, .e.g `.cpu.dra.claimName` vs `.cpu.claimName`.  I'm trying to understand what would having the embedded field help us do?

veps/sig-compute/10-dra-devices/vep.md
iholder101: We've discussed this in the sig-compute.

TLDR: this is entirely implementation details, worst case we can support both old and new approach is k8s decide to make breaking changes, then get rid of the old approach next release. I don't think this is a real concern as long as it's not related to APIs.

iholder101:
Thank you @alaypatel07! 

/lgtm

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: I agree, having `dra` field does not add value, but I want to keep this document honest with the POC that was demoed https://github.com/alaypatel07/kubevirt/commit/2711f8428076395625bd62b7f1412168ec575c8b 

Inspired from our discussion, I have added Appendix C, to make the auto field more generic
```
### C. Managed Resource Claims
```

veps/sig-compute/10-dra-devices/vep.md
varunrsekar: Yes unfortunately we’d require the DRA driver to support multiple versions of kubevirt if we decide to go with unqualified “mdevUUID” and it changes. 

On a side note, vGPU mdevs is getting phased out in favor of directly using sriov VFs in newer kernels.

But This brings up an interesting question: DRA driver vendor conformance w/ kubevirt. As part of this, should we come up with conformance tests so that it’s easier to determine what works and what doesn’t?

vladikr:
The new Appendix C on managedClaim is well done and strengthens the case for beta. 
It clearly demonstrates that future UX improvements (auto-generated claims for alignment, multi-device scenarios, etc.) can be added additively without changing the ClaimRequest that we're graduating.
I would only make the managedClaim name clearer (maybe autoClaim),  but that can be debated in the future.

Thank you @alaypatel07 

/approve
/hold
@rthallisey I'll let you remove the hold when you think it's ready.

alaypatel07:
> The new Appendix C on managedClaim is well done and strengthens the case for beta.
It clearly demonstrates that future UX improvements (auto-generated claims for alignment, multi-device scenarios, etc.) can be added additively without changing the ClaimRequest that we're graduating.
I would only make the managedClaim name clearer (maybe autoClaim), but that can be debated in the future.

> @rthallisey I'll let you remove the hold when you think it's ready.

Ryan and I had an indepth sync up on Appendix C. I wanted to capture the follow-up discussion we had around the `managedClaim` / possible `autoClaim` direction.

My understanding is that this future feature would sit in a similar problem space as kubelet’s CPU/device management: it would translate higher-level workload intent and policy into concrete resource allocation shape. The important distinction is that, for KubeVirt, the intent can span multiple VMI-level resources: GPU, NIC, CPU, host devices, and potentially storage-related lifecycle needs.

1. **Storage/lifecycle policies may become interesting later.**
   One possible future policy is detaching the `ResourceClaim` lifecycle from the virt-launcher pod lifecycle for usecases like VM suspend/resume, where storage needs to say even if pod is deleted. That likely requires Kubernetes support first, some kind of a `ResourceClaim` spec-level capability to allow resource claims to exists even after pod is gone. If/when Kubernetes builds that support, the `managedClaim`/`autoClaim` shape looks flexible enough for KubeVirt to expose it as part of a storage related policy.

2. **Policy conflicts and invalid combinations are future controller/admission logic.**
   If KubeVirt owns claim generation, KubeVirt also owns the business logic for deciding which policies make sense for which devices. For example, an L3-cache-alignment policy should not be valid for a GPU. Those rules would belong in KubeVirt admission and/or virt-controller reconciliation, to enforce.

3. **This should still allow integration with future external policy components.**
   If the community later builds a generic component for expressing these policies, it may surface either through Pod spec fields or through an external CR/controller model. A KubeVirt-managed claim layer should be able to integrate with either by generating or referencing the appropriate underlying objects, while keeping the VMI-side consumer shape stable.

So I think the appendix is useful because it shows that the `ClaimRequest` API we are graduating is the stable low-level reference: `(claimName, requestName)`. Future UX layers such as `managedClaim` or `autoClaim` can be debated and added later without changing that consumer contract.

I agree that the name is still open. `autoClaim` may communicate the intent better than `managedClaim`, but I think that naming discussion can happen in the future VEP: https://github.com/kubevirt/enhancements/issues/300 :)

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -128,7 +128,18 @@ type VirtualMachineInstanceSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	// +optional
-	ResourceClaims []k8sv1.PodResourceClaim `json:"resourceClaims,omitempty"`
+	ResourceClaims []VirtualMachineInstanceResourceClaim `json:"resourceClaims,omitempty"`
+}
+
+// VirtualMachineInstanceResourceClaim wraps the upstream k8sv1.PodResourceClaim to allow
+// KubeVirt-specific extensions while preserving the upstream wire format.
+type VirtualMachineInstanceResourceClaim struct {
```
alaypatel07, line 136: cc @rthallisey @iholder101 @vladikr 

This change is needed for Appendix-C, PTAL.

xref: https://github.com/kubevirt/enhancements/pull/292#issuecomment-4348130444

0xFelix: By wrapping the `k8sv1.PodResourceClaim` with this type we keep compatibility with the existing field and for existing users?

veps/sig-compute/10-dra-devices/vep.md
```diff
@@ -142,17 +156,60 @@ type GPU struct {
...
+	// ResourceClaim/ResourceClaimTemplate spec.devices.requests array this
+	// claim request corresponds to.
+	// +kubebuilder:validation:MinLength=1
+	RequestName string `json:"requestName,omitempty"`
 }
 ```
 
+### ClaimRequest String Change Rationale
+
+Changing `ClaimRequest.claimName` and `requestName` from `*string` to `string`
```
0xFelix, line 175: Will this change be compatible with existing clusters on upgrade, i.e. with protobuf serialization?

0xFelix (commented):
I'd be okay with breaking changes on upgrades, but we should document it if that is the case.

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: CRDs don't have protobuf serializations. The json/yaml serialization will be valid with existing clusters

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: Correct, please look at `### `vmi.spec.resourceClaims[]` Wrapper Type Rationale` section for more details

veps/sig-compute/10-dra-devices/vep.md
0xFelix: Thanks, wasn't sure.

alaypatel07:
@0xFelix JSON/YAML files will not break on upgrades, only go clients, when they vendor the 1.9 kubevirt api if they are using DRA API, they need to change the go structs (this is a very rare occurrence). I have added the justification in the Rationale for each change, pease let me know if this works. 

> I'd be okay with breaking changes on upgrades

0xFelix:
/lgtm

All clear, thank you!

veps/sig-compute/10-dra-devices/vep.md
alaypatel07: > But This brings up an interesting question: DRA driver vendor conformance w/ kubevirt. As part of this, should we come up with conformance tests so that it’s easier to determine what works and what doesn’t?

This is a good idea, we can discuss more during GA? Not sure if beta should be blocked on this.

Issue #294, opened 2026-04-23
VEP 293: Experimental Migration Options


Aseeef:
**Primary contact (assignee)**:

/assign Aseeef

**Current Feature Stage**: New

**Feature Gate**: `AdvancedExperimentalMigrationOptions`

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/295

**Timeline**:

* Alpha release target (x.y): 1.9.0
* Beta release target (x.y): Never
* Stable release target (x.y): Never

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/295 
  * [ ] Code PRs: TODO

> [!IMPORTANT]
> Please keep this description up to date.

Issue #291, opened 2026-04-22
VEP 291: Live Migration - Dynamic Downtime Tuning and Iteration-Based Convergence


michalskrivanek:
**Primary contact (assignee)**:
/assign @michalskrivanek

**Current Feature Stage**:
Alpha

**Feature Gate**:
AdvancedLiveMigration

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/249

**Timeline**:

- Alpha release target: v1.10
- Beta release target: TBD
- Stable release target: TBD
- Alpha
  * VEP PRs: https://github.com/kubevirt/enhancements/pull/249
  * Code PRs: https://github.com/kubevirt/kubevirt/pull/17480
  * Docs PRs:

### Related VEPs

This enhancement is part of a family of interrelated live migration improvements:

- **VEP 246: Live Migration Compression** (#246) — Adds zstd compression for migration streams to reduce bandwidth consumption. Synergizes well with downtime tuning; benchmark data shows 4.1x faster migration and 3.6x lower downtime when combined.
- **VEP 248: Live Migration Stall Detection & Convergence** (#248 / PR #252) — Proposes stall detection and intelligent switchover timing. Complementary: stall detection can inform when to begin the downtime ramp-up, potentially replacing the static `startAfterIteration` parameter.

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
@michalskrivanek can you please re-target this one for 1.10 if you are planing to work on this

PR #295, opened 2026-04-23
VEP 293: Experimental Migration Options

Aseeef:
This VEP introduces a new API field gated behind a feature gate dedicated for experimental and low-level api field-testing.

### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/294
**SIG label**: /sig compute

### What this PR does

Initially proposed in an existing VEP, this was split into its own, mini-VEP because 2-3 other existing VEPs are likely to make use of this functionality: 
* https://github.com/kubevirt/enhancements/issues/291
* https://github.com/kubevirt/enhancements/issues/246
* https://github.com/kubevirt/enhancements/issues/248

### DIFF
--- a/veps/sig-compute/experimental-migration-options.md
+++ b/veps/sig-compute/experimental-migration-options.md
@@ -1,0 +1,144 @@
+# VEP 293: Advanced Experimental Migration Options
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.9
+- This VEP targets beta for version: N/A (perpetually alpha)
+- This VEP targets GA for version: N/A (perpetually alpha)
+
+### Release Signoff Checklist
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes adding a new perpetually Alpha feature gate, `AdvancedExperimentalMigrationOptions`, to KubeVirt. When enabled, this feature gate reads a special experimental configuration section from the `MigrationPolicy` CR (`migrations.kubevirt.io/v1alpha1`). This allows users to experiment with low-level migration-related tunables and experimental APIs without custom builds, and without providing any stability guarantees for these configurations.
+
+`MigrationPolicy` is chosen using **VirtualMachineInstance** and **Namespace** label selectors, so experimental settings on a policy affect only **matching** VMIs’ migrations.
+
+## Motivation
+
+Tuning migrations is a complex task that often simply cannot be adequately achieved in a lab/test environment. In many cases, it is difficult to determine good default values for low-level migration tunables, and different workloads might require distinct settings. To test these low-level tunables effectively without forcing users to rely on custom builds, we need a dedicated area for experimentation. This allows KubeVirt to expose experimental migration features and settings that may or may not graduate into more top-level API fields in the future, while making zero guarantees about their stability.
+
+It is important to note that not everything under the experimental section is added with the intention of graduating it to a stable API. Some fields are added fully knowing we will never officially support them. They are simply low-level tunables that we might want to experiment with using different values.
+
+## Goals
+
+- Expose a configuration space for low-level migration-related tunables and experimental APIs.
+- Provide existing default values for cases when defaults don't work, allowing users to experiment with other values.
+- Give users the ability to test new migration features without custom builds.
+- Ensure strict safeguards (e.g., dropping new experimental fields on create/update when disabled, and using an admission controller to block invalid use) while still allowing cluster upgrades.
+
+## Non Goals
+
+- Moving this feature gate to Beta or GA. `AdvancedExperimentalMigrationOptions` will perpetually and permanently be an Alpha feature gate.
+- Providing stability guarantees for any options placed in the experimental section.
+
+## Definition of Users
+
+- **Cluster Administrators:** Who wish to volunteer their user-specific workloads by experimenting with migration tunables.
+- **KubeVirt Developers:** Who need to expose new experimental features for gathering community feedback before stabilizing them in higher-level APIs.
+
+## User Stories
+
+- As a cluster administrator, I want to modify low-level migration tunables so I can experiment with values for my specific workload when defaults do not work in order to share this information with KubeVirt developers.
+  * As an advanced cluster operator, I want to fine-tune low-level migration performance parameters when I know my workload better than the defaults, accepting that these knobs can expose implementation details and are not treated as supported product configuration.
+- As a developer, I want to expose a new migration API temporarily or experimentally so I can gather real-world data from users without committing to a stable API.
+
+## Repos
+
+- `kubevirt/api`
+- `kubevirt/kubevirt`
+
+## Design
+
+### Feature Gate and Configuration
+
+We will add a new feature gate called `AdvancedExperimentalMigrationOptions`.
+This feature gate is explicitly designed to remain **perpetually and permanently Alpha**. 
+
+When this feature gate is enabled, a special configuration section will be read. This section is exposed via `migrations.kubevirt.io/v1alpha1` `MigrationPolicy` resources. The config exposes a configuration that already has default values for cases when defaults don't work, or when we want to experiment with potentially better defaults on user clusters.
+
+This feature gate is independent from other migration features. Other features might independently expose some tunables via this experimental feature gate. For example, GA features may still expose some tunables via this option.
+
+### Safeguards and Admission Control
+
+Because these are purely experimental and unstable tunables that sometimes expose implementation details, we impose strict conditions on their usage via an admission policy:
+
+- Values inside the experimental section are entirely ignored if the `AdvancedExperimentalMigrationOptions` feature gate is not enabled.
+- If the feature gate is disabled, new changes cannot be pushed to the experimental section. Existing values will still be persisted but remain inactive and uneditable until the feature gate is re-enabled.
+- Specifically, to preserve data on update (e.g. during a cluster upgrade where the feature gate might be temporarily disabled by default), if the existing object already has a value in the field, it is preserved. If it does not, new usage is cleared. This is handled by the admission controller.
+
+> Note: This approach is also inline with how Kubernetes handles experimental APIs (see https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version), preserving data on update while reserving the right to break the schema at any time.
+>
+> However, while K8s applies gates per feature (thanks to nifty codegen that automatically generates code with a simple API marker like `// featuregate=MyFeatureGate` comment), in KubeVirt this code needs to be hand written. So we opt to use **one** perpetually-Alpha gate—`AdvancedExperimentalMigrationOptions`—for all migration-related `experimental.*` configuration, with admission written explicitly for that single gate.
+
+Rationale for this stricter barrier (versus only ignoring `experimental` at runtime while the gate is off): alignment with Kubernetes patterns that tie experimental API surface to a feature gate, and requiring the gate to be enabled is an **explicit** operator acknowledgement that these fields are unstable—stronger than documentation alone, which users may skip. Nevertheless, the API documentation (godoc comments) for the `experimental` field will still explicitly state that the schema, semantics, and existence of the fields under the field are unstable.
+
+## API Examples
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: KubeVirt
+metadata:
+  name: kubevirt
+  namespace: kubevirt
+spec:
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - AdvancedExperimentalMigrationOptions
+```
+
+```yaml
+apiVersion: migrations.kubevirt.io/v1alpha1
+kind: MigrationPolicy
+metadata:
+  name: test-migration-policy
+spec:
+  experimental:  # Section ignored with out feature gate
+    compression: "zstd"  # Example only
+    maxDowntimeMs: 1050  # Example only
+```
+
+## Alternatives
+
+- Exposing these experimental values via annotations. However, utilizing a properly structured `experimental` API explicitly denotes that these features are for testing purposes, making the schema clear without the opacity of unstructured annotations. Unstructured annotations also lack validations and creates the potential for security lapses which is why K8s no longer prefers using annotations for experimental API.
+- Use of hooks/plugins. While a strong alternative in theory, in practice this would require a serious rework of Live Migration code. However, resolving concerns and bottlenecks in Live Migration is an immediate pain-point.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+### Release Signoff Checklist
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes adding a new perpetually Alpha feature gate, `AdvancedExperimentalMigrationOptions`, to KubeVirt. When enabled, this feature gate reads a special experimental configuration section from both the KubeVirt CR (cluster-wide defaults) and the `MigrationPolicy` CR (`migrations.kubevirt.io/v1alpha1`). This allows users to experiment with low-level migration-related tunables and experimental APIs without custom builds, and without providing any stability guarantees for these configurations.
```
michalskrivanek: it "leaks" into VMI as well throught the inlcusion of migrationpolicy. I do not think it matters much, and I think it is a good thing, but since you're enumerations it's worth mentioning?

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Motivation
+
+Tuning migrations is a complex task. In many cases, it is difficult to determine good default values for low-level migration tunables, and different workloads might require distinct settings. To test these low-level tunables effectively without forcing users to rely on custom builds, we need a dedicated area for experimentation. This allows KubeVirt to expose experimental migration features and settings that may or may not graduate into more top-level API fields in the future, while making zero guarantees about their stability.
+
+It is important to note that not everything under the experimental section is added with the intention of graduating it to a stable API. Some fields are added fully knowing we will never officially support them. They are simply low-level tunables that we might want to experiment with using different values.
```
michalskrivanek, line 28: the last sentence is my reasoning for "advanced" as well. There will be options(I'd argue it's already the case fo "max downtime") that we will want to keep around, yet they may be pretty low level and difficult to explain, and this way we keep them "away" from the top level.
This would be similar to ESXi's >30 advanced tunables that power users can change, hidden from the main UI.

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Goals
+
+- Expose a configuration space for low-level migration-related tunables and experimental APIs.
+- Provide existing default values for cases when defaults don't work, allowing users to experiment with other values.
+- Give users the ability to test new migration features without custom builds.
+- Ensure strict safeguards (e.g., dropping new experimental fields on create/update when disabled, and using an admission controller to block invalid use) while still allowing cluster upgrades.
```
michalskrivanek, line 35: is this required though? There's always a schema validation, I do not see a drawback in allowing setting values/fields when gate is off. In https://github.com/kubevirt/kubevirt/pull/17509/commits/42a097e78397f5ad25d09699725cc3c31a7dbf08 this section is simply ignored. Do we really need anything more complicated than that?

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Definition of Users
+
+- **Cluster Administrators:** Who wish to volunteer their user-specific workloads by experimenting with migration tunables.
+- **KubeVirt Developers:** Who need to expose new experimental features for gathering community feedback before stabilizing them in higher-level APIs.
+
+## User Stories
+
+- As a cluster administrator, I want to modify low-level migration tunables so I can experiment with values for my specific workload when defaults do not work in order to share this information with KubeVirt developers.
+- As a developer, I want to expose a new migration API temporarily or experimentally so I can gather real-world data from users without committing to a stable API.
```
michalskrivanek, line 51: *if* we'd agree with `advanced` as well then this shall include the story "advanced users wanting to fine-tune migration performance" or so...

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+- Exposing these experimental values via annotations. However, utilizing a properly structured `experimental` API explicitly denotes that these features are for testing purposes, making the schema clear without the opacity of unstructured annotations.
+- Forcing administrators to compile KubeVirt with custom values. This is prohibitive and hinders user testing and feedback.
+
+## Scalability
+
+Exposing this configuration API does not inherently impact scalability, although the specific low-level tunables modified via this API might.
+
+## Update/Rollback Compatibility
+
+Update compatibility is explicitly **not supported** for the experimental fields themselves. While upgrading the cluster is allowed and the system will attempt to preserve existing experimental data during the update, the schema of these fields may change or be removed completely between versions without a migration path. If an experimental field is removed or its schema changes incompatibly, the API server will simply drop the data or fail validation on subsequent edits.
```
michalskrivanek, line 123: +1

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+
+As per Kubernetes API guidelines, if an experimental field is removed or graduates out of the `experimental` section into a stable section:
+- If a field is removed, its name and protobuf tag will not be reused. A tombstone comment will be left in the corresponding Go struct to explicitly denote its deprecation and removal.
```
michalskrivanek: IIUC there's no protobuf changes, the policy is json encoded.

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,141 @@
...
+- If a field graduates to a stable section, a migration path may be provided to automatically transition the experimental configuration into the new stable section, or the experimental field may simply be dropped requiring manual user intervention to populate the new stable field.
```
michalskrivanek, line 127: I think this is fine. Presumably we're doing this to find the optimal value, and when we do the user shouldn't need to set it...

0xFelix:
Could a separate CRD be an alternate solution to the 'here be dragons FG'? We can throw away a CRD more easily than getting rid of fields in the stable API.

veps/sig-compute/experimental-migration-options.md
Aseeef: The justification for intentionally adding this barrier is:
* Better alignment with K8s. K8s also protects experimental API with a feature gate.
* Enable the feature gate serves as an explicit acknowledgement that these fields are experimental and can change at anytime (as opposed to documentation only which user may or may not read).

veps/sig-compute/experimental-migration-options.md
Aseeef: You are right. Fixed!

Aseeef:
> Could a separate CRD be an alternate solution to the 'here be dragons FG'? We can throw away a CRD more easily than getting rid of fields in the stable API.

I'll put that as an alternative. But I don't see a clear advantage to this. Since while fields under `experimental` can change, we aren't planning to get rid of `experimental` itself. So equivalently, if implemented as a separate CRD, we wouldn't be throwing away the CRD itself. So I do prefer the current approach because it is simpler in terms of code changes. Does that make sense?

0xFelix:
Still there would be deprecated/no-op fields accumulating underneath the `experimental` key over the time. Why don't we experiment with new features outside of the stable API in an additional CRD and move them to the stable API once we are confident they work the way we want them to? The experimental API could stay in alpha (e.g. `v1alpha1, v1alpha2, ...`), and therefore it would be easier to remove unwanted fields from it again. Alpha APIs are designed for rapid iteration, they offer more flexibility and aren't bound by the stricter guidelines of the Kubernetes Deprecation Policy

veps/sig-compute/experimental-migration-options.md
michalskrivanek: I have no objection to a feature gate, I meant the line 33 only - why overcomplicate things with admission controller or upgrade handling.

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,150 @@
...
+- `kubevirt/kubevirt`
+
+## Design
+
+### Feature Gate and Configuration
+
+We will add a new feature gate called `AdvancedExperimentalMigrationOptions`.
+This feature gate is explicitly designed to remain **perpetually and permanently Alpha**. 
+
+When this feature gate is enabled, a special configuration section will be read. This section is exposed cluster-wide within the KubeVirt CR under `migrationConfiguration.experimental.*`, and additionally can be scoped via `migrations.kubevirt.io/v1alpha1` `MigrationPolicy` resources. The config exposes a configuration that already has default values for cases when defaults don't work, or when we want to experiment with potentially better defaults on user clusters.
```
iholder101: I think we should drop the Kubevirt CR section and only keep migration policies

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,150 @@
...
+  configuration:
+    developerConfiguration:
+      featureGates:
+        - AdvancedExperimentalMigrationOptions
+    migrationConfiguration:
+      experimental:
+        # These are examples of low-level tunables
+        compression: "zstd"
+        maxDowntimeMs: 1050
+        ewmaAlpha: 0.4
```
iholder101: Do we need to agree on which tunables we expose in advance?

iholder101 (commented):
Thanks @Aseeef.

I think this VEP makes sense.

In an ideal world we wouldn't need these at all, or would be able to support them via Plugins (see VEP 190). In an offline conversation @EdDev said he will investigate this aspect and we can think about it further in the next cycle.

I just want to emphasize that I think these knobs will be exposed only temporarily, and that I see the `experimental` field being eventually removed. Some of the reasons for this position are:
1. Kubevirt should take the responsibility of providing the sensible tunings according to a declarative policy. Ideally, the admin just needs to tell us what he wants to achieve and we should do the hard work behind the scenes.
2. We want to decouple ourselves from a specific hypervisor.

iholder101:
> Could a separate CRD be an alternate solution to the 'here be dragons FG'? We can throw away a CRD more easily than getting rid of fields in the stable API.

Interesting idea. @vladikr raised a good point though - how would be use this new CRD? Will it be referenced by migration policies? Because this kind of loses the whole point, right?

Let me know what you think.

Although I like this direction, I think that using MigrationPolicies (which are still alpha) for this cycle specifically is not such a bad idea. The only downside is if we wanted to graduate MigrationPolicies while the experimental knobs are still there, but I think we can defer this desicion to the future.

WDYT?

veps/sig-compute/experimental-migration-options.md
Aseeef: These are just examples. We can discuss specific tunable later (the intention of this VEP wasn't to discuss which to expose).

iholder101:
Another thing to keep in mind is that you're still missing an assigned reviewer on https://github.com/orgs/kubevirt/projects/21

veps/sig-compute/experimental-migration-options.md
michalskrivanek: in the current simple(simplistic?) POC implementation in https://github.com/kubevirt/kubevirt/pull/17509/changes/42a097e78397f5ad25d09699725cc3c31a7dbf08 it "shows up" in `MigrationConfiguration` due to the way the configuration is being passed to virt-launcher currently, IIUC virt-handler nor virt-launcher know anything about MigrationPolicies

virt-controller: merges MigrationPolicy → MigrationConfiguration (v1) → writes to vmi.Status.MigrationState.MigrationConfiguration
virt-handler: reads vmi.Status.MigrationState.MigrationConfiguration → builds MigrationOptions → gRPC to virt-launcher

iholder101: We've talked about this in the sig-compute call. Let's stick to Migration Policies only

Aseeef:
FYI, as discussed I have removed the changes to the cluster-wide CR from the VEP.

iholder101:
Thank you @Aseeef!

I think this is good to go from my POV.
/lgtm

FYI @EdDev

iholder101:
> Thank you @Aseeef!
> 
> I think this is good to go from my POV. /lgtm
> 
> FYI @EdDev

That being said, some additional thoughts:
- Which knobs exactly do we want to expose?
- According to the [feature-lifecycle policy](https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#release-stage-transition-table) the deadline for alpha VEPs is between 1 and 2 releases. IOW after two releases this should be removed from the API.
- What's the path forward?

As suggested elsewhere, I think the outcome of this VEP should be that we'll discuss, on follow-up VEPs, how to integrate it nicely within our APIs according to the data you'd collect.

Aseeef:
> That being said, some additional thoughts:
> * Which knobs exactly do we want to expose?

This is discussed individually in:
https://github.com/kubevirt/enhancements/issues/291
https://github.com/kubevirt/enhancements/issues/246
https://github.com/kubevirt/enhancements/issues/248

> * According to the [feature-lifecycle policy](https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#release-stage-transition-table) the deadline for alpha VEPs is between 1 and 2 releases. IOW after two releases this should be removed from the API.

Noted.

> * What's the path forward?

Especially in the context of how we want future migration tuning to work (i.e. the declarative config), @vladikr described it well here: https://github.com/kubevirt/enhancements/pull/252#issuecomment-4352695839.

TLDR: use data collected to configure some number of "pre-defined profiles" for these tunables based on typical use-cases. User tells us what they need or what kind of workload they are running. And we configure a "profile" based on that.

However, even long term, I do think we need some relatively accessible way to override certain low level, hard-coded constants like the ones proposed in the linked VEPs or existing ones like `parallelMultifdMigrationThreads` or other QEMU defaults. Tunables that we don't expect most users to ever touch, but we still want readily accessible for when we want to be able to test on customer clusters for workloads or hardware we may not have access to.

veps/sig-compute/experimental-migration-options.md
```diff
@@ -0,0 +1,144 @@
...
+        - AdvancedExperimentalMigrationOptions
+```
+
+```yaml
+apiVersion: migrations.kubevirt.io/v1alpha1
+kind: MigrationPolicy
+metadata:
+  name: test-migration-policy
+spec:
+  experimental:  # Section ignored with out feature gate
```
EdDev, line 104: I'll start with mentioning that I think that during an alpha CRD stage, there can be freedom to play with the manifest.
However, this kind of knob should not go to Beta and will have to be dropped.
I also suggest specific knobs are place it in a meaningful hierarchy, giving it a chance to have better meaning and hint less of the experimental nature of it. Said that, it is up to you.

This is why I think no VEP is needed for this at all. Maybe a clarification in the feature-lifecycle to allow this freedom explicitly, would be the best.

The FG could have been introduced in one of the specific features VEP in which you add this knob.

EdDev (commented):
Sorry for the late reply to this one.

Summarize my non-blocking opinion: We do not need this VEP, you can introduce the FG in the other VEPs which introduce specific knobs.
The API of the alpha CRD is free to abuse, no need to reason for it in general.

I will post an update to the feature lifecycle to express that explicitly.

iholder101:
@Aseeef can we close this VEP?

PR #299, opened 2026-04-29
VEP 109: vGPU Migration Enable Update

Aseeef:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/109
**SIG label**: /sig compute

### What this PR does

Updated VEP to include specific details about when migration between different Virtual GPU Manager is permitted. Previously VEP incorrectly stated that it never works but the reality is a bit more nuanced... Also linked the source of truth: NVIDIA.

Finally, added a Target releases section.

### DIFF
--- a/veps/sig-compute/vgpu_live-migration.md
+++ b/veps/sig-compute/vgpu_live-migration.md
@@ -4,9 +4,18 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.8
-- This VEP targets beta for version:
-- This VEP targets GA for version:
+<!--
+A PR must update this section during the planning phase of a given release in order to track it.
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/vgpu_live-migration.md
```diff
@@ -4,9 +4,18 @@
...
+<!--
+A PR must update this section during the planning phase of a given release in order to track it.
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
+For example, during the planning phase for version v1.123, do **not** target beta for v.124 in advance.
+-->
+
+- This VEP targets alpha for version: v1.9
```
lyarwood, line 16: @xpivarc I assume you are okay with this merging after the deadline without an exception as you moved it to track before hand right?

Aseeef: By the way for context this VEP implementation already merged. It was merged too late for v1.8 though.

This PR just updates that technically it is merging into v1.9, and second clarifies some things relevant for Beta (v1.10 or later).

xpivarc (commented):
/approve

Issue #82, opened 2025-08-11
VEP 82: Plugin-based generalization of KubeVirt's virtualization stack


harshitgupta1337:
**Primary contact (assignee)**: 

/assign @harshitgupta1337

**Current Feature Stage**:

New

**Feature Gate**:

`PluginVirtualizationStacks`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.8.0
* Beta release target (x.y): 1.9.0
* Stable release target (x.y): 1.10.0

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #311, opened 2026-05-14
Pluggable Virt-Stack Components

harshitgupta1337:
### VEP Metadata

**Tracking issue**: #82 
**SIG label**: /sig compute

### What this PR does

This VEP proposes a refactoring of KubeVirt to decouple it from the default virtualizations stack components, i.e., QEMU and LibVirt, while also adopting a plugin-model where the backend logic for alternate virtualization stacks will not be included in the core `kubevirt/kubevirt` repository.

This is a next step to the Hypervisor Generalization VEP (#98 ) which introduced in-tree support for multiple hypervisors, along with backend logic for MSHV.

This VEP is a continuation of the discussion carried out in a previous VEP: https://github.com/kubevirt/enhancements/pull/83


iholder101:
@harshitgupta1337 should this be reopened?

harshitgupta1337:
/reopen
Reopening this PR as this is the main one tracking the design on pluggable virtualization stacks.

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,153 @@
...
+Furthermore, although libvirt provides a useful unified Domain definition across virtualization stacks, it is redundant in the presence of KubeVirt’s own VirtualMachineInstance definition.
+Additionally, KubeVirt only utilizes around 20% (~57/292) of all the Libvirt APIs available –leaving Libvirt significantly underutilized while virt-launcher still incurs the overhead of running the Libvirt daemon.
+
+- Maintaining multiple virtualization stacks in-tree is unmaintainable: Implementing support for different virtualization stacks directly in the core KubeVirt repository would significantly increase code complexity, testing burden, and maintenance overhead. A plugin-based architecture allows each virtualization stack implementation to be developed and maintained independently, reducing the maintenance burden on the core KubeVirt project.
+
+### Areas of Tight Coupling Between KubeVirt and Libvirt/QEMU/KVM
```
enp0s3, line 31: @harshitgupta1337 I think there is also coupling on the metrics level, some of the VM related metrics are being instrumented based on libvirt DomainStats.

harshitgupta1337:
@jean-edouard @vladikr @EdDev @xpivarc @iholder101 @enp0s3 
Can you pls review the section in the VEP that compares the two options of (1) retaining LibVirt in the virt-stack, and (2) having virt-launcher directly talk to the VMM?

At moment, my design proposal assumes no LibVirt - and I have enumerated reasons for it. However, I also understand that retaining LibVirt would make this refactoring smaller and could be the better suited next step toward virt-stack generalization. In the future, we could think about removing LibVirt entirely.

Please let me know your thoughts. Thanks.

harshitgupta1337:
@xpivarc  @vladikr @iholder101 @EdDev @andreabolognani @jean-edouard @enp0s3 
I have made all the changes from my side. Please review when you get a chance.

lyarwood (commented):
Apologies, I thought this was the implementation PR when I saw 20 commits and a target branch of `kubevirt:main`!

lyarwood (commented):
Thanks for reminding me again about this @harshitgupta1337, quick initial note before a more thorough review in the coming days.

The VEP covers the runtime plugin boundaries well but doesn't address how the admission webhook/validation layer becomes pluggable or stack-aware. Many validations assume libvirt/QEMU (machine types, SMM/UEFI, CPU feature flags etc.) and would be wrong or irrelevant for Firecracker/OpenVMM.

More broadly, the api.Domain open question understates the problem, libvirt/QEMU semantics are baked in throughout our CRDs: Machine.Type is documented as "QEMU machine type", CPU.Model references the libvirt CPU model map, CPUFeature.Policy uses libvirt policy names verbatim, Disk.IO says "which QEMU disk IO mode should be used", and cluster config has OVMFPath/EmulatedMachines/MachineType as first-class fields. The v1 API is effectively a libvirt abstraction, the VEP needs to call this explicitly IMHO, even if it doesn't aim to address it directly straight away.

harshitgupta1337:
> The VEP covers the runtime plugin boundaries well but doesn't address how the admission webhook/validation layer becomes pluggable or stack-aware. Many validations assume libvirt/QEMU (machine types, SMM/UEFI, CPU feature flags etc.) and would be wrong or irrelevant for Firecracker/OpenVMM.

Agreed. Admission webhooks also need to be made pluggable. I will work on adding that to the VEP design.

> 
> More broadly, the api.Domain open question understates the problem, libvirt/QEMU semantics are baked in throughout our CRDs: Machine.Type is documented as "QEMU machine type", CPU.Model references the libvirt CPU model map, CPUFeature.Policy uses libvirt policy names verbatim, Disk.IO says "which QEMU disk IO mode should be used", and cluster config has OVMFPath/EmulatedMachines/MachineType as first-class fields. The v1 API is effectively a libvirt abstraction, the VEP needs to call this explicitly IMHO, even if it doesn't aim to address it directly straight away.

Definitely. The v1 API is tied to LibVirt semantics. I have an open question related to API which is specific to requesting non-LibVirt-QEMU features on VMs created using alternate virt-stacks. You have raised a more fundamental point though, which I will also call out as an open question.

lyarwood:
> > The VEP covers the runtime plugin boundaries well but doesn't address how the admission webhook/validation layer becomes pluggable or stack-aware. Many validations assume libvirt/QEMU (machine types, SMM/UEFI, CPU feature flags etc.) and would be wrong or irrelevant for Firecracker/OpenVMM.
> 
> Agreed. Admission webhooks also need to be made pluggable. I will work on adding that to the VEP design.

Awesome thanks!

> > More broadly, the api.Domain open question understates the problem, libvirt/QEMU semantics are baked in throughout our CRDs: Machine.Type is documented as "QEMU machine type", CPU.Model references the libvirt CPU model map, CPUFeature.Policy uses libvirt policy names verbatim, Disk.IO says "which QEMU disk IO mode should be used", and cluster config has OVMFPath/EmulatedMachines/MachineType as first-class fields. The v1 API is effectively a libvirt abstraction, the VEP needs to call this explicitly IMHO, even if it doesn't aim to address it directly straight away.
> 
> Definitely. The v1 API is tied to LibVirt semantics. I have an open question related to API which is specific to requesting non-LibVirt-QEMU features on VMs created using alternate virt-stacks. You have raised a more fundamental point though, which I will also call out as an open question.

Yeah thanks, we are starting to see projects attempt to wrap kubevirt with more abstract VM APIs but I wonder if this VEP means we could start thinking about a v2 agnostic VirtualMachine API finally :D

jean-edouard:
Thanks for the thorough write-up @harshitgupta1337, and sorry for the late response.
TL;DR: I'm a design 2 guy all the way

KubeVirt already has an API for abstracting the hypervisor: libvirt. We put significant effort into integrating it into containers, and now is the time to reap the benefits of that investment rather than build a second abstraction layer that duplicates its purpose.

Everything that happens *before* running VMs (node labeling, capability discovery, pod rendering, admission webhooks) I'm happy to put behind a pluggable API. But the VM lifecycle boundary should remain libvirt. The cost of writing a libvirt driver for a new VMM is the correct cost of joining an ecosystem that chose libvirt as its abstraction layer, and it's far cheaper in aggregate than having every plugin reimplement lifecycle management, migration, and event handling from scratch.

I'd suggest separating "pluggable control-plane pieces" (good, let's do it) from "remove libvirt" (separate, much larger discussion that shouldn't gate multi-hypervisor progress).

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,538 @@
...
+
+- Hardcoded Libvirt/QEMU/KVM-specific values in control-plane components: Code of components such as virt-controller and virt-handler contain hardcoded references to one of Libvirt, QEMU or KVM/MSHV. For instance, for computing the memory overhead of the virt-launcher's components, the `LauncherHypervisorResources` interface implementations for both KVM and MSHV assume the presence of a LibVirt daemon and QEMU process in the virt-launcher. Another example is when the virt-handler is called to update the `memlock` limit of a VM's VMM process, and it explicitly looks for a `virtqemud` or `qemu-system-x86` process.
+
+- Libvirt/QEMU-specific guest agent status determination: Virt-Handler checks the state of the Libvirt channel "org.qemu.guest_agent.0” to determine if the guest agent is connected. It also assumes that the guest agent is QEMU Guest Agent and compares the list of supported commands with the set of required commands to determine if the agent is supported.
```
andreabolognani, line 43: That's just the name that libvirt uses by default for the channel. You could have the same name if you used QEMU directly, or you could configure a different name via libvirt.

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,538 @@
...
+
+#### Decision
+
+We choose Design 1, making the entire `virt-launcher` pluggable. The principal reasons are:
+
+- **Support for stacks without Libvirt drivers:** OpenVMM and Firecracker do not have Libvirt drivers, and future virtualization stacks may make the same choice. Requiring Libvirt would exclude such stacks or require their maintainers to first build and maintain a Libvirt driver, significantly raising the cost of integrating with KubeVirt.
+
+- **Independent feature delivery:** Even when a Libvirt driver exists, new VMM capabilities must first be represented in Libvirt's API and domain XML before a KubeVirt plugin can use them. Direct integration allows plugin maintainers to expose stack features and fixes on their own release cadence, without waiting for changes to propagate through an additional project and abstraction layer.
+
+- **Mismatch with KubeVirt's process model:** Libvirt is designed to manage multiple domains on a host, whereas each KubeVirt `virt-launcher` pod runs a dedicated Libvirt instance that manages a single VMI. KubeVirt therefore pays for a general-purpose, multi-domain management daemon without using one of its primary architectural benefits.
```
andreabolognani, line 448: This downside is overstated IMO. libvirt can manage multiple instances, but is quite happy managing a single one too.

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,538 @@
...
+
+- **Resource overhead:** A Libvirt daemon and its supporting processes consume memory in every `virt-launcher` pod. Removing this mandatory layer lets lightweight VMM plugins preserve their resource-density advantages and makes the per-VMI overhead proportional to the selected stack.
+
+- **Reduced mandatory trusted code:** Libvirt is implemented primarily in C and adds a large, memory-unsafe codebase to every launcher pod. Design 1 does not guarantee that plugins are memory-safe, but it avoids requiring this particular component and allows a plugin to minimize its runtime dependencies and attack surface.
```
andreabolognani, line 452: While it is true that libvirt being implemented in C leaves it open to memory safety issues, it's also a very mature codebase that's still actively maintained. Security issues are addressed promptly. And in any case, per your proposal the KubeVirt core would still use libvirt to manage QEMU instances, so the overall risk would not really be reduced by having plugins avoid it.

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,538 @@
...
+
+- Node Labeling involves invoking Libvirt’s QEMU driver: During the initialization of KubeVirt’s virt-handler component on a given node, it generates labels for that node based on the node’s virtualization capabilities. The virt-handler component queries node’s virtualization capabilities by running a virt-launcher container and invoking multiple Libvirt APIs, which in turn query the QEMU VMM. In addition to virtualization capabilities, the node topology is also queried from Libvirt, although that is not tied to the virtualization stack nor used for node labeling.
+
+- Hardcoded Libvirt/QEMU/KVM-specific values in control-plane components: Code of components such as virt-controller and virt-handler contain hardcoded references to one of Libvirt, QEMU or KVM/MSHV. For instance, for computing the memory overhead of the virt-launcher's components, the `LauncherHypervisorResources` interface implementations for both KVM and MSHV assume the presence of a LibVirt daemon and QEMU process in the virt-launcher. Another example is when the virt-handler is called to update the `memlock` limit of a VM's VMM process, and it explicitly looks for a `virtqemud` or `qemu-system-x86` process.
```
andreabolognani, line 41: Here and throughout the document, libvirt is occasionally referred to as "LibVirt". That is an incorrect capitalization of the project's name. Please make sure the "v" is always lower-case.

veps/sig-compute/82-plugin-virtstack/vep.md
```diff
@@ -0,0 +1,538 @@
...
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Target version is explicitly mentioned and approved
+- [ ] (R) Graduation criteria filled
+
+## Overview
+
+At present KubeVirt can only be used to create and manage virtual machines (VMs) via Libvirt on the QEMU virtual-machine monitor (VMM), with only the hypervisor being configurable between KVM and MSHV. However, this design is limited and hard to maintain because in the wider industry, there are an increasing number of VMMs and hypervisors available - such as Cloud Hypervisor, OpenVMM (VMMs) and HyperV (hypervisor). Adding and maintaining support for so many backends in-tree in the core KubeVirt repository would make the code difficult to maintain. Furthermore, the dependence of KubeVirt on LibVirt forces any VMM that could be used to have a LibVirt driver, which is untenable.
```
andreabolognani, line 13: This feels like an overly biased take. Of course I'm not exactly unbiased myself O:-) But claiming that implementing libvirt drivers is "untenable" doesn't really stand up to scrutiny IMO. Many VMMs have libvirt drivers already, *including* two of those you use as examples (Cloud Hypervisor and HyperV). So going down that route is demonstrably feasible.

andreabolognani:
@harshitgupta1337 added some comments.

I won't be able to make it to today's meeting about the proposal, but I believe that @jean-edouard's take and mine are largely overlapping, so I'm confident the position will be adequately represented.

harshitgupta1337:
> Thanks for the thorough write-up @harshitgupta1337, and sorry for the late response. TL;DR: I'm a design 2 guy all the way

No worries at all and thanks a lot for your feedback.

> Everything that happens _before_ running VMs (node labeling, capability discovery, pod rendering, admission webhooks) I'm happy to put behind a pluggable API. But the VM lifecycle boundary should remain libvirt. 

Libvirt interaction is almost completely within the `virt-launcher`, and only exposed via the Command and Notify APIs. That's exactly the plugin boundary I am also proposing in this VEP. If we make the rest of core KubeVirt (control-plane and parts of virt-handler) pluggable, won't we be able to support a VMM that doesn't have Libvirt driver? 

Note that we are not going to make Libvirt driver obsolete. VMMs that have Libvirt driver can be much easily onboarded to KubeVirt by leveraging the current QEMU-based `virt-launcher` implementation.

Issue #312, opened 2026-05-15
VEP 312: Declarative VMState PVC


alromeros:
**Primary contact (assignee)**:

/assign @alromeros 

**Current Feature Stage**:

New

**Feature Gate**: 

DeclarativeVMState

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/313

**Timeline**:

* Alpha release target (x.y): v1.11
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:
This VEP introduces a declarative API for managing VMState PVCs (backend storage) through a new field in VMSpec. This enhancement addresses several key use cases:

- **VM Migration with State Preservation**: Enable migrating VMs to new clusters while preserving TPM/EFI state.
- **VM Templating**: Create VM templates with pre-configured EFI/TPM setup for consistent firmware settings.

> [!IMPORTANT]
> Please keep this description up to date.

PR #313, opened 2026-05-15
VEP #312: Declarative VMState PVC

alromeros:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/312
**SIG label**: /sig storage

### What this PR does

This PR introduces a VEP to support declarative creation and lifecycle management of backend storage PVCs (VMState PVCs) through a new API field in VMSpec. This enables reusing existing VMState PVCs across VMs, facilitating migration and VM templating workflows.

### DIFF
--- a/veps/sig-storage/312-declarative-vmstate/vep.md
+++ b/veps/sig-storage/312-declarative-vmstate/vep.md
@@ -1,0 +1,362 @@
+# VEP #312: Declarative VMState PVC
+
+## VEP Status Metadata
+
+### Target releases
+
+- Alpha: v1.10
+- Beta: TBD
+- GA: TBD
+
+### Release Signoff Checklist
+
+- [x] (R) Enhancement issue created, links to this VEP dir
+- [ ] (R) Alpha target approved
+- [ ] (R) Beta target approved
+- [ ] (R) GA target approved
+
+## Overview
+
+This VEP introduces a declarative API field (`virtualMachineState`) in the VMI spec to reference or template VMState PVCs. It lets a VMState PVC be reused across VMs (for migration or templating) and gives explicit control over storage properties like size and storage class, replacing the implicit PVC creation that happens today when TPM, EFI, or CBT is enabled.
+
+## Motivation
+
+VMState PVCs are created implicitly when persistent TPM, persistent EFI, or CBT is enabled. Users cannot reuse an existing VMState PVC in a new VM, pre-provision PVC size (problematic for CBT metadata growth) or declaratively manage all VM resources (breaks GitOps workflows). The PVC is created with a randomly generated name and is not exposed anywhere in the VM manifest, so external tools like migration controllers or backup operators have no straightforward way to discover or manage it.
+
+Exposing the PVC in the API makes it discoverable and allows tools to manage it through standard Kubernetes patterns. This doesn't change what is stored or widen any security boundary, since the data was already persisted and PVCs remain namespace-scoped. The implicit path stays as the default for users who don't need PVC reuse, while the declarative path is opt-in for workflows like cloning, migration, templating and backup/restore.
+
+## Goals
+
+- Declarative API field to reference existing VMState PVCs or template new ones
+- VM migration with full state preservation (TPM, EFI, CBT)
+- Full VM cloning and templating support
+- Explicit control over PVC properties (size, name, storageClass)
+- Backward compatibility with implicit PVC creation
+- VM-agnostic PVC filesystem structure (no embedded UUIDs or VM identifiers)
+- Establish a consistent nomenclature for "VirtualMachineState" across Kubevirt and other projects. Previously, terms like "persistent state", "backend storage" and "VirtualMachineState" were used interchangeably.
+
+## Non Goals
+
+- Auto-migration of existing VMs to the declarative API
+- Removing implicit PVC creation (coexistence is required)
+- Multiple VMState PVCs per VM
+
+## Definition of Users
+
+- **VM owners:** manage VM lifecycle and state across clusters
+- **Platform operators:** provision storage classes and enforce resource policies
+- **GitOps tooling:** declaratively manage all VM resources via Git
+
+## User Stories
+
+- **Migration:** Reuse an existing VMState PVC in a new VM after cross-cluster migration or VM recreation, preserving TPM/EFI state without relying on implicit PVC naming or label manipulation
+- **Templating:** Create VM template with pre-configured EFI setup, reused across instantiated VMs
+- **Cloning:** Clone VM with existing TPM state to preserve guest OS trust
+- **GitOps:** Define all VM resources (including the VirtualMachineState PVC) in Git, apply declaratively
+- **CBT:** Pre-provision large VMState PVC to avoid CBT metadata space exhaustion
+
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+### API
+
+This VEP introduces a new `virtualMachineState` field of type `VirtualMachineStateSpec` in `VirtualMachineInstanceSpec`:
+
+```go
+type VirtualMachineStateSpec struct {
+    // Template the controller uses to create PVCs it owns.
+    // Used for the initial PVC when Source is unset, and for every new PVC
+    // the controller creates later for operations like live migration.
+    // +optional
+    VolumeClaimTemplate *v1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
+    // Source is an existing PVC to adopt as the starting state.
+    // +optional
+    Source *VirtualMachineStateSource `json:"source,omitempty"`
+}
+
+type VirtualMachineStateSource struct {
+    // Name of the source PVC.
+    Name string `json:"name"`
+    // A struct wrapping a single `name` today so the source can grow later if necessary (other kinds like volume snapshots, namespace, adoption policy) without a breaking change.
+}
+```
+
+The two fields are not mutually exclusive. At least one must be set:
+- `volumeClaimTemplate` only: controller creates and owns the PVC.
+- `source` only: adopt an existing PVC as-is.
+- both: adopt `source` as the starting PVC, and use the template for every PVC the controller creates afterwards (this is what enables RWO live migration on the adoption path).
+
+VolumeMode must be Filesystem, other values are rejected.
+
+On the `volumeClaimTemplate` path the controller owns the name via `GenerateName` (like generic ephemeral volumes) and records the live name in status. `volumeClaimTemplate.metadata.name` is rejected and users who need a fixed name pre-create the PVC and set `source`.
+
+Each VMState PVC gets two labels, whether it was created from a template or adopted via `source`. One is a lock that tells if the PVC is in use, so two VMs don't write the same state at once. The other maps the PVC to its VM, so the controller can find it again if it crashes before writing status. The lock is the source of truth for concurrent use: another VM may adopt the PVC only when the current holder is not running.
+
+Status reports the live PVC via `VolumeStatus` and is the source of truth for its name, which may differ from `source` after a migration. 
+
+```go
+  // VirtualMachineStateVolume tracks the live VirtualMachineState PVC
+	// (UEFI, TPM, CBT state).
+	// +nullable
+	// +optional
+	VirtualMachineStateVolume *VolumeStatus `json:"virtualMachineStateVolume,omitempty" optional:"true"`
+```
+
+Virt-controller populates this for both declarative and implicit paths.
+
+### Validation
+
+- At least one of `source` or `volumeClaimTemplate` must be set. They are not mutually exclusive
+- Immutability: `virtualMachineState` fields are immutable after VM creation, except for the template's storage capacity and storageClassName
+- Volume mode: referenced PVCs must use Filesystem volume mode
+- Template name: `volumeClaimTemplate.metadata.name` is rejected.
+- PVC existence: if a referenced PVC does not exist or is deleted while the VM exists, virt-controller sets a `VirtualMachineStatePVCNotFound` condition on the VM status.
+- Exclusive ownership: if another running VM already references the same VirtualMachineState PVC, virt-controller prevents the VM from starting and reports a condition.
+
+### VirtualMachineState PVC Structure
+
+The current layout is VM-specific: TPM state lives under a UUID-named directory and EFI vars carry the VM name in the filename, which prevents PVC reuse across VMs.
+
+This VEP introduces a canonical, VM-agnostic layout:
+
+```
+/
+├── tpm/                  # Canonical TPM root (replaces UUID-named directory)
+│   └── tpm2/             # swtpm-managed subdirectory (preserved as-is)
+│       └── tpm2-00.permall
+├── swtpm-localca/        # Local CA for swtpm EK certificate generation
+├── efi/
+│   └── efi_vars.fd
+├── cbt/
+└── meta/
+```
+
+The `swtpm-localca/` directory is only written during first boot; on subsequent boots and adoptions, `swtpm_setup` runs with `--not-overwrite` and skips certificate generation.
+
+### Libvirt Path Mapping
+
+Libvirt expects TPM state at `<stateroot>/<vm-uuid>/tpm2/` and does not expose a configurable state path in domain XML. The VMState PVC is mounted as a whole volume (no SubPath) at `VMStatePVCMountPath`, and virt-launcher creates an ephemeral symlink before starting libvirt:
+
+```
+PathForSwtpm(vmi)/<vm-uuid>  →  VMStatePVCMountPath/tpm/
+```
+
+Libvirt resolves the UUID directory via the symlink and uses existing state. EFI does not need a symlink as the NVRAM path is directly configurable in domain XML.
+
+This keeps privileges and side effects to a minimum. Creating the symlink with `ln -s` needs no capabilities, and virt-launcher can read it and reach the target volume without a custom SELinux policy. The symlink is ephemeral: it disappears when the pod terminates, and once the initial state is written the PVC itself is never modified.
+
+### Legacy PVC Migration
+
+Opting in on an existing VM, by adding `virtualMachineState` that points at a legacy PVC, triggers a one-time normalization: virt-launcher rewrites the legacy layout to the canonical one on first boot. TPM and EFI are handled independently, each guarded by its own idempotency gate so a second boot does nothing.
+
+Virt-launcher identifies a legacy VMState PVC by its `persistent-state-for` label, which carries the original VM name (legacy PVCs also use the `persistent-state-for` GenerateName prefix). Virt-controller checks for that label before treating a referenced PVC as a legitimate legacy PVC.
+
+TPM migration is skipped if `/tpm/` already exists. Otherwise virt-launcher scans the top level for subdirectories, ignores `lost+found`, and keeps only UUID-formatted names (`^[0-9a-f]{8}-...-[0-9a-f]{12}$`). A single match is renamed to `/tpm/` through an atomic two-phase rename (`→ /tpm.migrating → /tpm`), so a crash mid-rename can't leave a half-migrated directory. If more than one UUID directory matches, virt-launcher can't tell which is the real one, so it emits a warning event on the VMI listing the candidates, skips TPM migration, and starts from fresh state, leaving the old directories in place for manual recovery. EFI-only PVCs have no UUID directories, so this step is a no-op.
+
+EFI migration is skipped if `efi/efi_vars.fd` already exists. Otherwise virt-launcher looks in `nvram/` for a `*_VARS.fd` file and renames it to `efi/efi_vars.fd`. The implicit path names that file `nvram/<vmname>_VARS.fd` after `vmi.Name`, which is what ties it to a single VM.
+
+As a concrete example, when VM-B adopts a legacy PVC that VM-A used (`persistent-state-for-abcde`):
+
+Before:
+```
+/
+├── aaa-111/              # VM-A's UUID directory
+│   └── tpm2/
+│       └── tpm2-00.permall
+├── nvram/
+│   └── vm-a_VARS.fd      # VM-A's name embedded in filename
+├── swtpm-localca/
+└── lost+found/
+```
+
+After:
+```
+/
+├── tpm/                  # renamed from aaa-111/
+│   └── tpm2/
+│       └── tpm2-00.permall
+├── efi/
+│   └── efi_vars.fd       # renamed from nvram/vm-a_VARS.fd
+├── swtpm-localca/
+├── cbt/
+├── meta/
+└── lost+found/
+```
+
+Once the renames are done virt-launcher creates any missing canonical directories (`/tpm/`, `/efi/`, and so on). From then on both gates are satisfied, so every subsequent boot skips migration entirely.
+
+### Implicit Persistence When `virtualMachineState` Is Set
+
+When `virtualMachineState` is set, any enabled TPM or EFI feature is treated as persistent even without `persistent: true`. Providing a state PVC already implies the state should be kept, so a separate `persistent: true` on each device is unnecessary:
+
+```yaml
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          tpm: {}              # automatically persistent
+        firmware:
+          bootloader:
+            efi: {}            # automatically persistent
+      virtualMachineState:
+        volumeClaimTemplate: {}
+```
+
+If `virtualMachineState` is set with no TPM/EFI/CBT, the PVC is created with empty canonical directories.
+
+Setting `persistent: false` on a device is not rejected. The device stays ephemeral and KubeVirt does not write its state to the PVC, same as the implicit API today. The PVC is still provisioned or adopted. If the device is later set persistent, or CBT is enabled, KubeVirt uses the PVC as usual.
+
+### State Adoption
+
+Adoption is the case where VM-B starts against a PVC that VM-A used before. For TPM state it works without touching the PVC at all:
+
+1. VM-A ran with symlink: `<stateroot>/aaa-111` → `VMStatePVCMountPath/tpm/`. swtpm wrote state to PVC through the symlink.
+2. VM-A stopped. Pod terminated, symlink gone. PVC unchanged.
+3. VM-B starts with same PVC. Virt-launcher creates symlink: `<stateroot>/bbb-222` → `VMStatePVCMountPath/tpm/`. Libvirt finds existing state through the new symlink.
+
+The PVC filesystem is identical before and after adoption for TPM and EFI state. Only the ephemeral symlink changes. EFI is even simpler and needs no symlink or rename, since the domain XML points straight at `VMStatePVCMountPath/efi/efi_vars.fd` no matter which VM is running, so VM-B just opens the same file VM-A wrote.
+
+The label on the PVC also lets the controller tell whether another VM already uses it, so adoption of a PVC still in use can be detected.
+
+CBT is the exception. Its metadata tracks block changes for one specific VM's disks and means nothing to a different VM, so virt-launcher clears `cbt/` whenever a different VM adopts the PVC. The cost is a single full backup on the next backup cycle.
+
+### Live Migration
+
+Live migration on RWO needs a new PVC on the destination, which the controller can only create if it has a template.
+
+**RWX:** source and target pods mount the same PVC. State moves over the libvirt migration stream, no second PVC.
+
+**RWO with a template:** the controller creates a new PVC on the destination from the template, migrates over the libvirt stream, and records the new name in `status.virtualMachineStateVolume`. This covers `volumeClaimTemplate` alone and the adoption path (`source` + `volumeClaimTemplate`).
+
+**RWO with `source` only (no template):** the controller has no spec to create a destination PVC, so live migration is blocked with a condition and the reason is surfaced on the VM. The VM still starts and cold-migrates. Users who want RWO live migration on an adopted PVC can add a template.
+
+After a migration the live PVC differs from `source`. Its name is authoritative in `status.virtualMachineStateVolume`, and the label on the PVC lets the controller recover the mapping if it crashes before writing status.
+
+### Garbage Collection
+
+- PVCs created from `volumeClaimTemplate` get an OwnerReference to the VM and are deleted with it. When a migration replaces one, Kubevirt deletes the old PVC too.
+- PVCs referenced via `source` may not have been created by Kubevirt, so it never deletes them. The user owns their lifecycle, including any left behind after a migration.
+
+### Behavior
+
+When `virtualMachineState` is set, it takes precedence over implicit creation. VMs without `virtualMachineState` continue using the implicit path unchanged.
+
+## API Examples
+
+### Template (owned PVC)
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: my-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          tpm: {}
+      virtualMachineState:
+        volumeClaimTemplate:
+          spec:
+            resources:
+              requests:
+                storage: 100Mi
+```
+
+### Adopt an existing PVC
+
+```yaml
+virtualMachineState:
+  source:
+    name: existing-vmstate-pvc
+```
+
+### Adopt an existing PVC with RWO live migration support
+
+The source PVC seeds the initial state; the template drives any PVC the controller creates on migration.
+
+```yaml
+virtualMachineState:
+  source:
+    name: existing-vmstate-pvc
+  volumeClaimTemplate:
+    spec:
+      storageClassName: fast
+      resources:
+        requests:
+          storage: 100Mi
+```
+
+### Migration with State Preservation
+
+```bash
+kubectl delete vm original-vm --cascade=orphan
+```
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: migrated-vm
+spec:
+  template:
+    spec:
+      domain:
+        devices:
+          tpm: {}
+      virtualMachineState:
+        source:
+          name: original-vm-state
+```
+
+## Alternatives
+
+### Metadata File on PVC to keep track of VM UUID and name
+
+Rejected due to clone/snapshot desync. Not Kubernetes-native.
+
+### Ephemeral Bind Mounts
+
+Rejected because it requires `CAP_SYS_ADMIN`. Cleanup is racy on crash. Symlinks provide identical semantics without these drawbacks.
+
+### Directory Renaming on PVC per Adoption
+
+Instead of symlinks, rename the UUID directory on the PVC to match the new VM's UUID each time a different VM adopts the PVC.
+
+Rejected for a few reasons. It mutates persistent storage on every adoption: the PVC filesystem changes each time a VM starts even though the underlying state (TPM keys, EFI vars) is unchanged, whereas the symlink approach leaves the PVC untouched after the initial write. It is also unsafe on crash, because a rename interrupted by a node crash or pod eviction can leave the directory under neither the old nor the new name, requiring a scan for partial renames; the symlink has no such window since it lives outside the PVC. Finally, it breaks snapshots and clones, since a snapshot taken while the directory still carries VM-A's UUID must be renamed before VM-B can use it, while canonical paths are usable by any VM as-is.
+
+## Update/Rollback Compatibility
+
+The `DeclarativeVMState` feature gate controls **API admission only**, but not PVC or api management. VMs that were created with `virtualMachineState` in their spec continue using the declarative path even if the gate is later disabled.
+
+## Functional Testing Approach
+
+- Template creates owned PVC with canonical structure, deleted with VM
+- PVC clone/snapshot: new VM boots with cloned state
+- Legacy migration: PVC auto-normalized to canonical when user adds `virtualMachineState`
+- Adoption of existing PVC: new VM boots with existing state.
+
+## Graduation Requirements
+
+### Alpha (v1.10)
+
+- `DeclarativeVMState` feature gate (opt-in)
+- `virtualMachineState` field in `VirtualMachineInstanceSpec`, `virtualMachineStateVolume` (`VolumeStatus`) in VM/VMI status
+
+### Beta
+
+- Feature gate enabled by default after 1-2 releases
+
+### GA
+
+- Feature gate removed
+
+## References
+
+- Enhancement issue: [VEP #312](https://github.com/kubevirt/enhancements/issues/312)
+- Upstream libvirt TPM documentation: https://libvirt.org/formatdomain.html#tpm-device


veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Definition of Users
+
+- **VM owners:** manage VM lifecycle and state across clusters
+- **Platform operators:** provision storage classes and enforce resource policies
+- **GitOps tooling:** declaratively manage all VM resources via Git
+
+## User Stories
+
+- **Migration:** Migrate VM to new cluster, preserving TPM state so guest OS trusts the platform
```
Acedus: Why isn't that the case right now with implicit creation?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+    // Storage size. Default: 10Mi without CBT, 1Gi with CBT as metadata grows with disk size and backup count.
+    // +optional
+    Storage *resource.Quantity `json:"storage,omitempty"`
+    // Access mode for the PVC.
+    // +optional
+    AccessMode *corev1.PersistentVolumeAccessMode `json:"accessMode,omitempty"`
+}
+```
+
+`VMStateTemplateSpec` is purpose-built rather than reusing `DataVolumeTemplate` because VMState PVCs are always Filesystem mode and never need CDI import/clone/upload. Users who need full PVC control can pre-create the PVC and reference it.
```
Acedus, line 96: Should we ensure that the provided PVC in the case of not using the VMStateTemplate option is a filesystem PVC?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+└── lost+found/
+```
+
+TPM migration renames `aaa-111/ → tpm/` (atomic two-phase via `tpm.migrating`). EFI migration renames `nvram/vm-a_VARS.fd → efi/efi_vars.fd`. Both are skipped on subsequent boots since their idempotency gates are already satisfied. After migration, VM-B and any future VM use the canonical paths without further PVC mutation.
+
+### Live Migration
+
+With RWX, both source and target pods mount the same VMState PVC during migration. State transfers via the libvirt migration stream and no second PVC is created.
+
+RWO live migration is not supported in the initial implementation. The current backend storage RWO mechanism creates and swaps PVCs with generated names, which conflicts with the deterministic PVC naming in the declarative API. RWO migration support is planned as a follow-up.
```
Acedus: Does that mean that using this feature will prevent the use case of RWO live migration with a VMState PVC?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+│   └── efi_vars.fd
+├── cbt/
+└── meta/
+```
+
+The `swtpm-localca/` directory is only written during first boot; on subsequent boots and adoptions, `swtpm_setup` runs with `--not-overwrite` and skips certificate generation.
+
+### Libvirt Path Mapping
+
+Libvirt expects TPM state at `<stateroot>/<vm-uuid>/tpm2/` and does not expose a configurable state path in domain XML. The VMState PVC is mounted as a whole volume (no SubPath) at `VMStatePVCMountPath`, and virt-launcher creates an ephemeral symlink before starting libvirt:
```
Acedus, line 151: Are you sure that's the case? https://libvirt.org/formatdomain.html#tpm-device I think we can specify source in the domain XML now which would save us the symlink workaround.

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+    // Mutually exclusive with PersistentVolumeClaim.
+    // +optional
+    Template *VMStateTemplateSpec `json:"template,omitempty"`
+    // References an existing PVC (independent lifecycle, no OwnerReference).
+    // Mutually exclusive with Template.
+    // +optional
+    PersistentVolumeClaim *corev1.LocalObjectReference `json:"persistentVolumeClaim,omitempty"`
+}
+
+type VMStateTemplateSpec struct {
```
Acedus, line 77: Can't we just embed `PersistentVolumeClaimSpec` and `ObjectMeta`?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+Virt-controller populates this for both declarative and implicit paths:
+
+```bash
+kubectl get vm my-vm -o jsonpath='{.status.vmState.claimName}'
+```
+
+### Validation
+
+- Mutual exclusivity: only one of `template` or `persistentVolumeClaim`
+- Immutability: only capacity can be modified after VM creation; other `vmState` fields are immutable
```
Acedus: What about scenarios where the user would like to change the VMState PVC's storage class (i.e., via storage live migration)?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+
+Virt-controller populates this for both declarative and implicit paths:
+
+```bash
+kubectl get vm my-vm -o jsonpath='{.status.vmState.claimName}'
+```
+
+### Validation
```
Acedus, line 120: I think we'd benefit from mutual exclusivity for the PVC ownership too, for example, if VM A is currently referencing VMState PVC A then VM B cannot reference it as well otherwise we risk state corruption for RWX VMState PVCs.

Acedus (commented):
Good start @alromeros!
I think explanation about CBT is a bit under-specified, though. For example, should the CBT subpath be wiped on VMState PVC handover between two VMs? (I'd say yes, even for CNLM/CCLM since the cost is only a full backup, but we can discuss it).

veps/sig-storage/312-declarative-vmstate.md
alromeros: MTC storage migrations currently fail because the recreated VM still references the old backend storage PVC. Any tool implementing VM migration today requires manual fixes (matching UUIDs, label manipulation, etc.) that could be avoided by simply allowing PVC reuse through a declarative spec.

veps/sig-storage/312-declarative-vmstate.md
alromeros: Yes, we should validate that the referenced PVC is a Filesystem volume. For legacy PVC adoption specifically, we also validate the `persistent-state-for` label to confirm it's a legitimate VMState PVC.

veps/sig-storage/312-declarative-vmstate.md
alromeros: That's for passthrought modeI which means passing a physical TPM device rom the host to the guest. In kubevirt we require swtpm emulation which expects the UUID path described in the vep.

veps/sig-storage/312-declarative-vmstate.md
alromeros: RWO live migration continues to work with the implicit API, no regression there. For the declarative API, the problem is that the current RWO mechanism creates and swaps PVCs with generated names, which conflicts with the deterministic naming. If RWO migration for the declarative path is a requirement for alpha, I can prioritize designing it.

veps/sig-storage/312-declarative-vmstate.md
alromeros: Good point.

veps/sig-storage/312-declarative-vmstate.md
alromeros: Instead of `Template`, right? We can consider it but might be unnecessary to ask for a full PVC configuration if not all fields are supported (namespace, volumeMode...).

veps/sig-storage/312-declarative-vmstate.md
alromeros: Good idea, we can look into it.

veps/sig-storage/312-declarative-vmstate.md
Acedus: No I mean, within template. The fields that aren't supported can be rejected at admission.

alromeros:
> Good start @alromeros! I think explanation about CBT is a bit under-specified, though. For example, should the CBT subpath be wiped on VMState PVC handover between two VMs? (I'd say yes, even for CNLM/CCLM since the cost is only a full backup, but we can discuss it).

Good point. I was actually thinking of just migrating the CBT data but I agree it should be wiped on adoption. I'll flesh out the CBT behavior in the VEP

veps/sig-storage/312-declarative-vmstate.md
awels: MTC would not be a consideration in this case, but the kubevirt-migration-controller will probably have the same issue.

akalenyu: I am pretty sure we test storage migration with backend storage
/cc @jpeimer

jpeimer: Correct, we do test VMs with vTPM - both Storage Live Migration (using kubevirt-migration-controller in 4.20+) and Cross-Clutser Live Migration (using MTV).

veps/sig-storage/312-declarative-vmstate.md
alromeros: I'm still torn between the two. It's true that your option provides native K8s muscle memory. If you are okay with it, let's leave it as is for now and see what the others think about this.

veps/sig-storage/312-declarative-vmstate.md
alromeros: Just reworded that section to focus on PVC reuse and how it facilitates migration instead of implying migration itself is broken.

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,360 @@
...
+- [ ] (R) Beta target approved
+- [ ] (R) GA target approved
+
+## Overview
+
+This VEP introduces a declarative API field (`vmState`) in the VMI Spec to explicitly reference or template VMState PVCs. This enables PVC reuse across VMs (migration, templating) and explicit control over storage properties (size, name, class), replacing implicit PVC creation when TPM/EFI/CBT features are enabled.
+
+## Motivation
+
+VMState PVCs are created implicitly when persistent TPM, persistent EFI, or CBT is enabled. Users cannot reuse an existing VMState PVC in a new VM, pre-provision PVC size (problematic for CBT metadata growth), or declaratively manage all VM resources (breaks GitOps workflows). The PVC is also created with a randomly generated name, making it difficult to discover and manage. This VEP introduces a declarative API to address these issues.
```
EdDev, line 24: > Users cannot reuse an existing VMState PVC in a new VM, pre-provision PVC size (problematic for CBT metadata growth), or declaratively manage all VM resources (breaks GitOps workflows).

This is expressing the current state, which I assume was intentional.
If there is a need now to do manage all these, then please express in detail why it is needed at the user/usage/operation level.

> The PVC is also created with a randomly generated name, making it difficult to discover and manage.

If Kubevirt creates them, then I do not understand why it is difficult.
If it is not exposed not on the manifest (as spec or status), then it can be done with a good reasoning.

> This VEP introduces a declarative API to address these issues.

nit: This is obvious, IMO it can be dropped from the motivation section.

EdDev (changes requested):
I'm trying to figure out the full story here.

It seems the suggestion is to move from non-exposed ref to backend PVC, to an explicit one. And this is needed in order to preserve the PVC and its content beyond one VM lifecycle.

Points/questions that pop up are:
- Originally the data was bounded to the VM lifecycle and now it needs to go beyond that. Is that safe?
- Expressing the same need through two mechanisms without a plan of replacing them needs reasoning. I think we prefer one path, not multiple ones to reach the same target state. If backward support is needed, that one path should solve it.
- How is the new persisted PVC option will be used by other tools? Can the full use case be expressed in more detail so we can see the big picture here?

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Goals
+
+- Declarative API field to reference existing VMState PVCs or template new ones
```
EdDev, line 28: If a user can be explicit about a backend existing PVC to be use, won't it introduce a security issue?

alromeros:
> I'm trying to figure out the full story here.
> 
> It seems the suggestion is to move from non-exposed ref to backend PVC, to an explicit one. And this is needed in order to preserve the PVC and its content beyond one VM lifecycle.
> 
> Points/questions that pop up are:
> 
>     * Originally the data was bounded to the VM lifecycle and now it needs to go beyond that. Is that safe?

This would only apply to the `persistentVolumeClaim` (reference) field, not the template. With template, the PVC has an OwnerReference and is GC'd with the VM, which is the same lifecycle to today's implicit path, just with explicit spec instead of the generated one.

The reference form does decouple PVC lifecycle from a single VM, but it's designed more as a hardening of a mechanism that already exists today. Nowadays, vmrestore or migration process already leave the vmstate PVC alive outside the vm lifecycle, and the only way to reuse it is to mess with the persistent-state-for label and even the tpm directory naming so the new VM is able to pick it up. So I wouldn't say we're extending the lifecycle boundary, more like we're making an existing gap safer.

>     * Expressing the same need through two mechanisms without a plan of replacing them needs reasoning. I think we prefer one path, not multiple ones to reach the same target state. If backward support is needed, that one path should solve it.

I don't think template and implicit creation are exactly two ways to reach the same thing. Implicit stays because it's zero-config, you just set persistent: true and forget about it, and a lot of existing VMs rely on exactly that. Template is for users who specifically need a predictable PVC name, control over size (this matters a lot for CBT, since metadata can grow), etc. That need exists even if nobody ever reuses the PVC across VMs.

But you're right that we should say something about this instead of just leaving the two mechanisms coexist. I don't think we should deprecate implicit creation now, it's too early and there's no migration path for existing VMs, but can add a note saying that converging both into one path is a possible future direction after GA.

>     * How is the new persisted PVC option will be used by other tools? Can the full use case be expressed in more detail so we can see the big picture here?

See https://github.com/kubevirt/virt-template/issues/71: cloning a VM with persistent vTPM/EFI into a VirtualMachineTemplate fails because the backend storage PVC is implicit and never shows up in the VM's volumes list, so the DataVolumeTemplate entry created to clone it gets rejectetd. With the declarative api, the virt-template can do a volumesnapshot of the vmstate pvc as it does with other VM volumes and declaratively include it in the vm spec through the new field.

veps/sig-storage/312-declarative-vmstate.md
alromeros: Reusing an existing VMState PVC comes up in more than one flow: cloning, snapshot/restore, migration, and templating all need it, and today the only way to do any of them is the hacky "guess the generated name and manipulate the persistent-state-for label" route. If the name or uuid of the VM change now the pvc adoption is completely broken even with the label manipulation as kubevirt expects hardcoded paths for tpm and efi.
 
On the naming point, you're right that "difficult to discover" is a weak claim (even though it's a complain we constantly get from backup partners). But the real problem is you can't reference it anywhere once you find it, it's not on the VM spec or status, so there's nothing to point another VM at or template. The refactoring will also include a new vm status field to help identifying the pvc even with the implicit api.

veps/sig-storage/312-declarative-vmstate.md
alromeros: I'm not sure if this would really increase the security gap. VMState adoption is limited to the same namespace as the VM. As mentioned in the VEP, we could restrict adoption to PVCs that already carry the persistent-state-for label, to be more safe. Other than that, the worst case is a VM ending up using the EFI or TPM state from another VM.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,360 @@
...
+
+This VEP introduces a declarative API field (`vmState`) in the VMI Spec to explicitly reference or template VMState PVCs. This enables PVC reuse across VMs (migration, templating) and explicit control over storage properties (size, name, class), replacing implicit PVC creation when TPM/EFI/CBT features are enabled.
+
+## Motivation
+
+VMState PVCs are created implicitly when persistent TPM, persistent EFI, or CBT is enabled. Users cannot reuse an existing VMState PVC in a new VM, pre-provision PVC size (problematic for CBT metadata growth), or declaratively manage all VM resources (breaks GitOps workflows). The PVC is also created with a randomly generated name, making it difficult to discover and manage. This VEP introduces a declarative API to address these issues.
+
+## Goals
+
+- Declarative API field to reference existing VMState PVCs or template new ones
```
awels: The EFI and TPM state have the firmware UUID in their paths, so if for some reason it gets attached to a different VM, the firmware UUID will be different.

But the point of this is that certain operations (like cloning and backup/restore) require some manipulation of the backend state PVC. Sometimes we want a new one (clone), and sometimes we need the same one on a different VM (restore for instance). There is no good way of expression this right now, this the main point of the VEP IMO.

veps/sig-storage/312-declarative-vmstate.md
awels: > If Kubevirt creates them, then I do not understand why it is difficult.
If it is not exposed not on the manifest (as spec or status), then it can be done with a good reasoning.

I think this is the critical issue. It is not exposed on the manifest currently, and that was an oversight on our part. The point of the VEP is to expose in the manifest so it can be discovered, and the life cycle can be managed differently.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,356 @@
...
+    // Mutually exclusive with PersistentVolumeClaim.
+    // +optional
+    Template *VMStateTemplateSpec `json:"template,omitempty"`
+    // References an existing PVC (independent lifecycle, no OwnerReference).
+    // Mutually exclusive with Template.
+    // +optional
+    PersistentVolumeClaim *corev1.LocalObjectReference `json:"persistentVolumeClaim,omitempty"`
+}
+
+type VMStateTemplateSpec struct {
```
awels: I think I agree with Adi here, just embed meta data for the annotations/labels, I also think name should probably be generated from the template and not specified by the user. If you are specifying the name, then it is not a template, but a specification. We could use GenerateName so we can specify a prefix.

And the PVC spec just to capture and changes on a kubernetes level without having to rework it every time something changs.

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,360 @@
...
+```bash
+kubectl get vm my-vm -o jsonpath='{.status.vmState.claimName}'
+```
+
+### Validation
+
+- Mutual exclusivity: only one of `template` or `persistentVolumeClaim`
+- Immutability: only capacity and storageClassName can be modified after VM creation; other `vmState` fields are immutable. StorageClassName changes enable storage live migration.
+- Volume mode: referenced PVCs must use Filesystem volume mode. Block mode is incompatible with the directory-based canonical layout.
+- Volume existence: referenced PVCs must exist at VM creation time
```
awels: I don't think we should be checking this. We can note the missing PVC at runtime in a condition instead I think.

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,360 @@
...
+- Exclusive ownership: a VMState PVC can only be referenced by one VM at a time. Virt-controller rejects a VM that references a VMState PVC already in use by another VM, preventing concurrent state corruption on RWX volumes.
```
awels: I don't think we want to check this at validation time either. We can detect it at runtime and prevent the VM from starting or something of that nature. Doing these type of checks during validation is not recommended since the state of the cluster can change.

veps/sig-storage/312-declarative-vmstate.md
```diff
@@ -0,0 +1,360 @@
...
+└── lost+found/
+```
+
+TPM migration renames `aaa-111/ → tpm/` (atomic two-phase via `tpm.migrating`). EFI migration renames `nvram/vm-a_VARS.fd → efi/efi_vars.fd`. Both are skipped on subsequent boots since their idempotency gates are already satisfied. After migration, VM-B and any future VM use the canonical paths without further PVC mutation.
+
+### Live Migration
+
+With RWX, both source and target pods mount the same VMState PVC during migration. State transfers via the libvirt migration stream and no second PVC is created.
+
+RWO live migration is not supported in the initial implementation. The current backend storage RWO mechanism creates and swaps PVCs with generated names, which conflicts with the deterministic PVC naming in the declarative API. RWO migration support is planned as a follow-up. If a migration-capable VM has a non-RWX VMState PVC, virt-controller sets a `VMStatePVCNotMigratable` condition.
```
awels: This is where I think the template would work. If you have the template, with a GenerateName, then RWO will work, since it will just generate a new PVC from that template and it can copy the data like now. It won't do CBT data copying anyway. In the end we just update the status with the new PVC name.

alromeros: That's a good idea. The question is whether we should allow the same for a regular name. How bad would it be for the name in the status to differ from the one originally set in the spec during initial templating?

veps/sig-storage/312-declarative-vmstate.md
alromeros: Makes sense, ack!

veps/sig-storage/312-declarative-vmstate/vep.md
alromeros: Okay makes sense, ack

veps/sig-storage/312-declarative-vmstate.md
awels: I would expect whomever is interested in that information to watch the status, and to notice if the name of the PVC changed. This is not any different on how it works right now with a RWO state PVC. This just surfaces that information the user.

alromeros: Great, ~~so I'm just wondering if we should allow for pvc name updates. I updated the rwo migration section saying that we should support manually triggering migration by updating the names in the specs but maybe we should just follow the current flow and let the controller handle it internally and just update its name in the status.~~
Edit: Just updated the RWO migration section to specify that it's triggered automatically by the controller. Included the manual trigger by updating the name mentioned as an alternative.

awels (approved):

akalenyu (commented):
What if backend state was just a volume in `spec.volumes[]` (or `utilityVolumes[]`) with its own source type?
This is similar to how K8s admission controllers [inject projected volumes for SA tokens](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume)

The API could follow the K8s ephemeral volume pattern which already has an inline `volumeClaimTemplate` for PVC creation with lifecycle management:

```yaml
# Templated — KubeVirt creates and owns the PVC
volumes:
- name: vm-state
  vmState:
    volumeClaimTemplate:
      spec:
        accessModes: [ReadWriteOnce]
        resources:
          requests:
            storage: 64Mi

# Or referencing an existing PVC — user manages lifecycle
volumes:
- name: vm-state
  vmState:
    claimName: my-existing-state-pvc
```

alromeros:
> What if backend state was just a volume in `spec.volumes[]` (or `utilityVolumes[]`) with its own source type? This is similar to how K8s admission controllers [inject projected volumes for SA tokens](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume)
> 
> The API could follow the K8s ephemeral volume pattern which already has an inline `volumeClaimTemplate` for PVC creation with lifecycle management:
> 
> ```yaml
> # Templated — KubeVirt creates and owns the PVC
> volumes:
> - name: vm-state
>   vmState:
>     volumeClaimTemplate:
>       spec:
>         accessModes: [ReadWriteOnce]
>         resources:
>           requests:
>             storage: 64Mi
> 
> # Or referencing an existing PVC — user manages lifecycle
> volumes:
> - name: vm-state
>   vmState:
>     claimName: my-existing-state-pvc
> ```

Regarding adding it to `utilityVolumes`, I considered this initially but decided against it because utility volumes are intended for volumes that the VM can use and discard for specific utilities/operations. The vmstate PVC is long-lived and required for the VM to work.

Regarding `spec.Volumes` I might actually prefer that approach, but I remember hearing some pushback about it during the initial vmstate PVC implementation, since the vmstate PVC isn't a disk that the guest consumes. I can't find anything related in the original design proposal or related PRs though https://github.com/kubevirt/community/pull/170.

I initially chose a new vmstate volume to keep it separate from the existing spec.volumes API and its lifecycle, but I can also see how having all persistent volumes under the same API could be the cleanest approach.

akalenyu:
> > What if backend state was just a volume in `spec.volumes[]` (or `utilityVolumes[]`) with its own source type? This is similar to how K8s admission controllers [inject projected volumes for SA tokens](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume)
> > The API could follow the K8s ephemeral volume pattern which already has an inline `volumeClaimTemplate` for PVC creation with lifecycle management:
> > ```yaml
> > # Templated — KubeVirt creates and owns the PVC
> > volumes:
> > - name: vm-state
> >   vmState:
> >     volumeClaimTemplate:
> >       spec:
> >         accessModes: [ReadWriteOnce]
> >         resources:
> >           requests:
> >             storage: 64Mi
> > 
> > # Or referencing an existing PVC — user manages lifecycle
> > volumes:
> > - name: vm-state
> >   vmState:
> >     claimName: my-existing-state-pvc
> > ```
> 
> Regarding adding it to `utilityVolumes`, I considered this initially but decided against it because utility volumes are intended for volumes that the VM can use and discard for specific utilities/operations. The vmstate PVC is long-lived and required for the VM to work.
> 
> Regarding `spec.Volumes` I might actually prefer that approach, but I remember hearing some pushback about it during the initial vmstate PVC implementation, since the vmstate PVC isn't a disk that the guest consumes. I can't find anything related in the original design proposal or related PRs though [kubevirt/community#170](https://github.com/kubevirt/community/pull/170).
> 
> I initially chose a new vmstate volume to keep it separate from the existing spec.volumes API and its lifecycle, but I can also see how having all persistent volumes under the same API could be the cleanest approach.

Yeah I think we might have some precedent to these special volume types that aren't visible in guest, not totally sure. This was the natural approach for me having watched k8s do the entire token dance. I think we may be able to learn from them in that we conditionally insert this volume when we absolutely must.

/cc @jean-edouard 
in case you recall the issues with that approach (and any thoughts in general :pray:)

jean-edouard:
Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.

Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes *are* consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.

+1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.

akalenyu:
> Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> 
> Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> 
> +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.

Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.

What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.

jean-edouard:
> > Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> > Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> > +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.
> 
> Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.
> 
> What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.

I'm confused, are we both talking about the ServiceAccount iso or something else? Because the SA token it contains is only ever used by (the k8s client inside) the guest, not by virt-launcher. As opposed to the contents of backend-storage that are exclusively used by libvirt/qemu inside of virt-launcher (to of course ultimately translate into content meant for the guest, but that's beside the point).

akalenyu:
> > > Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> > > Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> > > +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.
> > 
> > 
> > Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.
> > What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.
> 
> I'm confused, are we both talking about the ServiceAccount iso or something else? Because the SA token it contains is only ever used by (the k8s client inside) the guest, not by virt-launcher. As opposed to the contents of backend-storage that are exclusively used by libvirt/qemu inside of virt-launcher (to of course ultimately translate into content meant for the guest, but that's beside the point).

I'm trying to draw a comparison with regular pod's https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume

jean-edouard:
> > > > Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> > > > Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> > > > +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.
> > > 
> > > 
> > > Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.
> > > What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.
> > 
> > 
> > I'm confused, are we both talking about the ServiceAccount iso or something else? Because the SA token it contains is only ever used by (the k8s client inside) the guest, not by virt-launcher. As opposed to the contents of backend-storage that are exclusively used by libvirt/qemu inside of virt-launcher (to of course ultimately translate into content meant for the guest, but that's beside the point).
> 
> I'm trying to draw a comparison with regular pod's https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume

My bad, should have clicked the link :facepalm:

alromeros:
> > > > > Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> > > > > Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> > > > > +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.
> > > > 
> > > > 
> > > > Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.
> > > > What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.
> > > 
> > > 
> > > I'm confused, are we both talking about the ServiceAccount iso or something else? Because the SA token it contains is only ever used by (the k8s client inside) the guest, not by virt-launcher. As opposed to the contents of backend-storage that are exclusively used by libvirt/qemu inside of virt-launcher (to of course ultimately translate into content meant for the guest, but that's beside the point).
> > 
> > 
> > I'm trying to draw a comparison with regular pod's https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume
> 
> My bad, should have clicked the link 🤦

It's fine for me to just allow `GenerateName` as that makes implementation and rwo support more consistent with the current implicit flow, but that would require rejecting `Name` at creation, which I'm not super fond of (users who want a specific name can always pre-create the pvc though).

Regarding the volumes list, I'm still torn between the two options. I like the precedents @akalenyu shared, but a vmstate PVC is not like other volumes: only one is allowed, it would need exceptions and validations across the whole code... I see how a specific field for it makes sense both during creation and status check.

Edit: Going back over this, I'm doubling down on having the separate `vmState` field. Ideally the `UtilityVolumes` would've been great to include non-ephemeral pvcs such as this one, as the api description explicitly mentions "`without having a matching disk in the domain`". But the api as it is so far is super dependent on the original implementation and is practically unusable for this case.

Reusing the main `volumes[]` list is still valid, but with the expectation of each volume there matching a disk (like we already discussed above), having to reject disk/fs references, having to add validations just to enforce the 1-item limit... It feels like we would be forcing it to fit.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,362 @@
...
+```go
+type VMStateVolumeSource struct {
+    // Creates a new PVC owned by this VM (garbage-collected on deletion).
+    // Mutually exclusive with PersistentVolumeClaim.
+    // +optional
+    Template *VMStateTemplateSpec `json:"template,omitempty"`
+    // References an existing PVC (independent lifecycle, no OwnerReference).
+    // Mutually exclusive with Template.
+    // +optional
+    PersistentVolumeClaim *corev1.LocalObjectReference `json:"persistentVolumeClaim,omitempty"`
```
alromeros: I should probably use a `PersistentVolumeClaimVolumeSource` here or simply a `claimName`

awels: Yes I think claimName is probably the most straight forward. I understand the LocalObjectReference, but it is never going to be anything but a PVC in the same namespace, so just giving the claimName should be sufficient.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,362 @@
...
+```bash
+kubectl get vm my-vm -o jsonpath='{.status.vmState.claimName}'
+```
+
+### Validation
+
+**Admission-time:**
+- Mutual exclusivity: only one of `template` or `persistentVolumeClaim`. Switching between the two could be relaxed later if needed
+- Immutability: `vmState` fields are immutable after VM creation, except for storage capacity, storageClassName, and maybe name if we end up allowing manual trigger of migration
+- Volume mode: referenced PVCs must use Filesystem volume mode
```
awels, line 114: Where are we enforcing this? I think we should do it in the virt-controller with a condition (same condition as below? In that case that needs a slightly different name, maybe ValidStatePVC?) Then if the PVC has the wrong volume mode we can set it to false with a reason of invalid volume mode block or something like that.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,362 @@
...
+└── lost+found/
+```
+
+TPM migration renames `aaa-111/ → tpm/` (atomic two-phase via `tpm.migrating`). EFI migration renames `nvram/vm-a_VARS.fd → efi/efi_vars.fd`. Both are skipped on subsequent boots since their idempotency gates are already satisfied. After migration, VM-B and any future VM use the canonical paths without further PVC mutation.
+
+### Live Migration
+
+**RWX:** Both source and target pods mount the same VMState PVC. State transfers via the libvirt migration stream and no second PVC is needed.
+
+**RWO:** For both paths, virt-controller creates a new PVC on the destination node (from the template or with matching specs), migrates state via the libvirt migration stream, and updates `status.vmState.claimName`. After creation, the actual PVC is tracked through `status.vmState.claimName`.
```
awels: I have a question here with a provided PVC name. A migration requires us to create a new PVC, which we can't do if the spec says we have to use this particular name (which already exists). How are we planning on reconciling that discrepancy? I am thinking maybe we should require that provided state PVCs are RWX? And only when you provide the template (or nothing at all) will it create RWO?

alromeros: I wouldn't reject the RWO + `claimName` combination. I'm currently torn between two approaches here: On one hand, if a user explicitly sets the `claimName`, they are taking full control of that PVC and should already know that live migrating RWO volumes is usually blocked. We could just block the migration with a condition.

On the other hand, we could take the "spec is intent, status is actual state" idea to the next level and trigger the migration and just create a new PVC on the destination using the template flow anyway. Then that new PVC will be reflected in the status.

To keep things simple for alpha, I'd personally say it's fine to block the migration if they use RWO with a `claimName`. We can always leave the auto-provisioning option for beta.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,351 @@
...
+> **Open question:** when a user sets `template.metadata.name`, reject it at admission or just ignore it and always generate?
+
+A new status fields on the VM status report VMState PVC state, following the same patterns as `VolumeStatus`:
+
+```go
+  // VirtualMachineStateVolume tracks the VirtualMachineState PVC status
+	// containing persistent VM state (UEFI, TPM, CBT)
+	// +nullable
+	// +optional
+	VirtualMachineStateVolume *VolumeStatus `json:"virtualMachineStateVolume,omitempty" optional:"true"`
```
alromeros, line 105: For context, see https://github.com/kubevirt/kubevirt/pull/16779. I initially considered creating a dedicated status struct for this, mainly because `VolumeStatus` has a bunch of fields that don't really apply to the vmstate PVC. Plus, eventually we might want to track vmstate-specific things here (like the tpm path, efi name or the original pvc name after a live migration). But for now, I'm following my original implementation and @akalenyu suggestion by reusing VolumeStatus.

veps/sig-storage/312-declarative-vmstate/vep.md
alromeros: I agree, but I think that's more of an implementation detail so I wouldn't set it in stone right now

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,351 @@
...
+### API
+
+This VEP introduces a new `virtualMachineState` field in `VirtualMachineInstanceSpec`:
+
+```go
+type VirtualMachineStateVolumeSource struct {
+    // Creates a new PVC owned by this VM.
+    // Mutually exclusive with ClaimName.
+    // +optional
+    VolumeClaimTemplate *v1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
```
alromeros, line 74: Here I was essentially reimplementing https://pkg.go.dev/k8s.io/api/core/v1#PersistentVolumeClaimTemplate

veps/sig-storage/312-declarative-vmstate/vep.md
akalenyu: we already extend volume status for extra per-volume info:
```go
// VolumeStatus represents information about the status of volumes attached to the VirtualMachineInstance.
type VolumeStatus struct {
...
	// If the volume is memorydump volume, this will contain the memorydump info.
	MemoryDumpVolume *DomainMemoryDumpInfo `json:"memoryDumpVolume,omitempty"`
	// ContainerDiskVolume shows info about the containerdisk, if the volume is a containerdisk
	ContainerDiskVolume *ContainerDiskInfo `json:"containerDiskVolume,omitempty"`
}
```

jean-edouard (approved):
Solid VEP, no notes!

awels (approved):

awels:
@jean-edouard if you could approve, lgtm from me as well.

jean-edouard:
> @jean-edouard if you could approve, lgtm from me as well.

I can't approve VEPs

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,351 @@
...
+
+- Feature gate enabled by default after 1-2 releases
+
+### GA
+
+- Feature gate removed
+
+## References
+
+- Enhancement issue: kubevirt/kubevirt#312
```
iholder101: nit

```suggestion
- Enhancement issue: https://github.com/kubevirt/enhancements/issues/312.
```

or


```suggestion
- Enhancement issue: kubevirt/enhancements#312
```

alromeros:
> > > > > > Nice work @alromeros! The API is a clear improvement over the current name/label matching mechanism.
> > > > > > Indeed, backend-storage cannot be a volume, since volumes are for guest consumption and backend-storage definitely isn't. The SA token analogy doesn't quite apply either, because projected volumes _are_ consumed by the guest, while backend-storage is consumed by virt-launcher/swtpm/OVMF on behalf of the guest. A dedicated `vmState` field at the spec level seems like the right place.
> > > > > > +1 to @awels on `GenerateName` vs `Name` in templates. If it's a template then `GenerateName` fits better. The actual PVC name should be discoverable through `status.vmState.claimName` regardless.
> > > > > 
> > > > > 
> > > > > Let me push back a bit on the projected volume analogy not applying. Projected SA token volumes aren't really consumed by the "guest" (the application in the pod), they're consumed by the k8s client SDK/libraries that know to look for a token at a well-known path.
> > > > > What I'm trying to say is that the guest **is** consuming the backend state (store keys), just like the app **is** consuming the token (auth), and there is similarity in that it's transparent to the guest.
...
> 
> Reusing the main `volumes[]` list is still valid, but with the expectation of each volume there matching a disk (like we already discussed above), having to reject disk/fs references, having to add validations just to enforce the 1-item limit... It feels like we would be forcing it to fit.

**Note for future reviewers/maintainers:** This topic was dicussed in a meeting with all involved reviewers and the consensus was to include a new field in the vmi spec and enable rwo live migration by making both fields not mutually exclusive as shown in the VEP.

veps/sig-storage/312-declarative-vmstate/vep.md
```diff
@@ -0,0 +1,351 @@
...
+Virt-controller populates this for both declarative and implicit paths.
+
+### Validation
+
+- At least one of `source` or `volumeClaimTemplate` must be set. They are not mutually exclusive
+- Immutability: `virtualMachineState` fields are immutable after VM creation, except for the template's storage capacity and storageClassName
+- Volume mode: referenced PVCs must use Filesystem volume mode
+- Template name: `volumeClaimTemplate.metadata.name` is rejected.
+- PVC existence: if a referenced PVC does not exist or is deleted while the VM exists, virt-controller sets a `VirtualMachineStatePVCNotFound` condition on the VM status.
+- Exclusive ownership: if another running VM already references the same VirtualMachineState PVC, virt-controller prevents the VM from starting and reports a condition.
```
vladikr, line 117: But now i don't understand.. 
On line 53, you say that """- **Templating:** Create VM template with pre-configured EFI setup, shared across instantiated VMs""" 
how these two live together?

alromeros: I admit the wording in line 53 is a bit odd. I was referring to this issue: https://github.com/kubevirt/virt-template/issues/71, where the intent was to have a pre-configured EFI used by all templated VMs. What I'm trying to convey here is that the efi configuration will be shared, not the VMState PVC itself (there's only one PVC per VM).

alromeros: Changed `shared` to `reused`.

veps/sig-storage/312-declarative-vmstate/vep.md
alromeros: ack!

vladikr:
Thanks for the latest update, I see that most of the topics we've chatted about offline are addressed. 
The two-label split is what we discussed - thanks!

Could we add the policy next to it so it doesn’t stay only on Slack? 
Something like:
The lock is the source of truth for concurrent use. another VM may adopt the PVC only when the current holder is not running. 
The mapping label is just crash recovery (which VM this PVC belongs to) and is not the lock.

I'm fine with leaving the identifier choice (uuid/name), caches, and how the lock is stored to the implementation PR. 
I just want that rule in the VEP so we don’t have to reconstruct it later.

vladikr:
Other than that, it looks good to me, so I'll approve
and we can remove the hold once this is comment is added.
/approve
/hold

vladikr:
/unhold 
Thanks!

akalenyu (approved):

PR #316, opened 2026-05-19
VEP #144: draft extension to propagate memlock limit via VMI status

lyarwood:
### VEP Metadata

**SIG label**: /sig compute

### What this PR does

Adds a draft proposal to extend VEP #144 (Memory Lock RLimit configuration) with a `MemLockLimit` status field on `MemoryStatus`.

Currently the memlock rlimit value is computed independently in virt-handler (for prlimit64) and virt-launcher (for domain XML `<memtune><hard_limit>`). This extension proposes computing it once in virt-controller and propagating via `vmi.Status.Memory.MemLockLimit`, eliminating parallel calculations.

This also addresses the gap between VEP #144's stated goal ("adjust memory lock limits **without impacting VMI scheduling capacity**") and the current implementation where `addedOverhead` flows into `GetMemoryOverhead()`, inflating pod memory requests.

### Key points

- **Status-only** -- no VMI spec changes, `addedOverhead` and `memLock` remain as inputs
- **Compute-once** -- virt-controller computes the final value from all inputs (VFIO device count, `addedOverhead`, `memLock` flag)
- **Decoupled from pod sizing** -- memlock limit does not inflate pod memory requests
- **Backwards compatible** -- consumers fall back to existing behaviour when status field absent

### Prior work

Builds on:
- kubevirt/kubevirt#17805 -- VFIO memlock scaling in virt-handler
- kubevirt/kubevirt#17857 -- shared `pkg/vfio` package and `<memtune><hard_limit>` in domain XML

### Timeline

- v1.10

### DIFF
--- a/veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
+++ b/veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
@@ -1,0 +1,155 @@
+# VEP #144 Extension: Propagate computed memlock limit via VMI status
+
+*Draft for discussion -- extends the existing VEP #144 (Memory Lock RLimit configuration)*
+
+> **Note:** This draft assumes [kubevirt/kubevirt#17805](https://github.com/kubevirt/kubevirt/pull/17805) (VFIO memlock scaling in virt-handler) and [kubevirt/kubevirt#17857](https://github.com/kubevirt/kubevirt/pull/17857) (shared `pkg/vfio` package and `<memtune><hard_limit>` in domain XML) have merged. #17805 is under review; #17857 is a draft dependent on it. The assumptions about `pkg/vfio`, `CountDevices()`, `CalculateMemlockLimit()`, and `<memtune><hard_limit>` support in the domain converter all come from these PRs.
+
+## Problem
+
+The memlock rlimit value is currently computed independently in multiple places:
+
+- **virt-handler** (`CalculateMemlockSize()`) -- for prlimit64 on the QEMU process
+- **virt-launcher** (domain converter) -- for `<memtune><hard_limit>` in the domain XML
+- **libvirt** (`qemuDomainGetMemLockLimitBytes()`) -- its own internal calculation
+
+These calculations can drift out of sync. Additionally, the inputs that determine the memlock value come from multiple sources:
+
+- Automatic VFIO device counting (GPUs, HostDevices, SRIOV) -- internally computed
+- VEP #144's `reservedOverhead.addedOverhead` -- user or webhook specified
+- VEP #144's `reservedOverhead.memLock` flag -- signals locking is required
+- Future sources (new device types, platform-specific requirements)
+
+There is no single place where the final memlock value is computed and no mechanism to propagate it to all consumers.
+
+## Proposal
+
+Add a **status-only** field `memLockLimit` to `MemoryStatus`. The value is computed once by virt-controller and consumed by both virt-handler and virt-launcher. No VMI spec changes.
+
+### New status field
+
+```go
+type MemoryStatus struct {
+    GuestAtBoot    *resource.Quantity `json:"guestAtBoot,omitempty"`
+    GuestCurrent   *resource.Quantity `json:"guestCurrent,omitempty"`
+    GuestRequested *resource.Quantity `json:"guestRequested,omitempty"`
+    MemoryOverhead *resource.Quantity `json:"memoryOverhead,omitempty"`
+    MemLockLimit   *resource.Quantity `json:"memLockLimit,omitempty"`
+}
+```
+
+### Compute-once flow
+
+```
+virt-controller
+  │
+  ├─ Inputs:
+  │    - VFIO device count from VMI spec (vfio.CalculateMemlockLimit)
+  │    - reservedOverhead.addedOverhead (VEP #144)
+  │    - reservedOverhead.memLock flag (VEP #144)
+  │    - Future: platform-specific overrides
+  │
+  └─ Writes vmi.Status.Memory.MemLockLimit
+       │
+       ├─ virt-handler reads → prlimit64 on QEMU process
+       │
+       └─ virt-launcher reads (via SyncVirtualMachine gRPC)
+            → <memtune><hard_limit> in domain XML
+```
+
+No pod annotation is needed. Unlike `MemoryOverheadAnnotationBytes` (which the pod needs at creation time for resource requests), the memlock value is consumed at runtime: virt-handler reads it from VMI status before the domain starts, and virt-launcher receives the full VMI object via gRPC.
+
+### How this unifies both approaches
+
+| Source | Triggers | Current path | With this extension |
+| :----- | :------- | :----------- | :------------------ |
+| VFIO devices (auto) | KubeVirt counts devices internally | Parallel calc in handler + launcher | Controller computes, writes to status |
+| `addedOverhead` (VEP #144) | User/webhook sets spec field | Flows into `GetMemoryOverhead()`, inflates pod request | Feeds controller's memlock calc, writes to status. Pod request handled separately. |
+| `memLock: Required` (VEP #144) | User/webhook sets flag | Triggers handler's memlock adjustment | Triggers controller to compute and write to status |
+
+The VMI spec fields (`addedOverhead`, `memLock`) remain as inputs -- they express user intent. The computed result lives only in status.
+
+### Memlock calculation in virt-controller
+
+```go
+func calculateMemLockLimit(vmi *v1.VirtualMachineInstance) *resource.Quantity {
+    // Start with automatic VFIO calculation
+    memlockBytes := vfio.CalculateMemlockLimit(vmi)
+
+    // Add user-specified overhead from VEP #144
+    if vmi.Spec.Domain.Memory != nil &&
+       vmi.Spec.Domain.Memory.ReservedOverhead != nil &&
+       vmi.Spec.Domain.Memory.ReservedOverhead.AddedOverhead != nil {
+        memlockBytes += vmi.Spec.Domain.Memory.ReservedOverhead.AddedOverhead.Value()
+    }
+
+    if memlockBytes == 0 {
+        // Check if memLock is explicitly required without devices
+        if util.RequiresLockingMemory(vmi) {
+            // Base memlock: guestMemory + standard overhead
+            memlockBytes = vfio.GetVirtualMemoryBytes(vmi) + vfio.MMIOOverheadBytes
+        }
+    }
+
+    if memlockBytes == 0 {
+        return nil
+    }
+    q := resource.NewQuantity(memlockBytes, resource.BinarySI)
+    return q
+}
+```
+
+### Consumer changes
+
+**virt-handler** `AdjustResources()`:
+
+```
+1. Read vmi.Status.Memory.MemLockLimit (new)
+2. Fall back to vmi.Status.MigrationState.TargetMemoryOverhead (existing, migration)
+3. Fall back to local GetMemoryOverhead() calculation (existing, legacy)
+```
+
+**virt-launcher** `MemoryConfigurator.configureMemLock()`:
+
+```
+1. Read vmi.Status.Memory.MemLockLimit → set <memtune><hard_limit>
+2. Fall back to vfio.CalculateMemlockLimit(vmi) (existing, for when status not yet populated)
+```
+
+## Impact on VEP #144
+
+### No breaking changes
+
+- `reservedOverhead.addedOverhead` continues to work as an input
+- `reservedOverhead.memLock` continues to trigger memlock adjustment
+- Existing behaviour preserved when `MemLockLimit` status field is absent
+
+### Improvement: decoupled from pod sizing
+
+Currently `addedOverhead` flows into `GetMemoryOverhead()` which inflates both the pod memory request and the memlock rlimit. With this extension, virt-controller can compute the memlock value separately from the pod sizing. `addedOverhead` could be split into:
+
+- Its contribution to pod memory request (for actual memory consumption)
+- Its contribution to the memlock limit (for VFIO address space ceiling)
+
+This aligns with the VEP's stated goal: "Adjust memory lock limits **without impacting VMI scheduling capacity**."
+
+## Benefits
+
+- **Single source of truth** -- memlock value computed once, consumed everywhere
+- **No parallel calculations** -- eliminates drift between handler, launcher, and libvirt
+- **Status-only** -- no VMI spec changes, no VEP needed for the status field
+- **Backwards compatible** -- consumers fall back to existing behaviour when status field absent
+- **Extensible** -- future memlock inputs feed into the same controller calculation
+- **Decouples memlock from pod sizing** -- addresses VEP #144's stated goal
+
+## Prior work
+
+This extension builds on two PRs that address the immediate multi-device VFIO memlock problem in v1.9:
+
+- [kubevirt/kubevirt#17805](https://github.com/kubevirt/kubevirt/pull/17805) -- scales virt-handler's memlock rlimit for multi-device VFIO passthrough. Extracts `CalculateMemlockSize()` and introduces `CountVFIODevices()` in `pkg/util`.
+- [kubevirt/kubevirt#17857](https://github.com/kubevirt/kubevirt/pull/17857) -- extracts the VFIO memlock formula into a shared `pkg/vfio` package and sets `<memtune><hard_limit>` in the domain XML so libvirt uses KubeVirt's value directly.
+
+These PRs solve the problem for v1.9 by sharing the formula between handler and launcher via `pkg/vfio`. This extension replaces the shared formula approach with a single computation in virt-controller propagated via VMI status, eliminating parallel calculations entirely.
+
+## Timeline
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+
+```go
+func calculateMemLockLimit(vmi *v1.VirtualMachineInstance) *resource.Quantity {
+    // Start with automatic VFIO calculation
+    memlockBytes := vfio.CalculateMemlockLimit(vmi)
+
+    // Add user-specified overhead from VEP #144
+    if vmi.Spec.Domain.Memory != nil &&
+       vmi.Spec.Domain.Memory.ReservedOverhead != nil &&
+       vmi.Spec.Domain.Memory.ReservedOverhead.AddedOverhead != nil {
```
bgartzi: Only if `vmi.Spec.Domain.Memory.ReservedOverhead.MemLock == v1.MemLockRequired`

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+No pod annotation is needed. Unlike `MemoryOverheadAnnotationBytes` (which the pod needs at creation time for resource requests), the memlock value is consumed at runtime: virt-handler reads it from VMI status before the domain starts, and virt-launcher receives the full VMI object via gRPC.
+
+### How this unifies both approaches
+
+| Source | Triggers | Current path | With this extension |
+| :----- | :------- | :----------- | :------------------ |
+| VFIO devices (auto) | KubeVirt counts devices internally | Parallel calc in handler + launcher | Controller computes, writes to status |
+| `addedOverhead` (VEP #144) | User/webhook sets spec field | Flows into `GetMemoryOverhead()`, inflates pod request | Feeds controller's memlock calc, writes to status. Pod request handled separately. |
```
bgartzi, line 66: If I understand correctly, `addedOverhead` would still fall into `GetMemoryOverhead()`, right?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+- `reservedOverhead.addedOverhead` continues to work as an input
+- `reservedOverhead.memLock` continues to trigger memlock adjustment
+- Existing behaviour preserved when `MemLockLimit` status field is absent
+
+### Improvement: decoupled from pod sizing
+
+Currently `addedOverhead` flows into `GetMemoryOverhead()` which inflates both the pod memory request and the memlock rlimit. With this extension, virt-controller can compute the memlock value separately from the pod sizing. `addedOverhead` could be split into:
+
+- Its contribution to pod memory request (for actual memory consumption)
+- Its contribution to the memlock limit (for VFIO address space ceiling)
```
bgartzi, line 133: I'm missing the most important part of this: How can we split its contribution to one or another?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+
+This aligns with the VEP's stated goal: "Adjust memory lock limits **without impacting VMI scheduling capacity**."
```
bgartzi, line 135: IIRC this was one of those goals that were initially proposed and pushed back but not removed from the merged version.

I think the reason this was not accepted in the end was that it was not safe enough to let processes in the virt-launcher lock memory that was not requested by the pod. Yes, this could be due to the N-memory accounting issue that should be alleviated by iommufd, or could be because actual memory is being locked by that process, which would actually go against what it advertised in the compute container memory requests/limits.

PS in which case k8s would take care of that pod, right?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Benefits
+
+- **Single source of truth** -- memlock value computed once, consumed everywhere
+- **No parallel calculations** -- eliminates drift between handler, launcher, and libvirt
+- **Status-only** -- no VMI spec changes, no VEP needed for the status field
+- **Backwards compatible** -- consumers fall back to existing behaviour when status field absent
```
bgartzi, line 161: plus it would add the field during a version upgrade, which would ease deprecating the legacy mechanisms at some point, right?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
...
+
+- Its contribution to pod memory request (for actual memory consumption)
+- Its contribution to the memlock limit (for VFIO address space ceiling)
+
+This aligns with the VEP's stated goal: "Adjust memory lock limits **without impacting VMI scheduling capacity**."
+
+## Benefits
+
+- **Single source of truth** -- memlock value computed once, consumed everywhere
+- **No parallel calculations** -- eliminates drift between handler, launcher, and libvirt
```
bgartzi, line 159: I really like this, as it would also help to reduce future drifts/workarounds such as the one required in this fix: https://github.com/kubevirt/kubevirt/pull/17712

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,155 @@
+# VEP #144 Extension: Propagate computed memlock limit via VMI status
+
+*Draft for discussion -- extends the existing VEP #144 (Memory Lock RLimit configuration)*
+
+> **Note:** This draft assumes [kubevirt/kubevirt#17805](https://github.com/kubevirt/kubevirt/pull/17805) (VFIO memlock scaling in virt-handler) and [kubevirt/kubevirt#17857](https://github.com/kubevirt/kubevirt/pull/17857) (shared `pkg/vfio` package and `<memtune><hard_limit>` in domain XML) have merged. #17805 is under review; #17857 is a draft dependent on it. The assumptions about `pkg/vfio`, `CountDevices()`, `CalculateMemlockLimit()`, and `<memtune><hard_limit>` support in the domain converter all come from these PRs.
```
bgartzi: https://github.com/kubevirt/kubevirt/pull/17805 is merged now :)

bgartzi (commented):
Thanks for the heads up @lyarwood, I like this.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: Thanks, I'll update the preamble.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: Right, I'll update the pseudo-code to make this conditional explicit. The `util.RequiresLockingMemory(vmi)` check at the bottom of the function already covers this case (it checks the `memLock` flag plus VFIO/realtime requirements), but it's worth being clear that the `addedOverhead` contribution only applies when locking is actually in play.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: That's the current behaviour, yes — and it's the problem this extension aims to fix. Today `addedOverhead` inflates both the pod memory request (via `GetMemoryOverhead()`) and the memlock limit, but these serve different purposes. The pod request determines scheduling capacity, while the memlock limit sets the address space ceiling for VFIO.

With this extension, the goal is to stop routing `addedOverhead` through `GetMemoryOverhead()` for memlock purposes entirely. The controller would consume it as an input to `MemLockLimit` in status, and `GetMemoryOverhead()` would only handle actual pod sizing. Whether `addedOverhead` should *also* still contribute to the pod request is a separate question — in many VFIO cases the overhead is address space, not real memory consumption.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: This is the right question and I agree the draft doesn't spell it out clearly enough. The split I have in mind:

- **Pod memory request**: `GetMemoryOverhead()` accounts for real memory consumption — guest RAM, QEMU overhead, virtio buffers, page tables, etc. This determines scheduling.
- **Memlock limit**: `MemLockLimit` in status accounts for the VFIO address space ceiling — guest RAM + MMIO BAR mappings + any user-specified `addedOverhead`. This determines what the kernel allows the process to lock.

The key insight is that VFIO BAR mappings inflate the memlock requirement without consuming proportional physical memory (especially with iommufd). So `addedOverhead` for VFIO use cases should feed into memlock but not necessarily into the pod request.

I'll update the draft with a concrete example showing the split for a multi-GPU VM.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: The safety concern was valid for the original VEP #144 discussion, but the landscape has shifted.

With type1 VFIO, multi-device vIOMMU passthrough was already broken before #17805 due to insufficient rlimits — the cgroup sizing gap was latent, not newly introduced. With iommufd replacing type1 for modern deployments, the memlock rlimit becomes an address space ceiling rather than a physical memory commitment, making the decoupling safe.

For any remaining type1 edge cases, the pod request would still need to account for actual pinned memory, but that's an existing gap that this extension makes explicit and fixable rather than hidden. The controller could factor in the IOMMU backend (type1 vs iommufd) when deciding whether `addedOverhead` should also inflate the pod request, rather than always inflating both.

And yes, if actual physical memory consumption exceeds the cgroup limit, the kubelet will OOM-kill the pod — that safety net doesn't change regardless of what we do with the rlimit.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: Exactly. Once `MemLockLimit` is populated in status during upgrades, consumers can switch to reading it exclusively. The fallback paths in virt-handler and virt-launcher become dead code for any cluster that's been through at least one reconciliation cycle, and we can remove them in a later release.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
lyarwood: Agreed, that's a good example. With a single computed value in status, fixes like #17712 become unnecessary — there's no parallel calculation to drift.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
bgartzi: I re-read the thing. It makes sense, thanks.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,176 @@
...
+  │
+  └─ Writes vmi.Status.Memory.MemLockLimit
+       │
+       ├─ virt-handler reads → prlimit64 on QEMU process
+       │
+       └─ virt-launcher reads (via SyncVirtualMachine gRPC)
+            → <memtune><hard_limit> in domain XML
+```
+
+No pod annotation is needed. Unlike `MemoryOverheadAnnotationBytes` (which the pod needs at creation time for resource requests), the memlock value is consumed at runtime: virt-handler reads it from VMI status before the domain starts, and virt-launcher receives the full VMI object via gRPC.
```
Barakmor1, line 59: Hey, is it guaranteed that the amount of memory that needs to be locked remains unchanged throughout the migration process? For example, could memory or CPU hotplug operations, or any other changes that do not require a VM restart, affect this value?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,176 @@
...
+
+### How this unifies both approaches
+
+| Source | Triggers | Current path | With this extension |
+| :----- | :------- | :----------- | :------------------ |
+| VFIO devices (auto) | KubeVirt counts devices internally | Parallel calc in handler + launcher | Controller computes, writes to status |
+| `addedOverhead` (VEP #144) | User/webhook sets spec field | Flows into `GetMemoryOverhead()`, inflates pod request | Feeds controller's memlock calc, writes to status. Pod request handled separately. |
+| `memLock: Required` (VEP #144) | User/webhook sets flag | Triggers handler's memlock adjustment | Triggers controller to compute and write to status |
```
Barakmor1, line 67: Can you please elaborate why we need this field?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+No pod annotation is needed. Unlike `MemoryOverheadAnnotationBytes` (which the pod needs at creation time for resource requests), the memlock value is consumed at runtime: virt-handler reads it from VMI status before the domain starts, and virt-launcher receives the full VMI object via gRPC.
+
+### How this unifies both approaches
+
+| Source | Triggers | Current path | With this extension |
+| :----- | :------- | :----------- | :------------------ |
+| VFIO devices (auto) | KubeVirt counts devices internally | Parallel calc in handler + launcher | Controller computes, writes to status |
+| `addedOverhead` (VEP #144) | User/webhook sets spec field | Flows into `GetMemoryOverhead()`, inflates pod request | Feeds controller's memlock calc, writes to status. Pod request handled separately. |
```
Barakmor1, line 66: Could you please elaborate a bit more on the motivation for this? I’m trying to understand why adjusting a VMI’s memory configuration can’t be used instead of introducing another field.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
bgartzi: I looked into libvirt. I wouldn't say CPU hotplug would need to affect this, but memory hotplug does:  `qemuDomainAttachMemory` https://gitlab.com/libvirt/libvirt/-/blob/fa0a4c899f032f3d013f1eec08bbbc654b06a6c2/src/qemu/qemu_hotplug.c#L2425 adjusts memlock limits.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
bgartzi: Initially we needed it because of two reasons: pod sizing and memlock limits were coupled through this API and we also needed a way to describe that  we needed to lock VM's memory (with/without overhead).

`memLock` defines whether the VM memory needs to be locked or not, no matter overhead is needed. Now that pod sizing is being decoupled from this mechanism, I agree it gets confusing, as it does not makes sense to have `NotRequired` and an actual value for `addedOverhead`.

Maybe for this proposal it would make more sense using something like:
```
.spec.domain.memory:
    lockMemory:
        overhead: 1Gi
```

So that users might also set `.spec.domain.memory.lockMemory: {}` (i.e. `!= nil`) if the VM requires to lock VM memory but not any overhead?

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
bgartzi: One of the motivations of this VEP extension is to be able to alter memlock limits without altering pod sizing (and so VM schedulability). Adjusting VM memory does affect that.

Plus if we were to adjust VM memory based on the devices using iommu we would hit a chicken and egg problem, as limits are predicted based on VM memory and number of iommu devices.

If we would assume that the iommu range is always 1Gi and simplify the API by letting users set memlock required/not required, we still have a problem, as devices might be implemented outside of kubevirt's core source code and we need a field in which the static + the dynamic limits (the ones based on number of devices) can get into.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
Barakmor1: So, is the plan to change this value during live migration?

iholder101:
Please see https://github.com/kubevirt/enhancements/issues/144#issuecomment-4659729815.

TLDR: I don't think there's a reason to include any API to control the prlimit. We can simply have a very generous prlimit and let libvirt do the rest.

veps/sig-compute/144-memory-lock-rlimits/memlock-limit-status-extension.md
Barakmor1: If this field is only meant as a temporary workaround until libvirt/QEMU exposes an API for determining the required locked memory, I'm concerned that it may become a long-term API. Once users start relying on it, removing it could be difficult, even if we no longer need it, leaving us with an ongoing maintenance burden.

lyarwood:
/close

hard_limit isn't something  we can actually use as kubevirt connects to libvirt in session mode, apologies for the noise!

Issue #233, opened 2026-03-17
VEP 233: Enforce Release Branch End-of-Life Policy


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood

**Current Feature Stage**:

New

**Feature Gate**:

N/A — this is a process and policy change, not a code feature.

**Responsible SIGs**:

Primary SIG:
/sig release

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/234

**Timeline**:

* Policy adoption target: v1.9
* Enforcement target: v1.10

* [ ] Phase 1: Policy Adoption
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/234
  * [ ] Documentation PRs (`docs/release.md`, `docs/release-branch-backporting.md`):
  * [ ] Announcement on `kubevirt-dev` mailing list:

**Additional context**:

KubeVirt currently has 59 release branches with no enforced EOL policy. CVE
fixes routinely require backports to 10+ branches. This VEP proposes adopting
a hard EOL policy modeled on Kubernetes, limiting active maintenance to the 3
most recent minor releases (~12-15 months per release).

> [!IMPORTANT]
> Please keep this description up to date.

lyarwood:
@fossedihelm @xpivarc as the sig-release OWNERS would you mind being marked as reviewers for this VEP proposal? The enhancement PR is at https://github.com/kubevirt/enhancements/pull/234.

fossedihelm:
@lyarwood Sure!

lyarwood:
VEP Freeze Exception posted to kubevirt-dev under https://groups.google.com/g/kubevirt-dev/c/a_5Abnz-fAM

PR #234, opened 2026-03-17
VEP #233: Enforce Release Branch End-of-Life Policy

lyarwood:
## Summary

- Proposes a hard EOL policy for KubeVirt release branches, limiting active maintenance to the **3 most recent minor releases** (~12-15 months per release)
- Aligns with the [Kubernetes patch release support period](https://kubernetes.io/releases/patch-releases/#support-period)
- Documents real CVE backport examples (CVE-2023-39325, CVE-2024-21626, CVE-2025-22869, CVE-2025-47913) showing the current unbounded maintenance burden
- Defines enforcement mechanisms: branch protection, CI removal, auto-close bot for cherry-picks to EOL branches
- Includes a transition plan with a one-release-cycle grace period

## Timeline

Policy announced with v1.9 GA (2026-07-22), enforcement begins at v1.10 GA (~2026-11).

```mermaid
gantt
    title VEP #233 — Release Branch EOL Timeline
    dateFormat YYYY-MM-DD
    axisFormat %b %Y

    section Releases
    v1.5 GA              :milestone, 2025-03-05, 0d
    v1.6 GA              :milestone, 2025-07-30, 0d
    v1.7 GA              :milestone, 2025-11-19, 0d
    v1.8 GA              :milestone, 2026-03-25, 0d
    v1.9 GA              :milestone, 2026-07-22, 0d
    v1.10 GA (estimated) :milestone, 2026-11-18, 0d

    section Support Windows (3 releases)
    v1.5 supported       :active, 2025-03-05, 2026-11-18
    v1.6 supported       :active, 2025-07-30, 2026-11-18
    v1.7 supported       :active, 2025-11-19, 2026-11-18
    v1.8 supported       :active, 2026-03-25, 2027-03-17
    v1.9 supported       :active, 2026-07-22, 2027-07-22

    section Policy Phases
    Phase 1: Announce     :crit, 2026-07-22, 2026-11-18
    Phase 2: Enforce      :crit, 2026-11-18, 2027-03-17

    section EOL Events
    v1.7 and earlier EOL  :milestone, crit, 2026-11-18, 0d
    v1.8 EOL              :milestone, crit, 2027-03-17, 0d
```

Tracking issue: https://github.com/kubevirt/enhancements/issues/233

/sig release

PR #317, opened 2026-05-20
VEP 233: Add vendor fork alternative for downstream extended maintenance

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/233
**SIG label**: /sig release

### What this PR does

Adds Alternative 6 to VEP #233, proposing that downstream vendors maintain
their own public forks (e.g., `openshift/kubevirt`) to take over maintenance
of release branches after they are EOL'd upstream.

This follows the established pattern used in the Kubernetes ecosystem where
vendors maintain public forks (e.g., `openshift/kubernetes`) for extended
branch support. Under this model:

- The upstream 3-release EOL policy proceeds unchanged
- When a branch is EOL'd upstream, vendors that need continued support
  maintain the corresponding branch in their own public fork
- CI is replicated from upstream, providing equivalent test coverage
- Each vendor maintains only the branches they need, without imposing on the
  upstream project or other vendors

This differs from Alternative 4 (vendor-driven two-tier) by moving extended
branches entirely out of `kubevirt/kubevirt`, keeping the upstream branch
count bounded to 3 and eliminating ambiguity about review responsibilities.

This is posted as a draft as we are outside of the VEP design window for
v1.9.0. This alternative was raised as an action item during the VEP #233
community discussion and is intended to be discussed alongside the existing
alternatives before the v1.10 enforcement deadline.

### DIFF
--- a/veps/sig-release/233-release-branch-eol-policy/vep.md
+++ b/veps/sig-release/233-release-branch-eol-policy/vep.md
@@ -370,6 +370,101 @@ VEP because:
 - An LTS model could be layered on top of this policy in a future VEP if
   demand exists
 
+### Alternative 6: Vendor-maintained public forks for extended maintenance
+
+Rather than maintaining extended-support branches within `kubevirt/kubevirt`
+(Alternative 4), downstream vendors that need longer support windows could
+maintain their own public forks of the repository. When a release branch
+reaches EOL upstream under the 3-release policy, the vendor's fork becomes
+the maintenance target for continued backports.
+
+This pattern is well-established in the Kubernetes ecosystem. For example,
+OpenShift maintains `openshift/kubernetes` as a public fork that carries
+downstream patches and extended branch maintenance independently of the
+upstream `kubernetes/kubernetes` project. The same model could be applied to
+KubeVirt by any vendor that needs to support releases beyond the upstream
+window (e.g., `openshift/kubevirt`, `suse/kubevirt`, etc.).
+
+Under this model:
+
+- The upstream 3-release EOL policy proceeds unchanged. When a release branch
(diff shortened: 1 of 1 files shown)


iholder101:
I think this alternative is essentially identical to what the VEP already suggests (supporting only 3 release branches), since vendors can always fork KubeVirt regardless of upstream policy.

What about the alternative mentioned in https://github.com/kubevirt/enhancements/pull/234#issuecomment-4495976123? The idea is to keep older branches open for merges on a best-effort basis - no upstream obligation, but downstream consumers who are willing to do the work can backport critical CVEs or bug fixes and benefit others in the same position. Do you mind adding it as an alternative as well?

lyarwood:
Apologies @iholder101, I've been out for 3 weeks.

> I think this alternative is essentially identical to what the VEP already suggests (supporting only 3 release branches), since vendors can always fork KubeVirt regardless of upstream policy.

True but it didn't suggest the idea of an open midstream repo for the forks as used by OpenShift so I thought it is still worth documenting.

> What about the alternative mentioned in [#234 (comment)](https://github.com/kubevirt/enhancements/pull/234#issuecomment-4495976123)? The idea is to keep older branches open for merges on a best-effort basis - no upstream obligation, but downstream consumers who are willing to do the work can backport critical CVEs or bug fixes and benefit others in the same position. Do you mind adding it as an alternative as well?

I've drafted https://github.com/kubevirt/enhancements/pull/335

lyarwood:
Superseded by #341, which reworks the VEP to use dedicated release branch maintainer groups with a maintainer-driven EOL model. This addresses the governance gap that motivated the midstream fork alternative.

Issue #210, opened 2026-02-17
VEP 210: CentOS Stream 10 Support


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood 

**Current Feature Stage**:

New

**Feature Gate**:

N/A

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig ci
/sig buildsystem
/sig storage
/sig network

**Enhancement link**:

TBD

**Timeline**:

* Alpha release target (x.y): v1.9
* Beta release target (x.y): v1.10
* Stable release target (x.y): v1.11

* [ ] Alpha v1.9
  * [ ] Code PRs: 
    * [ ] https://github.com/kubevirt/kubevirt/pull/16582
    * [ ] https://github.com/kubevirt/kubevirtci/pull/1594

lyarwood:
Hi @dhiller @xpivarc @dollierp @enp0s3 - as members of sig-buildsystem reviewers, VEP 210 currently has no VEP reviewer assigned on the [1.9 Enhancements Tracking project](https://github.com/orgs/kubevirt/projects/21). Would any of you be interested in taking on the VEP reviewer role for this proposal?

dollierp:
> Would any of you be interested in taking on the VEP reviewer role for this proposal?

Hi @lyarwood,

I added myself as reviewer since I reviewed it a few weeks ago.

lyarwood:
## Deferring alpha target from v1.9 to v1.10

I'm proposing to defer this VEP's alpha target from v1.9 to v1.10. The key blockers:

1. **k8s 1.36 regressions on CentOS Stream 9**: Two critical-urgent issues have been discovered on the current CS9-based k8s 1.36 provider that are unrelated to the CS10 transition but block stabilizing the k8s 1.36 lanes:
   - kubevirt/kubevirt#17792 — virtiofs serviceAccountToken permission denied (0% pass rate on k8s 1.36)
   - kubevirt/kubevirt#17724 — datavolume migration to filesystem volumes failing ~21% on k8s 1.36

2. **kubevirtci CS10 provider not landed**: kubevirt/kubevirtci#1665 (move k8s 1.36 to CS10) is approved but hasn't merged, and the k8s 1.36 regressions above need to be resolved on CS9 first before we can confidently switch to CS10.

With the v1.9 beta.0 deadline imminent, there isn't time to resolve these issues and land the CS10 provider in this cycle.

### Updated timeline

[...]

PR #318, opened 2026-05-20
VEP #210: defer alpha target from v1.9 to v1.10

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/210
**SIG label**: /sig compute

### What this PR does

Defers the CentOS Stream 10 transition (VEP #210) alpha target from v1.9 to v1.10 due to unresolved k8s 1.36 regressions on CentOS Stream 9:

- kubevirt/kubevirt#17792 — virtiofs serviceAccountToken permission denied (0% pass rate on k8s 1.36)
- kubevirt/kubevirt#17724 — datavolume migration to filesystem volumes failing ~21% on k8s 1.36

These regressions need to be resolved before the CentOS Stream 10 provider (kubevirt/kubevirtci#1665) can land, and with v1.9 beta.0 imminent there isn't time this cycle.

The full timeline shifts by one release:

| Phase | Previous | New |
|-------|----------|-----|
| Alpha | v1.9 | v1.10 |
| Beta | v1.10 | v1.11 |
| GA | v1.11 | v1.12 |

The first CS10 provider lane shifts from k8s-1.36 to k8s-1.37. The per-lane transition model is otherwise unchanged. A note has been added about k8s-1.36 (CS9) extending ~6 weeks past CentOS Stream 9 EOL (2027-05-31) during phase-out.

kubevirt/kubevirtci#1665 has been moved to draft — it will need to be adapted for the k8s-1.37 provider in the v1.10 cycle.

/cc @dollierp

### DIFF
--- a/veps/sig-compute/210-centos-stream-10/vep.md
+++ b/veps/sig-compute/210-centos-stream-10/vep.md
@@ -4,9 +4,9 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.9
-- This VEP targets beta for version: v1.10
-- This VEP targets GA for version: v1.11
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
 
 ### Release Signoff Checklist
 
@@ -24,7 +24,7 @@ This VEP proposes transitioning KubeVirt from CentOS Stream 9 to CentOS Stream 1
 1. **Userspace**: The container images (virt-handler, virt-launcher, virt-operator, virt-controller, virt-api, etc.) and the builder image used to compile KubeVirt binaries.
 2. **Host OS**: The kubevirtci cluster providers used for development and CI testing.
 
-Rather than a big-bang switchover, the transition follows a per-lane model: from v1.9, new Kubernetes version lanes are introduced on CentOS Stream 10 for both userspace and host OS, while existing older lanes remain on CentOS Stream 9 until those Kubernetes versions naturally reach end of life and are dropped from CI. CentOS Stream 9 is supported during v1.8, v1.9, and v1.10, with support removed in v1.11 once all remaining CS9 lanes have aged out. This avoids doubling CI capacity and provides a gradual, low-risk migration path.
+Rather than a big-bang switchover, the transition follows a per-lane model: from v1.10, new Kubernetes version lanes are introduced on CentOS Stream 10 for both userspace and host OS, while existing older lanes remain on CentOS Stream 9 until those Kubernetes versions naturally reach end of life and are dropped from CI. CentOS Stream 9 is supported during v1.8, v1.9, v1.10, and v1.11, with support removed in v1.12 once all remaining CS9 lanes have aged out. This avoids doubling CI capacity and provides a gradual, low-risk migration path.
 
 CentOS Stream 9 reaches end of life on 2027-05-31, requiring a planned migration well before that date.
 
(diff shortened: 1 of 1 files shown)


veps/sig-compute/210-centos-stream-10/vep.md
```diff
@@ -4,9 +4,9 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.9
-- This VEP targets beta for version: v1.10
-- This VEP targets GA for version: v1.11
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version: v1.11
+- This VEP targets GA for version: v1.12
```
iholder101: Can you keep these empty?
In the relevant release cycle a PR to add these should be created, that's how we can discuss the bump in the relevant cycle.

lyarwood:
@iholder101 @dollierp apologies for not getting back to this,  updated and hopefully something we can deliver during 1.10 this time around.

veps/sig-compute/210-centos-stream-10/vep.md
EdDev: AFAIK there is no such path. Kubevirt does not support rollback.
Many times in this section the option to rollback is mentioned as a theoretical/conceptual option. I.e. for a specific feature, if  rollback or upgrade cancellation would be supported in the future, there are no special care needed to handle it.

I think, that in the context of this section, a word or two can be given for the case of a VM attempting to migrate from a new node to an older one (10 to 9).

EdDev, line 277: As the base OS is changing, an upgrade to v1.10 may imply special requirement. While I am raising this at the testing section, it may have to be considered in other parts as well.
Specifically, what happens with users that upgrade from v1.9 to v1.10 (the only supported direct upgrade path)? Are they require to prepare the nodes with a new OS in advance?
What happens if the nodes are with the older CentOS Stream 9?

And are we planning to test this? I know we have the virt-operator tests which check upgrades, but it is not clear that they will cover this.

EdDev (commented):
Thank you.

It all makes sense and it is described clearly.
I am missing only the upgrade path for v1.9 users to v1.10 and onward.

I missed the original VEP document, so I am raising this now for your consideration:
Looking at how Kubernetes handles base OS and container image transitions, this seems to fit better under ssig-release rather than sig-compute. In upstream Kubernetes, transitioning the base OS (such as moving from Debian to Distroless) was handled by SIG-Release via [KEP-1729](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/1729-rebase-images-to-distroless), because these transitions heavily impact CVE management, build pipelines, and release engineering rather than core virtualization logic.

veps/sig-compute/210-centos-stream-10/vep.md
lyarwood: These are already blank on this branch — alpha is v1.10, beta and GA are left empty to be filled in separate PRs per cycle.

veps/sig-compute/210-centos-stream-10/vep.md
lyarwood: Good catch — the rollback claim has been removed. The section now explicitly states that KubeVirt does not support rollback.

On the VM migration between nodes with different base OS versions: during the transition window (v1.10–v1.11) both EL9 and EL10 userspace builds are maintained, so two mixed combinations can occur in practice — EL9 host with EL10 containers, and EL10 host with EL9 containers. The VEP notes both scenarios are expected to work. Explicitly testing the full dual matrix in CI is not planned due to resource constraints (this was settled during the original VEP discussion), so the rolling-upgrade mixed-node migration scenario is acknowledged but not covered by a dedicated CI requirement.

lyarwood: Both EL9 and EL10 userspace builds are maintained throughout the transition — the per-lane model means users on older Kubernetes version lanes (k8s-1.35, k8s-1.36) continue to receive EL9-based images when upgrading to v1.10. No forced OS change occurs at the point of upgrade; only the new k8s-1.37 lane ships EL10 userspace. A Functional Testing bullet has been added to make this explicit.

On host OS requirements: that is a platform concern (upstream Kubernetes / OpenShift) rather than something this VEP asserts. The VEP describes the container userspace transition and notes the mixed host/container scenarios that arise in practice during the window.

lyarwood:
> I missed the original VEP document, so I am raising this now for your consideration: Looking at how Kubernetes handles base OS and container image transitions, this seems to fit better under ssig-release rather than sig-compute. In upstream Kubernetes, transitioning the base OS (such as moving from Debian to Distroless) was handled by SIG-Release via [KEP-1729](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/1729-rebase-images-to-distroless), because these transitions heavily impact CVE management, build pipelines, and release engineering rather than core virtualization logic.

Thanks @EdDev, I don't think sig-release really own anything outside of the release schedule but I might be wrong. Happy to move this to sig-ci or elsewhere if folks think sig-compute isn't the right choice here.

lyarwood:
@dollierp Would you mind reviewing this VEP given we are pushing ahead with implementation at the moment in project-infra and kubevirtci :)

veps/sig-compute/210-centos-stream-10/vep.md
```diff
@@ -4,9 +4,9 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.9
-- This VEP targets beta for version: v1.10
-- This VEP targets GA for version: v1.11
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
```
dollierp, line 8: beta target seems to be `v1.11`

veps/sig-compute/210-centos-stream-10/vep.md
```diff
@@ -4,9 +4,9 @@
...
+- This VEP targets GA for version:
```
dollierp, line 9: GA target seems to be `v1.12`

veps/sig-compute/210-centos-stream-10/vep.md
lyarwood: We want folks to revisit the doc each cycle to get approval for the next step so leaving this empty is correct for now.

veps/sig-compute/210-centos-stream-10/vep.md
lyarwood: As above.

dollierp (approved):
Looks great!

/lgtm

xpivarc (commented):
/approve

Issue #321, opened 2026-05-28
VEP 321: Online VMSnapshot Using QEMU External Snapshots


noamasu:
**Primary contact (assignee)**:

/assign @noamasu 

**Current Feature Stage**:

New

**Feature Gate**:

ExternalVMSnapshot

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**: https://github.com/kubevirt/enhancements/pull/322

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:
This VEP introduces a new VMSnapshot flow using QEMU external snapshots to decouple the freeze duration from CSI driver speed. The existing VMSnapshot flow remains unchanged as the default.

> [!IMPORTANT]
> Please keep this description up to date.

fabiand:
I think this is a quite good idea.

vladikr:
@noamasu can you please retarget this to 1.10 and update the status

PR #322, opened 2026-05-28
VEP #321: Online VMSnapshot Using QEMU External Snapshots

noamasu:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/321 
**SIG label**: /sig storage

### What this PR does

Introduces a new online VMSnapshot flow using QEMU external snapshots to decouple the guest filesystem freeze from CSI VolumeSnapshot creation.

The current flow keeps the VM frozen while CSI snapshots are taken, making the freeze duration dependent on CSI driver speed. This causes Windows VMs to fail with VSS timeout and disrupts applications on any OS.

The new flow uses a QEMU transaction to atomically establish the snapshot point at the libvirt level in sub-second time, then takes CSI VolumeSnapshots asynchronously while the VM runs unfrozen.

### DIFF
--- a/veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
+++ b/veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
@@ -99,6 +99,9 @@ needs to be established.
   manage their own CSI snapshots currently use freeze/unfreeze hooks. Exposing
   the libvirt-level snapshot to these vendors via a dedicated CRD
   (e.g. `VirtualMachineSnapshotRequest`) could be addressed in a follow-up VEP
+- Memory state capture. QEMU external snapshots can also save the full VM
+  state (memory, CPU, devices) alongside disk snapshots, enabling exact
+  restore to the running state. This could also be addressed in a follow-up VEP
 
 
 ## Definition of Users
@@ -144,6 +147,9 @@ disruption
 
 ### New flow
 
+The new flow is enabled by setting `spec.externalSnapshot: true` on the
+VMSnapshot. If not set, the current flow is used.
+
 ```
 Phase 0: Create scratch PVC, hotplug as UtilityVolume
 Phase 1: Freeze -> QEMU transaction (all disks, atomic, ~20ms) -> Unfreeze
@@ -158,17 +164,18 @@ Freeze duration is sub-second regardless of disk count.
 
 Before taking the snapshot, the snapshot controller:
 
-1. Creates a PVC (`snap-scratch-{uid}`) sized to hold qcow2 overlay files
+1. Sets the `OverlaySnapshotActive` condition on the VMI.
+2. Creates a PVC (`snap-scratch-{uid}`) sized to hold qcow2 overlay files
    for all disks during the overlay window. The size can be set explicitly
    via `spec.overlayScratchSize` on the VMSnapshot. If not configured, a
    default size is used
    (see [Scalability](#scalability) for details).
-2. Hotplugs the PVC as a UtilityVolume to the VMI via JSON patch on
+3. Hotplugs the PVC as a UtilityVolume to the VMI via JSON patch on
    `spec.utilityVolumes` (same mechanism as [VEP #90 Utility Volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md),
    same pattern as the incremental backup push-target PVC in [VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md)).
-3. Adds a Kubernetes finalizer `snapshot.kubevirt.io/overlay-protection`
+4. Adds a Kubernetes finalizer `snapshot.kubevirt.io/overlay-protection`
    on the scratch PVC to prevent its deletion while overlays are active.
-4. Waits for the volume to reach `VolumeReady` or `HotplugVolumeMounted` status.
+5. Waits for the volume to reach `VolumeReady` or `HotplugVolumeMounted` status.
 
 ### Phase 1: QEMU atomic snapshot (overlay creation)
 
@@ -255,7 +262,7 @@ requeues for retry.
 ### Phase 4: Cleanup
 
 Once all disks are back on their base images:
-- Clear the `snapshot.kubevirt.io/overlay-active` annotation on the VMI
+- Remove the `OverlaySnapshotActive` condition from the VMI
 - Detach the utility volume via JSON patch
 - Remove the finalizer from the scratch PVC
 - Delete the scratch PVC
@@ -274,6 +281,7 @@ sequenceDiagram
     User->>SC: Create VMSnapshot
 
     Note over SC,VL: Phase 0: Scratch Volume Setup
+    SC->>SC: Set OverlaySnapshotActive condition on VMI
     SC->>SC: Create scratch PVC + finalizer
     SC->>VH: Hotplug as UtilityVolume
     VH->>VL: Mount scratch volume
@@ -315,7 +323,7 @@ sequenceDiagram
     VL-->>SC: All disks back on base images
 
     Note over SC,VH: Phase 4: Cleanup
-    SC->>SC: Clear overlay-active annotation
+    SC->>SC: Remove OverlaySnapshotActive condition
     SC->>VH: Detach UtilityVolume
     SC->>SC: Remove finalizer, delete scratch PVC
 
@@ -325,31 +333,26 @@ sequenceDiagram
 
 ### Overlay state tracking
 
-An annotation `snapshot.kubevirt.io/overlay-active` is set on the VMI when
-overlays are active (after Phase 1) and cleared after successful commit
-(after Phase 3). This annotation is checked by:
+A VMI condition `OverlaySnapshotActive` is set at the start of Phase 0 and
+removed after successful commit (Phase 3). This condition is checked by:
 
-- **Migration**: blocks `startMigration()` - migration with overlays would
-  break the backing chain on the target node
-- **Backup**: blocks `BackupVirtualMachine()` - backup checkpoints on an
-  overlay become invalid after commit
+- **Migration**: blocks `startMigration()` - the overlay references the
+  base image on the source node which would not exist on the target
+- **Backup**: blocks `BackupVirtualMachine()` - concurrent backup and
+  snapshot overlay operations would conflict
 - **Volume unplug**: blocks `removeVolumeRequestHandler()` - unplugging a
   disk with an active overlay would orphan the overlay
 - **Disk resize**: skips resize in `syncDisks()` - resizing during overlay
   could cause size mismatch between overlay and base
 - **VM destroy**: `KillVMI()` aborts active block jobs before
   `DestroyFlags()` - prevents orphaned block jobs
-- **VM crash / pod eviction**: the overlay + base together remain consistent
-  (no data loss). Once the VM comes back up, the pending commit resumes and
-  completes the cleanup. The scratch PVC finalizer ensures it survives across
-  restarts.
-
-### Fallback
-
-If `ExternalSnapshot` fails for any reason (no guest agent, libvirt error,
-unsupported configuration), the controller falls back to the current
-sequential CSI path automatically, as the new flow is regarded as an
-optimization, not a requirement.
+- **VM crash / pod eviction**: the scratch PVC finalizer ensures it
+  survives across restarts. On startup, virt-launcher inspects the actual
+  disk state (checks for overlay files on the scratch PVC). If overlays
+  are found, the VM starts in paused mode with the overlay still active
+  and a live block-commit is performed before unpausing. If the scratch PVC is lost (despite the finalizer), the
+  base images are consistent (application-consistent at the snapshot
+  point) but missing writes from the overlay window.
 
 ### Restore (unchanged)
 
@@ -360,7 +363,7 @@ VolumeSnapshots the same way it does today. No changes to the restore path.
 
 ## API Examples
 
-### VMSnapshot with overlay scratch size override
+### VMSnapshot with external snapshot enabled
 
 ```yaml
 apiVersion: snapshot.kubevirt.io/v1beta1
@@ -372,18 +375,24 @@ spec:
     apiGroup: kubevirt.io
     kind: VirtualMachine
     name: my-windows-vm
-  overlayScratchSize: "8Gi"  # optional, overrides default calculation
+  externalSnapshot: true          # optional, defaults to false (current flow)
+  overlayScratchSize: "8Gi"       # optional, overrides default scratch size
 ```
 
-### Overlay-active annotation on VMI during snapshot
+### VMI condition during overlay snapshot
 
 ```yaml
 apiVersion: kubevirt.io/v1
 kind: VirtualMachineInstance
 metadata:
   name: my-windows-vm
-  annotations:
-    snapshot.kubevirt.io/overlay-active: "my-snapshot"
+status:
+  conditions:
+  - type: OverlaySnapshotActive
+    status: "True"
+    reason: SnapshotInProgress
+    message: "Snapshot my-snapshot has active overlays"
+    lastTransitionTime: "2026-05-28T12:00:00Z"
 ```
 
 ### Scratch PVC with finalizer
@@ -435,6 +444,8 @@ blocks that changed since the last snapshot.
   2x permanent storage (a full-size target PVC per disk for the lifetime of
   the VM)
 - Additional PVC management complexity (sizing, lifecycle, storage class)
+- Uses the backup API (`virDomainBackupBegin`) for snapshot purposes,
+  conflating two APIs that are intentionally kept separate in libvirt/QEMU
 
 ### Tuning external-snapshotter parameters
 
--- a/veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
+++ b/veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
@@ -1,0 +1,554 @@
+# VEP #321: Online VMSnapshot Using QEMU External Snapshots
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.11
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+
+## Overview
+
+The current online VMSnapshot flow keeps the VM frozen while CSI snapshots are
+taken, making the freeze duration entirely dependent on CSI driver speed and
+Kubernetes snapshot infrastructure throughput. This VEP introduces a new flow
+that uses a QEMU external snapshot transaction to atomically establish the
+snapshot point at the libvirt level in sub-millisecond time, unfreeze the guest
+immediately, then take CSI VolumeSnapshots of the now-read-only base images
+asynchronously while the VM runs normally. After CSI snapshots are complete,
+block-commit the overlays back into the base images and clean up. The existing
+VMSnapshot flow remains unchanged and serves as the default.
+
+This makes the freeze duration independent of CSI driver speed, disk count, or
+Kubernetes API throughput.
+
+
+## Motivation
+
+The online VMSnapshot flow freezes the guest filesystem and creates
+CSI VolumeSnapshots for each of the VM's disks. The VM stays frozen until all
+snapshots have their `CreationTime` set, meaning the freeze duration depends
+entirely on how fast the CSI driver and the Kubernetes snapshot infrastructure
+can process the requests. Even for a single disk, a slow CSI driver can exceed
+acceptable freeze times. For VMs with multiple disks, tuning the
+external-snapshotter's `--kube-api-qps`/`--kube-api-burst`/`--worker-threads` can
+reduce the Kubernetes API throttling overhead, but even with optimal settings the
+bottleneck moves to the CSI driver and storage layer, which KubeVirt has no control
+over. This creates several problems:
+
+### 1. Windows VSS hard timeout
+
+Windows has a [non-configurable 10-second limit](https://learn.microsoft.com/en-us/windows/win32/vss/overview-of-processing-a-backup-under-vss)
+on how long the filesystem can be held frozen during shadow copy creation. When
+this limit is exceeded, the VSS provider returns
+[`VSS_E_HOLD_WRITES_TIMEOUT` (0x80042314)](https://support.microsoft.com/en-us/topic/time-out-errors-occur-in-volume-shadow-copy-service-writers-and-shadow-copies-are-lost-during-backup-and-during-times-when-there-are-high-levels-of-input-output-69abf5d3-eadd-9a9a-416a-d1a5752dbef4),
+VSS writers enter Failed state, and the snapshot is crash-consistent rather
+than application-consistent. If the CSI
+driver or Kubernetes snapshot infrastructure takes more than 10 seconds to
+process the snapshots, the freeze window exceeds this limit.
+
+### 2. Long freeze is dangerous on any OS
+
+Even without the Windows 10-second limit, keeping a filesystem frozen for
+extended periods is harmful on Linux as well:
+
+- **`fsfreeze` is designed for brief use**: intended for the short window
+  needed to take a storage-level
+  snapshot (typically <100ms for LVM/dm snapshots). Holding it for extended
+  periods while waiting for CSI round-trips is far outside the intended use case.
+- **Application disruption**: applications running inside the VM (databases,
+  web servers, etc.) experience blocked I/O during the freeze. This
+  can cause client disconnects, replication lag, and service interruptions,
+  especially for latency-sensitive workloads commonly migrated to KubeVirt.
+
+A snapshot operation should not keep the guest frozen while waiting for
+external infrastructure (CSI drivers, Kubernetes controllers) to respond.
+The freeze should only last as long as the hypervisor-level snapshot point
+needs to be established.
+
+
+## Goals
+
+- Reduce the VM freeze window during snapshot to sub-second, regardless of
+  CSI driver latency or number of disks
+- Make VMSnapshot work reliably for Windows VMs without exceeding the
+  VSS 10-second limit
+- Restore from snapshots taken with the new flow works identically to today -
+  the VolumeSnapshot CRs produced are the same, no changes to the restore path
+
+## Non Goals
+
+- Replacing CSI VolumeSnapshots with a different snapshot mechanism. We still
+  use CSI VolumeSnapshots for the actual storage-level snapshot
+- Offline VM snapshots. This VEP focuses on online (running) VM snapshots
+- Incremental backup integration. CBT/incremental backup is a separate feature
+  ([VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md))
+  that coexists with this proposal
+- Backup vendor integration. External backup providers (Velero, Kasten) that
+  manage their own CSI snapshots currently use freeze/unfreeze hooks. Exposing
+  the libvirt-level snapshot to these vendors via a dedicated CRD
+  (e.g. `VirtualMachineSnapshotRequest`) could be addressed in a follow-up VEP
+
+
+## Definition of Users
+
+- **VM owners:** Users who create VMSnapshots of their running VMs, especially
+  VMs where the CSI driver or disk count causes the freeze to exceed acceptable
+  durations
+- **Cluster administrators:** Operators who manage KubeVirt and need to
+  understand the new snapshot flow for troubleshooting
+
+
+## User Stories
+
+As a Windows VM owner, I want to take a VMSnapshot without VSS writers failing
+so that my backups are application-consistent and I can migrate from VMware
+to KubeVirt
+
+As a Linux VM owner running a database, I want the snapshot freeze window to
+be as short as possible so that my database connections don't time out and my
+application doesn't experience visible hangs during snapshot
+
+As a VM owner who snapshots regularly, I want VMSnapshot to work reliably
+regardless of how many disks the VM has or how fast the CSI driver is so that
+I can take snapshots without worrying about freeze timeouts or guest application
+disruption
+
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+
+## Design
+
+### Current flow (what changes)
+
+```
+1. Freeze guest filesystem (QEMU guest agent)
+2. Create N VolumeSnapshot CRs
+3. Wait for all CreationTime to be set - VM stays frozen
+4. Unfreeze guest filesystem
+```
+
+### New flow
+
+```
+Phase 0: Create scratch PVC, hotplug as UtilityVolume
+Phase 1: Freeze -> QEMU transaction (all disks, atomic, ~20ms) -> Unfreeze
+Phase 2: Create N VolumeSnapshot CRs (VM is unfrozen, no time pressure)
+Phase 3: Block-commit overlays back to base images
+Phase 4: Cleanup scratch volume
+```
+
+Freeze duration is sub-second regardless of disk count.
+
+### Phase 0: Scratch volume setup
+
+Before taking the snapshot, the snapshot controller:
+
+1. Creates a PVC (`snap-scratch-{uid}`) sized to hold qcow2 overlay files
+   for all disks during the overlay window. The size can be set explicitly
+   via `spec.overlayScratchSize` on the VMSnapshot. If not configured, a
+   default size is used
+   (see [Scalability](#scalability) for details).
+2. Hotplugs the PVC as a UtilityVolume to the VMI via JSON patch on
+   `spec.utilityVolumes` (same mechanism as [VEP #90 Utility Volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md),
+   same pattern as the incremental backup push-target PVC in [VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md)).
+3. Adds a Kubernetes finalizer `snapshot.kubevirt.io/overlay-protection`
+   on the scratch PVC to prevent its deletion while overlays are active.
+4. Waits for the volume to reach `VolumeReady` or `HotplugVolumeMounted` status.
+
+### Phase 1: QEMU atomic snapshot (overlay creation)
+
+Once the scratch volume is mounted, the snapshot controller calls a new
+`ExternalSnapshot` subresource on the VMI. Inside virt-launcher:
+
+```go
+dom.FSFreeze(nil, 0)
+dom.CreateSnapshotXML(snapshotXML,
+    DOMAIN_SNAPSHOT_CREATE_DISK_ONLY |
+    DOMAIN_SNAPSHOT_CREATE_ATOMIC |
+    DOMAIN_SNAPSHOT_CREATE_NO_METADATA)
+dom.FSThaw(nil, 0)
+```
+
+This is a single [`virDomainSnapshotCreateXML`](https://libvirt.org/html/libvirt-libvirt-domain-snapshot.html#virDomainSnapshotCreateXML)
+call that sends one QEMU `transaction` QMP command
+containing a [`blockdev-snapshot`](https://qemu-project.gitlab.io/qemu/interop/live-block-operations.html)
+action for every snapshotable disk. QEMU executes them all atomically in
+sub-millisecond time. Non-snapshotable disks (cloud-init, ephemeral) are
+marked with `snapshot='no'` in the XML.
+
+After this call:
+- The original disk images (disk.img / block devices) are **read-only** - the
+  snapshot point
+- New VM writes go to qcow2 overlay files on the scratch volume
+- The VM is unfrozen and running normally
+
+
+### Phase 2: CSI VolumeSnapshots
+
+The snapshot controller creates VolumeSnapshot CRs for each PVC - exactly
+the same as today. The only difference is that the VM is **unfrozen** during
+this phase. The CSI snapshots capture the read-only base images, which contain
+the exact frozen-point-in-time data regardless of how long the CSI processing
+takes.
+
+**Safety mechanisms:** While the VM is unfrozen during this phase, the overlay
+files on the scratch PVC continue to grow with every write. Two mechanisms
+prevent the scratch PVC from filling up:
+
+- **Overlay usage monitoring:** virt-launcher monitors scratch volume usage
+  in the background (same pattern as the freeze auto-unfreeze safety timer).
+  If usage crosses a threshold (e.g. 80%), it notifies the snapshot controller,
+  which aborts Phase 2, deletes pending VolumeSnapshot CRs, triggers
+  block-commit to restore the VM, and marks the VMSnapshot as Failed with a
+  clear error indicating insufficient scratch space.
+
+- **Timeout:** Phase 2 is bounded by the VMSnapshot's existing
+  `FailureDeadline` (default: 5 minutes, configurable per-snapshot). If any
+  VolumeSnapshot has not received its `CreationTime` within this deadline,
+  the same abort flow is triggered.
+
+The scratch PVC is sized to accommodate the maximum possible writes within
+the `FailureDeadline` (see [Scalability](#scalability)).
+
+### Phase 3: Block-commit
+
+After all VolumeSnapshots have `CreationTime` set, the snapshot controller
+calls a new `CommitSnapshot` subresource on the VMI. Inside virt-launcher,
+for each disk:
+
+1. **Check state**: Is the disk on overlay? If already on base, skip (idempotent).
+   Is there an existing block job? Resume waiting instead of starting a new one.
+2. **Start commit**: `dom.BlockCommit(disk, "", "", 0, ACTIVE | DELETE)` ([ref](https://libvirt.org/kbase/merging_disk_image_chains.html))
+3. **Wait for READY**: Parse domain XML for `<mirror ready='yes'>` attribute,
+   which reflects libvirt's internal state after receiving QEMU's
+   [`BLOCK_JOB_READY`](https://libvir-list.redhat.narkive.com/KKDmcDw5/libvirt-rfc-exposing-ready-bool-of-query-block-jobs-or-qmp-block-job-ready-event) event.
+4. **Pivot**: `dom.BlockJobAbort(disk, PIVOT)` with retry (10x, 200ms backoff).
+5. **Verify**: Read domain XML and confirm the disk source is back on the
+   original base image. If still on overlay, return error - do NOT proceed
+   to cleanup.
+
+After ALL disks are committed and verified, a final check reads the full
+domain XML and confirms every disk source is back on the base image.
+
+The block-commit is idempotent and resumable: already committed disks are
+skipped, active block jobs from a previous attempt are resumed. After all
+disks are committed, the domain XML is verified to confirm every disk is
+back on the base image before proceeding to cleanup. If commit fails, the
+scratch PVC is preserved (finalizer prevents deletion) and the controller
+requeues for retry.
+
+### Phase 4: Cleanup
+
+Once all disks are back on their base images:
+- Clear the `snapshot.kubevirt.io/overlay-active` annotation on the VMI
+- Detach the utility volume via JSON patch
+- Remove the finalizer from the scratch PVC
+- Delete the scratch PVC
+
+### New flow sequence diagram
+
+```mermaid
+sequenceDiagram
+    actor User
+    participant SC as Snapshot Controller
+    participant VH as virt-handler
+    participant VL as virt-launcher
+    participant QEMU
+    participant CSI
+
+    User->>SC: Create VMSnapshot
+
+    Note over SC,VL: Phase 0: Scratch Volume Setup
+    SC->>SC: Create scratch PVC + finalizer
+    SC->>VH: Hotplug as UtilityVolume
+    VH->>VL: Mount scratch volume
+    VL-->>SC: Volume Ready
+
+    Note over SC,QEMU: Phase 1: Atomic Snapshot
+    SC->>VH: ExternalSnapshot subresource
+    VH->>VL: gRPC
+    VL->>QEMU: FSFreeze
+    QEMU-->>VL: OK
+    Note over VL,QEMU: FS FROZEN
+    VL->>QEMU: CreateSnapshotXML (DISK_ONLY | ATOMIC)
+    Note over QEMU: QEMU transaction:<br/>all disks switched to<br/>overlay in ~20ms
+    QEMU-->>VL: OK
+    VL->>QEMU: FSThaw
+    QEMU-->>VL: OK
+    VL-->>SC: disk.img now read-only, overlays active
+
+    Note over SC,CSI: Phase 2: CSI VolumeSnapshots (VM unfrozen)
+    Note over VL: VM running normally,<br/>writes go to overlays
+    SC->>CSI: Create VolumeSnapshot per PVC
+    Note over CSI: disk.img is read-only,<br/>content frozen in time
+    CSI-->>SC: All VolumeSnapshot CreationTimes set
+
+    Note over SC,QEMU: Phase 3: Block-Commit
+    SC->>VH: CommitSnapshot subresource
+    VH->>VL: gRPC
+
+    loop For each disk
+        VL->>QEMU: BlockCommit (overlay to base)
+        Note over QEMU: Sync overlay data<br/>into base image
+        QEMU-->>VL: mirror ready='yes'
+        VL->>QEMU: BlockJobAbort(PIVOT)
+        QEMU-->>VL: OK
+        VL->>VL: Verify disk back on base
+    end
+
+    VL->>VL: verifyAllDisksOnBase()
+    VL-->>SC: All disks back on base images
+
+    Note over SC,VH: Phase 4: Cleanup
+    SC->>SC: Clear overlay-active annotation
+    SC->>VH: Detach UtilityVolume
+    SC->>SC: Remove finalizer, delete scratch PVC
+
+    SC-->>User: Snapshot Ready
+    Note over User,CSI: VolumeSnapshot CRs identical to current flow
+```
+
+### Overlay state tracking
+
+An annotation `snapshot.kubevirt.io/overlay-active` is set on the VMI when
+overlays are active (after Phase 1) and cleared after successful commit
+(after Phase 3). This annotation is checked by:
+
+- **Migration**: blocks `startMigration()` - migration with overlays would
+  break the backing chain on the target node
+- **Backup**: blocks `BackupVirtualMachine()` - backup checkpoints on an
+  overlay become invalid after commit
+- **Volume unplug**: blocks `removeVolumeRequestHandler()` - unplugging a
+  disk with an active overlay would orphan the overlay
+- **Disk resize**: skips resize in `syncDisks()` - resizing during overlay
+  could cause size mismatch between overlay and base
+- **VM destroy**: `KillVMI()` aborts active block jobs before
+  `DestroyFlags()` - prevents orphaned block jobs
+- **VM crash / pod eviction**: the overlay + base together remain consistent
+  (no data loss). Once the VM comes back up, the pending commit resumes and
+  completes the cleanup. The scratch PVC finalizer ensures it survives across
+  restarts.
+
+### Fallback
+
+If `ExternalSnapshot` fails for any reason (no guest agent, libvirt error,
+unsupported configuration), the controller falls back to the current
+sequential CSI path automatically, as the new flow is regarded as an
+optimization, not a requirement.
+
+### Restore (unchanged)
+
+The VolumeSnapshot CRs produced by the new flow are identical to what the
+current flow produces - they capture the same disk.img content at the same
+frozen point-in-time. The restore controller creates PVCs from these
+VolumeSnapshots the same way it does today. No changes to the restore path.
+
+## API Examples
+
+### VMSnapshot with overlay scratch size override
+
+```yaml
+apiVersion: snapshot.kubevirt.io/v1beta1
+kind: VirtualMachineSnapshot
+metadata:
+  name: my-snapshot
+spec:
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: my-windows-vm
+  overlayScratchSize: "8Gi"  # optional, overrides default calculation
+```
+
+### Overlay-active annotation on VMI during snapshot
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-windows-vm
+  annotations:
+    snapshot.kubevirt.io/overlay-active: "my-snapshot"
+```
+
+### Scratch PVC with finalizer
+
+```yaml
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: snap-scratch-ae375815
+  labels:
+    snapshot.kubevirt.io/scratch: my-snapshot
+  finalizers:
+  - snapshot.kubevirt.io/overlay-protection
+spec:
+  accessModes: [ReadWriteOnce]
+  resources:
+    requests:
+      storage: 6Gi
+```
+
+
+## Alternatives
+
+### Push-mode backup to separate PVC
+
+Instead of switching VM disks to overlays, use KubeVirt's existing push-mode
+backup API (which uses `virDomainBackupBegin` under the hood) to copy
+consistent point-in-time disk data to separate target PVCs. QEMU preserves
+old blocks before overwriting them while a background job copies the data
+out. The snapshot controller would then take CSI VolumeSnapshots of those
+target PVCs and attach them to the VMSnapshot.
+
+If the target PVCs are kept persistently and combined with CBT (Changed Block
+Tracking), the first VMSnapshot of a VM performs a full copy of every disk to
+the target PVCs, but any subsequent VMSnapshot of the same VM only copies the
+blocks that changed since the last snapshot.
+
+**Pros:**
+- No overlays on the live disk chain, no block-commit, no live merge risk
+- Avoids the complexity of managing overlay lifecycle entirely
+- If target PVCs are kept persistently and combined with CBT, subsequent
+  snapshots only push the delta (changed blocks), making them fast
+
+**Cons:**
+- Without persistent target PVCs, every snapshot requires a full data copy
+  of every disk, which can be heavy I/O for large VMs and take minutes to
+  hours
+- With persistent target PVCs, the full copy only happens once, but requires
+  2x permanent storage (a full-size target PVC per disk for the lifetime of
+  the VM)
+- Additional PVC management complexity (sizing, lifecycle, storage class)
+
+### Tuning external-snapshotter parameters
+
+Increase `--kube-api-qps`, `--kube-api-burst`, and `--worker-threads` on
+the external-snapshotter sidecar to reduce Kubernetes API throttling during
+multi-disk snapshot creation.
+
+**Pros:**
+- No code changes in KubeVirt
+- Can meaningfully reduce the freeze window for multi-disk VMs where API
+  throttling is the bottleneck
+
+**Cons:**
+- Only addresses the Kubernetes API overhead, not the CSI driver latency.
+  Once throttling is eliminated, the bottleneck moves to the storage layer
+  which KubeVirt has no control over
+- A single slow CSI driver can still exceed the VSS 10-second limit
+  regardless of tuning
+- Requires cluster-level configuration changes that may not be feasible
+  in all environments
+- Does not solve the fundamental problem: the VM stays frozen while
+  waiting for external infrastructure
+
+
+## Scalability
+
+- **Scratch PVC sizing.** Users can set `spec.overlayScratchSize` on the
+  VMSnapshot to control the scratch PVC size directly. If not set, the
+  following default calculation is used:
+
+  ```
+  min(FailureDeadline × 125MB/s × 2, total_disk_size × 1.1)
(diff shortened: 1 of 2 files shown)


fabiand:
A really good idea.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,556 @@
...
+marked with `snapshot='no'` in the XML.
+
+After this call:
+- The original disk images (disk.img / block devices) are **read-only** - the
+  snapshot point
+- New VM writes go to qcow2 overlay files on the scratch volume
+- The VM is unfrozen and running normally
+
+The `QUIESCE` flag is not used because KubeVirt's QEMU guest agent command
+allowlist blocks direct `guest-fsfreeze-freeze` calls from libvirt. Instead,
```
michalskrivanek: what are you referring to?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,556 @@
...
+    SC->>VH: Detach UtilityVolume
+    SC->>SC: Remove finalizer, delete scratch PVC
+
+    SC-->>User: Snapshot Ready
+    Note over User,CSI: VolumeSnapshot CRs identical to current flow
+```
+
+### Overlay state tracking
+
+An annotation `snapshot.kubevirt.io/overlay-active` is set on the VMI when
```
michalskrivanek: why annotation? why not a VMI status or something "closer" to the virt-launcher VM lifecycle code?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,556 @@
...
+  (no data loss). Once the VM comes back up, the pending commit resumes and
+  completes the cleanup. The scratch PVC finalizer ensures it survives across
+  restarts.
+
+### Fallback
+
+If `ExternalSnapshot` fails for any reason (no guest agent, libvirt error,
+unsupported configuration), the controller falls back to the current
+sequential CSI path automatically, as the new flow is regarded as an
+optimization, not a requirement.
```
michalskrivanek: what if I prefer the snapshot to fail instead? if the long fsfreeze is problematic for me then i may not want a silent fallback. And vice versa, the flows are significantly different, and risks are significantly different too - live merge can go wrong for so many reasons - that I think this deserves a failsafe even when GA

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,556 @@
...
+
+### Tuning external-snapshotter parameters
+
+Increase `--kube-api-qps`, `--kube-api-burst`, and `--worker-threads` on
+the external-snapshotter sidecar to reduce Kubernetes API throttling during
+multi-disk snapshot creation.
+
+**Pros:**
+- No code changes in KubeVirt
+- Can meaningfully reduce the freeze window for multi-disk VMs where API
```
michalskrivanek, line 458: is there a recommended tuning? that should be done in any case, no? reducing the freeze window makes sense for the new snapshot too - less data to merge back.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
michalskrivanek: especially since the conflicting operations below are so problematic it needs to be properly synchronized all over the place. E.g. you should probably block the phase 0 already if any of the below is active, not _after_ Phase 1 , that's too late and racy.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,556 @@
...
+- **Backup**: blocks `BackupVirtualMachine()` - backup checkpoints on an
+  overlay become invalid after commit
+- **Volume unplug**: blocks `removeVolumeRequestHandler()` - unplugging a
+  disk with an active overlay would orphan the overlay
+- **Disk resize**: skips resize in `syncDisks()` - resizing during overlay
+  could cause size mismatch between overlay and base
+- **VM destroy**: `KillVMI()` aborts active block jobs before
+  `DestroyFlags()` - prevents orphaned block jobs
+- **VM crash / pod eviction**: the overlay + base together remain consistent
+  (no data loss). Once the VM comes back up, the pending commit resumes and
```
michalskrivanek: this probably needs more elaboration on possible failures. (and one more reason I think we shouldn't rely on annotation at all, we should try to discover the current state on startup).
I.e. if VM starts again then something must discover there's pending merge to do and start it, and if the overlays are gone or the job can't be restarted for whatever reason we're screwed. We're also may not be able to merge back the overlay within a reasonable time even if you try to abort before the scratch gets full, what then?

specifically for crash/shutdown I wonder if it doesn't deserve a pause mode start just to finish the merge.

michalskrivanek (commented):
That's a great proposal addressing one of the big pain points. However the risks stemming from live merge shouldn't be underestimated, we need to make the failure scenarios and mitigations clear

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: yeah this is a bit unclear sorry, i was refferring to use libvirt's FSFreeze/FSThaw rather than using qemu-ga commands directly.. actually `VIR_DOMAIN_SNAPSHOT_CREATE_QUIESCE` flag is supported, but we dont use it to have control over the freeze/thaw lifecycle and error handling... but regardless, i think i can just remove this line.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: qps and burst were increased to from 5/10 to 30/60 here https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/261,
(for csi-snapshotter and snapshot-controller containers and relevant for multi-disks snapshot)

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: thanks, i agree, a status condition (set at the start of phase 0) makes more sense than an annotation for this

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: what do you say about adding a field to the VMSnapshot spec (for example `spec.snapshotMode`) that lets the user choose between the two flows? That way each flow fails independently, and both remain available at GA?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: yeah, the pause mode start idea is definitely safer, maybe we can even run the block-commit offline before the domain starts? i think virt-launcher already does offline qemu-img operations during pre-start, so the same pattern could merge the overlays back before booting

i also agree that we should discover the actual state on startup rather than relying purely on the condition ofc... since the scratch PVC has a finalizer for this, the overlay files should technically survive restarts, do you think there is a safer approach?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,554 @@
...
+```
+
+This is a single [`virDomainSnapshotCreateXML`](https://libvirt.org/html/libvirt-libvirt-domain-snapshot.html#virDomainSnapshotCreateXML)
+call that sends one QEMU `transaction` QMP command
+containing a [`blockdev-snapshot`](https://qemu-project.gitlab.io/qemu/interop/live-block-operations.html)
+action for every snapshotable disk. QEMU executes them all atomically in
+sub-millisecond time. Non-snapshotable disks (cloud-init, ephemeral) are
+marked with `snapshot='no'` in the XML.
+
+After this call:
```
Acedus, line 201: How does this work with disks that already have a qcow2 overlay such as in the case of CBT?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,554 @@
...
+the same as today. The only difference is that the VM is **unfrozen** during
+this phase. The CSI snapshots capture the read-only base images, which contain
+the exact frozen-point-in-time data regardless of how long the CSI processing
+takes.
+
+**Safety mechanisms:** While the VM is unfrozen during this phase, the overlay
+files on the scratch PVC continue to grow with every write. Two mechanisms
+prevent the scratch PVC from filling up:
+
+- **Overlay usage monitoring:** virt-launcher monitors scratch volume usage
```
Acedus, line 220: What if the threshold is crossed too fast and the PVC fills out completely? What would the implications be on the VM?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,554 @@
...
+
+### Overlay state tracking
+
+An annotation `snapshot.kubevirt.io/overlay-active` is set on the VMI when
+overlays are active (after Phase 1) and cleared after successful commit
+(after Phase 3). This annotation is checked by:
+
+- **Migration**: blocks `startMigration()` - migration with overlays would
+  break the backing chain on the target node
+- **Backup**: blocks `BackupVirtualMachine()` - backup checkpoints on an
```
Acedus: Not sure I understand, does that mean that post CSI snapshot the backup chain (e.g., checkpoints / bitmaps) become invalid?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
Acedus: +1 with current behavior as default to A. maintain compatibility and B. the proposed approach outlined in this VEP is more expensive (i.e., requires a scratch space PVC).

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,554 @@
...
+the target PVCs, but any subsequent VMSnapshot of the same VM only copies the
+blocks that changed since the last snapshot.
+
+**Pros:**
+- No overlays on the live disk chain, no block-commit, no live merge risk
+- Avoids the complexity of managing overlay lifecycle entirely
+- If target PVCs are kept persistently and combined with CBT, subsequent
+  snapshots only push the delta (changed blocks), making them fast
+
+**Cons:**
```
Acedus, line 439: Conflates two APIs that are purposefully kept separate.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
michalskrivanek: +1
snapshotMode sounds good, as long as it's fully controllable/selectable

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
michalskrivanek: "pause" might be easier to manage (not sure how long the offline operation can take before other kubevirt components decides to kill it since the VM is not there:)

yeah i think the overlays are good enough, we can't do much if someone deletes them. but other than that the rest should be discoverable

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: > that should be done in any case, no? reducing the freeze window makes sense for the new snapshot too - less data to merge back.

thats right, this improves csi-snapshot taking time regardless if old or this new suggested flow is used.. the goal of adding this alternative to the VEP was to highlight that with the current design (fs frozen during csi snapshot), there are ways to tweak and make sure that k8s/kubevirt is not limiting factor in CSI snapshot speed, moving it to be solely dependent on the csi-driver and the storage layer

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: > yeah i think the overlays are good enough, we can't do much if someone deletes them. but other than that the rest should be discoverable

ok, i will add that if overlays are not found(worst case), the vm will come up with the application-consistent state at snapshot taking

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: thanks for pointing this out, looked into it.. from my understanding, libvirt handles this. when creating an external snapshot, `qemuSnapshotDiskBitmapsPropagate` adds matching bitmaps to the snapshot overlay as part of the same QMP, so dirty tracking should continue on the overlay and during block-commit, `qemuBlockBitmapsHandleCommitFinish` merges the overlay's bitmaps back into disk.img

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: by default the VM pauses and stays paused until either the space issue is resolved or the VM is restarted. our overlay usage monitoring in the vep is designed to abort before this point (at 80% usage) to keep the vm running and never reach paused vm (anyways, trigger abort if VM is paused as well).  let me know if you think this monitoring is worth it or not

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: actually (according to [the other comment](https://github.com/kubevirt/enhancements/pull/322#discussion_r3379488408)) we expect the bitmaps to remain valid, the issue here is just with concurrent backup and snapshot.. fixing

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: eventually went with `spec.externalSnapshot: true` (false is default) - let me know if you still prefer `spec.snapshotMode `

michalskrivanek:
Not as part of this VEP, but I would be very excited if you could consider adding memory to the snapshot as well to have complete fully consistent snapshots in future. hooks are very specific and "difficult" to use with VMs.
We're finally getting to a point where it will be possible.

noamasu:
@michalskrivanek sure, added in non-goals https://github.com/kubevirt/enhancements/pull/322/changes#diff-4249140aa5c5ea58cd3ea79b52f81912aa5f87e96287335abe54ebd3d5334167R102-R104 (under Memory state capture)
let me know if you think this deserves a more detailed/different wording

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+```
+
+Freeze duration is sub-second regardless of disk count.
+
+### Phase 0: Scratch volume setup
+
+Before taking the snapshot, the snapshot controller:
+
+1. Sets the `OverlaySnapshotActive` condition on the VMI.
+2. Creates a PVC (`snap-scratch-{uid}`) sized to hold qcow2 overlay files
```
mhenriks, line 168: uid of what?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+  VolumeSnapshot has not received its `CreationTime` within this deadline,
+  the same abort flow is triggered.
+
+The scratch PVC is sized to accommodate the maximum possible writes within
+the `FailureDeadline` (see [Scalability](#scalability)).
+
+### Phase 3: Block-commit
+
+After all VolumeSnapshots have `CreationTime` set, the snapshot controller
+calls a new `CommitSnapshot` subresource on the VMI. Inside virt-launcher,
```
mhenriks, line 238: Is`CommitSnapshot` synchronous or async? If it is something that may take a while consider async. I assume `ExternalSnapshot` would be fast

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+  could cause size mismatch between overlay and base
+- **VM destroy**: `KillVMI()` aborts active block jobs before
+  `DestroyFlags()` - prevents orphaned block jobs
+- **VM crash / pod eviction**: the scratch PVC finalizer ensures it
+  survives across restarts. On startup, virt-launcher inspects the actual
+  disk state (checks for overlay files on the scratch PVC). If overlays
+  are found, the VM starts in paused mode with the overlay still active
+  and a live block-commit is performed before unpausing. If the scratch PVC is lost (despite the finalizer), the
+  base images are consistent (application-consistent at the snapshot
+  point) but missing writes from the overlay window.
```
mhenriks, line 355: Need to go into more detail on failure recovery. Some random thoughts:

When QEMU dies, the VMI enters Failed phase. With `RunStrategyAlways` or `RerunOnFailure`, the VM controller deletes the Failed VMI and creates a new one from the VM spec. How do we know to mount the scratch volume and apply overlays, etc?

What happens if we crash during block commit? Some disks may have pivoted, some not. Is it possible that there is still a qcow for a pivoted disk? How do we know not to pivot again because that could cause data corruption? Or does qemu protect from this somehow?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+  in all environments
+- Does not solve the fundamental problem: the VM stays frozen while
+  waiting for external infrastructure
+
+
+## Scalability
+
+- **Scratch PVC sizing.** Users can set `spec.overlayScratchSize` on the
+  VMSnapshot to control the scratch PVC size directly. If not set, the
+  following default calculation is used:
```
mhenriks, line 477: Not just sizing, have to think about performance going from writing to N disks down to 1. Could become a bottleneck especially in database workloads that have log+data disks

mhenriks (commented):
Very cool feature! Excited to see how this goes. Mostly concerned about failure handling and recovery

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: > Need to go into more detail on failure recovery.

I agree, as soon as i have a better picture on how to mitigate all of recovery edgecases, i will add a more detailed section. 

> When QEMU dies, the VMI enters Failed phase. With RunStrategyAlways or RerunOnFailure, the VM controller deletes the Failed VMI and creates a new one from the VM spec. How do we know to mount the scratch volume and apply overlays, etc?

hmm didnt thought about it, thats a good point. still trying to find ways to overcome this... I wonder if extending utility volumes to support persistent VM-level state is possible, or using a second backend-storage-like PVC that survives VMI recreation could work here? or (not ideal) maybe we can let the snapshot controller deal with re-attaching the utilityvolume?

the base images are always at an app-consistent state (froze before switching to overlays). if the crash happens before any commit, the VM can boot normally from that point. If it happens during commit, the CSI VolumeSnapshots already capture the consistent state, so a restore brings the VM back. Either way, we have a very recent application-consistent point to recover from, considering a worst case scenario of losing the overlays

anyways, still thinking it through..

> What happens if we crash during block commit? Some disks may have pivoted, some not. Is it possible that there is still a qcow for a pivoted disk? How do we know not to pivot again because that could cause data corruption? Or does qemu protect from this somehow?

(if we are able to preserve the overlays considering the former) The commit is idempotent since each disk is checked via the XML before committing. from what i saw, QEMU also rejects a pivot on a disk with no active block job, so i think a double-pivot isn't possible. i will remove the `DELETE` from the blockcommit command so overlay files will be deleted only when the pvc is deleted (the scratch PVC cleanup handles that),

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: yes, `ExternalSnapshot` is indeed fast so it stays synchronous and `CommitSnapshot` is going to be async (following a VMI status condition).

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: sry, just meant a unique suffix to generate a distinct pvc name... will change it to {vmsnapshot-name}-snap-scratch

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: I was thinking about having a separate PVC for each overlay originally, but eventually went with the same implementation as backup (also uses a single utility volume for all disks), considering the scratch PVC is short-lived so the IO performance hit is bounded. that said, if we decide to go with one PVC per disk instead, we need to keep in mind environments with PVC count limits. what do you think we we should do here?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: @mhenriks,  maybe we can set a vm.status.activeOverlay (or annotation) field (with the PVC name) when the external snapshot overlays are created and stays set until the cleanup phase.
If the VMI crashes while `activeOverlay` is set, the VM controller would see it on the next VMI creation and include the scratch PVC as a utility volume directly in the VMI spec so the new VMI is created with it, not patched after. (will also require changes in the create admitter) and run qemu-img commit (offline) before the domain starts?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+Kubernetes snapshot infrastructure throughput. This VEP introduces a new flow
+that uses a QEMU external snapshot transaction to atomically establish the
+snapshot point at the libvirt level in sub-millisecond time, unfreeze the guest
+immediately, then take CSI VolumeSnapshots of the now-read-only base images
+asynchronously while the VM runs normally. After CSI snapshots are complete,
+block-commit the overlays back into the base images and clean up. The existing
+VMSnapshot flow remains unchanged and serves as the default.
+
+This makes the freeze duration independent of CSI driver speed, disk count, or
+Kubernetes API throughput.
```
EdDev, line 34: This is the motivation.
Please provide here a short summary of what this VEP is all about and the change it suggests.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
```diff
@@ -0,0 +1,565 @@
...
+```
+1. Freeze guest filesystem (QEMU guest agent)
+2. Create N VolumeSnapshot CRs
+3. Wait for all CreationTime to be set - VM stays frozen
+4. Unfreeze guest filesystem
+```
+
+### New flow
+
+The new flow is enabled by setting `spec.externalSnapshot: true` on the
```
EdDev, line 150: Are you suggesting a new API field? If you do, please express it before entering the flow so we know what is new and what exists already.

Note that booleans are not recommended for API usage.
Guidelines about this are in the working, see https://github.com/kubevirt/kubevirt/pull/18612  for more details.

EdDev (commented):
This is a very partial review, especially on the requirement section and API.

I found the motivation very clear, thanks.

It is targeted for v1.11, so I guess more work is planned here.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
noamasu: yes, this will be a new api field. it is expressed in the API Examples section, but i can definitely mention it before the flow as well
> Note that booleans are not recommended for API usage.

thanks, Noted! i will use `spec.snapshotMethod` instead

iholder101:
> It is targeted for v1.11, so I guess more work is planned here.

@EdDev just FYI that the target was changed to v1.10.

akalenyu:
/cc @noamasu 
we briefly discussed this during sig-storage meeting, do we want to try asking for a small exception?
(vep freeze next week, this is targeted 1.10)

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots.md
mhenriks: The new "Crash recovery" section addresses this well: vm.status.snapshotInProgress + VMSnapshotContent.status.snapshotMethod drive the VM controller to re-attach the scratch PVC and re-run recovery on VMI recreation, and offline qemu-img commit being idempotent means there's no need to track per-disk pivot state (safe to blindly re-commit every overlay found). Recovery-of-recovery failure and scratch-PVC loss are also covered. LGTM on this thread.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,684 @@
...
+```
+
+### Overlay state tracking
+
+A VMI condition `OverlaySnapshotActive` is set at Phase 1 and removed after
+commit (Phase 3). Only set during external snapshots, the existing flow is
+unaffected. The condition is checked by:
+
+- **Migration**: blocks `startMigration()` - the overlay references the
+  base image on the source node which would not exist on the target
```
mhenriks, line 353: Not sure if we already block vmsnapshots when a migration is in progress but I assume we will want to do that now?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,684 @@
...
+     125MB/s is a conservative estimate of maximum sustained write throughput,
+     and `× 2` is a safety margin to account for write bursts.
+
+  2. **Disk capacity cap** (`total_disk_size × 1.1`): `total_disk_size` is the
+     sum of all snapshotable disk virtual sizes. The overlay can never exceed
+     this, since writes are bounded by the guest's addressable space, plus
+     extra 10% for metadata.
+
+  For example, a VM with 10 × 10Gi disks and default 5-minute deadline:
+  `min(5min × 125MB/s × 2, 100Gi × 1.1) = min(75Gi, 110Gi) = 75Gi`.
```
mhenriks: mixing units here I think if using MB should also use GB?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: right, we currently do not block vmsnapshots when migration happens (disk layout doesn't change, so no need to block), but for this proposed flow blocking is needed, since qemu writes to overlay qcow2 files that reference the base images by path. migrating the domain would break this backing chain on the target node

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: noted! Switched to Mi/Gi. (kept the same numbers for simplicity)

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,684 @@
...
+
+
+## Update/Rollback Compatibility
+
+The feature is additive and behind a feature gate. On upgrade, the existing
+VMSnapshot flow continues to work unchanged until the feature gate is enabled.
+On rollback, disabling the feature gate reverts VMSnapshot to the current
+sequential CSI path. No data migration is needed. Snapshots taken with the
+new flow are restorable by any version, since the VolumeSnapshot CRs produced
+are standard Kubernetes objects with no format changes.
```
stefanhaRH: Rollback is only possible when there are no active snapshot operations in any VM. In an error scenario where `qemu-img commit` is failing, downgrade cannot be performed since the scratch PVC is still around and the VM resources contained `External` snapshot metadata that the old version does not support.

One way out would be to drop the overlay data (scratch PVC), remove the `External` snapshot metadata from the VM resources, and then downgrade so VMs launch from the last point-in-time snapshot.

stefanhaRH:
Kubernetes and CSI storage features may encounter issues with external snapshots since VM PVCs no longer have a 1:1 relationship with the VM's disks. For example, will CSI plugin I/O statistics still be accurate during the snapshot creation period when writes are redirected to the scratch PVC, and could this lead to false monitoring alerts if I/O to the PVC suddenly cuts out?

stefanhaRH:
Is there an interaction with live storage migration (moving data from one PVC to another PVC while the guest is running)?

stefanhaRH:
How will snapshot requests in rapid succession be handled? The VEP does not discuss chaining, so I think the intention is to only have 1 snapshot operation active at any given time? Will the user experience a delay on new snapshot operations or will they fail when there is already a snapshot in progress?

stefanhaRH:
I didn't see anything that spells out what happens when the PVC snapshot is still being created by the CSI plugin when a recovery scenario starts (e.g. virt-launcher Pod dies). Recovery must wait for the snapshots to be created - maybe this should be explicit in the VEP if it is not already.

stefanhaRH:
If the user decides to remove the VM (especially in a case where there was a failure and the VM is offline with an active snapshot operation), does the VEP need text to describe what happens in that case? If the user wants to keep the VM's PVCs but delete the KubeVirt resources, then it isn't safe to do that until the snapshot has been committed back to the PVCs (recent writes to the disk would be lost).

stefanhaRH:
Are there any concerns with directing all writes to a single scratch PVC during snapshot creation? Performance will be different from the disk PVCs and concentrating it all onto 1 scratch PVC could make a noticeable QoS difference (e.g. applications like etcd that have strict latency requirements).

Additionally, the disk PVC's storage class might offer something like a replication factor that is different from that offered by the single scratch PVC. I'm not sure if users would be comfortable with that, so maybe there should be 1 scratch PVC per storage class used by the disk PVCs?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: > the VM resources contained External snapshot metadata that the old version does not support.

We use `CreateSnapshotXML` with the `VIR_DOMAIN_SNAPSHOT_CREATE_NO_METADATA` flag, so no metadata is stored in libvirt's snapshot tree (or persisted to disk), This avoids state we would need to reconcile across pod restarts and crashes, so we track the overlay lifecycle (at kubevirt level) instead. for recovery, `qemu-img commit` operates on the qcow2 backing chain directly and does not require this snapshot metadata

> One way out would be to drop the overlay data (scratch PVC), remove the External snapshot metadata from the VM resources, and then downgrade so VMs launch from the last point-in-time snapshot.

I'll add a note clarifying that if `qemu-img` commit fails (overlays not yet committed), restarting the VM boots it from the base images (app-consistent state from the last CSI snapshot point), losing the post-snapshot overlay writes.

noamasu:
thanks for the valuable feedback @stefanhaRH!

> Kubernetes and CSI storage features may encounter issues with external snapshots since VM PVCs no longer have a 1:1 relationship with the VM's disks. For example, will CSI plugin I/O statistics still be accurate during the snapshot creation period when writes are redirected to the scratch PVC, and could this lead to false monitoring alerts if I/O to the PVC suddenly cuts out?

Good point. If monitoring is configured for per-PVC I/O, this could definitely trigger false alerts. it will probably happen in any external snapshot approach that redirects writes. maybe since ours is relatively short lived, it might go unnoticed? also, from what i saw, PVC alerts usually focus on disk space rather than I/O throughput so i wonder how common these type alerts might be.

> Is there an interaction with live storage migration (moving data from one PVC to another PVC while the guest is running)?

Thanks, Yes, both are block-level operations on the same disk and can conflict. We should block volume migration while `OverlaySnapshotActive` is set (same as compute migration and backup). I'll add it.

> How will snapshot requests in rapid succession be handled? The VEP does not discuss chaining, so I think the intention is to only have 1 snapshot operation active at any given time? Will the user experience a delay on new snapshot operations or will they fail when there is already a snapshot in progress?

This is actually existing behavior (prior to this VEP). `vm.status.snapshotInProgress` acts as a lock, only one snapshot can be active per VM at a time so a second snapshot request will wait (not fail) until the current one completes


> I didn't see anything that spells out what happens when the PVC snapshot is still being created by the CSI plugin when a recovery scenario starts (e.g. virt-launcher Pod dies). Recovery must wait for the snapshots to be created - maybe this should be explicit in the VEP if it is not already.

Right, good catch. so if a crash happens, committing the overlays before snapshots complete at CSI level would break app-consistency. We can either wait for the CSI `VolumeSnapshots` to complete (get `CreationTime`) before committing, or fail the `VMSnapshot` entirely. I'll address this in the crash recovery section.

> If the user decides to remove the VM (especially in a case where there was a failure and the VM is offline with an active snapshot operation), does the VEP need text to describe what happens in that case? If the user wants to keep the VM's PVCs but delete the KubeVirt resources, then it isn't safe to do that until the snapshot has been committed back to the PVCs (recent writes to the disk would be lost).

The VM is already protected by a finalizer (`snapshot.kubevirt.io/snapshot-source-protection`) during an active snapshot, so it can't be deleted until the snapshot completes. This is existing behavior that will apply to external snapshots as well.

> Are there any concerns with directing all writes to a single scratch PVC during snapshot creation? Performance will be different from the disk PVCs and concentrating it all onto 1 scratch PVC could make a noticeable QoS difference (e.g. applications like etcd that have strict latency requirements).
Additionally, the disk PVC's storage class might offer something like a replication factor that is different from that offered by the single scratch PVC. I'm not sure if users would be comfortable with that, so maybe there should be 1 scratch PVC per storage class used by the disk PVCs?

This was discussed with @mhenriks [here](https://github.com/kubevirt/enhancements/pull/322#discussion_r3398909711) (still open). in short we went with a single PVC to match the backup (CBT) flow, and also since low PVC count limit per node can be a real challenge.
i agree that If a VM has disks across multiple storage classes we could at least create one scratch PVC per storage class. im currently testing this approach and will update the VEP after some tests.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,698 @@
...
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+
+## Design
+
+### Current flow (what changes)
+
+```
+1. Freeze guest filesystem (QEMU guest agent)
```
Acedus: nit: formatting of old flow should match new flow.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,554 @@
...
+the same as today. The only difference is that the VM is **unfrozen** during
+this phase. The CSI snapshots capture the read-only base images, which contain
+the exact frozen-point-in-time data regardless of how long the CSI processing
+takes.
+
+**Safety mechanisms:** While the VM is unfrozen during this phase, the overlay
+files on the scratch PVC continue to grow with every write. Two mechanisms
+prevent the scratch PVC from filling up:
+
+- **Overlay usage monitoring:** virt-launcher monitors scratch volume usage
```
Acedus: It still suffers from a timing issue, the cancellation here would be best effort. The question is what are the implications for the worst case scenario and whether we can recover from it.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,698 @@
...
+  which aborts Phase 2, deletes pending VolumeSnapshot CRs, triggers
+  block-commit to restore the VM, and marks the VMSnapshot as Failed with a
+  clear error indicating insufficient scratch space.
+
+- **Timeout:** Phase 2 is bounded by the VMSnapshot's existing
+  `FailureDeadline` (default: 5 minutes, configurable per-snapshot). If any
+  VolumeSnapshot has not received its `CreationTime` within this deadline,
+  the same abort flow is triggered.
+
+The scratch PVC is sized to accommodate the maximum possible writes within
```
Acedus: nit: I think this is a redundant statement given the former reference to the topic.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,698 @@
...
+the `FailureDeadline` (see [Scalability](#scalability)).
+
+### Phase 3: Block-commit
+
+After all VolumeSnapshots have `CreationTime` set, the snapshot controller
+calls a new asynchronous `CommitSnapshot` subresource on the VMI. Inside
+virt-launcher, for each disk:
+
+1. **Check state**: Is the disk on overlay? If already on base, skip (idempotent).
```
Acedus: Not sure I understand this sentence, does it mean _an_ overlay instead of _on_? If so, what about VMs with CBT where certain disks already have an overlay?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,698 @@
...
+  VolumeSnapshot has not received its `CreationTime` within this deadline,
+  the same abort flow is triggered.
+
+The scratch PVC is sized to accommodate the maximum possible writes within
+the `FailureDeadline` (see [Scalability](#scalability)).
+
+### Phase 3: Block-commit
+
+After all VolumeSnapshots have `CreationTime` set, the snapshot controller
+calls a new asynchronous `CommitSnapshot` subresource on the VMI. Inside
```
Acedus: nit: I presume you mean RPC and not subresource? I don't see a reason for it to be a new subresource.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,698 @@
...
+
+### Overlay state tracking
+
+A VMI condition `OverlaySnapshotActive` is set at Phase 1 and removed after
+commit (Phase 3). Only set during external snapshots, the existing flow is
+unaffected. The condition is checked by:
+
+- **Migration**: blocks `startMigration()` - the overlay references the
+  base image on the source node which would not exist on the target
+- **Backup**: blocks `BackupVirtualMachine()` - concurrent backup and
```
Acedus, line 354: We should probably do the same on the CBT side :)

Acedus (commented):
Really nice stuff @noamasu! Clearly a lot of thought went into this design. Left a few minor comments but I'm inclined to lgtm the design.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: worst case the monitor misses the threshold and QEMU hits ENOSPC, pausing the VM. (tested in POC) abort block-commit kicks in, merges all overlays back to base, VM resumes once disks are back on base and snapshot controller cleans the scratch PVC and the snapshot fails. no data loss, just no I/O during the pause window.

https://github.com/noamasu/enhancements/blob/b23c1901589af9549a533e402209357b2c5fcc50/veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md?plain=1#L219-L233
will update the wording there shortly

Acedus: Sounds good!

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,565 @@
...
+  in all environments
+- Does not solve the fundamental problem: the VM stays frozen while
+  waiting for external infrastructure
+
+
+## Scalability
+
+- **Scratch PVC sizing.** Users can set `spec.overlayScratchSize` on the
+  VMSnapshot to control the scratch PVC size directly. If not set, the
+  following default calculation is used:
```
mhenriks: I think it is okay to do a single pvc for alpha and get feedback for beta/ga

mhenriks:
I'm good with this for alpha
/lgtm

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: yes, its a bit unclear, fixed the wording. From what I saw, CBT overlays use the raw image as a data-file so they sit outside the backing chain, meaning block-commit stops at the CBT qcow2 and never touches it. Does that match your understanding?

Acedus (approved):
Thanks @noamasu!

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+
+1. Sets the `OverlaySnapshotActive` condition on the VMI.
+2. Creates a PVC (`{vmsnapshot-name}-snap-scratch`) sized to hold qcow2 overlay files
+   for all disks during the overlay window. The scratch PVC inherits the
+   storage class from the VM's first disk. The size can be set explicitly
+   via `spec.overlayScratchSize`
+   on the VMSnapshot. If not configured, a default size is used
+   (see [Scalability](#scalability) for details).
+3. Hotplugs the PVC as a UtilityVolume to the VMI via JSON patch on
+   `spec.utilityVolumes` (same mechanism as [VEP #90 Utility Volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md),
```
iholder101: broken link

```suggestion
   `spec.utilityVolumes` (same mechanism as [VEP #90 Utility Volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/90-utility-volumes/vep.md),
```

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+   same pattern as the incremental backup push-target PVC in [VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md)).
```
iholder101:
```suggestion
   same pattern as the incremental backup push-target PVC in [VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/25-incremental-backup/vep.md)).
```

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+### New flow
+
+The new flow is enabled by setting `spec.snapshotMethod: External` on the
+VMSnapshot. If not set, the default is `Direct` (current flow). See
+[API Examples](#api-examples) for all new fields.
+
+```
+Phase 0: Create scratch PVC, hotplug as UtilityVolume
+Phase 1: Freeze -> QEMU transaction (all disks, atomic, ~20ms) -> Unfreeze
+Phase 2: Create N VolumeSnapshot CRs (VM is unfrozen, no time pressure)
```
iholder101, line 158: With external snapshots the guest is unfreezed ~20ms after the QMP transaction, before the CSI VolumeSnapshot is durable. For VSS-integrated guests this means the app may get its "backup complete" signal and truncate logs before the snapshot is confirmed, giving weaker app-consistency than the classic flow.

I don't think it should be blocking for alpha, but I suggest documenting this as a known alpha limitation and revisiting for beta.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+takes.
+
+**Safety mechanisms:** While the VM is unfrozen during this phase, the overlay
+files on the scratch PVC continue to grow with every write. Two mechanisms
+prevent the scratch PVC from filling up:
+
+- **Overlay usage monitoring:** virt-launcher monitors the scratch PVC
+  in the background. If usage crosses a certain threshold (for example 70%),
+  it triggers block-commit to merge all overlays back to base. The snapshot
+  controller then marks the VMSnapshot as Failed.
```
iholder101: If the 70% emergency block-commit fires before the CSI VolumeSnapshot is durable, we abort with no real snapshot taken. That's fine as long as it fails loudly: does the emergency-abort path guarantee the VMSnapshot is marked Failed (never Succeeded)? If so, no concern. Worth stating explicitly in the doc so a restore never points at a snapshot that was never taken.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+   `spec.utilityVolumes` (same mechanism as [VEP #90 Utility Volumes](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/utility-volumes.md),
+   same pattern as the incremental backup push-target PVC in [VEP #25](https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md)).
+4. Adds a Kubernetes finalizer `snapshot.kubevirt.io/overlay-protection`
+   on the scratch PVC to prevent its deletion while overlays are active.
+5. Waits for the volume to reach `VolumeReady` or `HotplugVolumeMounted` status.
+
+### Phase 1: QEMU atomic snapshot (overlay creation)
+
+Once the scratch volume is mounted, the snapshot controller calls a new
+`ExternalSnapshot` RPC on the VMI. Inside virt-launcher:
```
iholder101: `ExternalSnapshot` doesn't define idempotency, unlike `CommitSnapshot`. On retry-after-timeout we could create a second overlay. Suggest mirroring `CommitSnapshot`: detect an existing overlay for this `VMSnapshot` and return success, specify retry semantics in the RPC contract.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+```
+
+Freeze duration is sub-second regardless of disk count.
+
+### Phase 0: Scratch volume setup
+
+Before taking the snapshot, the snapshot controller:
+
+1. Sets the `OverlaySnapshotActive` condition on the VMI.
+2. Creates a PVC (`{vmsnapshot-name}-snap-scratch`) sized to hold qcow2 overlay files
```
iholder101: Deterministic name `{vmsnapshot-name}-snap-scratch` collides on delete+recreate with a leftover PVC. WDYT about using a unique name?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+
+After ALL disks are committed and verified, a final check reads the full
+domain XML and confirms every disk source is back on the base image.
+
+The block-commit is idempotent and resumable: already committed disks are
+skipped, active block jobs from a previous attempt are resumed. After all
+disks are committed, the domain XML is verified to confirm every disk is
+back on the base image before proceeding to cleanup. If commit fails, the
+scratch PVC is preserved (finalizer prevents deletion) and the controller
+requeues for retry.
```
iholder101, line 272: Active block-commit may never reach READY if guest writes outpace the commit (active-layer starvation). There's no timeout or convergence guarantee here, and `OverlaySnapshotActive` stays set meanwhile, indefinitely blocking migration/backup/resize.

WDYT about: briefly re-freeze the guest to let it reach READY and pivot, then unfreeze?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+apiVersion: snapshot.kubevirt.io/v1beta1
+kind: VirtualMachineSnapshot
+metadata:
+  name: my-snapshot
+spec:
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: my-windows-vm
+  snapshotMethod: External        # optional, defaults to Direct (current flow)
```
iholder101: `snapshotMethod: External/Direct` names the QEMU mechanism, not what the user gets. WDYT about naming by the guarantee, e.g. `snapshotMode: Online` (VM stays responsive) vs `Standard`. The values should express the tradeoff, not the implementation.

I also think that `snapshotMethod` is not a great name either, but not sure I have a better idea :) naming is hard.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
```diff
@@ -0,0 +1,699 @@
...
+kind: VirtualMachineInstance
+metadata:
+  name: my-windows-vm
+status:
+  conditions:
+  - type: OverlaySnapshotActive
+    status: "True"
+    reason: SnapshotInProgress
+    message: "Snapshot my-snapshot has active overlays"
+    lastTransitionTime: "2026-05-28T12:00:00Z"
```
iholder101, line 502: `OverlaySnapshotActive` leaks the qcow2-overlay implementation detail into the API. WDYT about a more neutral, reusable name like type: `VolumeSnapshotInProgress` with a clear reason/message?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: thanks! fixing

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: The snapshot point is the QEMU transaction, not the CSI snapshot. After the transaction the base images are read-only, meaning the CSI snapshot captures exactly that frozen state regardless of when it happens. neew writes (including any post-thaw log truncation) go to the overlay, so the app-consistency guarantee holds at the transaction point. Want me to add a sentence in the doc explaining this, or do you think it's clear enough as is?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: yes, currently if emergency block-commit fires, VMSnapshot is marked Failed. i'll make it clearer.

btw it is possible to technically add logic that checks if the CSI already created VolumeSnapshot and tag the VMSnapshot as succeeded, but i would prefer to just fail it and let the user try another one.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: yes, we can use a unique name, i'll change it

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: "briefly" means for how long? i mean, long freezes are exactly what this VEP is trying to avoid eventually. when i tested block commit and pivot under super heavy I/O it took ~3min to converged and complete the block commit back to the base image. while convergence is not gurantied for now, I've opened an RFE https://redhat.atlassian.net/browse/RHEL-178640 (stated in the VEP) asking to expose copy-mode parameter for block-commit QMP as well, so new writes will be directed to both the overlay and the base, which will help with convergence

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: i thought of online vs standard orininally, but techincailly, both flows are of an Online VM. i'll try to come up with better namings and send some options here soon

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
iholder101: That's great.

However, what happens if it never converges - is there a timeout that fails the snapshot and clears `OverlaySnapshotActive` so lifecycle ops aren't blocked forever?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
iholder101: Agreed, the point-in-time is consistent. Minor edge is that VSS signals "backup complete" on unfreeze, so apps may truncate logs before CSI is durable. Small enough to just note as a known limitation and revisit for beta. A sentence in the doc would be great.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: i can add a 5min timeout (seems right according to my heavy IO tests) and if it still doesn't converge, we pause the VM until convergence instead of freeze, WDYT?

iholder101: Yup, sounds good to me!

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: i agree, ExternalSnapshot should also be idempotent.

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: btw https://github.com/kubevirt/enhancements/pull/322#discussion_r3398405822

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: i was thinking about different direction, like:
`freezeStrategy: Hold/Instant`
or
`snapshotMode: External/Direct` (just `snapshotMethod` to `snapshotMode`)

The reason I like snapshotMode better is because as stated in the non-goals section, using external snapshots will also let us do a full snapshot of the VM State including memory and all later on, so we could extend it to `snapshotMode: External/Direct/Full` for example. let me know if you have better alternatives to `External/Direct`, this is indeed very hard!

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: I agree, i just wonder about the existing VM status `vm.status.snapshotInProgress`. its not on the VMI, but sill can cause confusion? I'd be good to just apply your suggestion as well, WDYT?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
iholder101: Okay, let's start with that, just please add to revisit the API naming as a criteria for beta. Makes sense?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
iholder101: Yeah good point.. maybe we can use this new condition to eventually deprecate this status field, hence improve it?

veps/sig-storage/321-online-vmsnapshot-qemu-external-snapshots/vep.md
noamasu: great idea, thanks!

iholder101:
Looks good for Alpha :) 

Thanks a lot @noamasu! 

/approve

iholder101:
The relevant people who have LGTMed this are offline, and it's the last day before the freeze.
Taking the liberty of re-LGTMing on their behalf.

@Acedus @mhenriks let me know if you have objections.

/lgtm

pipo:
Sorry for necromancing this after it was merged. I only learned about this proposal recently.

I've got a couple observations from libvirt's point of view. Other than those the design makes sense from PoV how libvirt designed this feature to be used.

My points are:

### Regarding 'Design -> Phase 1':

  The libvirt API has also 'VIR_DOMAIN_SNAPSHOT_CREATE_QUIESCE' flag which
  uses the guest agent to freeze the filesystems automatically, which would
  theoretically decrease the time the filesystems are frozen slightly.
  
  The disadvantage of that flag is that it absorbs all error states from the
  guest agent interaction under the snapshot API, so if you need better
  control about the state of the VM, using the fsfreeze API explicitly 
  as proposed is a better idea.


### Regarding 'Design -> Phase 3, point 3'

  Querying XML and parsing is expensive. Libvirt emits an event when a commit
  job reaches the ready state. Kubevirt should listen for the event instead.
  
  Look for event id 'VIR_DOMAIN_EVENT_ID_BLOCK_JOB_2', which reports the job
  status.
  
### Regarding 'Crash recovery'

  During startup you could also just point the VM XML config to the overlay
  file and do the commit while the VM is started instead of doing it via
  qemu-img commit
  
  
### Regarding 'Alternatives'

  Use of the backup API doesn't allow taking snapshot of memory (it was
  described as a non-goal currently)
  
  
### Regarding 'Scalability'

  The block-commit throughput can be also limited by starting commit
  sequentially to avoid concurent reads/writes. Specifically on mechanical
  hard drives this may be beneficial.

noamasu:
Hi Peter, thanks a lot for going through this even though it's already merged. Its great to get a read on this from the libvirt side. This is new territory for kubevirt so making sure we're doing things the right way is worth the extra eyes!

> Regarding 'Design -> Phase 1':
>The libvirt API has also 'VIR_DOMAIN_SNAPSHOT_CREATE_QUIESCE' flag which
>uses the guest agent to freeze the filesystems automatically, which would
>theoretically decrease the time the filesystems are frozen slightly.
>
>The disadvantage of that flag is that it absorbs all error states from the
>guest agent interaction under the snapshot API, so if you need better
>control about the state of the VM, using the fsfreeze API explicitly
>as proposed is a better idea.

I saw VIR_DOMAIN_SNAPSHOT_CREATE_QUIESCE when working on constructing the command but eventually preferred explicit fsfreeze calls, for the same reasons you mentioned


> Regarding 'Design -> Phase 3, point 3'
Querying XML and parsing is expensive. Libvirt emits an event when a commit
job reaches the ready state. Kubevirt should listen for the event instead.
Look for event id 'VIR_DOMAIN_EVENT_ID_BLOCK_JOB_2', which reports the job
status.

Good point, thanks! I'll listen for `VIR_DOMAIN_EVENT_ID_BLOCK_JOB_2` instead.

>Regarding 'Crash recovery'
> During startup you could also just point the VM XML config to the overlay
> file and do the commit while the VM is started instead of doing it via
> qemu-img commit

the domain XML gets rebuilt from scratch when the VMI is recreated, so we'd need to point each disk at its overlay there. We already do something similar for cbt, and we can verify the mapping from the backing file in the overlay itself, so it looks reasonable. Thinking of keeping the offline commit for alpha and doing this for beta. WDYT?

> Regarding 'Alternatives'
Use of the backup API doesn't allow taking snapshot of memory (it was
described as a non-goal currently)

Right. full VM snapshots with memory would be an interesting valuable follow up to this effort for sure!


> Regarding 'Scalability'
The block-commit throughput can be also limited by starting commit
sequentially to avoid concurent reads/writes. Specifically on mechanical
hard drives this may be beneficial.

we already run them sequentially. Convergence is a worry for us though, since a guest writing faster than the commit copies means the job never reaches ready. I filed an RFE to expose copy-mode on block-commit like mirror has https://redhat.atlassian.net/browse/RHEL-178640 that would guarantee convergence. not sure if anyone from libvirt/qemu looked at it yet.

iholder101:
Thanks @pipo for these thoughts!

Just one minor comment regarding

> Thinking of keeping the offline commit for alpha and doing this for beta. 

@noamasu just note that major development is expected to land during alpha. Ideally, the alpha -> beta graduation needs to as minimal as possible. See https://github.com/kubevirt/enhancements/blob/main/docs/feature-lifecycle.md for more info.

noamasu:
@iholder101, thanks for clarifying. I'll prioritize this to be ready for alpha as i think the scope of this extra logic is probably not too large. I'll make sure to update if I come across any blocking issues.

pipo:
[...]

> > Regarding 'Crash recovery'
> > During startup you could also just point the VM XML config to the overlay
> > file and do the commit while the VM is started instead of doing it via
> > qemu-img commit
> 
> the domain XML gets rebuilt from scratch when the VMI is recreated, so we'd need to point each disk at its overlay there. We already do something similar for cbt, and we can verify the mapping from the backing file in the overlay itself, so it looks reasonable. Thinking of keeping the offline commit for alpha and doing this for beta. WDYT?

I'm not familiar with Kubevirt's design but IMO it would make sense to have live-commit only if you want to have the simpler design because you'll only have the one completion code.

Additionally the benefit is that the VM can start right away and the commit operation will proceed while the guest OS is already running.


> > Regarding 'Alternatives'
> > Use of the backup API doesn't allow taking snapshot of memory (it was
> > described as a non-goal currently)
> 
> Right. full VM snapshots with memory would be an interesting valuable follow up to this effort for sure!

I'm also planning on adding support for qemu's 'background-snapshot' feature which allows minimizing the downtime of the VM needed to take the memory snapshot (currently the VM has to be paused). That'd make the memory snapshot also more attractive for you as the VM will be paused only very briefly to take the disk snapshot in sync with the memory snapshot.


> > Regarding 'Scalability'
> > The block-commit throughput can be also limited by starting commit
> > sequentially to avoid concurent reads/writes. Specifically on mechanical
> > hard drives this may be beneficial.
> 
> we already run them sequentially. Convergence is a worry for us though, since a guest writing faster than the commit copies means the job never reaches ready. I filed an RFE to expose copy-mode on block-commit like mirror has https://redhat.atlassian.net/browse/RHEL-178640 that would guarantee convergence. not sure if anyone from libvirt/qemu looked at it yet.

Heh, before writing the comment I was checking which block jobs actually support the synchronous mode and didn't mention it only because commit doesn't have it.

Nevertheless I'll watch the issue you've linked and I'll expose the knob if qemu adds something.

Issue #323, opened 2026-06-01
VEP 323: Validation admission policy


iholder101:
**Primary contact (assignee)**:

/assign @iholder101

**Current Feature Stage**:

New

**Feature Gate**:

N/A (policy change, no feature gate)

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig control-plane
/sig storage

**Enhancement link**:

TBD

**Timeline**:

* Alpha release target (x.y): N/A (policy change, no kubevirt version dependency)
* Beta release target (x.y): N/A
* Stable release target (x.y): N/A

* [ ] VEP PRs:
	* [ ] https://github.com/kubevirt/enhancements/pull/324.
* [ ] Code PRs:
* [ ] Docs PRs:

**Additional context**:

This is a meta-VEP establishing a validation policy (kubebuilder markers vs VAPs vs webhooks).
See discussion at https://github.com/kubevirt/kubevirt/issues/16708.

vladikr:
I did a POC here: https://github.com/kubevirt/kubevirt/pull/16874

PR #324, opened 2026-06-01
Meta-VEP: Validation admission policy

iholder101:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/323.
**SIG label**:

/sig compute
/sig storage
/sig network

### What this PR does
Proposes a three-tier validation policy that uses the right mechanism for each class of validation,
with [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) (VAPs) as the primary mechanism for most validations.

### DIFF
--- a/veps/meta-VEPs/323-validation-admission-policy/vep.md
+++ b/veps/meta-VEPs/323-validation-admission-policy/vep.md
@@ -1,0 +1,283 @@
+# VEP #323: Validation Admission Policy
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: N/A (policy change, no kubevirt version dependency)
+- This VEP targets beta for version: N/A
+- This VEP targets GA for version: N/A
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- ~[ ] (R) Target version is explicitly mentioned and approved~
+- ~[ ] (R) Graduation criteria filled~
+
+## Overview
+
+This meta-VEP captures the outcome of the [validation policy discussion](https://github.com/kubevirt/kubevirt/issues/16708).
+
+KubeVirt currently relies on admission webhooks for nearly all validation logic.
+This VEP proposes a three-tier validation policy that uses the right mechanism for each class of validation,
+with [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) (VAPs)
+as the primary mechanism for most validations.
+
+## Motivation
+
+KubeVirt's validation is almost entirely implemented via admission webhooks.
+While webhooks are flexible, they come with significant drawbacks:
+
+- **Operational burden**: webhooks require a running, reachable service. If virt-api is unavailable,
+  all mutating and validating operations on KubeVirt resources are blocked.
+- **Complexity**: webhook code is imperative Go, which is harder to review, test, and reason about
+  than declarative validation rules.
+- **Scalability**: every admission request requires a network round-trip to the webhook service.
+- **Maintenance cost**: as KubeVirt grows, the webhook codebase grows with it, making it harder to maintain.
+
+Kubernetes introduced [ValidatingAdmissionPolicies](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/)
+(GA since Kubernetes 1.30) as a declarative, in-process alternative to admission webhooks.
+VAPs use [CEL](https://kubernetes.io/docs/reference/using-api/cel/) expressions evaluated directly
+by the API server - no network hop, no external service dependency.
+
+Additionally, kubebuilder validation markers (e.g., `+kubebuilder:validation:Minimum`, `+kubebuilder:validation:Enum`)
+offer OpenAPI schema-level validation for simple, invariant field constraints.
+
+This VEP establishes when to use each mechanism.
+
+## Goals
+
+- Establish a clear, documented policy for choosing between kubebuilder markers, VAPs, and webhooks.
+- Migrate existing webhook validations to VAPs iteratively over time.
+- Reduce KubeVirt's dependency on admission webhooks for validation.
+
+## Non Goals
+
+- Define a specific timeline or release target for migrating all existing validations.
+  Migration should be done iteratively with no pressure.
+- Discuss the specifics of individual validations (each migration can be handled in its own PR).
+- Replace mutating admission webhooks - this VEP only covers validation.
+
+## Definition of Users
+
+- **KubeVirt developers**: who implement and review validation logic.
+- **KubeVirt reviewers/approvers**: who review PRs that add or modify validation.
+- **Cluster administrators**: who manage KubeVirt deployments and may need to understand validation behavior.
+
+## User Stories
+
+- As a KubeVirt developer, I want a clear guideline for where to implement validation logic
+  so that I don't have to make ad-hoc decisions on each PR.
+- As a KubeVirt reviewer, I want a documented policy to point to when reviewing PRs
+  that add validation in the wrong place.
+- As a cluster administrator, I want validations to work even if the virt-api webhook service
+  is temporarily unavailable.
+
+## Repos
+
+- kubevirt/kubevirt
+
+## Design
+
+### Three-tier validation policy
+
+Validation in KubeVirt should use the following mechanisms, in order of preference:
+
+#### 1. Kubebuilder markers (OpenAPI schema validation)
+
+Use kubebuilder validation markers **only** for simple, single-field constraints
+that are certain to never change. Examples:
+
+- A string field that must not be empty
+- An integer field that must not be negative
+
+These constraints are baked into the CRD's OpenAPI schema. Once published,
+they are effectively permanent - changing them is a breaking API change.
+Therefore, use them conservatively and only for true invariants.
+
+**Do not** use kubebuilder markers for constraints that may evolve over time. Examples of what to avoid:
+
+- `+kubebuilder:validation:Minimum` / `+kubebuilder:validation:Maximum` - acceptable ranges tend to change as
+  requirements evolve (e.g., a port field might later need to allow 0 for auto-assignment).
+- `+kubebuilder:validation:Enum` - enum values almost always grow over time. Adding a new value
+  to a kubebuilder enum is a breaking schema change.
+
+Use VAPs for these validations instead.
+
+**Note on existing markers**: KubeVirt already has kubebuilder validation markers that do not meet
+the above criteria. Removing them would be a breaking API change, so they will remain as-is.
+This policy applies to new markers going forward.
+
+#### 2. ValidatingAdmissionPolicies (VAPs)
+
+Use VAPs for **most** validations. This includes:
+
+- Multi-field validations (e.g., "if field A is set, field B must also be set")
+- Complex single-field validations that may evolve over time
+- Cross-field consistency checks
+- Validations that depend on the object's current state (`oldSelf` vs `self`)
+- Immutability checks
+- Any validation that is not a true invariant
+
+VAPs are declarative, evaluated in-process by the API server, and can be updated
+without code changes. They are the right default for nearly all validation logic.
+
+#### 3. Admission webhooks
+
+Use webhooks **only** as a last resort, when VAPs cannot express the validation. Examples:
+
+- Validations that require API calls to other resources (e.g., checking if a feature gate is enabled)
+- Validations that require complex procedural logic not expressible in CEL
+- Validations with dependencies on external state
+
+When a webhook is necessary, the PR should document why a VAP is insufficient.
+
+### VAP lifecycle management
+
+VAPs are managed by virt-operator and continuously reconciled, just like other
+operator-managed resources (RBAC, Deployments, etc.).
+
+This means:
+- VAPs are created and updated by virt-operator as part of the normal reconciliation loop.
+- Manual modifications to operator-managed VAPs are reverted on the next reconciliation cycle.
+- Upgrades naturally apply VAP changes as part of the operator's desired-state reconciliation.
+
+If an administrator needs to customize validation behavior, the proper mechanism is
+a configuration knob in the KubeVirt CR, not direct VAP modification.
+The operator then uses these configuration values when generating the VAPs.
+This ensures admin intent is captured declaratively and survives upgrades.
+
+The specifics of which configuration knobs to expose are left to individual feature PRs.
+
+### Migration strategy
+
+Existing webhook validations should be migrated to VAPs iteratively.
+There is no fixed timeline or release target for completing the migration.
+
+When migrating a validation:
+1. Implement the VAP equivalent.
+2. Add functional tests for the VAP-based validation.
+3. Remove the corresponding webhook validation logic.
+
+Each migration can be handled in its own PR. There is no need to migrate all validations at once.
+
+## API Examples
+
+No API changes are proposed. VAPs are standard Kubernetes resources deployed by virt-operator.
+
+## Alternatives
+
+### VAP reconciliation strategies
+
+This VEP proposes continuous reconciliation by virt-operator (option 1 below).
+Two alternatives were considered:
+
+#### Option 1: Continuous reconciliation (chosen)
+
+virt-operator creates and continuously reconciles all VAPs, reverting any manual changes.
+Admin customization goes through the KubeVirt CR.
+
+**Pros:**
+- Upgrades are clean - operator ensures VAPs match the desired state for the current version.
+- Consistent with how KubeVirt manages other resources (RBAC, Deployments).
+- No drift between expected and actual validation state.
+- Consistent with the ecosystem - Gatekeeper, Kyverno, and OpenShift operators all use this pattern.
+
+**Cons:**
+- Administrators cannot directly modify VAPs for quick, ad-hoc changes.
+- Every customization requires a KubeVirt CR knob, which means code changes and a release cycle.
+
+#### Option 2: No reconciliation
+
+VAPs are applied during install and upgrades but are not continuously reconciled.
+Administrators are free to modify them between upgrades.
+
+**Pros:**
+- Maximum admin flexibility - any VAP can be modified at any time.
+- Simpler operator logic (no reconciliation loop for VAPs).
+
+**Cons:**
+- Upgrades become risky. If the operator needs to update a VAP that the admin has modified,
+  there is no clean way to merge the changes. The operator must either overwrite (losing admin changes)
+  or skip (leaving outdated validation logic).
+- Interrupted upgrades may leave VAPs in an inconsistent state with no self-healing mechanism.
+- Validation drift is invisible - there is no signal that the running VAPs differ from
+  what the current KubeVirt version expects.
+- No project in the Kubernetes ecosystem uses this pattern for operator-managed VAPs.
+
+#### Option 3: Opt-out annotation
+
+virt-operator reconciles by default, but skips VAPs annotated with
+`kubevirt.io/managed: "false"`. Administrators can opt out specific VAPs from reconciliation.
+
+**Pros:**
+- Reconciliation by default for the common case.
+- Administrators can customize specific VAPs when needed.
+
+**Cons:**
+- Opted-out VAPs face the same upgrade problem as option 2 - the administrator
+  must manually review and merge upstream changes on upgrade, or risk running
+  outdated validation logic.
+- Provides a false sense of control - most administrators will not track upstream
+  VAP changes across releases.
+- Adds complexity to the operator (annotation checking, partial reconciliation).
+- Creates a support burden - debugging validation issues requires checking which
+  VAPs are managed and which are not.
+
+### Kubebuilder markers for client-side dry-run
+
+An alternative to the conservative use of kubebuilder markers proposed above is to
+use them more broadly, including for validations that may change over time,
+in order to enable client-side dry-run (`kubectl apply --dry-run=client`).
+
+This was rejected because kubebuilder markers are baked into the CRD's OpenAPI schema.
+Once published, changing them is a breaking API change - clients that depend on the schema
+will break. Server-side dry-run (`kubectl apply --dry-run=server`) activates all validation
+including VAPs, and should be used instead.
+
+Kubebuilder markers should therefore be reserved for truly immutable field constraints
+where we are confident the validation will never change.
+
+### Webhooks only (status quo)
+
+Continuing to use admission webhooks for all validation logic.
+
+**Pros:**
+- No migration effort.
+- Full flexibility (arbitrary Go code).
+
+**Cons:**
+- All the drawbacks listed in [Motivation](#motivation) persist and grow with the codebase.
+- KubeVirt does not benefit from the operational improvements that VAPs provide.
+
+## Scalability
+
+VAPs are evaluated in-process by the API server. This eliminates the network round-trip
+required by webhooks, improving validation latency and reducing the load on virt-api.
+
+No scalability issues are expected.
+
+## Update/Rollback Compatibility
+
+VAPs are continuously reconciled by virt-operator. On upgrade, the operator applies
+the new version's VAPs as part of its normal reconciliation loop. On rollback,
+the previous version's operator restores its own VAPs.
+
+No update/rollback compatibility issues are expected.
+
+## Functional Testing Approach
+
+VAP-based validations should be tested via functional tests that exercise the API server's
+admission path, verifying that invalid requests are rejected with the expected error messages.
+
+Existing webhook validation tests can be reused with minimal changes, since the test
+assertions (create/update a resource and expect rejection) remain the same regardless
+of the underlying mechanism.
+
+Unit testing of CEL expressions is available using a vendored CEL evaluation library
+introduced in [#17790](https://github.com/kubevirt/kubevirt/pull/17790).
+
+## Implementation History
+


veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+- Multi-field validations (e.g., "if field A is set, field B must also be set")
+- Complex single-field validations that may evolve over time
+- Cross-field consistency checks
+- Validations that depend on the object's current state (`oldSelf` vs `self`)
+- Immutability checks
+- Any validation that is not a true invariant
+
+VAPs are declarative, evaluated in-process by the API server, and can be updated
+without code changes. They are the right default for nearly all validation logic.
```
mhenriks: This says VAPs "can be updated without code changes," but the lifecycle section (and Option 1's own Cons, lines 189–190) says all customization goes through a KubeVirt CR knob, which is code plus a release. Those point in opposite directions. I think "no code changes" is true of a VAP as a raw Kubernetes object, but not of one that's operator-reconciled the way this VEP proposes. Reconciliation reverts manual edits, so the only durable change is a code change. Could the benefit be restated to match the management model? As written it reads as a flexibility the design actually gives up.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+by the API server - no network hop, no external service dependency.
+
+Additionally, kubebuilder validation markers (e.g., `+kubebuilder:validation:Minimum`, `+kubebuilder:validation:Enum`)
+offer OpenAPI schema-level validation for simple, invariant field constraints.
+
+This VEP establishes when to use each mechanism.
+
+## Goals
+
+- Establish a clear, documented policy for choosing between kubebuilder markers, VAPs, and webhooks.
```
EdDev, line 52: Can we include in this VEP guidance or rules on when to have the validation only in the controller reconciliation only?
I believe it is a good opportunity to question the imperative approach taken on some of the validations.

veps/meta-VEPs/323-validation-admission-policy/vep.md
EdDev, line 1: I would consider classifying it under "control-plan" or "compute" SIGs, and not limit it to just the policy.
We can phase it and build a policy, followed with the steps to implement.

In a simplified form, moving from a policy to implementation should be simple.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+- Migrate existing webhook validations to VAPs iteratively over time.
+- Reduce KubeVirt's dependency on admission webhooks for validation.
+
+## Non Goals
+
+- Define a specific timeline or release target for migrating all existing validations.
+  Migration should be done iteratively with no pressure.
```
EdDev, line 59: If we define phases and steps, we can define it well enough to see actual results pretty fast. After a policy is agreed and established, it should be pretty straight forward to convert most validations to declarative ones.

The review commitment would be the challenging part, but I think it can be reduced by providing good unit testing.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+- **KubeVirt developers**: who implement and review validation logic.
+- **KubeVirt reviewers/approvers**: who review PRs that add or modify validation.
+- **Cluster administrators**: who manage KubeVirt deployments and may need to understand validation behavior.
+
+## User Stories
+
+- As a KubeVirt developer, I want a clear guideline for where to implement validation logic
+  so that I don't have to make ad-hoc decisions on each PR.
+- As a KubeVirt reviewer, I want a documented policy to point to when reviewing PRs
+  that add validation in the wrong place.
```
EdDev, line 74:
```suggestion
- As a KubeVirt reviewer, I want a documented policy that I can use to assure validations land in the most appropiate location.
```

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+## Design
+
+### Three-tier validation policy
+
+Validation in KubeVirt should use the following mechanisms, in order of preference:
+
+#### 1. Kubebuilder markers (OpenAPI schema validation)
+
+Use kubebuilder validation markers **only** for simple, single-field constraints
+that are certain to never change. Examples:
```
EdDev, line 91: I suggest to list only things we know is a fit. Negative integers is a great example.

Do you know if there are actual Kubernetes guidelines in this area?

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+- `+kubebuilder:validation:Minimum` / `+kubebuilder:validation:Maximum` - acceptable ranges tend to change as
+  requirements evolve (e.g., a port field might later need to allow 0 for auto-assignment).
+- `+kubebuilder:validation:Enum` - enum values almost always grow over time. Adding a new value
+  to a kubebuilder enum is a breaking schema change.
+
+Use VAPs for these validations instead.
+
+**Note on existing markers**: KubeVirt already has kubebuilder validation markers that do not meet
+the above criteria. Removing them would be a breaking API change, so they will remain as-is.
```
EdDev, line 110: I am not sure if removing them will be a breaking change.
If the validation will move to a different mechanism, I think we could remove them after a few versions.

Maybe we need to think and plan it a bit.
On the other hand, maybe it is not worth the effort.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### 3. Admission webhooks
+
+Use webhooks **only** as a last resort, when VAPs cannot express the validation. Examples:
+
+- Validations that require API calls to other resources (e.g., checking if a feature gate is enabled)
+- Validations that require complex procedural logic not expressible in CEL
+- Validations with dependencies on external state
+
+When a webhook is necessary, the PR should document why a VAP is insufficient.
```
EdDev, line 135:
```suggestion
Consider admission webhooks as an exception. The VEP/implementation should reason why the other alternatives methods are insufficient.
```

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+- Any validation that is not a true invariant
+
+VAPs are declarative, evaluated in-process by the API server, and can be updated
+without code changes. They are the right default for nearly all validation logic.
+
+#### 3. Admission webhooks
+
+Use webhooks **only** as a last resort, when VAPs cannot express the validation. Examples:
+
+- Validations that require API calls to other resources (e.g., checking if a feature gate is enabled)
```
EdDev, line 131: > checking if a feature gate is enabled

I have an idea on how we can drop these as well.
But we can take this to a follow up in order not to overload this VEP.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+operator-managed resources (RBAC, Deployments, etc.).
+
+This means:
+- VAPs are created and updated by virt-operator as part of the normal reconciliation loop.
+- Manual modifications to operator-managed VAPs are reverted on the next reconciliation cycle.
+- Upgrades naturally apply VAP changes as part of the operator's desired-state reconciliation.
+
+If an administrator needs to customize validation behavior, the proper mechanism is
+a configuration knob in the KubeVirt CR, not direct VAP modification.
+The operator then uses these configuration values when generating the VAPs.
```
EdDev, line 149: Is this an extension to the current code-base implementation?
If it just preserves what already exists, a real example would be nice to understand the source of this.
If it is a new extended options, I would recomend it for a follow up.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+This ensures admin intent is captured declaratively and survives upgrades.
+
+The specifics of which configuration knobs to expose are left to individual feature PRs.
+
+### Migration strategy
+
+Existing webhook validations should be migrated to VAPs iteratively.
+There is no fixed timeline or release target for completing the migration.
```
EdDev, line 157: In practice, this hints that it may never finish.
I would prefer a plan and targeting the implementation of each per version.

I find the main challenge in curating the policy and less in the implenentation.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+When migrating a validation:
```
EdDev, line 159:
```suggestion
When migrating a validation from webhook to policy based:
```

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+1. Implement the VAP equivalent.
+2. Add functional tests for the VAP-based validation.
```
EdDev, line 161:
```suggestion
1. Add unit tests for the existing validation in the form of a manifest (yaml) input.
2. Implement the VAP equivalent.
3. Assure the same input manifest passes the VAP-based validation.
```

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+3. Remove the corresponding webhook validation logic.
```
EdDev, line 162: - Is this happening in the same version bump?
- There is a need to assure the VEP is deployed before webhooks are removed. Can this be done?

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+#### Option 1: Continuous reconciliation (chosen)
+
+virt-operator creates and continuously reconciles all VAPs, reverting any manual changes.
+Admin customization goes through the KubeVirt CR.
+
+**Pros:**
+- Upgrades are clean - operator ensures VAPs match the desired state for the current version.
+- Consistent with how KubeVirt manages other resources (RBAC, Deployments).
+- No drift between expected and actual validation state.
+- Consistent with the ecosystem - Gatekeeper, Kyverno, and OpenShift operators all use this pattern.
```
EdDev, line 186: Is the vendor specific operator necesery here?

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+VAPs are evaluated in-process by the API server. This eliminates the network round-trip
+required by webhooks, improving validation latency and reducing the load on virt-api.
+
+No scalability issues are expected.
+
+## Update/Rollback Compatibility
+
+VAPs are continuously reconciled by virt-operator. On upgrade, the operator applies
+the new version's VAPs as part of its normal reconciliation loop. On rollback,
+the previous version's operator restores its own VAPs.
```
EdDev, line 266: - In practice, there is no support for rollback.
- However, a scenario in which we will be interested in dropping a VAP (or replacing it with another) raises the question on how and when old ones are removed. It will fit this section IMO.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+No update/rollback compatibility issues are expected.
+
+## Functional Testing Approach
```
EdDev, line 270: Why not unit tests?
We do not really need to check that VAPs work, we can assume they are.
If we need to test anything, is the CEL or other content they have been programmed to apply.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+VAP-based validations should be tested via functional tests that exercise the API server's
+admission path, verifying that invalid requests are rejected with the expected error messages.
+
+Existing webhook validation tests can be reused with minimal changes, since the test
+assertions (create/update a resource and expect rejection) remain the same regardless
+of the underlying mechanism.
```
EdDev, line 277: Such tests are too expensive and probably do not worth having.
I suggest we focus our efforts on unit tests unless something critical in the path is missed.

We could create some integration tests, which are somewhere in between the unit and functional ones. But I do not think it is worth it.

EdDev (changes requested):
Thank you!

I like this direction very much.
Going this path will improve significantly projects, especially Kubevirt which is IMO over-using webhooks. 🙏

veps/meta-VEPs/323-validation-admission-policy/vep.md
mhenriks: On the guidelines question: the [API change doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) ("On compatibility") covers this, and it's directional. Tightening (narrowing a range, dropping an enum value) is the breaking case; adding an enum value needs care for client/component skew, but that's true wherever the rule lives — so I don't think it justifies ruling out Enum wholesale.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+Use VAPs for **most** validations. This includes:
+
+- Multi-field validations (e.g., "if field A is set, field B must also be set")
+- Complex single-field validations that may evolve over time
+- Cross-field consistency checks
+- Validations that depend on the object's current state (`oldSelf` vs `self`)
+- Immutability checks
+- Any validation that is not a true invariant
+
```
mhenriks, line 123: This is the piece I'd most want pinned down. If you classify the existing admitters by what each check actually needs, it comes out roughly: ~55% self-contained (object only — single-field, cross-field, immutability), ~25% gated on cluster config/feature flags, ~20% genuinely procedural or needing live state. (Rough classification, but the shape is clear)

Tier 2 explicitly lists cross-field, multi-field, and immutability checks as VAP cases ("the right default for nearly all validation logic"), and Tier 1 is restricted to single-field invariants — so the self-contained majority (most of which is cross-field, not single-field) lands in VAPs by definition. On the VMI alone that's 100+ validation expressions becoming VAPs, and the doc doesn't say how they're organized into policy objects. @vladikr raised exactly this in [#16708](https://github.com/kubevirt/kubevirt/issues/16708) ("a large VAP with many validations will become cluttered and hard to maintain," and "maybe per admitter?" — unresolved). One big policy vs. many small ones trade maintainability against a pile of policies and bindings for virt-operator to reconcile.

My actual suggestion: for self-contained checks, default to the schema (markers / CRD CEL) rather than VAPs — which sidesteps the decomposition question for the bulk of them, since the rule lives with the field. Reserve VAPs for the config/feature-gate cases (the ~25%), where a param-backed policy genuinely fits.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+they are effectively permanent - changing them is a breaking API change.
+Therefore, use them conservatively and only for true invariants.
+
+**Do not** use kubebuilder markers for constraints that may evolve over time. Examples of what to avoid:
+
+- `+kubebuilder:validation:Minimum` / `+kubebuilder:validation:Maximum` - acceptable ranges tend to change as
+  requirements evolve (e.g., a port field might later need to allow 0 for auto-assignment).
+- `+kubebuilder:validation:Enum` - enum values almost always grow over time. Adding a new value
+  to a kubebuilder enum is a breaking schema change.
+
```
mhenriks, line 106: I'd not discourage structural markers (Enum/Minimum/Maximum) here — we already ship plenty (Enum=kvm;hyperv-direct, Enum=VersionTLS10;..., Minimum, Pattern), so a blanket "use conservatively" sits awkwardly with current practice, and these self-contained constraints are exactly what the schema is for.

On the breaking-change rationale: the [API change guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) ("On compatibility") are directional. Tightening (narrowing a range, dropping an enum value) makes valid values invalid and is the breaking case — and it's breaking wherever the rule lives, including a VAP, so evolvability isn't an argument for VAP over a marker. Adding an enum value does need care for client/component skew, so "breaking" isn't unreasonable — but "enums almost always grow" is too absolute (many encode a fixed set). And moving an enum to a VAP doesn't help: the skew risk is that an older component doesn't understand a new value, which is true wherever the validation lives. The VAP just hides the value from the schema — losing discoverability without reducing the risk. With ratcheting (default since 1.30, caveats noted) the simple cases are reasonably safe to add even to existing fields.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+in order to enable client-side dry-run (`kubectl apply --dry-run=client`).
+
+This was rejected because kubebuilder markers are baked into the CRD's OpenAPI schema.
+Once published, changing them is a breaking API change - clients that depend on the schema
+will break. Server-side dry-run (`kubectl apply --dry-run=server`) activates all validation
+including VAPs, and should be used instead.
+
+Kubebuilder markers should therefore be reserved for truly immutable field constraints
+where we are confident the validation will never change.
+
```
mhenriks, line 242: I'd reframe this around discoverability rather than dry-run. The durable benefit of schema validation isn't client-side dry-run — it's that the constraint is visible: kubectl explain, generated clients, and offline validators all read the schema, and a VAP is invisible to all of them. That affects how easily someone understands a type's constraints without hunting through separate policy objects (see @xpivarc's comment [here](https://github.com/kubevirt/kubevirt/issues/16708#issuecomment-3839924590) about "dispersing where one needs to look to understand API"). Worth weighing whenever a rule could live in either place.

mhenriks (commented):
Thanks for writing this up, @iholder101. I'm aligned on the direction — less webhook dependence, more declarative validation. One theme and a few specifics, left inline:

The theme: most of our validation is self-contained (needs only the object being validated), and I think its natural default home is the schema (markers + CRD CEL), not a VAP. The VEP leans the other way. A few specifics:

- Structural markers (Enum/Minimum/Maximum) shouldn't be discouraged — we already ship them, and the breaking-change rationale is more directional than the doc states.
- Defaulting self-contained checks to VAPs creates a decomposition problem the VEP doesn't address — how ~100+ checks carve into policy objects.
- Discoverability — schema constraints show in kubectl explain/tooling; VAPs don't.

To be clear I'm not arguing against VAPs — there's a real ~quarter of our validations (feature-gate guards) that fit them well. It's specifically the self-contained majority I'd default to the schema. None of this blocks the direction; the reconciliation-options analysis is solid and the CEL test harness ([#16721](https://github.com/kubevirt/kubevirt/issues/16721)) addresses the testing concern.

veps/meta-VEPs/323-validation-admission-policy/vep.md
Acedus: +1 on a specific list, there aren't that many markers, too.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+#### 1. Kubebuilder markers (OpenAPI schema validation)
+
+Use kubebuilder validation markers **only** for simple, single-field constraints
+that are certain to never change. Examples:
+
+- A string field that must not be empty
+- An integer field that must not be negative
+
+These constraints are baked into the CRD's OpenAPI schema. Once published,
+they are effectively permanent - changing them is a breaking API change.
```
Acedus, line 97: Permanent implies it can never change, which is not entirely true...

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+
+VAPs are managed by virt-operator and continuously reconciled, just like other
+operator-managed resources (RBAC, Deployments, etc.).
+
+This means:
+- VAPs are created and updated by virt-operator as part of the normal reconciliation loop.
+- Manual modifications to operator-managed VAPs are reverted on the next reconciliation cycle.
+- Upgrades naturally apply VAP changes as part of the operator's desired-state reconciliation.
+
+If an administrator needs to customize validation behavior, the proper mechanism is
```
Acedus, line 147: I don't think we should allow expression modification in any way. Enforcement controls are fine though (maybe), e.g., controlling `Deny/Warn/Audit`.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+Admin customization goes through the KubeVirt CR.
+
+**Pros:**
+- Upgrades are clean - operator ensures VAPs match the desired state for the current version.
+- Consistent with how KubeVirt manages other resources (RBAC, Deployments).
+- No drift between expected and actual validation state.
+- Consistent with the ecosystem - Gatekeeper, Kyverno, and OpenShift operators all use this pattern.
+
+**Cons:**
+- Administrators cannot directly modify VAPs for quick, ad-hoc changes.
```
Acedus, line 189: As stated in a previous comment, I don't necessarily see this as a con. If a user is trying to work around a validation, it hints at a gap in our supported configuration or validation logic. Also I'm not aware of escape hatches for VAPs being commonplace in the wild (AFAIK, it's an install-time boolean decision of whether to deploy them or not in most cases).

Acedus (commented):
@EdDev and @mhenriks have already done an awesome job reviewing this iteration of the VEP, only have comments on admin interaction with KubeVirt's admission control.

veps/meta-VEPs/323-validation-admission-policy/vep.md
michalskrivanek: combined with the current dislike for any knobs this makes it a major issue. There needs to be a way how to at least control `Deny/Warn/Audit` indeed, otherwise this is just blocking admins completely.
Not sure why Option 3 below was dismissed, that would work too.

EdDev:
> Thanks for writing this up, @iholder101. I'm aligned on the direction — less webhook dependence, more declarative validation. One theme and a few specifics, left inline:
> 
> The theme: most of our validation is self-contained (needs only the object being validated), and I think its natural default home is the schema (markers + CRD CEL), not a VAP. The VEP leans the other way. A few specifics:
> 
> * Structural markers (Enum/Minimum/Maximum) shouldn't be discouraged — we already ship them, and the breaking-change rationale is more directional than the doc states.
> * Defaulting self-contained checks to VAPs creates a decomposition problem the VEP doesn't address — how ~100+ checks carve into policy objects.
> * Discoverability — schema constraints show in kubectl explain/tooling; VAPs don't.
> 
> To be clear I'm not arguing against VAPs — there's a real ~quarter of our validations (feature-gate guards) that fit them well. It's specifically the self-contained majority I'd default to the schema. None of this blocks the direction; the reconciliation-options analysis is solid and the CEL test harness ([#16721](https://github.com/kubevirt/kubevirt/issues/16721)) addresses the testing concern.

@mhenriks , the discussion on this matter is raised frequently.
I tried to outline the arguments and analyze why it is not an option.
I hope @iholder101 will add it to this PR: https://github.com/iholder101/kubevirt-enhancements/pull/1

Feel free to open the discussion on the above PR.

veps/meta-VEPs/323-validation-admission-policy/vep.md
```diff
@@ -0,0 +1,283 @@
...
+- `+kubebuilder:validation:Enum` - enum values almost always grow over time. Adding a new value
+  to a kubebuilder enum is a breaking schema change.
+
+Use VAPs for these validations instead.
+
+**Note on existing markers**: KubeVirt already has kubebuilder validation markers that do not meet
+the above criteria. Removing them would be a breaking API change, so they will remain as-is.
+This policy applies to new markers going forward.
+
+#### 2. ValidatingAdmissionPolicies (VAPs)
```
EdDev, line 113: I have updated https://github.com/iholder101/kubevirt-enhancements/pull/1 to include a section about the option to embed CEL (usually simple one) into the CRD as an alternative to a VAP.

In theory, this means we have another mechanism here which is server-side-only and combines the advantages of both worlds and drops most cons.

veps/meta-VEPs/323-validation-admission-policy/vep.md
EdDev: Due to their effect on the clients, in practice a change will have a chance to break clients. See https://github.com/iholder101/kubevirt-enhancements/pull/1 for more information.

veps/meta-VEPs/323-validation-admission-policy/vep.md
EdDev: The discussion here was mainly in comparison to the validation we perform today through CRD markers and admission webhooks.
These are usually not configurable, however in some cases they do depend on external inputs (e.g. the Kubevirt CR).
I am guessing that @iholder101 attempted to solve these cases here.

The extensibility option is touching a totally different area.
I am unsure if it fits in this VEP, as this proposal is not reducing the existing options (i.e. an admin could not change the validation on-the-fly).

We can treat it as a follow up IMO.

mhenriks:
> @mhenriks , the discussion on this matter is raised frequently. I tried to outline the arguments and analyze why it is not an option. I hope @iholder101 will add it to this PR: [iholder101#1](https://github.com/iholder101/kubevirt-enhancements/pull/1)
> 
> Feel free to open the discussion on the above PR.

Thanks @EdDev — this is a solid writeup and I don't have much to dispute in it; we're largely agreeing, and +1 to folding it in. The CEL section in particular makes the case for CEL _in the schema_ (`x-kubernetes-validations`) — server-executed, discoverable via `kubectl explain`, replacing webhooks — which is exactly what I'm proposing for self-contained checks. And on enums I agree with the volatile-field argument (the `PodPhase` example); my point was only the narrower one that we shouldn't discourage `Enum`/`Minimum`/`Maximum` _wholesale_, since some fields encode genuinely fixed sets where a marker is both correct and discoverable.

One thing your skew argument actually reinforces: if rigid enums are brittle because old clients/components can't handle a new value, a VAP doesn't fix that — the value is still unrecognized downstream, it's just hidden from the schema. So the resolution your doc points to is CEL in the schema, not a VAP — which is exactly the schema-over-VAP default I'm arguing for on the VEP.

I'd just note my two open points on the VEP sit a layer below the philosophy this covers, so I think they still belong on the VEP rather than the philosophy PR:

1. **Schema/CEL vs VAP as the default for self-contained checks.** This doc argues for declarative validation in the schema, which to me points toward defaulting these to CRD CEL rather than routing them to VAPs — that's the core of my comment on the VEP.
2. **How the checks that do become VAPs decompose into policy objects** — one large policy vs. many small ones, and the VEP doesn't specify which. Still unresolved.

Neither is a declarative-vs-rigid question, so I don't think the writeup closes them — it's good groundwork that makes them sharper. Happy to discuss the philosophy itself on your PR, but I think these two need to resolve on the VEP.

veps/meta-VEPs/323-validation-admission-policy/vep.md
mhenriks: +1
This is the piece I think is missing from the tiers as written. CRD CEL (`x-kubernetes-validations`) is the natural home for most of what Tier 2 lists: cross-field, immutability/transition (`oldSelf`), and complex single-field rules are all expressible in the schema, not just in a VAP. It also largely dissolves the decomposition concern I raised below — if self-contained checks live with the field as CRD CEL, there's no "how do 100+ checks carve into VAP objects" problem for the bulk of them.

So I'd treat this as more than an alternative to a VAP — arguably the _default_ for self-contained validation, with VAP as the step up when you need cluster config. Worth pulling into the VEP's tiering directly rather than leaving it in iholder101#1, since it changes where most validations land.

veps/meta-VEPs/323-validation-admission-policy/vep.md
mhenriks: Yeah VAPs should be able to get the KubeVirt CR via `paramRef`

PR #325, opened 2026-06-01
docs: Add guidelines for when a VEP is required

iholder101:
### What this PR does
Adds guidelines for when a VEP is required

Related to https://github.com/kubevirt/community/issues/285.

### DIFF
--- a/docs/when-is-a-vep-required.md
+++ b/docs/when-is-a-vep-required.md
@@ -1,0 +1,59 @@
+# When Is a VEP Required?
+
+## Summary
+
+Not every change to KubeVirt requires a VEP (Virtualization Enhancement
+Proposal). This document provides guidelines for contributors and
+reviewers to determine when a VEP is needed.
+
+## A VEP Is Required When a Change
+
+1. **Adds or modifies a public API** - new CRD fields, new CRDs, adding
+   new possible field values (e.g. enums), or changes to existing API
+   semantics or defaults.
+2. **Introduces new user-facing behavior that requires a feature gate** -
+   new VM/VMI behaviors, new configuration options, new status
+   conditions, or anything a user could notice without reading the code.
+3. **Adds new behavior under an existing feature gate** - this requires a
+   new VEP or an update to the existing one.
+4. **Changes component architecture** - adding a new sidecar, splitting a
+   controller, introducing a new daemon, or adding new inter-component
+   communication (e.g. a new gRPC service).
+5. **Requires coordination across multiple PRs or SIGs** due to
+   complexity and scope.
+
+## A VEP Is NOT Required For
+
+1. Bug fixes, unless they are extraordinarily complex.
+2. Test improvements or new test coverage.
+3. Documentation changes.
+4. Dependency bumps.
+5. Build, CI, or infrastructure changes.
+6. Trivial code cleanups (typos, dead code removal).
+
+## Gray Area - Demands Judgment
+
+1. **Large internal refactors** - no API change, but high blast radius
+   and cross-team coordination needs.
+2. **New internal mechanisms** (e.g. a new webhook, a new internal
+   controller) that don't surface in the API but shape future design.
+3. **Changes to operator/deployment topology** that don't affect the user
+   API but affect how KubeVirt is installed or upgraded.
+
+## Rationale
+
+The cost of creating a VEP is deliberately low - a short document with
+motivation, goals, and a high-level design. When in doubt about whether
+a change warrants a VEP, prefer writing one. This applies even for
+features that pre-date the VEP process or for changes that seem small
+in scope. It is better to invest a small upfront effort in a VEP and
+get early feedback than to implement a change and be surprised by
+objections during review. A lightweight VEP ensures the change is
+visible to the community, reviewed by the right SIGs, and tracked
+through the feature lifecycle.
+
+## Note
+
+These are guidelines, not rigid rules. Approvers and maintainers may
+require a VEP at their discretion for any change they consider
+significant enough to warrant one.


docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
...
+reviewers to determine when a VEP is needed.
+
+## A VEP Is Required When a Change
+
+1. **Adds or modifies a public API** - new CRD fields, new CRDs, adding
+   new possible field values (e.g. enums), or changes to existing API
+   semantics or defaults.
+2. **Introduces new user-facing behavior that requires a feature gate** -
+   new VM/VMI behaviors, new configuration options, new status
+   conditions, or anything a user could notice without reading the code.
```
michalskrivanek, line 16: for 1+2, I would think a docs/ file would actualy be better in many "smaller" cases.

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
...
+   API but affect how KubeVirt is installed or upgraded.
+
+## Rationale
+
+The cost of creating a VEP is deliberately low - a short document with
+motivation, goals, and a high-level design. When in doubt about whether
+a change warrants a VEP, prefer writing one. This applies even for
+features that pre-date the VEP process or for changes that seem small
+in scope. It is better to invest a small upfront effort in a VEP and
+get early feedback than to implement a change and be surprised by
```
michalskrivanek, line 50: It all makes sense, but it does assume VEPs are actually reviewed timely. Seeing 44 opened PRs right now doesn't feel like that, though. Granted, compared to 329 opened PRs on kubevirt/kubevirt it's smaller, but still, pushing towards VEPs for everything means more work and bigger pile to go through.

michalskrivanek (commented):
It's a good idea, however the added bureaucracy and nitpicking on wording in some cases so far makes me wonder if this shouldn't go hand in hand with push on significant simplification. e.g. no draft code needed, graduation left more open, etc. Constantly going back to VEPs to keep them in line with implementation sounds also questionable to me, I'd rather prefer docs/ in sync TBH.

docs/when-is-a-vep-required.md
EdDev: An API is a contract with its consumers, be it internal (kubevirt components) or external (clients integrating with kubevirt).
I do not know how to classify a contract change as "small".

Small is also not fast, this project has seen small changes consuming a large amount of resources for little benefit.
As I see it, VEPs can be small, clear and convincing, allowing a centralized place to discuss and manage API and functionality.

This decision has been taken when this VEP repo has been created, I do not think it is under dispute at the moment.

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
...
+## A VEP Is Required When a Change
+
+1. **Adds or modifies a public API** - new CRD fields, new CRDs, adding
+   new possible field values (e.g. enums), or changes to existing API
+   semantics or defaults.
+2. **Introduces new user-facing behavior that requires a feature gate** -
+   new VM/VMI behaviors, new configuration options, new status
+   conditions, or anything a user could notice without reading the code.
+3. **Adds new behavior under an existing feature gate** - this requires a
+   new VEP or an update to the existing one.
```
EdDev, line 18: Maybe this point with the previous one can be combined to "Behavior Change".
The fact a FG is needed or not is discussed in the VEP.

docs/when-is-a-vep-required.md
EdDev: It could be noted that sometimes, due to the trivial nature of a change, a VEP tracker may be enough. But these are usually very rare and I would consider them exceptions.

Examples: A new value to an existing trivial field.
When the discussion is about exposing it or not is minimal and pretty trivial. Small discussions can be recorded on the issue itself.

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
...
+6. Trivial code cleanups (typos, dead code removal).
+
+## Gray Area - Demands Judgment
+
+1. **Large internal refactors** - no API change, but high blast radius
+   and cross-team coordination needs.
+2. **New internal mechanisms** (e.g. a new webhook, a new internal
+   controller) that don't surface in the API but shape future design.
+3. **Changes to operator/deployment topology** that don't affect the user
+   API but affect how KubeVirt is installed or upgraded.
```
EdDev, line 41: I think "Changes component architecture" and " Requires coordination across multiple PRs or SIGs" cover these, and it does not seem that gray to me.

Said that, if we provide an exception path, we can use it without trying to classify the non trivial cases.

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
...
+   complexity and scope.
+
+## A VEP Is NOT Required For
+
+1. Bug fixes, unless they are extraordinarily complex.
+2. Test improvements or new test coverage.
+3. Documentation changes.
+4. Dependency bumps.
+5. Build, CI, or infrastructure changes.
+6. Trivial code cleanups (typos, dead code removal).
```
EdDev, line 32: I would classify it under refactoring.
With a note that in this context, refactoring refers to syntactic changes or semantics that do not change units logic (e.g. dependency inversion, design patterns improvements, etc).

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
+# When Is a VEP Required?
+
+## Summary
+
+Not every change to KubeVirt requires a VEP (Virtualization Enhancement
+Proposal). This document provides guidelines for contributors and
+reviewers to determine when a VEP is needed.
+
+## A VEP Is Required When a Change
```
EdDev, line 9: I would consider two additional reasons:
1. Release related changes (e.g. changing the format, changing the base OS of the components which effects upgrades).
2. Large scale changes which need to be tracked and managed over a large amount of PRs and/or releases.

docs/when-is-a-vep-required.md
EdDev: Adding new features, changing behavior or performing large changes to a production level project requires more attention and a stronger bar to be accepted.
I think the project has recognized that new functionality is harder to get in at this stage. However, we are trying to mitigate this through other means (e.g. assigning more attention/resources on VEPs and an extensibility to Kubevirt).

docs/when-is-a-vep-required.md
```diff
@@ -0,0 +1,59 @@
+# When Is a VEP Required?
+
+## Summary
+
+Not every change to KubeVirt requires a VEP (Virtualization Enhancement
```
EdDev, line 5: Can we go with the positive path first? :)

docs/when-is-a-vep-required.md
EdDev, line 1: How about `vep-qualification-criteria.md` ?

Other options are:
- `vep_scope.md`
- `vep_eligibility.md`

EdDev (changes requested):
Thank you!

Issue #326, opened 2026-06-04
VEP 326: directsync disk cache mode


michalskrivanek:
**Primary contact (assignee)**:

/assign @michalskrivanek 

**Current Feature Stage**:

GA

**Feature Gate**:

N/A

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): -
* Beta release target (x.y): -
* Stable release target (x.y): 1.10

* [] Alpha
  * [ ] VEP PRs: 
  * [ ] Code PRs: 
  * [ ] Docs PRs:

* [ ] Beta
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

* [X] GA
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/327
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/17718
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #327, opened 2026-06-04
VEP 326: directsync cache mode

michalskrivanek:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/326
**SIG label**:  /sig storage

### What this PR does
exposes qemu directsync cache mode

a lot of discussion is in https://github.com/kubevirt/kubevirt/pull/17718 already

### DIFF
--- a/veps/sig-storage/327-cache-modes/vep.md
+++ b/veps/sig-storage/327-cache-modes/vep.md
@@ -1,0 +1,171 @@
+# VEP #326: Expose `directsync` disk cache mode
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets GA for version: v1.11
+
+### Rationale for targeting GA directly
+
+The `directsync` cache mode is not a new hypervisor feature. It has been stable, well-understood primitive in QEMU and libvirt for over a decade:
+
+- `directsync` has been supported by QEMU since 2011 ([QEMU commit 92196b2f](https://github.com/qemu/qemu/commit/92196b2f5664d5defa778b1b24df56e2239b5e93)) and treated as migration-safe for shared storage by libvirt since 2017 ([libvirt commit fed9cc8](https://github.com/libvirt/libvirt/commit/fed9cc8)).
+- Libvirt has supported live migration for all cache modes (including those using the host page cache) since 2019 ([libvirt commit 4514abb](https://github.com/libvirt/libvirt/commit/4514abb)).
+- OpenStack Nova has exposed all five cache modes as first-class configuration options (`disk_cachemodes`) for years ([Nova libvirt config](https://docs.openstack.org/nova/latest/configuration/config.html#libvirt.disk_cachemodes)).
+- Community users have independently requested `directsync` and `unsafe` modes: [kubevirt#14265](https://github.com/kubevirt/kubevirt/issues/14265), [kubevirt#14363](https://github.com/kubevirt/kubevirt/pull/14363), [harvester#7906](https://github.com/harvester/harvester/issues/7906).
+
+This proposal does not introduce a new API field; it adds one value to an existing `DriverCache` enum that already exposes three of libvirt's five cache modes. The validation, defaulting, and I/O path logic already exist. A feature gate would add process overhead without meaningful risk reduction for values that are over a decade old in the underlying stack.
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt currently exposes three disk cache modes to users: `none`, `writethrough`, and `writeback`. The underlying hypervisor (QEMU/libvirt) supports two additional modes -- `directsync` and `unsafe` -- that address specific performance and data-safety trade-offs not covered by the existing three. This VEP proposes adding the `directsync` one as first-class `DriverCache` value in the KubeVirt API, along with the validation, defaulting, and test coverage required to support it.
+
+## Motivation
+
+Users running database workloads (e.g. MSSQL, PostgreSQL) inside VMs often need precise control over the I/O path to balance performance and durability. The existing three modes leave gaps:
+
+- **No direct I/O mode with writethrough guest reporting**: `none` bypasses the host page cache (via `O_DIRECT`) but reports writeback cache to the guest; the hypervisor issues `fdatasync()` only when the guest sends a flush. `directsync` also bypasses the host page cache but reports writethrough to the guest, so every guest write is treated as synchronous and committed to storage before completion is reported.
+
+Additionally, `directsync` is safe for shared/multi-attach disks (it uses `O_DIRECT`, ensuring data coherency), but KubeVirt currently rejects all shareable disks unless `cache=none`.
+
+### Performance evidence
+
+HammerDB TPC-C benchmarks against Microsoft SQL Server on Windows Server 2025 (32 vCPU, 440 GiB memory, virtio with 8 I/O multiqueues, 1 GiB hugepages, Intel Sapphire Rapids) show a consistent ~43-48% throughput improvement when switching from `cache=none` to `cache=directsync` using a temporary hook (see https://github.com/kubevirt/kubevirt/pull/17718):
+
+
+| Virtual users            | 100        | 200        | 300        | 400        | 500        |
+| ------------------------ | ---------- | ---------- | ---------- | ---------- | ---------- |
+| `cache=none` (TPM)       | 1,023,457  | 1,041,870  | 1,004,201  | 999,718    | 1,007,184  |
+| `cache=directsync` (TPM) | 1,464,588  | 1,494,152  | 1,486,263  | 1,432,949  | 1,464,549  |
+| **Improvement**          | **+43.1%** | **+43.4%** | **+48.0%** | **+43.3%** | **+45.4%** |
+
+
+The improvement comes from how MSSQL interacts with the guest-visible cache mode. With `cache=none`, the guest sees a writeback cache and may issue frequent flush commands for durability; each guest-initiated flush is translated to a host `fdatasync()`. With `cache=directsync`, the guest sees a writethrough cache, so MSSQL can rely on the writethrough guarantee instead of issuing its own flush/FUA traffic. That reduces guest-driven sync overhead on the host I/O path even though `directsync` commits each write synchronously via `O_DSYNC`.
+
+## Goals
+
+- Add `directsync` as valid value for `DriverCache`.
+- Validate that `directsync` is only used on storage that supports direct I/O (same check as `none`).
+- Allow `directsync` as a valid cache mode for shareable disks.
+- Validate that all cache modes, including the new one, function correctly through functional tests covering migration and data integrity.
+
+## Non Goals
+
+- Changing the default cache mode (remains `none` for block devices, auto-detected for file-based).
+- Modifying the KubeVirt CR-level `DiskVerification` or cluster-wide cache override mechanism.
+
+## Definition of Users
+
+- **VM administrators** who need to tune disk I/O for specific guest workloads (databases).
+- **Platform operators** who manage shared storage and need `O_DIRECT`-safe cache modes beyond `none`.
+
+## User Stories
+
+- As a VM administrator running MSSQL inside a VM on NVMe-backed PVCs, I want to set `cache: directsync` so that writes bypass the host page cache (maintaining data coherency for shared disks) and the guest sees writethrough semantics, avoiding guest-driven flush/FUA overhead on workloads like MSSQL, because my storage already has fast sync.
+- As a platform operator, I want to use `directsync` on shared/multi-attach block volumes so that multiple VMs can safely access the same disk with data coherency, without being limited to `cache: none`.
+
+## Repos
+
+- kubevirt/kubevirt
+
+## Design
+
+### New API constants
+
+Two new `DriverCache` values are added:
+
+```go
+const (
+    CacheNone         DriverCache = "none"
+    CacheWriteThrough DriverCache = "writethrough"
+    CacheWriteBack    DriverCache = "writeback"
+    CacheDirectSync   DriverCache = "directsync"   // new
+)
+```
+
+### Cache mode behavior
+
+The five modes differ in three dimensions: host page cache usage, guest-visible cache reporting, and per-write sync behavior. These are standard hypervisor cache semantics (currently implemented via QEMU/libvirt):
+
+
+| Mode           | Host page cache | Guest sees writeback | Sync per write             | Direct I/O |
+| -------------- | --------------- | -------------------- | -------------------------- | ---------- |
+| `none`         | No              | Yes                  | No (only on guest flush)   | Yes        |
+| `writethrough` | Yes             | No                   | Yes (fdatasync per write)  | No         |
+| `writeback`    | Yes             | Yes                  | No (only on guest flush)   | No         |
+| `directsync`   | No              | No                   | Yes (O_DSYNC)              | Yes        |
+| `unsafe`       | Yes             | Yes                  | No (guest flushes ignored) | No         |
+
+
+**Key distinctions:**
+
+- `none` vs `directsync`: Both bypass the host page cache via direct I/O. The difference is that `none` reports writeback cache to the guest; the hypervisor issues `fdatasync()` only when the guest sends a flush. `directsync` reports writethrough to the guest, so every guest write is committed via `O_DSYNC`.
+- `writeback` vs `unsafe`: Both use the host page cache and report writeback to the guest. The difference is that `unsafe` ignores guest-initiated flush commands, never calling `fdatasync()`. This means maximum throughput but data may not be in sync on host crash, even if the guest explicitly flushed.
+- **Shareable disk safety**: Direct I/O modes (`none`, `directsync`) bypass the host page cache, preventing stale reads when multiple VMs access the same block device. Modes that use the host cache (`writeback`, `writethrough`, `unsafe`) are unsafe for shared access because each host may cache different views of the data.
+
+### Validation changes
+
+**Direct I/O validation:**
+
+- When `directsync` is requested, the same direct I/O support check used for `none` is applied. If the backing storage does not support direct I/O and the mode was explicitly requested, admission fails.
+
+**Shareable disk validation:**
+
+- The existing restriction that only allows `cache=none` for shareable disks is relaxed to also accept `directsync`, since it also uses direct I/O and is safe for shared access.
+
+**Admission webhook:**
+
+- The webhook validates that the provided cache string is one of the four known values.
+
+### Live migration
+
+**`directsync`:** Fully compatible with live migration. Libvirt has treated `directsync` as migration-safe for shared storage since [2017](https://github.com/libvirt/libvirt/commit/fed9cc8). It uses direct I/O like `none`, so no host page cache state needs to be flushed or transferred. The cache mode is preserved on the destination.
+
+## Alternatives
+
+1. **Expose individual cache sub-properties**: The hypervisor internally represents cache modes as a combination of boolean flags (direct I/O, writeback reporting, flush suppression). Exposing these individually would give maximum flexibility but adds API complexity and requires users to understand hypervisor internals. The five named modes cover all practical combinations.
+2. **Use I/O mode as a proxy for direct I/O**: The I/O mode (`native`, `threads`) is orthogonal to cache mode. `native` requires direct I/O but doesn't control the caching behavior. These are separate concerns and should remain separate API fields.
+3. **Do nothing**: Users can already work around missing modes by patching the hypervisor domain configuration via hooks or plugins, but this is fragile, not validated, and not portable across migrations.
+4. **Expose `unsafe` as well**: This has been rejected during PR and VEP review.
+
+## Scalability
+
+No scalability impact. Cache mode is a per-disk property resolved at VMI creation time. No new controllers, watchers, or API calls are introduced.
+
+## Update Compatibility
+
+- **Forward compatible**: Existing VMIs with `none`, `writethrough`, or `writeback` are unaffected. New VMIs specifying `directsync` are only valid on clusters running the version that introduces these values.
+- **API versioning**: The new constants are additive to the `DriverCache` type. No existing constants are modified or removed.
+
+## Functional Testing Approach
+
+- **Migration with cache modes**: Functional tests verify that VMIs using the VM with `directsync` cache mode successfully complete live migration with PVC-backed shared block disks, and that data written before migration is readable after migration.
+- **Validation**: Unit tests cover admission validation for all four cache modes, including shareable disk restrictions and direct I/O fallback behavior.
+
+## Implementation History
+
+- 2026-06-04: Initial PR adding `directsync` and `unsafe` to API types, validation, and test coverage ([kubevirt/kubevirt#17718](https://github.com/kubevirt/kubevirt/pull/17718)).
+- 2026-06-04: VEP created ([kubevirt/enhancements#327](https://github.com/kubevirt/enhancements/pull/327)).
+- 2026-08-06: `unsafe` cache mode split into separate implementation PR with live migration blocker ([kubevirt/kubevirt#18664](https://github.com/kubevirt/kubevirt/pull/18664)).
+- 2026-09-07: removed `unsafe`
+
+## Graduation Requirements
+
+### GA (v1.11)
+
+- `directsync`  accepted as valid `DriverCache` values.
+- Direct I/O validation for `directsync` matches `none`.
+- `directsync` allowed for shareable disks.
+- Unit tests for all validation paths.
+- Functional tests for live migration with the new cache mode.
+- Documentation updated in kubevirt.io user guide.
+- `virtctl` accepts the new mode in cache-related flags (e.g. hotplug volume).
+
--- a/veps/sig-storage/cache_modes.md
+++ b/veps/sig-storage/cache_modes.md
@@ -1,0 +1,178 @@
+# VEP #326: Expose `directsync` and `unsafe` disk cache modes
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets beta for version: v1.10
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- (R) Alpha target version is explicitly mentioned and approved
+- (R) Beta target version is explicitly mentioned and approved
+- (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt currently exposes three QEMU disk cache modes to users: `none`, `writethrough`, and `writeback`. QEMU supports two additional modes -- `directsync` and `unsafe` -- that address specific performance and data-safety trade-offs not covered by the existing three. This VEP proposes adding `directsync` and `unsafe` as first-class `DriverCache` values in the KubeVirt API, along with the validation, defaulting, and test coverage required to support them.
+
+## Motivation
+
+Users running database workloads (e.g. MSSQL, PostgreSQL) inside VMs often need precise control over the I/O path to balance performance and durability. The existing three modes leave gaps:
+
+- **No "safe + bypass host cache" option without guest flush penalties**: `none` bypasses the host page cache (via `O_DIRECT`) but reports writeback cache to the guest, causing QEMU to issue `fdatasync()` on every guest write. `directsync` also bypasses the host cache but issues `fdatasync()` only on guest-initiated flush commands, which can significantly reduce I/O overhead for flush-rare workloads on storage with fast sync (NVMe, BBU-backed RAID).
+- **No "maximum throughput, relaxed durability" option**: Some workloads (CI pipelines, ephemeral VMs, batch processing) tolerate data loss on host crash in exchange for maximum I/O throughput. `unsafe` disables all host-side flush processing and caches aggressively, which is currently not expressible.
+
+Additionally, `directsync` is safe for shared/multi-attach disks (it uses `O_DIRECT`, ensuring data coherency), but KubeVirt currently rejects all shareable disks unless `cache=none`.
+
+### Performance evidence
+
+HammerDB TPC-C benchmarks against Microsoft SQL Server on Windows Server 2025 (32 vCPU, 440 GiB memory, virtio with 8 I/O multiqueues, 1 GiB hugepages, Intel Sapphire Rapids) show a consistent ~43-48% throughput improvement when switching from `cache=none` to `cache=directsynca` using a temporary hook (see https://github.com/kubevirt/kubevirt/pull/17718):
+
+
+| Virtual users            | 100        | 200        | 300        | 400        | 500        |
+| ------------------------ | ---------- | ---------- | ---------- | ---------- | ---------- |
+| `cache=none` (TPM)       | 1,023,457  | 1,041,870  | 1,004,201  | 999,718    | 1,007,184  |
+| `cache=directsync` (TPM) | 1,464,588  | 1,494,152  | 1,486,263  | 1,432,949  | 1,464,549  |
+| **Improvement**          | **+43.1%** | **+43.4%** | **+48.0%** | **+43.3%** | **+45.4%** |
+
+
+The improvement comes from how MSSQL interacts with the guest-visible cache mode. With `cache=none`, the guest sees a writeback cache and QEMU issues `fdatasync()` on every write (FUA emulation). With `cache=directsync`, the guest sees a writethrough cache; MSSQL can rely on the writethrough guarantee and reduce the frequency of explicit flush commands, significantly lowering the `fdatasync()` overhead on the host I/O path.
+
+## Goals
+
+- Add `directsync` and `unsafe` as valid values for `v1.DriverCache`.
+- Validate that `directsync` is only used on storage that supports `O_DIRECT` (same check as `none`).
+- Allow `directsync` as a valid cache mode for shareable disks.
+- Ensure all five cache modes survive live migration with the correct libvirt XML.
+- Provide unit and e2e test coverage for the new modes.
+
+## Non Goals
+
+- Changing the default cache mode (remains `none` for block devices, auto-detected for file-based).
+- Modifying the KubeVirt CR-level `DiskVerification` or cluster-wide cache override mechanism.
+
+## Definition of Users
+
+- **VM administrators** who need to tune disk I/O for specific guest workloads (databases, CI, batch).
+- **Platform operators** who manage shared storage and need `O_DIRECT`-safe cache modes beyond `none`.
+
+## User Stories
+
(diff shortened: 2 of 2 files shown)


Acedus (commented):
As per the discussion on https://github.com/kubevirt/kubevirt/pull/17718, the VEP doesn't raise implementation concerns for me, the design is straightforward and the validation/migration changes are well-scoped. What is missing, however, is concrete justification in the VEP itself (that _isn't_ to say that usecases provided aren't good, though!). The performance data backing directsync (MSSQL flushing improvements on NVMe-backed storage) was shared on the PR via internal references (e.g., CNV-50763, RHEL-71987) but doesn't appear in this VEP. A VEP should be self-contained so I'd like to see that data included here, ideally in a publicly accessible form.

Onto the merits themselves: directsync has a clear, data-backed usecase and is strictly safer than none (yes, O_DIRECT, stricter sync). I'm comfortable with it. unsafe is a harder sell since while the stated usecases (CI and ephemeral VMs) are reasonable, I'd appreciate seeing external clients requesting this in the wild before committing it to the API.

michalskrivanek:
@Acedus thank you so much for a review!

> As per the discussion on [kubevirt/kubevirt#17718](https://github.com/kubevirt/kubevirt/pull/17718), the VEP doesn't raise implementation concerns for me, the design is straightforward and the validation/migration changes are well-scoped. What is missing, however, is concrete justification in the VEP itself (that _isn't_ to say that usecases provided aren't good, though!). The performance data backing directsync (MSSQL flushing improvements on NVMe-backed storage) was shared on the PR via internal references (e.g., CNV-50763, RHEL-71987) but doesn't appear in this VEP. A VEP should be self-contained so I'd like to see that data included here, ideally in a publicly accessible form.

We're not trying to demonstrate that it is a superior configuration for a general case. The performance characteristic is very specific to setup and workload. The whole "justification" really boils down to a simple "there are workloads that are more efficient when they control their own fdatasync behavior". I'll copy&paste some data, but TBH that's not really that interesting, the main point is that Kubevirt should allow the flexibility for various workloads

> Onto the merits themselves: directsync has a clear, data-backed usecase and is strictly safer than none (yes, O_DIRECT, stricter sync). I'm comfortable with it. unsafe is a harder sell since while the stated usecases (CI and ephemeral VMs) are reasonable, I'd appreciate seeing external clients requesting this in the wild before committing it to the API.

The major argument for `unsafe` was mentioned there too, we use it in kubevirtci for years, so that's both a very valid usecase for other users, and it is also extensively exercised code path already.

fabiand:
> The performance data backing directsync (MSSQL flushing improvements on NVMe-backed storage) was shared on the PR via internal references (e.g., CNV-50763, RHEL-71987) but doesn't appear in this VEP. A VEP should be self-contained so I'd like to see that data included here, ideally in a publicly accessible form.

+1

I htink as VEP authors we need to figure out the best way to provide this data. And also the background of how this got aggreagted (in this case it as several months of multiple ppl and sidecars in order to get to this small change).

> I'd appreciate seeing external clients requesting this in the wild before committing it to the API.

How do you imagine this to work if the clients can not be disclosed?

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,168 @@
...
+
+- **Forward compatible**: Existing VMIs with `none`, `writethrough`, or `writeback` are unaffected. New VMIs specifying `directsync` or `unsafe` are only valid on clusters running the version that introduces these values.
+- **Rollback**: If a cluster is rolled back to a version that does not recognize `directsync` or `unsafe`, existing VMIs with these modes will fail validation on restart/migration. Running VMIs are unaffected until they are stopped. Administrators should update VMI specs before rollback.
+- **API versioning**: The new constants are additive to the `DriverCache` type. No existing constants are modified or removed.
+
+## Functional Testing Approach
+
+### E2e tests
+
+- **Migration with cache modes**: A `DescribeTable` tests that VMIs with `writeback`, `writethrough`, `directsync`, and `unsafe` cache modes successfully complete live migration and preserve the correct cache mode in the post-migration libvirt domain XML.
```
akalenyu, line 150: I believe we were considering actually looking up the data in the guest post migration, does that not have value?

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,168 @@
...
+- **Shared block disk migration**: A `DescribeTable` tests migration of PVC-backed shared block disks with `none` and `directsync`, verifying cache mode preservation.
```
akalenyu, line 151: So this test seems to operate entirely on the XML level, so, I wonder if it has any value over a unit test or some libvirt e2e test outside kubevirt? Maybe we're missing a key step in the guest like ensuring no write lock? I just want to make sure we are testing the original scenario which inspired this test. I doubt it was just about XML correctness

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,168 @@
...
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt currently exposes three QEMU disk cache modes to users: `none`, `writethrough`, and `writeback`. QEMU supports two additional modes -- `directsync` and `unsafe` -- that address specific performance and data-safety trade-offs not covered by the existing three. This VEP proposes adding `directsync` and `unsafe` as first-class `DriverCache` values in the KubeVirt API, along with the validation, defaulting, and test coverage required to support them.
+
+## Motivation
+
+Users running database workloads (e.g. MSSQL, PostgreSQL) inside VMs often need precise control over the I/O path to balance performance and durability. The existing three modes leave gaps:
```
akalenyu, line 25: Can't argue with that!

akalenyu (commented):
Thank you for following up with a VEP!

akalenyu:
@michalskrivanek note sign off is missing on the commit :pray:

Acedus:
> The major argument for unsafe was mentioned there too, we use it in kubevirtci for years, so that's both a very valid usecase for other users, and it is also extensively exercised code path already.

I'll be blunt, the real problem isn't that unsafe is technically dangerous or anything, it's just that the user story in this VEP is weak. kvci using it internally is a convenience for our own test infra, not evidence of user demand. I can come up with _some_ usecases that may require ephemeral VMs for legacy purposes (e.g., Windows CI runners on a Linux k8s cluster, driver validation, etc.). Without a real user story, I don't personally see why it shouldn't exist as a plugin.

> How do you imagine this to work if the clients can not be disclosed?

That's a fair point, and I recognize engineers vouching for downstream demand does carry weight. But for an upstream VEP to be self-contained, we need something tangible... an anonymized benchmark, a blog post, a conference talk, even a Github issue under a pseudonym describing the problem. Not because we doubt the internal data exists, but because a VEP that justifies an API addition with "trust us, clients need this" sets a precedent that makes it hard for the community to evaluate future proposals on their own merits.

Having said all that, I'm still inclined to lgtm this, at the very least for directsync.

mhenriks:
> an anonymized benchmark, a blog post, a conference talk, even a Github issue under a pseudonym describing the problem

@Acedus this evidence exists:

- kubevirt/kubevirt#14265 — external RFE for exactly `unsafe` and `directsync`. When a maintainer asked for the use case, the reporter described the problem concretely: slow disks make VMs slow enough that etcd times out and nodes go NotReady; `cache=unsafe` on plain libvirt solves it, tradeoff knowingly accepted for ephemeral VMs.
- harvester/harvester#7906 — the same request filed downstream in Harvester.
- kubevirt/kubevirt#14363 — an external contributor's PR implementing exactly these two modes.

The user story isn't weak. It's the standard ephemeral/CI pattern, and the demand for it is on public record: a user describing a concrete problem, and a contributor who went as far as implementing it.

On the plugin suggestion: that's effectively the status quo, and @fabiand already described what it costs. A lot. The VEP exists because that path doesn't work, not because nobody tried it.

On "committing it to the API": the cache field already exists and exposes three of libvirt's five values, so this is less new API surface than completing a partial enum.

veps/sig-storage/cache_modes.md
michalskrivanek: it does to some degree ....though if we're presuming we're losing data and neither qemu nor libvirt CI catches that it's kind of....eh...but fair enough, if anything can go wrong it will surely will, one day:)
So yes, it's simple and we can do that. It's in https://github.com/kubevirt/kubevirt/pull/17889

this line wasn't to say we shouldn't also do other checks, it's just htat we have this post-migration xml check for the other modes already, so just extend it for the new ones...

akalenyu:
@vladikr @xpivarc I think we can proceed with this, and possibly unhold the impl.?
https://github.com/kubevirt/kubevirt/pull/17718#issuecomment-5014782118

akalenyu (approved):

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- **Shared block disk migration**: A `DescribeTable` tests migration of PVC-backed shared block disks with `none` and `directsync`, verifying cache mode preservation.
+
+## Implementation History
+
+- 2025-06-04: Initial PR adding `directsync` and `unsafe` to API types, validation, and test coverage (kubevirt/kubevirt#17718).
+- 2025-06-04: VEP created.
+
+## Graduation Requirements
+
+### Alpha
```
EdDev, line 160: The VEP metadata points to Beta and bypasses Alpha. Please elaborate on the reasoning that merits skipping Alpha.

Also, as a pre-GA release, it requires a FG.
The only way to avoid a FG is to target it for GA, which may be reasonable if you provide strong reasoning.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- **No "maximum throughput, relaxed durability" option**: Some workloads (CI pipelines, ephemeral VMs, batch processing) tolerate data loss on host crash in exchange for maximum I/O throughput. `unsafe` disables all host-side flush processing and caches aggressively, which is currently not expressible.
+
+Additionally, `directsync` is safe for shared/multi-attach disks (it uses `O_DIRECT`, ensuring data coherency), but KubeVirt currently rejects all shareable disks unless `cache=none`.
+
+### Performance evidence
+
+HammerDB TPC-C benchmarks against Microsoft SQL Server on Windows Server 2025 (32 vCPU, 440 GiB memory, virtio with 8 I/O multiqueues, 1 GiB hugepages, Intel Sapphire Rapids) show a consistent ~43-48% throughput improvement when switching from `cache=none` to `cache=directsynca` using a temporary hook (see https://github.com/kubevirt/kubevirt/pull/17718):
```
EdDev, line 34:
```suggestion
HammerDB TPC-C benchmarks against Microsoft SQL Server on Windows Server 2025 (32 vCPU, 440 GiB memory, virtio with 8 I/O multiqueues, 1 GiB hugepages, Intel Sapphire Rapids) show a consistent ~43-48% throughput improvement when switching from `cache=none` to `cache=directsync` using a temporary hook (see https://github.com/kubevirt/kubevirt/pull/17718):
```

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+
+
+The improvement comes from how MSSQL interacts with the guest-visible cache mode. With `cache=none`, the guest sees a writeback cache and QEMU issues `fdatasync()` on every write (FUA emulation). With `cache=directsync`, the guest sees a writethrough cache; MSSQL can rely on the writethrough guarantee and reduce the frequency of explicit flush commands, significantly lowering the `fdatasync()` overhead on the host I/O path.
+
+## Goals
+
+- Add `directsync` and `unsafe` as valid values for `v1.DriverCache`.
+- Validate that `directsync` is only used on storage that supports `O_DIRECT` (same check as `none`).
+- Allow `directsync` as a valid cache mode for shareable disks.
+- Ensure all five cache modes survive live migration with the correct libvirt XML.
```
EdDev, line 51: What does
> with the correct libvirt XML.

mean?

This proposal is about two new cache modes, why it refers all of them and not only these two?

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- Provide unit and e2e test coverage for the new modes.
```
EdDev, line 52: This is not a feature goal. Test coverage is to be discussed in its own specific spection.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- `writeback` vs `unsafe`: Both use the host page cache and report writeback to the guest. The difference is that `unsafe` ignores guest-initiated flush commands (`BDRV_O_NO_FLUSH`), never calling `fdatasync()`. This means maximum throughput but data loss on host crash even if the guest explicitly flushed.
+- **Shareable disk safety**: `O_DIRECT` modes (`none`, `directsync`) bypass the host page cache, preventing stale reads when multiple VMs access the same block device. Modes that use the host cache (`writeback`, `writethrough`, `unsafe`) are unsafe for shared access because each VM's host may cache different views of the data.
+
+### Validation changes
+
+`**SetDriverCacheMode` in `converter.go`:**
```
EdDev, line 112: Please avoid file and function names, unless it is an API field or value. The code is changing all the time and the VEP is expected to describe a higher level abstract. This gives the freedom to implement the VEP per the current code status and not get back here for updates (e.g. after codebase refactoring).

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- The existing check `cache != none → error` is relaxed to `cache != none && cache != directsync → error`, since `directsync` also uses `O_DIRECT` and is safe for shared access.
+
+**Admission webhook:**
+
+- The existing `DriverCache` validation already accepts any string value that maps to a valid QEMU cache mode. The webhook validates that the provided cache string is one of the five known values.
+
+### Live migration
+
+All five cache modes are compatible with QEMU live migration. The cache mode is a property of the disk driver, not of the migration state, and is preserved in the libvirt domain XML on the target host. No special migration handling is needed.
+However they deserve to be tested, there has been issues in the past.
```
EdDev, line 128: Can you ref any? If not, this sentence leaves the readers wandering if it works or not.

Perhaps you can just mention that PoC evaluations show that migration succeeds and that coverage is planned as part of Alpha/Beta.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+3. **Do nothing**: Users can already work around missing modes by patching libvirt domain XML via hooks, but this is fragile, not validated, and not portable across migrations.
+
+## Scalability
+
+No scalability impact. Cache mode is a per-disk property resolved at VMI creation time. No new controllers, watchers, or API calls are introduced.
+
+## Update/Rollback Compatibility
+
+- **Forward compatible**: Existing VMIs with `none`, `writethrough`, or `writeback` are unaffected. New VMIs specifying `directsync` or `unsafe` are only valid on clusters running the version that introduces these values.
+- **Rollback**: If a cluster is rolled back to a version that does not recognize `directsync` or `unsafe`, existing VMIs with these modes will fail validation on restart/migration. Running VMIs are unaffected until they are stopped. Administrators should update VMI specs before rollback.
```
EdDev, line 143: Rollback is not supported by Kubevirt in general. You can mention the expected behavior here from a theoretical aspect, but emphasize that is not expected to be used as of today.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- **API versioning**: The new constants are additive to the `DriverCache` type. No existing constants are modified or removed.
+
+## Functional Testing Approach
+
+### E2e tests
```
EdDev, line 148: I would recommend calling these "Functional".
IMO e2e is confusing, but this is a nit.

veps/sig-storage/cache_modes.md
EdDev: Checking the hypervisor domain configuration is violating layering and abstractions. Please avoid it, as we have been in the process of reducing our tests dependency on these.

If there is data which needs to be checked, check the VMI status or access the guest and perform some check. If the VMI status cannot express the status directly or indirectly, and accessing the guest does not help, you may suggest a new status (field, condition, etc).

Regarding the test details, IMO you should not use implementation details like "DescribeTable", it is not useful here. And I do not see the value of discussing the other cache options when this VEP focuses on 2 specifically.

veps/sig-storage/cache_modes.md
EdDev: I do not really understand the "cache mode preservation" thing.
The only interest part IMO is that the live-migration succeeds and that the functionality is preserved. Maybe checking that writing and reading is working smoothly with no curroptions.

I'll ignore on how you check something here, but will warn that in the functional tests we avoid checking the underlying hypervisor configuration directly.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+
+- 2025-06-04: Initial PR adding `directsync` and `unsafe` to API types, validation, and test coverage (kubevirt/kubevirt#17718).
+- 2025-06-04: VEP created.
+
+## Graduation Requirements
+
+### Alpha
+
+- `directsync` and `unsafe` accepted as valid `DriverCache` values.
+- `O_DIRECT` validation for `directsync` matches `none`.
```
EdDev, line 163: Can you express this  in a more abstract language?

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+- `directsync` allowed for shareable disks.
+- Unit tests for all validation paths.
+- E2e tests for migration with all five cache modes.
```
EdDev, line 166:
```suggestion
- Functional tests for live-migration with the new `directsync` and `unsafe` cache modes.
```

EdDev (changes requested):
Thank you.

The VEP makes good arguments on adding two new values to an existing cache mode field.
I find the motivation and reasoning clear and convincing.

Beyond my inline points, I would add that the VEP content uses QEMU and Libvirt frequently, and I would prefer to see a more generic use of "hypervisor" as a whole and mention that these are specific to it in one place. Other future hypervisors may or may not support these. I think the VEP can recognize the project direction of abstracting away the hypervisor as a long term goal by using the relevant generic language.

veps/sig-storage/cache_modes.md
```diff
@@ -0,0 +1,178 @@
...
+
+- The existing check `cache != none → error` is relaxed to `cache != none && cache != directsync → error`, since `directsync` also uses `O_DIRECT` and is safe for shared access.
+
+**Admission webhook:**
+
+- The existing `DriverCache` validation already accepts any string value that maps to a valid QEMU cache mode. The webhook validates that the provided cache string is one of the five known values.
+
+### Live migration
+
+All five cache modes are compatible with QEMU live migration. The cache mode is a property of the disk driver, not of the migration state, and is preserved in the libvirt domain XML on the target host. No special migration handling is needed.
```
EdDev, line 127: Based on the discussions initiated in the implementation PR [here](https://github.com/kubevirt/kubevirt/pull/17718#issuecomment-5132150153), 
the live-migration is not safe with the unsafe caching mode.

Please consider extracting it out from this VEP.

michalskrivanek: right, VEP text is not in sync now. `unsafe` has been moved to its own https://github.com/kubevirt/kubevirt/pull/18664 with the live migration protection.

veps/sig-storage/cache_modes.md
michalskrivanek: let's try the GA way, added reasoning

veps/sig-storage/cache_modes.md
michalskrivanek: updated to more general language

veps/sig-storage/cache_modes.md
michalskrivanek: done

veps/sig-storage/cache_modes.md
michalskrivanek: done

veps/sig-storage/cache_modes.md
michalskrivanek: validation of all cache modes (not just the new ones) were requested because the validation changes touch shared I/O path logic, past issues have required hypervisor patches specifically for KubeVirt's migration behavior ([example](https://lists.libvirt.org/archives/list/devel@lists.libvirt.org/thread/WWMSGL3EZ7EMBOQRBB4WFKTYSDAVHOTD/)), and generally the VM configuration is being [modified by kubevirt during migration](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/live-migration-source.go#L190)

veps/sig-storage/cache_modes.md
michalskrivanek: rewritten

veps/sig-storage/cache_modes.md
michalskrivanek: removed

veps/sig-storage/cache_modes.md
michalskrivanek: removed

veps/sig-storage/cache_modes.md
michalskrivanek: removed, leaving xml preservation to unit tests only

michalskrivanek:
@EdDev can you take another look, I believe all your comments have been addressed

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Overview
+
+KubeVirt currently exposes three disk cache modes to users: `none`, `writethrough`, and `writeback`. The underlying hypervisor (QEMU/libvirt) supports two additional modes -- `directsync` and `unsafe` -- that address specific performance and data-safety trade-offs not covered by the existing three. This VEP proposes adding `directsync` and `unsafe` as first-class `DriverCache` values in the KubeVirt API, along with the validation, defaulting, and test coverage required to support them.
+
+## Motivation
+
+Users running database workloads (e.g. MSSQL, PostgreSQL) inside VMs often need precise control over the I/O path to balance performance and durability. The existing three modes leave gaps:
+
+- **No "safe + bypass host cache" option without guest flush penalties**: `none` bypasses the host page cache (via `O_DIRECT`) but reports writeback cache to the guest, causing QEMU to issue `fdatasync()` on every guest write. `directsync` also bypasses the host cache but issues `fdatasync()` only on guest-initiated flush commands, which can significantly reduce I/O overhead for flush-rare workloads on storage with fast sync (NVMe, BBU-backed RAID).
```
mhenriks: You've got `none` and `directsync` reversed here. With `none` the hypervisor issues fdatasync() only on guest-initiated flush commands. `directsync` also bypasses the host cache but reports writethrough to the guest, so every guest write is treated as synchronous

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+| Virtual users            | 100        | 200        | 300        | 400        | 500        |
+| ------------------------ | ---------- | ---------- | ---------- | ---------- | ---------- |
+| `cache=none` (TPM)       | 1,023,457  | 1,041,870  | 1,004,201  | 999,718    | 1,007,184  |
+| `cache=directsync` (TPM) | 1,464,588  | 1,494,152  | 1,486,263  | 1,432,949  | 1,464,549  |
+| **Improvement**          | **+43.1%** | **+43.4%** | **+48.0%** | **+43.3%** | **+45.4%** |
+
+
+The improvement comes from how MSSQL interacts with the guest-visible cache mode. With `cache=none`, the guest sees a writeback cache and QEMU issues `fdatasync()` on every write (FUA emulation). With `cache=directsync`, the guest sees a writethrough cache; MSSQL can rely on the writethrough guarantee and reduce the frequency of explicit flush commands, significantly lowering the `fdatasync()` overhead on the host I/O path.
```
mhenriks: With cache=none, the guest sees a writeback cache, so the hypervisor issues fdatasync() only when the guest sends a flush

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+### Cache mode behavior
+
+The five modes differ in three dimensions: host page cache usage, guest-visible cache reporting, and per-write sync behavior. These are standard hypervisor cache semantics (currently implemented via QEMU/libvirt):
+
+
+| Mode           | Host page cache | Guest sees writeback | Sync per write             | Direct I/O |
+| -------------- | --------------- | -------------------- | -------------------------- | ---------- |
+| `none`         | No              | Yes                  | Yes (FUA emulation)        | Yes        |
```
mhenriks: No sync per write for `none`

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+| `writethrough` | Yes             | No                   | Yes (FUA emulation)        | No         |
+| `writeback`    | Yes             | Yes                  | No (only on guest flush)   | No         |
+| `directsync`   | No              | No                   | Yes (FUA emulation)        | Yes        |
+| `unsafe`       | Yes             | Yes                  | No (guest flushes ignored) | No         |
+
+
+**Key distinctions:**
+
+- `none` vs `directsync`: Both bypass the host page cache via direct I/O. The difference is that `none` reports writeback cache to the guest, causing the hypervisor to emit `fdatasync()` after every write via FUA emulation. `directsync` reports writethrough to the guest, so the guest application controls when explicit flushes are issued. Both modes call `fdatasync()` per write, but `directsync` allows workloads that manage their own flush behavior to reduce redundant sync overhead.
```
mhenriks: `none` does not cause the hypervisor to emit `fdatasync()`. That is what `directsync` does

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+- `writeback` vs `unsafe`: Both use the host page cache and report writeback to the guest. The difference is that `unsafe` ignores guest-initiated flush commands, never calling `fdatasync()`. This means maximum throughput but data may not be in sync on host crash, even if the guest explicitly flushed.
+- **Shareable disk safety**: Direct I/O modes (`none`, `directsync`) bypass the host page cache, preventing stale reads when multiple VMs access the same block device. Modes that use the host cache (`writeback`, `writethrough`, `unsafe`) are unsafe for shared access because each host may cache different views of the data.
+
+### Validation changes
+
+**Direct I/O validation:**
+
+- When `directsync` is requested, the same direct I/O support check used for `none` is applied. If the backing storage does not support direct I/O, the mode falls back to `writethrough` (matching `none`'s existing fallback to `writeback`).
```
mhenriks: Currently, it is an error if `none` is explicitly requested but the disk does support directio. The fallback only applies when cache mode is not explicitly specified. `directsync` should behave similarly

veps/sig-storage/327-cache-modes/vep.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Motivation
+
+Users running database workloads (e.g. MSSQL, PostgreSQL) inside VMs often need precise control over the I/O path to balance performance and durability. The existing three modes leave gaps:
+
+- **No "safe + bypass host cache" option without guest flush penalties**: `none` bypasses the host page cache (via `O_DIRECT`) but reports writeback cache to the guest, causing QEMU to issue `fdatasync()` on every guest write. `directsync` also bypasses the host cache but issues `fdatasync()` only on guest-initiated flush commands, which can significantly reduce I/O overhead for flush-rare workloads on storage with fast sync (NVMe, BBU-backed RAID).
+- **No "maximum throughput, relaxed durability" option**: Some workloads (CI pipelines, ephemeral VMs, batch processing) tolerate data loss on host crash in exchange for maximum I/O throughput. `unsafe` disables all host-side flush processing and caches aggressively, which is currently not expressible.
```
mhenriks: +1 to keeping `unsafe` may just want to flesh out the motivation/user stories and risks

mhenriks (commented):
I hope we can get consensus on keeping `unsafe` otherwise everything looks pretty good

veps/sig-storage/327-cache-modes/vep.md
michalskrivanek: right

veps/sig-storage/327-cache-modes/vep.md
michalskrivanek: right

veps/sig-storage/327-cache-modes/vep.md
michalskrivanek: that's better indeed

veps/sig-storage/327-cache-modes/vep.md
michalskrivanek: tried...

veps/sig-storage/327-cache-modes/vep.md
michalskrivanek: I guess that can happen when you keep editing this over and over....

michalskrivanek:
Hi @mhenriks, thank you for the review, addressed comments, PTAL.

mhenriks:
/lgtm
Thanks @michalskrivanek!

michalskrivanek:
note: on Kubevirt VEP Discussion today it has been suggested to split further to `directsync` and `unsafe` VEPs.

(I would still love to see an answer/reasoning to dismiss the `unsafe` use-cases mentioned, oh well...)

veps/sig-storage/cache_modes.md
michalskrivanek: removed, and rewritten

veps/sig-storage/cache_modes.md
michalskrivanek: fair enough, as per discussion in the PR this test is now being removed instead of extended

michalskrivanek:
It's been a long discussion, and I thank everyone for the thoughtful feedback on the nuances of these modes.

I'd push back gently on splitting the VEP further. The implementation is already split: directsync in [kubevirt#17718](https://github.com/kubevirt/kubevirt/pull/17718), unsafe in [kubevirt#18664](https://github.com/kubevirt/kubevirt/pull/18664). That's the right place to nail down specifics — migration blocking, validation, tests, and any remaining technical objections. Those PRs can move at different speeds without blocking each other.

The VEP serves a different purpose. It's one proposal for the two cache modes users have been asking for ([kubevirt#14265](https://github.com/kubevirt/kubevirt/issues/14265)) that KubeVirt doesn't expose today, even though they're well-established in QEMU/libvirt. They address different trade-offs — stricter sync with direct I/O vs. maximum throughput with relaxed durability — but they're part of the same question: how much control over the disk I/O path should KubeVirt offer through the existing cache field. Pulling unsafe out of the VEP would leave an incomplete picture of what we're proposing and why, while the implementation split already addresses the "land one, defer the other" concern.

Happy to keep refining the unsafe sections — migration blocking, user stories, risks — and to address point-by-point feedback in #18664. But I'd rather not fork the VEP into two documents for modes that belong to the same API field and the same design discussion.

Does that work for folks who asked to defer unsafe? The VEP can describe both; only the impl PRs need to merge on different timelines.

cc @xpivarc, @vladikr

vladikr:
> It's been a long discussion, and I thank everyone for the thoughtful feedback on the nuances of these modes.
> 
> I'd push back gently on splitting the VEP further. The implementation is already split: directsync in [kubevirt#17718](https://github.com/kubevirt/kubevirt/pull/17718), unsafe in [kubevirt#18664](https://github.com/kubevirt/kubevirt/pull/18664). That's the right place to nail down specifics — migration blocking, validation, tests, and any remaining technical objections. Those PRs can move at different speeds without blocking each other.
> 
> The VEP serves a different purpose. It's one proposal for the two cache modes users have been asking for ([kubevirt#14265](https://github.com/kubevirt/kubevirt/issues/14265)) that KubeVirt doesn't expose today, even though they're well-established in QEMU/libvirt. They address different trade-offs — stricter sync with direct I/O vs. maximum throughput with relaxed durability — but they're part of the same question: how much control over the disk I/O path should KubeVirt offer through the existing cache field. Pulling unsafe out of the VEP would leave an incomplete picture of what we're proposing and why, while the implementation split already addresses the "land one, defer the other" concern.
> 
> Happy to keep refining the unsafe sections — migration blocking, user stories, risks — and to address point-by-point feedback in #18664. But I'd rather not fork the VEP into two documents for modes that belong to the same API field and the same design discussion.
> 
> Does that work for folks who asked to defer unsafe? The VEP can describe both; only the impl PRs need to merge on different timelines.
> 
> cc @xpivarc, @vladikr

Thanks @michalskrivanek 
My concerns around "unsafe" are really because this is unsafe... the node goes down and you lost your workload, not just live migration.
Until we have a really solid story around this, I'd prefer users who really want this functionality use the plugins route. 

> Pulling unsafe out of the VEP would leave an incomplete picture of what we're proposing and why, while the implementation split already addresses the "land one, defer the other" concern.

I really don't mind that... we don't strive for completeness (I think ([kubevirt#14265](https://github.com/kubevirt/kubevirt/issues/14265)) was motivated by completeness rather than a real use case that considers alternatives) - we don't support all vmm/hypervisor features just because they are there.

I don't mind if the "unsafe" option is mentioned in this vep, but the language around it shouldn't sound like we've agreed to offer this in the future in this shape.

michalskrivanek:
> Thanks @michalskrivanek My concerns around "unsafe" are really because this is unsafe... the node goes down and you lost your workload, not just live migration. Until we have a really solid story around this, I'd prefer users who really want this functionality use the plugins route.

the node goes down and you lose your workload with `writeback` too. 
Plugins are what people use today because they have no other option. They’re brittle, harder to validate, and don’t give kubevirt a place to enforce guardrails.
To me clearly "owning" it is a more solid story than just turn people away to plugins.
 
> > Pulling unsafe out of the VEP would leave an incomplete picture of what we're proposing and why, while the implementation split already addresses the "land one, defer the other" concern.
> 
> I really don't mind that... we don't strive for completeness (I think ([kubevirt#14265](https://github.com/kubevirt/kubevirt/issues/14265)) was motivated by completeness rather than a real use case that considers alternatives) - we don't support all vmm/hypervisor features just because they are there.

I don't want to vouch for someone else's use cases too much, but the comments in https://github.com/kubevirt/kubevirt/issues/14265#issuecomment-2733142663 are specific, not just mere completeness. 

> I don't mind if the "unsafe" option is mentioned in this vep, but the language around it shouldn't sound like we've agreed to offer this in the future in this shape.

when discussing in [this](https://github.com/kubevirt/kubevirt/pull/17718#issuecomment-5205284668) thread 
the VEP was changed to target GA immediately. I wonder if softening this to target alpha only would make it more acceptable for keeping both, with more time to build confidence by beta or GA.

but regardless - ok, understood, we're not there yet in terms of being convincing enough as is. Thanks a lot for thoughts.

akalenyu:
Summary from the sig-storage call:
@vladikr correct me if I'm wrong, I think if we eliminate the hard promises about unsafe, we can hopefully move faster at least on the directsync PR. Thanks @michalskrivanek for sticking with this!

michalskrivanek:
I can drop `unsafe` entirely. I fail to see how is that *any* better for users, the arguments were presented repeatedly, the only argument against was that we don't want users to shoot themselves in the foot. By pushing them to plugins not allowing us to block the _only_ more unsafe scenario than `writeback`, we're making sure that when they shoot they won't miss.

michalskrivanek:
dropped

akalenyu (approved):

vladikr:
Thank you!
/approve

Issue #293, opened 2026-04-23
VEP 293: Extend Hot Plug APIs


rthallisey:
### Overview
The goal for this VEP is to make hot plug more extensible than it is today.  Solving this will allow for end users to replace parts of the hot plug solution with out-of-tree custom solutions, in order to handle various edge cases.

### High-level Approach
Volume hot plug breaks down to 3 steps:
1. Volume delivery
2. Volume attach (hot plug)
3. Volume detach (hot unplug)

KubeVirt provides 1 API to control this flow, however this can be broken down into 3 separate APIs.

### All Volumes can be Unplugged
Second, hot plugged volumes are tracked separately than "regular" volumes.  As part of this change, remove this distinction allowing users more fine-grained control over the volumes and their lifecycle.

**Primary contact (assignee)**:

/assign @Sreeja1725 

**Responsible SIGs**:

SIG-storage

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig scale

**Enhancement link**:

<tbd>

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

akalenyu:
Just to summarize the sig-storage call, we tend to think the end goal will be to get hotplug support natively from k8s, which in turn would also allow the granularity outlined here.
That said, we will wait for the VEP and will feel much better about proceeding with it knowing that a k8s KEP path is being pursued in parallel. Thank you @Sreeja1725 @alaypatel07 for tackling this!

xpivarc:
@rthallisey Please add VEP link

rthallisey:
@xpivarc added

PR #328, opened 2026-06-08
VEP 293: Extend Hotplug APIs

Sreeja1725:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/293
**SIG label**:  /sig storage

### What this PR does

Adds an ability to decompose hotplugging.

### DIFF
--- a/veps/sig-storage/node-local-hotplug.md
+++ b/veps/sig-storage/node-local-hotplug.md
@@ -1,0 +1,452 @@
+# VEP #293: Extend Hotplug APIs
+
+## VEP Status Metadata
+
+### Target releases
+
+-  This VEP targets alpha for version: v1.10
+-  This VEP targets beta for version: TBD
+-  This VEP targets GA for version: TBD
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [+] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [] (R) Alpha target version is explicitly mentioned and approved
+- [] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes a node-local gRPC service inside virt-handler that drives disk attach and detach for a running VirtualMachineInstance. The service splits the existing monolithic hotplug flow into three explicit phases — Delivery, Attach, Detach.
+
+It does not change KubeVirt's existing AddVolume subresource or the HotplugVolume API marker; both remain. The proposal is additive.
+
+## Motivation
+
+KubeVirt hotplug today works but is monolithic and Pod-heavy for PVC-backed volumes. The mechanism works by:
+
+1. **Launching an attachment Pod** on the target node, which mounts the PVC so the volume is available on the host.
+2. **Manipulating bind mounts on the host** to expose that volume into the `virt-launcher` Pod's mount namespace, since Kubernetes does not allow mutating a Pod's `volumes` spec after creation.
+3. **Hotplugging the disk into the VM** via libvirt/QEMU once the volume is visible inside `virt-launcher`.
+
+While this design successfully leverages standard Kubernetes components and reuses as many Kubernetes apis as possible to limit the burden on kubevirt to maintain custom code, the end-to-end flow is hard-wired inside KubeVirt and exposed through a single API surface. Every step — volume delivery, attachment into the VM, and detachment — is driven by the same built-in controllers, which makes it difficult to customize or replace any individual stage for workloads that have specialized requirements.
+
+This VEP proposes an extension of the hotplug APIs that decomposes the current monolithic flow into discrete, independently replaceable stages.
+
+## Goals
+
+  * **Make volume hotplug extensible** by allowing end users and out-of-tree components to replace individual stages of the hotplug flow with custom implementations, so that various edge cases (e.g., node-local hotplug, specialized delivery mechanisms) can be handled without changes to KubeVirt core.
+  * **Decompose the single hotplug API into three distinct APIs** aligned with the three logical stages of the flow: volume delivery, volume attach (hot plug), and volume detach (hot unplug).
+  * Provide a node-local gRPC API for attach and detach.
+  * Allow Phase-1 volume delivery to be pluggable — raw host paths, attachment pods for PVC/DV, CSI mounts, hostPath, future implementations.
+
+## Non Goals
+* Replacing the AddVolume subresource. It continues to serve callers that want one-shot delivery+attach.
+* Implementing PVC / DataVolume Phase-1 delivery in this VEP. That belongs in a follow-on VEP or is upto the user. 
+* Modifying libvirt attach/detach semantics. We rely entirely on the existing converter + syncDisks path.
+* Live migration of node-local volumes.
+* Replacing the current hot plug solution using pods.
+
+## Definition of Users
+
+* **Cluster Administrators / Debug Tooling:** Attach a one-off block device or file to a running VM for diagnostics, data recovery, or ad-hoc operational tasks without restarting the guest.
+* **CSI Driver Sidecars:** After completing `NodeStageVolume`, wire the already-staged path directly into a running VM via the node-local API, avoiding a round-trip through `virt-controller` and an attachment Pod.
+  * **On-Node DaemonSets:** Expose a host device or file as a disk on a co-located VMI — for example, to surface node-local hardware, host-managed state, or operator-provided artifacts to the VM.
+
+## User Stories
+
+* As a KubeVirt developer or storage integrator, I want a low-latency hotplug path that bypasses the attachment-Pod lifecycle, so that attaching and detaching disks on a running VM completes quickly and does not generate apiserver/etcd write load proportional to the number of hotplug operations.
+* As a cluster administrator, I want CSI sidecars or on-node DaemonSets that already have a staged volume on the node to bind that staged path into a running VM through a node-local API, so that node-local components can integrate with VM storage without a round-trip through `virt-controller` or an attachment Pod.
+* As an operator debugging a running VM, I want to expose a host device or file to the VM as a disk directly, without creating an attachment Pod or provisioning a PVC, so that I can collect diagnostics, inject a rescue image, or surface a one-off device without apiserver side effects or guest restarts.
+* As a vendor delivering vendor-specific block devices to nodes via my own on-node controller, I want to attach the device to a running VMI as a libvirt disk without forcing the device through a PVC
+
+# Repos
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+# Design
+
+## High-level Approach
+
+Volume hotplug today is a single, KubeVirt-managed pipeline exposed through one API. Conceptually, however, it breaks down into three independent steps:
+
+1. **Volume delivery** — making the underlying storage available on the target node (today, by launching an attachment Pod that binds the PVC and exposes it on the host).
+2. **Volume attach (hot plug)** — surfacing the delivered volume inside the running `virt-launcher` Pod and hotplugging the corresponding disk into the VM via libvirt/QEMU.
+3. **Volume detach (hot unplug)** — reversing the above: detaching the disk from the VM, removing it from the `virt-launcher` Pod, and tearing down any delivery resources.
+
+This VEP splits KubeVirt's single hotplug API into three separate APIs — one per stage — so that each stage can evolve independently and can be driven (or replaced) by an out-of-tree controller. For example, a node-local hotplug implementation can substitute a lightweight delivery step that skips the attachment Pod altogether, while reusing the standard attach and detach paths.
+
+### Phase Pipeline
+
+The end-to-end flow is modeled as a three-phase pipeline. Phase 1 (Delivery) is pluggable and may run off-node or on-node; Phases 2 and 3 (Attach / Detach) are the node-local surface introduced by this VEP and run inside `virt-handler`:
+
+```text
++-------------------------+     +-------------------------+     +-------------------------+
+| Phase 1: Volume Delivery|     | Phase 2: Volume Attach  |     | Phase 3: Volume Detach  |
+| (pluggable)             | --> |                         | --> |                         |
+| off-node OR on-node     |     |                         |     |                         |
+|                         |     | AttachDevice RPC        |     | DetachDevice RPC        |
+| Produces:               |     |  - mknod or bind-mount  |     |  - patch spec           |
+|  - absolute host path   |     |    into virt-launcher   |     |    (remove device)      |
+|                         |     |  - patch spec           |     |  - wait libvirt detach  |
+|                         |     |    (NodeLocalDevice)    |     |  - cleanup, conditional |
+|                         |     |  - patch status         |     |    on delivery source   |
+| Examples:               |     |    (HotplugVolMounted)  |     |                         |
+|  - raw (admin-staged)   |     |                         |     |                         |
+|  - attachment Pod       |     | virt-handler reconcile  |     | virt-handler reconcile  |
+|  - CSI                  |     | runs syncDisks -->      |     | runs syncDisks -->      |
+|  - DataVolume           |     | libvirt AttachDevice    |     | libvirt DetachDevice    |
+|  - hostPath / local PV  |     |                         |     |                         |
++-------------------------+     +-------------------------+     +-------------------------+
+```
+
+#### Phase 1 — Volume Delivery 
+
+Delivery is the only phase that can run either off-node (controllers, attachment Pods, CSI provisioners) or on-node (DaemonSets, sidecars), and it is intentionally pluggable. Its only contract with the rest of the pipeline is a single output consumed by Phase 2: an **absolute host path** where the backing storage is visible on the target node. Phase 2 `stat`s the path to determine whether it is a block device or a regular file, and picks the corresponding staging step (`mknod` or bind-mount) accordingly.
+
+Reference implementations include raw admin-staged paths, the existing attachment-Pod flow (preserved for PVC/DataVolume-backed hotplug), CSI-staged paths, and `hostPath` / local PVs.
+
+#### Phase 2 — Attach 
+
+Phase 2 is the node-local surface added by this VEP. A caller invokes `AttachDevice` on `virt-handler`'s gRPC server with the Phase-1 output. `virt-handler` then:
+
+1. Records intent — adds the volume to `spec.volumes`, the disk to `spec.domain.devices.disks`, and creates the `status.volumeStatus[i]` entry with the `Bound` condition set `True, reason: VolumeBound` via `buildAttachIntentPatch`.
+2. Creates the device inside the target `virt-launcher` Pod — `mknod` for block, bind-mount for file — so the disk is visible from the VM's viewpoint.
+3. Flips the `MountedToPod` condition to `True, reason: MountedToPod` via `buildAttachMountedPatch` and returns success on the RPC. The eventual `phase: Ready` transition happens asynchronously when libvirt reports the disk live, via the existing reconcile (see [Status Conditions](#status-conditions) below).
+
+The existing `virt-handler` reconcile loop picks up the spec change and drives the libvirt `AttachDevice` call via the unchanged `syncDisks` converter path — no libvirt semantics are altered.
+
+#### Phase 3 — Detach 
+
+Phase 3 mirrors Phase 2. A caller invokes `DetachDevice`, and `virt-handler`:
+
+1. Marks the entry as `phase: Detaching, reason: VolumeDetaching` and removes the volume / disk from the VMI spec in the same patch (`applyDetachPatch`). The `VolumeStatus` entry remains visible during the detach so observers can tell it is in progress.
+2. Waits for libvirt to complete the detach (again via the existing `syncDisks` path, which issues `DetachDevice` to libvirt).
+3. Cleans up the node-side artifacts (bind mounts, device nodes) and — conditional on the delivery source's declared lifecycle owner — optionally tears down Phase-1 resources, then removes the `VolumeStatus` entry.
+
+### Status Conditions
+
+The existing `VolumeStatus` shape (`phase`, `reason`, `message`, `target`) is **not** changed. `phase: Ready, reason: VolumeReady, target: vd<x>` continues to be written by the existing `updateVolumeStatusesFromDomain` reconcile when libvirt reports the disk live, exactly as it does today for attachment-pod volumes. On the way out, the `DetachDevice` RPC writes `phase: Detaching, reason: VolumeDetaching` so observers can distinguish a volume that is being unplugged from one that is still live:
+
+```yaml
+volumeStatus:
+- name: scratch
+  phase: Ready
+  reason: VolumeReady
+  message: "Successfully attach hotplugged volume scratch to VM"
+  target: vdb
+  hotplugVolume: {}
+```
+
+To track the node-local stages — the work the `AttachDevice` RPC does before libvirt is involved — a `conditions []VolumeCondition` field is added to `VolumeStatus`. Two condition types are defined; both are written by the RPC, neither replaces or duplicates `phase`:
+
+| Condition type | Set `True` by | When | `reason` | Notes |
+|---|---|---|---|---|
+| `Bound` | `AttachDevice` RPC, before staging (`buildAttachIntentPatch`) | Validation passed; volume + disk added to spec; about to `mknod` / bind-mount | `VolumeBound` | Intent recorded so a crash here does not lose the request — recovery resumes from the existing entry. |
+| `MountedToPod` | `AttachDevice` RPC, after staging (`buildAttachMountedPatch`) | `mknod` or bind-mount into `virt-launcher` succeeded; libvirt attach not yet started | `MountedToPod` | The RPC returns `success: true` at this point; libvirt attach proceeds asynchronously, after which `phase` flips to `Ready` via the existing reconcile. |
+
+The full lifecycle (existing `phase` plus the new `Bound` / `MountedToPod` conditions) progresses through the following steps:
+
+1. **`AttachDevice` RPC accepted** → `buildAttachIntentPatch` creates the entry with `Bound = True, reason: VolumeBound, message: "host path validated; bind-mount pending"`.
+2. **Staging completes** (`MountFile` / `MountBlock` succeeds) → `buildAttachMountedPatch` sets `MountedToPod = True, reason: MountedToPod, message: "device exposed under launcher hotplug-disks dir"`. The RPC returns `success: true` here.
+3. **Libvirt attach completes asynchronously** — `virt-handler` reconcile → `SyncVirtualMachine` → `virt-launcher` → `syncDisks` → `libvirt.AttachDeviceFlags` succeeds.
+4. **Existing reconcile flips `phase`** — `updateVolumeStatusesFromDomain` sees `Target != ""` in the domain XML and writes `phase: Ready, reason: VolumeReady, target: vd<x>` (unchanged code path).
+5. **`DetachDevice` RPC accepted** → `applyDetachPatch` removes the volume from `spec.volumes` and the disk from `spec.domain.devices.disks`, and updates the `VolumeStatus` entry to `phase: Detaching, reason: VolumeDetaching, message: "detach in progress; awaiting libvirt"` (the entry stays visible during the detach).
+6. **Libvirt detach completes asynchronously** — same reconcile chain calls `libvirt.DetachDeviceFlags`. The domain informer no longer reports the disk and `Service.waitDomainDiskAbsent` unblocks.
+7. **Host-side staging removed and entry deleted** — `UnmountFile` / `UnmountBlock` tears down the bind-mount or device node and the cgroup allow rule, and the `VolumeStatus` entry is removed from `vmi.Status.VolumeStatus`.
+
+## Volume Delivery
+
+Before a disk can be attached to a running VM, the storage backing it has to be present on the node — visible either as a block device or as a file at some path. This VEP does not define how that storage gets there. The user picks one of two approaches, depending on where the storage lives:
+
+* **Use a controller that provisions the volume for you.** If the storage does not already exist on the node — for example, a PVC that needs to be created, or a remote volume that needs to be attached and mounted — the user runs a controller to create the volume and make it available on the node. This can be a standard CSI driver, or a custom controller that does the same job. Once the volume is visible on the node at a known path, the controller calls `AttachDevice` to hand it off to the VM.
+
+* **Use a volume that is already on the node.** If the storage is already there — a host directory, a loopback file, a local disk, a vendor-managed device — no provisioning is needed. It can be given to the VM directly via `AttachDevice`.
+
+Both approaches use the same attach path once the volume is delivered. The choice is about where the storage comes from; this VEP does not prefer one over the other.
+
+## API Examples
+
+New `VolumeSource`
+
+```go
+// staging/src/kubevirt.io/api/core/v1/schema.go
+type VolumeSource struct {
+    // ... existing fields ...
+    NodeLocalDevice *NodeLocalDeviceSource `json:"nodeLocalDevice,omitempty"`
+}
+type NodeLocalDeviceSource struct {
+    Path   string                `json:"path"`                 // required: host path
+    Type   *string               `json:"type,omitempty"`      // optional
+}
+```
+The volume source captures only what the converter needs to build a libvirt disk source. It is set by the gRPC service — users do not author it directly.
+
+gRPC Service
+
+```go
+service NodeLocalHotplug {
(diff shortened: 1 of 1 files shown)


veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,488 @@
+# VEP #293: Extend Hotplug APIs
+
+## VEP Status Metadata
+
+### Target releases
+
+-  This VEP targets alpha for version:
```
rthallisey: Add a target for alpha.  Probably 1.10

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,488 @@
...
+2. **Manipulating bind mounts on the host** to expose that volume into the `virt-launcher` Pod's mount namespace, since Kubernetes does not allow mutating a Pod's `volumes` spec after creation.
+3. **Hotplugging the disk into the VM** via libvirt/QEMU once the volume is visible inside `virt-launcher`.
+
+While this design successfully leverages standard Kubernetes components, the end-to-end flow is hard-wired inside KubeVirt and exposed through a single API surface. Every step — volume delivery, attachment into the VM, and detachment — is driven by the same built-in controllers, which makes it difficult to customize or replace any individual stage for workloads that have specialized requirements (for example, node-local volumes that do not need an attachment Pod, or environments that want to plug in a custom volume-delivery mechanism). 
+
+**Attachment-Pod hotplug is slow and apiserver-heavy.** Every hotplug and unplug is driven through a full Pod lifecycle — create, schedule, kubelet start, CSI stage/mount, Ready, bind-mount, libvirt hotplug (and the reverse on unplug). That design has several classes of overhead:
```
rthallisey: The current path does not provide a path for low latency volume attachment.  I would agree.  However there are benefits to the current solution, like not having to manage the CSI stage/mount layer.  So I would word this a little differently because we still need to support current solution, something like:
"the current solution reuses as many Kubernetes apis as possible to limit the burden on kubevirt to maintain custom code.  This VEP aims to not change kubevirt's maintenance burden, but to provide extensions points for the end user at each stage of hotplug.  This will unlock new use cases for admins.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,488 @@
...
+  * **Unify the volume lifecycle** by removing the distinction between hot-plugged volumes and "regular" volumes, so that any volume attached to a running VM can be plugged and unplugged through the same APIs.
+  * Provide a node-local gRPC API for attach and detach.
+  * Allow Phase-1 volume delivery to be pluggable — raw host paths, attachment pods for PVC/DV, CSI mounts, hostPath, future implementations.
+
+## Non Goals
+* Replacing the AddVolume subresource. It continues to serve callers that want one-shot delivery+attach.
+* Implementing PVC / DataVolume Phase-1 delivery in this VEP. That belongs in a follow-on VEP. 
+* Modifying libvirt attach/detach semantics. We rely entirely on the existing converter + syncDisks path.
+* Live migration of node-local volumes.
+
```
rthallisey, line 51: Replacing the current hot plug solution using pods.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,488 @@
...
+## User Stories
+
+* As a KubeVirt developer or storage integrator, I want a low-latency hotplug path that bypasses the attachment-Pod lifecycle, so that attaching and detaching disks on a running VM completes quickly and does not generate apiserver/etcd write load proportional to the number of hotplug operations.
+* As a cluster administrator, I want CSI sidecars or on-node DaemonSets that already have a staged volume on the node to bind that staged path into a running VM through a node-local API, so that node-local components can integrate with VM storage without a round-trip through `virt-controller` or an attachment Pod.
+* As an operator debugging a running VM, I want to expose a host device or file to the VM as a disk directly, without creating an attachment Pod or provisioning a PVC, so that I can collect diagnostics, inject a rescue image, or surface a one-off device without apiserver side effects or guest restarts.
+* As a vendor delivering vendor-specific block devices to nodes via my own on-node controller, I want to attach the device to a running VMI as a libvirt disk without forcing the device through a PVC
+
```
rthallisey, line 64: As a cluster admin, I want to be able to lifecycle my VM without having to think about managing my hot plugged volumes differently than non hot plugged volumes.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,488 @@
...
+This VEP splits KubeVirt's single hotplug API into three separate APIs — one per stage — so that each stage can evolve independently and can be driven (or replaced) by an out-of-tree controller. For example, a node-local hotplug implementation can substitute a lightweight delivery step that skips the attachment Pod altogether, while reusing the standard attach and detach paths.
+
+### Phase Pipeline
+
+The end-to-end flow is modeled as a three-phase pipeline. Phase 1 (Delivery) is pluggable and may run off-node or on-node; Phases 2 and 3 (Attach / Detach) are the node-local surface introduced by this VEP and run inside `virt-handler`:
+
+```text
++-------------------------+     +-------------------------+     +-------------------------+
+| Phase 1: Delivery       |     | Phase 2: Attach         |     | Phase 3: Detach         |
+| (pluggable)             | --> | (this VEP)              | --> | (this VEP)              |
```
rthallisey: (pluggable) and (this VEP) are confusing to me.  This VEP enables Phase 1 to be pluggable, so how is that not also included in this VEP?

Sreeja1725: Targeting Phase-1 in a separate VEP. Yeah removed (this VEP) as it is confusing.

veps/sig-storage/node-local-hotplug.md
Sreeja1725: Agree, updated

mhenriks:
@Sreeja1725 Thanks for submitting this! It appears to be a backdoor API into virt-handler to add/remove volumes. Interesting idea but it seems to be satisfying a very niche use case that I'm not sure has broad appeal in the community. Can you imagine an ecosystem building up around this grpc API? What would it look like? 

Can we maybe build on existing APIs to get you some of what you i.e. faster hotplug when you need it? Maybe the existing hotplug API can be somehow split up such that user can add a volume with `desiredPhase: AttachedToNode` or `MountedToPod`? Then when it is needed by the VM set `desiredPhase: Ready` which will hotplug the volume into the VM?

michalskrivanek:
as an argument for tighter/integrated control I also want to mention hotplug has security implications.
The "regular" volumes are handled via standard/common k8s mechanisms, while the hotplugged volumes are where we diverge significantly, basically replacing/extending the permission model in a not-so-nice way where we modify the pod permissions on the fly (currently via overwriting pod's device access filters). This is inherently risky as whatever mistake we make can lead to priviledge escalation.
I'm not saying this is a wrong idea, just that we need to be extra careful here, and that it might be a good argument _against_ generic extensibility.

rthallisey:
> This is inherently risky as whatever mistake we make can lead to priviledge escalation.
>
@michalskrivanek as a slight push back to this, KubeVirt has managed volumes very well. If kubevirt's hot plug wasn't stable or safe, the proposal wouldn't be able to build on top of those mechanism. So we're assuming we can reuse that work.  Agreed that we need to be very careful.

awels:
There are a variety of issues with the current hotplug implementation, see the variety of PRs to solve some of those issues (like the problem when a hotplugged volume is tied to a node for whatever reason, and the VM restarts, and it gets scheduled somewhere else causing the hotplugging to fail). That being said the mechanism is in wide use without major reported issues, so as Ryan said it is stable and safe from what we can tell.

I want to also point out that we are moving away from the addvolume/removevolume endpoints in favor of a declarative approach where you just modify the VM spec to indicate you want to hotplug a volume. This is useful for things like inserting/ejecting a cdrom image for instance. It does not appear this VEP is in conflict with that approach.

I don't mind the approach of splitting the entire flow into three separate steps that can be overriden by specific implementations if needed. This is likely useful if different architectures have different requirements from libvirt/qemu when generating the domain XML, or in this case if you want an alternate implementation of attaching a volume to a node.

I would still worry that the alternate implementation would violate the kubernetes management of volumes, but if the user of that implementation is willing to take that risk, that is on them. The default implementation uses the attachment pods to ensure that we can properly account for the volumes from a kubernetes perspective.

I want to point out that the asymetric nature of hotplugged volumes vs normal volumes is entirely intentional. As we all known pods are immutable so there is no mechanism to dynamically add/remove a volume from a pod. But since we are bypassing the immutability of a pod by doing some things in the background, we should really keep the asymetric nature to indicate hey this volume is actually doing something weird, be careful what you do with it.

That being said, I don't like the idea of detaching a 'normal' volume at all. Since there is no actual way to remove the volume from the virt-launcher pod. We can make it disappear from the VM domain XML and no longer be visible inside the guest VM, but we can't use that volume anywhere else since it is still attached to the virt-launcher pod.

TL:DR;
I like the splitting of the process into the 3 distinct steps, not a fan of detaching normal volumes because it won't really do anything and confuse the user as to why they can't use the detached volume somewhere else.

rthallisey:
@mhenriks - the ecosystem around the grpc API would be for the advanced user. I don't think the target is broad appeal.  

Many of the advanced use cases require breaking further away from K8s APIs. If kubevirt were to invest in supporting these, I think the cost will out weight the gain.

This proposal focuses on protecting kubevirt from veering to far from K8s. I think that's an advantage of this approach, while still providing the ability to implement these features.

The long term solution is still to have K8s support hot plug. Having that solution would build the ecosystem and can replace this feature eventually.  However, we don't when that will be. 

>That being said, I don't like the idea of detaching a 'normal' volume at all. Since there is no actual way to remove the volume from the virt-launcher pod. We can make it disappear from the VM domain XML and no longer be visible inside the guest VM, but we can't use that volume anywhere else since it is still attached to the virt-launcher pod.
>
@awels that's fair.  I think an alternative idea would be keeping two separate fields, but allow a hotplugged volume be added in the initial VMI spec.  This is the case at CREATE time where you expect to hotunplug the same volume later.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,458 @@
...
+|  - hostPath / local PV  |     |                         |     |                         |
++-------------------------+     +-------------------------+     +-------------------------+
+```
+
+#### Phase 1 — Volume Delivery 
+
+Delivery is the only phase that can run either off-node (controllers, attachment Pods, CSI provisioners) or on-node (DaemonSets, sidecars), and it is intentionally pluggable. Its only contract with the rest of the pipeline is an output tuple consumed by Phase 2:
+
+* **Absolute host path** — where the backing storage is visible on the target node.
+* **Format** — `block` or `file`, so Phase 2 knows whether to `mknod` or bind-mount.
```
awels: Do we really need this information to be passed from phase to phase? Based on the file path we should be able to determine if the file is file or block. I guess maybe to eliminate the code to infer the difference it is simpler to pass that information. But I don't think it is technically needed.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,458 @@
...
+sequenceDiagram
+    client ->> NLH gRPC server: AttachDevice(volName, device_path, format ...)
+    NLH gRPC server ->> NLH gRPC server: Validate (VMI running, path, bus)
+    NLH gRPC server ->> NLH gRPC server: checks if AlreadyStaged?
+    NLH gRPC server ->> K8S apiserver: Update spec (volume, disk)
+    NLH gRPC server ->> K8S apiserver: Update status conditions (set phase to volumeBound)
+    NLH gRPC server ->> NLH gRPC server: mknod (block) OR bind-mount (file)into hotplug-disks dir
+    NLH gRPC server ->> K8S apiserver: Update status (phase to volume Mounted)
+    NLH gRPC server ->> client: AttachDeviceResponse{success:true}
+    K8S apiserver ->> VirtaHandler: Reconciles(on VMI spec change)
```
awels:
```suggestion
    K8S apiserver ->> VirtHandler: Reconciles(on VMI spec change)
```

awels (commented):
I am good with this, one question and a typo. Neither are blocking for accepting the VEP IMO.

veps/sig-storage/node-local-hotplug.md
Sreeja1725: I agree with you, format is not technically needed.
I have added it thinking it will be self-describing and can fail fast without deriving it. But yeah removed it.

awels: Yeah it is always the question should we keep the API surface as small as possible or make it a little more explicit, to allow in this case faster failures. I think smaller in this case is slightly better, but as I said I would not block it either way.

awels:
> @awels that's fair. I think an alternative idea would be keeping two separate fields, but allow a hotplugged volume be added in the initial VMI spec. This is the case at CREATE time where you expect to hotunplug the same volume later.

Note you can do that right now. There a is a hotpluggable field on volumes that indicates you want it to be hotplugged at the startup of the VM, and thus be able to unplug it while the VM is still running.

mhenriks:
I'm really not keen on an approach that has no integration points with the k8s api. 

Have you thought about going with the VEP #190 plugin model? Could add a couple new hooks - `PreVolumeAttach` and `PostVolumeDetach`. The user adds a hotpluggable volume to the VM as they do today, but instead of creating an attachment Pod, `PreVolumeAttach` is called and stages the volume somewhere on the node.

This would avoid a new standalone gRPC service, keep the user-facing hotplug API unchanged, and build on existing plugin infrastructure rather than running parallel to it.

I think the pluging approach is worthy of serious consideration and should at a minimum be included in the "alternatives" section

mhenriks:
Until plugin alternative is discussed
/hold

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,452 @@
...
+
+This VEP proposes an extension of the hotplug APIs that decomposes the current monolithic flow into discrete, independently replaceable stages.
+
+## Goals
+
+  * **Make volume hotplug extensible** by allowing end users and out-of-tree components to replace individual stages of the hotplug flow with custom implementations, so that various edge cases (e.g., node-local hotplug, specialized delivery mechanisms) can be handled without changes to KubeVirt core.
+  * **Decompose the single hotplug API into three distinct APIs** aligned with the three logical stages of the flow: volume delivery, volume attach (hot plug), and volume detach (hot unplug).
+  * Provide a node-local gRPC API for attach and detach.
+  * Allow Phase-1 volume delivery to be pluggable — raw host paths, attachment pods for PVC/DV, CSI mounts, hostPath, future implementations.
```
EdDev, line 43: These goals seem to be implementations/solutions. Perhaps the last one is closer to a goal if the "phase" is removed.

veps/sig-storage/node-local-hotplug.md
```diff
@@ -0,0 +1,452 @@
...
+
+Delivery is the only phase that can run either off-node (controllers, attachment Pods, CSI provisioners) or on-node (DaemonSets, sidecars), and it is intentionally pluggable. Its only contract with the rest of the pipeline is a single output consumed by Phase 2: an **absolute host path** where the backing storage is visible on the target node. Phase 2 `stat`s the path to determine whether it is a block device or a regular file, and picks the corresponding staging step (`mknod` or bind-mount) accordingly.
+
+Reference implementations include raw admin-staged paths, the existing attachment-Pod flow (preserved for PVC/DataVolume-backed hotplug), CSI-staged paths, and `hostPath` / local PVs.
+
+#### Phase 2 — Attach 
+
+Phase 2 is the node-local surface added by this VEP. A caller invokes `AttachDevice` on `virt-handler`'s gRPC server with the Phase-1 output. `virt-handler` then:
+
+1. Records intent — adds the volume to `spec.volumes`, the disk to `spec.domain.devices.disks`, and creates the `status.volumeStatus[i]` entry with the `Bound` condition set `True, reason: VolumeBound` via `buildAttachIntentPatch`.
```
EdDev, line 115: IIUC the suggestion is for virt-handler to expose a gRPC API at the node level and allow this attach (and then detach) operation, which in turn is changing the VMI spec.
If this is indeed the case, then we have two main issues here which seem like blockers to me:
- Bypassing and possibly colliding with the VMI (and perhaps VM) controller flow. 
- Mutating the VMI spec by virt-handler.

Perhaps the first one can be solved with some complex logic, but I do not see how the second one can.

EdDev (changes requested):
I like and support the idea to extend Kubevirt functionality in general.

The work on extending Kubevirt has seen progress lately, currently through a plugin enhancement. It collects scenarios, and this seem a good one to add to it.

For the suggested approach, I would suggest to take existing hook mechanics from Kubernetes and see which one fits the storage hotplug case.
The current K8S runtime extendable solutions are: Device Plugin, DRA and NRI.
They differ from the approach introduced by this VEP in their integration point. The extensions are introduced in the middle or end of the flow, with low or no impact on how users express the desired state. In other words, one would hotplug a new storage device in a declarative manner as today and under the hood, "drivers" can implement that desire in their preferred way.

Extending functionality by allowing to express it through two different APIs is something I think we should avoid.

rthallisey:
Pulling from VEP 190...
>This proposal is building on the philosophy of frameworks like NRI (Node Resource Interface). It introduces a structured hooking-based plugin mechanism to replace the existing "sidecar hook" model. The new mechanism will enable a safer, holistic, maintainable and powerful way of integrating out-of-tree functionality into KubeVirt.
>
I think VEP 190 is solving the same use cases.  If there was an NRI-like framework, it would fill the same role as the gRPC server this KEP proposes. The use case is an out-of-tree extension, highly custom, pertaining to volume management. That way, when a VMI spec has a `hotplug` volume, custom plugin code is run to fulfill the work while kubevirt maintains the accounting.

@Sreeja1725 can you take a stab at designing with the structured plugin method?

Sreeja1725:
Thanks @mhenriks @EdDev @rthallisey - for the feedback. it's clear and agree with the direction. Will explore and redesign the VEP around VEP 190.

akalenyu (commented):
The plugin approach seems like the right path to take to achieve the goals outlined here 🙏 

Possibly unrelated, but I do wonder if you ever explored alternatives to achieve performance gains in the hotplug path that would get them closer to a no pod implementation.

For example, using a clean `pause` container for hotplug purposes instead of a full blown virt-launcher that does some socket manipulation (today's impl.). I think there is nonzero gain from going from today's hotplug pod to a real placeholder pod. 

Another (admittedly, uglier) option is something like ~~KWOK~~ to bypass the "real" container creation for that pod.
EDIT: KWOK doesn't really fit, I think something like a dummy `RuntimeClass` may be worth exploring
```yaml
apiVersion: node.k8s.io/v1
kind: RuntimeClass
metadata:
  name: no-container
handler: fake-runtime
```
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: dummy-pod
spec:
  runtimeClassName: no-container
```

rthallisey:
@akalenyu in the proposal, Sreeja would add a flexible volume attach option.  This would allow volumes to reach nodes using different methods than the current one (attachment pod).  In theory you can use any any method, including no Pod at all.

EdDev:
> Thanks @mhenriks @EdDev @rthallisey - for the feedback. it's clear and agree with the direction. Will explore and redesign the VEP around VEP 190.

:+1: 
I'm going to get into the plugin effort next week probably, it has potential to solve many requests we see coming lately.
I expect it to keep on improving and change, per the scenarios and needs.

akalenyu:
> @akalenyu in the proposal, Sreeja would add a flexible volume attach option. This would allow volumes to reach nodes using different methods than the current one (attachment pod). In theory you can use any any method, including no Pod at all.

Yes I understand, I am suggesting that some of the delta between current..desired can be bridged without the VEP, or at least I thought this angle was interesting enough to bring up. BTW, wouldn't you just opt out of any particular volume attach steps and just have a hostpath pv sitting there ready? a reference end goal impl. would be nice to see if it's not a secret :stuck_out_tongue:

akalenyu:
Is this still targeted 1.10 with plugin approach? Just note VEP freeze coming up next week

Sreeja1725:
I have explored the structured plugin integration, and here are my thoughts.

With the hook-based approach, the attach path still has to go through all the virt-components — virt-api, virt-controller, API spec/status patching — and then virt-handler reconciles and does the hot-attach. This adds unnecessary latency for a node-local attach where the device is already prepared on the node.

The hook-based path makes sense for extending the normal attach flow (PVC-backed volumes, attachment pods), but not for node-local volumes where the goal is to eliminate the API roundtrip entirely — the reconcile loop exists to manage volume lifecycle that node-local volumes simply don't have.

The current plugin model is unidirectional: virt-handler calls the plugin. For node-local-hotplug may be we want it in reverse - the plugin calls into virt-handler to trigger an immediate attach. The plugin writes its intent (volume name, device path, bus) to the socket or as an annotation and calls a minimal `TriggerReconcile(vmiName)`. Virt-handler reads the intent and runs the attach immediately using its existing code, owning everything from there — validation, idempotency, libvirt attach, and status update. and the k8s api is only touched after the attach is done, as an async statys write by virt-handler and not on the critical path.

cc: @rthallisey

rthallisey:
@iholder101 do you see a path for this to work with VEP 190?  The other alternative we can explore is VEP 311.

mhenriks:
> the goal is to eliminate the API roundtrip entirely

If this is the goal then I guess I am fundamentally opposed to this VEP. I think all resource updates should driven by updates done through the k8s apiserver. By updating the VM resources directly or via subresource api

rthallisey:
This proposal focuses on making the hot plug steps pluggable. Volume attach, hotplug, and hot unplug can be extended, so out of tree solutions can implement these steps in their own way.  I feel we're aligned on the goal.

The challenge is the implementation of that plugin mechanism in order to satisfy some use cases. Sreeja highlighted that is the current sticking point here. One of the use cases we've been discussing is: a high performance application requires a sub 1-second volume attach operation for a local volume. The volume is native to the node and the VM has been running for some unspecified period of time. At some point, the application desires a volume and requests it. Instead of reaching out to the API server and back down to the node, the request is made to the local virt-handler. 

If I were to simply the idea, it would use the same steps that exist today.  The virt-handler runs `syncDisks`, libvirt attaches the disk, then virt-handle updates ` VMI.status.volumeStatus`.  The change is to how `syncDisks` gets called. Sreeja will work on setting up a community meeting to discuss further.

vladikr:
Hey everyone.
Sorry, i'm late reviewing this VEP, but this problem is not new to me as we've discussed this multiple times with @alaypatel07 


> > the goal is to eliminate the API roundtrip entirely
> 
> If this is the goal then I guess I am fundamentally opposed to this VEP. I think all resource updates should driven by updates done through the k8s apiserver. By updating the VM resources directly or via subresource api

I agree with @mhenriks . If the goal is to take attach/detach off the apiserver path, I am not in favor of this. 
Disks on a VM should appear because the VM/VMI spec changed, not because something on the node called virt-handler.

On the other hand, I'm very much in favor of the node local volume usecase. 
I do not think the apiserver round-trip is what makes today’s path slow. 
From my pov, the slow part is the extra Pod we create to get the disk onto the node.
 I mean, once the vmi spec says that the volume should be already there and the path is local to virt-handler then the `syncDisks` is a local reconcile. 


**What I am proposing instead**

We should keep the intent on the VM/VMI spec. The request is still “add this volume to the VM.” 
The only part we need to change is how the disk gets onto the node.

For a file that is already on the node, use HostDisk we already have. 

```yaml
volumes:
- name: rescue
  hostDisk:
    path: /var/run/vendor/rescue.img
    type: Disk   # file must already exist
```

Someone updates the VM / VMI, the same way declarative hotplug already works. 
A user can do it or node agent or controller, after it has created the file under an allowlisted path. 
That agent already talks to the apiserver; it does not need a new socket on virt-handler.

**What KubeVirt does with that**

1. virt-controller sees `hostDisk` and skips the helper Pod.
2. virt-handler on the VM’s node sees the new volume, checks that `path` exists, exposes it to the VM Pod, and uses the same libvirt attach as today.
3. It updates `volumeStatus` when the disk is in the guest.

Remove the volume from the spec to unplug. If virt-handler or the VM Pod restarts, the volume is still in spec, so virt-handler attaches again.

At VM create time, HostDisk stays as it is today (a hostPath on the VM Pod). 
That kind is still not unpluggable. Only HostDisk that was added while the VM is already running would use the hotplug path above.

This needs the existing HostDisk feature gate, plus an allowlist on the KubeVirt CR (which ServiceAccounts may set `hostDisk`, and which path prefixes). 
Treat it as hostPath. 
And we should block live-migration for this. at least for now.

Let's leave the PVCs alone in this VEP...
If we later skip the helper Pod for a PVC, the rule has to be equally explicit, for example a local/block PV whose path is already on the PV object and whose node matches the VM. 
I would not put that in the first version.

vladikr:
@rthallisey , I wonder if there is another way.
If this is really about moving data, perhaps it does not require plugging a new device.
Can we use virtiofs? 
We did something similar in VEP165 (Containerpath Volumes)
The VM starts with a virtiofs (emptyDir, hostPath, or a PVC). The guest mounts it once. 
After that, whatever is running on the node can add files to that directory and delete them when it is done. 
The guest just sees files appear and disappear. 
That would be a good fit if the workload can use a folder.

vladikr:
Another idea.., 

if the node-local case is really needed and cannot wait on a spec update for every plug
then I'd suggest a `discover me` volume. 

I mean we (KubeVirt) did this trick a lot in the past: put the slot in spec first, fill it in later.

DataVolumes are an empty claim that becomes a disk when import finishes. 
Volume populators are the same idea. 
We add empty PCI controllers to the domain so there is somewhere to plug a device later. 
ContainerDisk is an emptyDir on the Pod; virt-handler bind-mounts the image into that directory before QEMU starts.

A node-local hotplug can be that, instead of a new call into virt-handler.
Declare the disk on the VM up front. Nothing is attached until a payload shows up on the node in a path KubeVirt owns! 
virt-handler notices it,  plugs it, and writes status. 
When the payload is gone, it unplugs. 

The spec never changes on the path. The vendor agent never calls virt-handler. It just publishes into a specific path.

So, I'd suggest something like

```yaml
volumes:
- name: inbox
  discover: {}
```

This keeps intent on the VM spec. 

The path is computed, not user-supplied.
This could be something like `/var/lib/kubevirt/discover/<namespace>/<vmi-name>/inbox`. 
That is the allowlist. An arbitrary `path:` on the spec is ordinary hostPath.

Publish has to be atomic. So write beside the target, `rename` into place. 
Unplug is `unlink` of that name.. 
virt-handler detaches before the bytes should disappear. 
Do not delete a file out from under a live QEMU disk. 

Live migration with just empty slot is fine, but when something is attached we should block.

What do you think?

mhenriks:
Great feedback @vladikr. I'm still thinking about `discover me` but I'd like to push back against extending `hostDisk` to support hotplug. `HostDisk` is very problematic from a security perspective as-is. Regular users should not be allowed to access an arbitrary file on the host.

Maybe we can extend hotplug to support PVCs with a `hostPath` backed PV though. The difference between this and `HostDisk` being that a cluster admin or some privileged component has to create the PV

awels:
> Great feedback @vladikr. I'm still thinking about `discover me` but I'd like to push back against extending `hostDisk` to support hotplug. `HostDisk` is very problematic from a security perspective as-is. Regular users should not be allowed to access an arbitrary file on the host.
> 
> Maybe we can extend hotplug to support PVCs with a `hostPath` backed PV though. The difference between this and `HostDisk` being that a cluster admin or some privileged component has to create the PV

Hotplug supports hostpath PVCs already, but it still goes through the entire attachment pod sequence for that. Which is the most problematic part from what I can tell from this VEP. The reason it goes through the attachment pod sequence is to account for the use of the volume in kubernetes. This is so other pods don't attempt to use the volume because it is in use by the attachment pod already (yes I know other pods on the same node can still use it).

mhenriks:
> > Great feedback @vladikr. I'm still thinking about `discover me` but I'd like to push back against extending `hostDisk` to support hotplug. `HostDisk` is very problematic from a security perspective as-is. Regular users should not be allowed to access an arbitrary file on the host.
> > Maybe we can extend hotplug to support PVCs with a `hostPath` backed PV though. The difference between this and `HostDisk` being that a cluster admin or some privileged component has to create the PV
> 
> Hotplug supports hostpath PVCs already, but it still goes through the entire attachment pod sequence for that. Which is the most problematic part from what I can tell from this VEP. The reason it goes through the attachment pod sequence is to account for the use of the volume in kubernetes. This is so other pods don't attempt to use the volume because it is in use by the attachment pod already (yes I know other pods on the same node can still use it).

Yeah I get this, but I think it may be possible to have a special hotplug path for hostpath/local PVs that does not create an attachment pod. I don't see why we have to care about "other pods don't attempt to use the volume" for hostPath/local

edit: unless ReadWriteOncePod

vladikr:
Hey, yeah.. thanks, I also dont really like the hostDisk idea :(
Let's focus on `discover: {}` ?
I think in spirit it's much closer to our existing patterns.
A namespace user can add the slot; without that node writer they just have an empty slot. Regular users never get a path: field.

s/writer/populater even

akalenyu:
@vladikr thank you for this great reiginition of the conversation, some really interesting ideas

> Maybe we can extend hotplug to support PVCs with a `hostPath` backed PV though.

These are supported today - I guess you mean that for those, we'd make some optimizations as @vladikr suggested with hostdisk?

I looked at an adjacent issue a few months ago and came to the conclusion virt-handler would have to be PV-aware (informer), which is not the greatest. The alternative is to expose sensitive PV data on vmi/elsewhere which was a no-go.

I think no pod will open up bookeeping issues but I wonder if those can be acceptable for the use case.

Question for @Sreeja1725 @rthallisey:
Would hostPath even be interesting? just want to make sure there is no CSI driver (and thus CSI pv/pvc) behind these.

rthallisey:
The `discover` idea could work. Maybe we can tweak it a little bit for the node local case. For example, a user would create the VMI with a hotplug disk in the spec. The disk would be empty or `local`, to be addressed by the handler later - no attachment pod.  At some point, a node local controller signals the virt-handler to run `syncDisks`, then handler hotplugs the specific local volume and updates the volume status in the vmi.

mhenriks:
> At some point, a node local controller signals the virt-handler to run syncDisks, then handler hotplugs the specific local volume and updates the volume status in the vmi.

I think it would be better if virt-handler simply watches a directory for files that match the vmi

rthallisey:
@mhenriks what if there's a local block device? Can the handler be given a device identity via a file and resolve that path?

vladikr:
> > At some point, a node local controller signals the virt-handler to run syncDisks, then handler hotplugs the specific local volume and updates the volume status in the vmi.
> 
> I think it would be better if virt-handler simply watches a directory for files that match the vmi

+1 , signaling the virt-handler is what we're trying to avoid.

vladikr:
> @mhenriks what if there's a local block device? Can the handler be given a device identity via a file and resolve that path?

@rthallisey yes for a local block device, as long as we do not invent a second path field.

The node-local agent can publish the device into the slot virt-handler already watches...
I think something ilike mknod with the right major/minor, or a bind-mount of /dev/disk/by-id/... onto /var/lib/kubevirt/discover/<ns>/<vmi>/<volume>. 
Dunno... we can discuss it.

Handler stats that name. If it is a block device it uses the same stage path we use for block hotplug. 
If it is a file, bind-mount.

I would not have the handler read a file that contains a device path and then resolve it. That is an arbitrary host path again.
If these devices are already CSI/local PVs, that is what @akalenyu asked... 
I think in this case we should be talking about skipping the attachment Pod for that PV, not using the discover.

rthallisey:
The way I see it, the node local volume hot plug use case would require several things:
- API field to express a volume is node local and tracked as a hotplug volume
  - The volume may or may not be backed by a PVC, depends on volume delivery method
- A way for the virt-handler to discover the node local disk after N seconds/minutes/hours
  - virt-handler would attach the discovered disk and update the volume status on the vmi

I think there's a way to address the issue and possibly take this KEP phases.  We can align further in the call @Sreeja1725 set up for this VEP.

Sreeja1725:
The meeting has been scheduled for Tuesday, September 8th at 4:00 PM CET and is available on the kubevirt calendar.

akalenyu:
What if we decompose the hotplug flow into named steps, keep the default logic in a shared file (always compiled), and use Go build tags to let downstream consumers override the step impl.?

```go
// attach_common.go — no build tag, always compiled
package hotplug

func attachViaAttachmentPod(ctx context.Context, vmi *v1.VirtualMachineInstance, vol *v1.Volume) error {
    // the current attachment pod flow
}
```

```go
//go:build !custom_volume_attach

// attach_default.go
package hotplug

func AttachVolume(ctx context.Context, vmi *v1.VirtualMachineInstance, vol *v1.Volume) error {
    return attachViaAttachmentPod(ctx, vmi, vol)
}
```

```go
//go:build custom_volume_attach

// attach_custom.go - provided by downstream folks
package hotplug

func AttachVolume(ctx context.Context, vmi *v1.VirtualMachineInstance, vol *v1.Volume) error {
    if isLocalVolume(vol) {
        return attachLocal(ctx, vmi, vol)
    }
    // fall back to default for everything else
    return attachViaAttachmentPod(ctx, vmi, vol)
}
```

PR #329, opened 2026-06-13
VEP 248: Syncing VEP with implementation & target beta

Aseeef:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/248
**SIG label**: sig-compute

### What this PR does

This PR updates VEP 248 using the lessons learned during implementation and testing. This update re-aligns the VEP with what the implementation ended up being.

- **New `stallProgressTimeout` knob (default 40s)** replaces repurposing legacy `progressTimeout` during alpha/beta.
  - Rationale: Reusing the same API field didn't make sense since different default values made sense for the legacy path vs the stall detector path.
- **GA plan documented:** `progressTimeout` repurposed as `stallProgressTimeout` with value override to 40s; entire `spec.experimental` section removed; most knobs baked into defaults.
- **New `switchoverTimeout` knob (default 60s)** — dedicated deadline after switchover is triggered; no longer piggybacks on `progressTimeout`.
  - Rationale: Previously we used `progressTimeout` for both switchover and completion timeouts but coupling the two did not make sense. (Note: "switchover" timeout means after we tell QEMU to switchover, how long do we give QEMU to actually switchover.)
- **New `completionTimeoutFactor` knob (default 2.0)** — generalizes the hard-coded ×2 completion budget check (formerly "Post-copy Network Risk Mitigation").
  - Rationale: Needed to be able to tune this to target the correct code paths in E2E tests. 
- **Multifd E2E workaround** via KUBEVIRT_MIGRATION_DISABLE_MULTIFD (test-only env; clears ParallelMigrationThreads so BandwidthPerMigration is honored).
  - Rationale: Due to a bug in QEMU, migration bandwidth cannot be adjusted with multifd enabled.
- **Relaxation logic corrected** — raises `bestRemainingBytes` to the **next-larger** distinct value (not "pop smallest").
  - Rationale: In K8s, now new fields are technically supposed to include units. 
- **Removed VEP 293** — References to VEP 293 removed, and descriptions about these experimental sections are now self-contained.
  - Rationale: This VEP never merged since it was considered redundant and an unnecessary overhead to track.
- **Entire "Logging & Visibility" section dropped** — no Prometheus metric changes, elevated logging, or shadow-mode observability in alpha. Associated API was also dropped.
  - Rationale: Prometheus wasn't the right tool for logging, and auditing at the level we wanted to. Nor is the existing KubeVirt source code wired for the kind of visibility we wanted. Making this work would have been a major effort that was not worth the return.
- **Set MaxDowntime in QEMU at migration start** if user sets `maxDowntimeMs` to a value below 300ms.
   - Rationale: honor the user set `maxDowntimeMs` even when it's below the typical 300ms. This effectively also allows users to specify tighter downtimes if they so choose.
- **Go-like pseudocode replaced** with higher-level pseudocode plus decision tables for switchover policy and completion timeout.
  - Rationale: It read too much like KubeVirt code and not enough like psuedocode making it hard to follow. Having proper high-level psuedocode would be valuable to reviewers reviewing the implementation PRs today. 
- **VFIO alpha behavior documented** — warning logged, post-copy skipped; full VFIO support is a beta requirement (QEMU 11.1+).
- **E2E deferred to Beta**; Alpha is unit tests only. E2E tunes via KUBEVIRT_MIGRATION_* env vars injected with a Mutating Admission Policy (ConfigMap envFrom), not spec.experimental and not an external webhook.
  - Rationale: experimental API goes away before GA; MAP is in-process and more stable than a test webhook.
- **E2E Plans** — Documented planned E2E scenarios.
- **Existing migration E2E tests flagged for update** at feature-gate graduation.
- **Default `stallProgressTimeout` lowered from 60s → 40s** with updated rationale.
  - Rationale: Actual post-implementation testing showed that the 60s timeout was too long.
- **Stall Relaxation Patience decay:** no minimum floor (earlier 1s cap removed).
  - Rationale: Earlier, a minimum floor of 1s was enforced, but this might restrict relaxation speed in some scenarios.
- **Pre-copy switchover semantics modified** — for the middle band (`maxDowntimeMs < impliedDowntimeMs ≤ maxDowntimeMs × precopyPossibleFactor`), the VEP now arms QEMU with soft stop-and-copy immediately instead of waiting for another iteration to “find a better” local minimum. 
  - Rationale: We defer the switchover logic to QEMU since there is no point in us trying to find local minima anymore since our threshold for what is acceptable is ANYTHING that can satisfy the max downtime.
- **Changes to graduation criteria**: Some additions in this section to make sure the various things we have agreed to regarding this VEP are not overlooked before graduating this VEP.
- **Target VEP for Beta**: Beta is now targeted for v1.10.
- **Other various tweaks** — made certain explanations more clear and aligned variable names with what the implementation is now using.
  - Rationale: Make it easier for someone reviewing the implementation PRs to compare against the VEP and orient themselves.

- The single most important/consequential change to review is the "Update/Rollback Compatibility" section.
- The second most significant change is Functional Testing Approach: Beta E2E uses env overrides + MAP; spec.experimental is for experimental-cluster tuning only, not E2E.

## Summary by CodeRabbit

* **Documentation**
  * Updated migration convergence guidance with revised stall-detection and switchover settings.
  * Added documented defaults for `stallProgressTimeout`, `switchoverTimeout`, and `completionTimeoutFactor`.
  * Clarified configuration locations and the transition from legacy `progressTimeout` behavior.
  * Updated dynamic-downtime, post-copy, VFIO, testing, compatibility, rollback, and release-readiness guidance.
  * Revised API examples, monitoring guidance, and migration policy recommendations.

### DIFF
--- a/veps/sig-compute/248-migration-convergence/migration-convergence.md
+++ b/veps/sig-compute/248-migration-convergence/migration-convergence.md
@@ -42,7 +42,7 @@ Why this enhancement is important
 
 When a VM's writable working set (WWS) is large compared to bandwidth, downtimes cannot be brought down low enough to trigger switchover automatically. Today, such cases are addressed by a static timeout derived from the VM size. However, this wastefully burns energy, compute, and bandwidth waiting for a conservative timeout:
     (a) **`completionTimeoutPerGiB`** defaults to **150s per GiB** (e.g. a 16 GiB VM can run **40+ minutes** before aborting). However, in most cases, a VM often stops making progress on the "remaining bytes" to transfer well before that.
-    (b) existing stall detection using `progressTimeout` doesn't work as it fails to consider that remaining bytes **oscillate** up and down with workload phases or bandwidth fluctuations and consequently resets progress timers anytime "remaining bytes" decreases.
+    (b) existing stall detection using the legacy `progressTimeout` ("remaining bytes did not decrease") doesn't work as it fails to consider that remaining bytes **oscillate** up and down with workload phases or bandwidth fluctuations and consequently resets progress timers anytime "remaining bytes" decreases.
 
 > __Workload Phase__
 Many applications experience different "workload phases" throughout their lifecycle. For example, a database application might occasionally have a "flush phase" where it stops everything it is doing and flushes all data to disk. The key feature of a "phase" is that the underlying workload characteristics change, exhibiting different dirty rates, CPU usage, memory etc. Attempting to take advantage of short-term phase changes to lower downtime (or time spent in post-copy) is a key objective of this proposal.
@@ -59,7 +59,7 @@ The desired outcome
 
 - Decrease total migration time, downtime, and/or the time spent in post-copy for VMs with high dirty rates.
 - Trigger post-copy or pause-and-copy when pre-copy has meaningfully stalled, rather than only after long generic timeouts, so migrations do not spin indefinitely once they have converged as far as they can.
-- Minimize user-side API changes and avoid exposing implementation details to end-users.
+- Minimize permanent user-facing API surface. **`maxDowntimeMs`** is the only new top-level field in alpha. Stall-detector tunables live in **`MigrationPolicy.spec.experimental`** — a temporary, alpha-only block whose name and scope signal that its fields are not stable API and will be removed at GA (see **API Changes**). Operators on experimental clusters may tune stall-detector hyper-parameters there while we decide what to promote or bake into defaults.
 - Ensure robustness against network fluctuations and failures.
     * Since migrations can last minutes, a temporary network drop should not trigger a switch-over as this is harmful especially in post-copy mode leading to data loss or significant performance degradation.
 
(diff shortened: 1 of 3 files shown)


veps/sig-compute/248-migration-convergence/migration-convergence.md
```diff
@@ -78,8 +78,8 @@ Limitations on the scope of the design
 Who is this feature set intended for
 -->
 
-Cluster Admins: Admins managing multi-node clusters.
-VM Admin: A VM owner/manager running their VM in a multi-node cluster.
+* Cluster Admins: Admins managing multi-node clusters.
+* VM Admin: A VM owner/manager running their VM in a multi-node cluster.
```
iholder101: Are these changes really necessary?

These kinds of changes that do not seem essential make this PR much harder to review...

Aseeef: I reverted all cosmetic changes.

veps/sig-compute/248-migration-convergence/vep.md
```diff
@@ -637,20 +563,32 @@ For example:
...
-Currently, in the absence of a fix from QEMU, we prefer either the first approach or the second (TBD). The third approach is completely dependent on the bandwidth the test cluster supports and should be avoided.
+Functional testing must reliably trigger stall under controlled conditions: a sustained high dirty-rate workload plus capped migration bandwidth so pre-copy plateaus predictably. Outcomes are verified via migration phase/mode and virt-launcher logs.
+
+Tests enable the `MigrationStallDetection` feature gate, drive dirty rates (e.g. `stress-ng`), and tune stall-detector hyper-parameters via `KUBEVIRT_MIGRATION_*` environment variables on virt-launcher. Injection uses a Kubernetes **Mutating Admission Policy** (MAP) that adds an `envFrom` ConfigMap to the compute container. Unlike traditional mutating webhooks, MAPs run in-process and are thus less susceptible to flakes, cheaper to run, and less complex to deploy (i.e., no SSL/secrets deployment).
+
+Env knobs used include:
+* `KUBEVIRT_MIGRATION_STALL_MARGIN`
+* `KUBEVIRT_MIGRATION_STALL_PROGRESS_TIMEOUT`
```
jean-edouard, line 573: I understand the use of these env vars for values that don't have an API, or for super-fast value changes, but I would expect most e2e tests to prefer CR API for simple cases, just to make sure it gets some coverage.

jean-edouard (commented):
At least some AI concerns seem valid, please address them.

veps/sig-compute/248-migration-convergence/vep.md
Aseeef: >  most e2e tests to prefer CR API for simple cases

Yes, I use CR API where I can. These env variables are mostly used to force a specific branch of the code path or to speed up how long the E2E tests last. As @iholder101 it does not make sense to expose knobs like this via API and so @vladikr and @iholder101 proposed to inject environment variables into virt-launcher to tune their behavior.

Aseeef:
> At least some AI concerns seem valid, please address them.

Should now be fixed:

https://github.com/kubevirt/enhancements/compare/4eee3b80137621af5d5406bd677605ab2c945da9..6e21b4e128e09f0c0161be7505c7ce1b3788c7fb

(Ignore the edits in `veps/sig-compute/109-vgpu-live-migration/vep.md` caused by a rebase to main)

jean-edouard:
The beta graduation criteria say "Revisit hyper-parameter fine-tuning and how trade-offs should be exposed to users," and @vladikr's condition at VEP 252 merge was: "Going forward, we must produce an adaptive convergence logic that doesn't require manual `stallMargin` or `ewmaAlpha`."
The old Prometheus/shadow-mode section was dropped (fair enough), but nothing replaced it. What signals will inform those decisions?
Without *some* data collection path documented here, it's hard to see how the hyper-parameter fine-tuning happens in practice.

Aseeef:
> Going forward, we must produce an adaptive convergence logic that doesn't require manual stallMargin or ewmaAlpha.

stallMargin and ewmaAlpha will not be exposed and would have no need for manual tuning. These will become constants with some reasonable default.

> What signals will inform those decisions?

This will have to be feedback from users and internal testing. Also, there is reason to believe in most cases, algorithm performance isn't super sensitive to the specific value of (most of) these knobs. Even if you don't have the perfect value, most reasonable values work good enough.

jean-edouard (approved):

veps/sig-compute/109-vgpu-live-migration/vep.md
```diff
@@ -14,7 +14,7 @@ For example, during the planning phase for version v1.123, do **not** target bet
 -->
 
 - This VEP targets alpha for version: v1.9
-- This VEP targets beta for version: TBD
+- This VEP targets beta for version: v1.10
```
xpivarc: Was the intention to update both 109 and 248 in one PR?

Aseeef: This was a rebase error on my end. Fixing...

jean-edouard (approved):

vladikr:
/approve
Thank you very much @Aseeef

Issue #333, opened 2026-06-16
VEP 333: Reusable Pod Metadata Input for virt-launcher Plugin Sidecars


fanzhangio:
**Primary contact (assignee)**:
/assign fanzhangio

**Current Feature Stage**:
New

**Feature Gate**:
TBD. The long-term feature gate should be confirmed by the owning SIG after the structured plugin API direction is accepted.

**Responsible SIGs**:
Primary SIG:
/sig compute
Additional SIGs (optional):
/sig network

**Enhancement link**:

TBD: link the VEP PR first, then replace with the merged VEP path.

Related discussion and prototype:

- Enhancement request: https://github.com/kubevirt/kubevirt/issues/18025
- Prototype implementation: https://github.com/kubevirt/kubevirt/pull/18026
- Related structured plugin direction: https://github.com/kubevirt/enhancements/tree/main/veps/sig-compute/190-kubevirt-structured-plugins

**Timeline**:

- Alpha release target (x.y): TBD by owning SIG
- Beta release target (x.y): TBD
- Stable release target (x.y): TBD

- [ ] Alpha
   - [ ] VEP PRs:
   - [ ] Code PRs:
   - [ ] Docs PRs:

**Additional context**:

This VEP is intentionally scoped to a reusable sidecar input named `pod-info`.

`pod-info` projects virt-launcher Pod labels and annotations into explicitly requesting sidecars through a Kubernetes DownwardAPI volume. It does not define a new plugin hook point and does not define cloud-init mutation behavior.

The primary long-term consumer is the structured plugin model. A separate follow-up VEP will propose a structured cloud-init hook, such as `PreCloudInitIso`, for cloud-init mutation use cases. That VEP may consume the `pod-info` input defined here.

Legacy hook-sidecar support, if implemented for alpha, is only a compatibility bridge for existing integrations and is not the beta/GA API path.

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
I believe this can already be addressed with the current Plugin mechanism (VEP-190) without a new VEP.

A Plugin can reference a MutatingAdmissionPolicy that adds a DownwardAPI volume projecting metadata.labels and metadata.annotations to the virt-launcher pod and mounts it into the sidecar container. This gives the sidecar access to all pod metadata - including CNI-provided annotations - without any new API, no RBAC, and no service account token.

[...]

PR #336, opened 2026-06-16
VEP-333: Reusable Pod Metadata Input for virt-launcher Plugin Sidecars

fanzhangio:
### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/333 
**SIG label**: /sig compute

### What this PR does
This VEP is intentionally scoped to a reusable sidecar input named pod-info.

pod-info projects virt-launcher Pod labels and annotations into explicitly requesting sidecars through a Kubernetes DownwardAPI volume. It does not define a new plugin hook point and does not define cloud-init mutation behavior.

The primary long-term consumer is the structured plugin model. A separate follow-up VEP will propose a structured cloud-init hook, such as PreCloudInitIso, for cloud-init mutation use cases. That VEP may consume the pod-info input defined here.

Legacy hook-sidecar support, if implemented for alpha, is only a compatibility bridge for existing integrations and is not the beta/GA API path.

### DIFF
--- a/veps/sig-compute/333-pod-metadata-for-sidecar/vep.md
+++ b/veps/sig-compute/333-pod-metadata-for-sidecar/vep.md
@@ -1,0 +1,238 @@
+# VEP #333: Reusable Pod Metadata Input for virt-launcher Plugin Sidecars
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


EdDev:
The [tracker](https://github.com/kubevirt/enhancements/issues/333) is closed, therefore this one also should be closed.
If this was left open intentionally, feel free to re-open,

/close

PR #339, opened 2026-06-17
Normalize VEP directory names across all sigs

fossedihelm:
Let’s tidy things up a bit by organising the repository into folders and subfolders with appropriate names.
Introduce a GitHub action enforcing this.
~Only sig-compute and sig-release are covered here.~

> [!IMPORTANT]
> This should be done at the beginning of the 1.10 cycle


.github/workflows/check-vep-naming.yaml
```diff
@@ -0,0 +1,25 @@
+name: Check VEP Naming Convention
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize]
```
iholder101: IIUC `pull_request_target` causes `github.sha` to resolve to the base branch HEAD, not the PR head - so `git diff` would diff base-to-base.

```suggestion
  pull_request:
    types: [opened, edited, reopened, synchronize]
```

.github/scripts/check-vep-naming.sh
iholder101, line 1: nit: I have to admit I don't like bash scripts, unless they're ridiculously short and concise, since I find them harder to debug, understand and maintain in general. But that might be a matter of taste :)

iholder101 (commented):
Thanks @fossedihelm! Nice!

Any reason to apply this only to sig-compute VEPs?

.github/workflows/check-vep-naming.yaml
fossedihelm: Thanks for catching it :pray:

.github/scripts/check-vep-naming.sh
fossedihelm: Me as well. I thought that for a very light work like this, installing the whole go would be an overkill.
But if u think it's better, I'll do that

fossedihelm:
> Any reason to apply this only to sig-compute VEPs?

Just to limit the size, allowing other sigs to align when they are ready (or when I decide that it's time to align :slightly_smiling_face: )

.github/scripts/check-vep-naming.sh
iholder101: Personally I'd just use Python probably.
But that's a nit, and I won't insist :) up to you.

fossedihelm: Here we go :)

fossedihelm:
@iholder101 Can you take another look? I would like to get this before v1.10 cycle starts.
Also, I covered all the other sigs.

@EdDev @orelmisan Can you take a look since it covers sig-network?
@akalenyu @mhenriks Can you take a look since it covers sig-storage?
@sradco Can you take a look since it covers sig-observability?
All of you can also look at the Github Action :) 
Thank you!

orelmisan (approved):
Thank you @fossedihelm!
SIG-network directories looks good!

.github/workflows/check-vep-naming.yaml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Validate VEP naming convention
+        env:
+          VEP_DIRS: veps/meta-VEPs veps/sig-compute veps/sig-network veps/sig-observability veps/sig-release veps/sig-storage
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.sha }}
+        run: python3 .github/scripts/check-vep-naming.py
```
iholder101: nit: no newline at end of file (feel free to ignore)

.github/scripts/check-vep-naming.py
```diff
@@ -0,0 +1,66 @@
...
+def check(vep_dirs, base_sha, head_sha):
+    errors = []
+
+    for vep_dir in vep_dirs:
+        for file in changed_files(vep_dir, base_sha, head_sha):
+            rel = file.removeprefix(f"{vep_dir}/")
+
+            if "/" not in rel:
+                errors.append(f"{file}: VEP files must be inside a <number>-<name>/ directory, not directly under {vep_dir}/")
+                continue
```
iholder101, line 29: A thought for the future (just so I won't forget) - we may want to add another action to check that the PR body has a reference to the tracking issue, and that the number in the path corresponds to that issue number.

.github/workflows/check-vep-naming.yaml
```diff
@@ -0,0 +1,25 @@
+name: Check VEP Naming Convention
+
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize]
```
iholder101:
```suggestion
    types: [opened, reopened, synchronize]
```

From [the docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request):
> synchronize: A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.
> edited: The title or body of a pull request was edited, or the base branch of a pull request was changed.

We can probably drop `edited` here since it triggers whenever the PR body/title are being modified. `synchronize` triggers when files are being modifies, so we should be good 👍

iholder101 (commented):
Thanks @fossedihelm! Looks great.
Left some minor comments, all of them are nits.

In addition, do you mind changing the PR title to be a bit more informative?

Thank you!
/lgtm

lyarwood:
@fossedihelm looks good, but I worry that this breaks links in VEP issues, implementation PRs and elsewhere. We could likely extend the VEP agent to detect these and either fix the links or comment in the issues? @iholder101 Do you think it's worth writing this up as a feature request against the agent?

iholder101:
> @fossedihelm looks good, but I worry that this breaks links in VEP issues, implementation PRs and elsewhere. We could likely extend the VEP agent to detect these and either fix the links or comment in the issues? @iholder101 Do you think it's worth writing this up as a feature request against the agent?

Nice catch! That's a valid concern indeed..

Regarding the agent - I think that (as said [here](https://www.youtube.com/watch?v=r3jZo20AH6c) BTW) we should ensure we choose the deterministic-code approach if we can. In this area, I think that with simple github action scripts we can do things like check if all of the links in a tracking issue are valid. That being said, I think that things like "ensure that all PRs that are related to this VEP are linked" are probably more difficult and should be dealt by the agent.

I'm planning to dedicate some time to make some major improvements to the agent. I'd appreciate opening issues to its repo and discussing new ideas!

iholder101 (approved):

xpivarc (commented):
/approve
/hold
Requires rebase

lyarwood (approved):
/lgtm
/hold cancel

PR #335, opened 2026-06-16
VEP #233: Add open-ended best-effort branches alternative

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/233
**SIG label**: /sig release

### What this PR does

Adds Alternative 6 to VEP #233, documenting the option of keeping older release branches open for merges on a **best-effort, no-obligation basis** after they exit the 3-release support window.

Under this model:
- The 3 most recent releases still receive full upstream support (unchanged)
- Older branches remain open for merges — anyone willing to do the work can propose cherry-picks
- No upstream obligation to initiate, review, or merge backports on these branches
- No CI requirement for best-effort branches

The alternative is rejected in the VEP because it provides little practical value over private forks without CI or review obligations, risks creating a false sense of support, and leaves the branch count unbounded.

This was requested by @iholder101 in https://github.com/kubevirt/enhancements/pull/317#issuecomment-4532806297.

### Special notes for your reviewer

This PR is independent of #317 (vendor forks alternative) and #334 (two-tier rework). All three are based on the original VEP branch and can be reviewed/merged separately.

PR #334, opened 2026-06-16
VEP #233: Rework proposal to use two-tier support model

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/233
**SIG label**: /sig release

### What this PR does

Promotes the two-tier support model (previously Alternative 4) to the primary proposal for VEP #233, replacing the 3-release-only model.

The two-tier model introduces:

- **Tier 1 — Full support (3 most recent releases)**: Unchanged from the original proposal — active upstream maintenance with full CI, matching the Kubernetes model
- **Tier 2 — Extended maintenance (next 2 older releases)**: Branches remain open for merges but do not receive proactive upstream backports. Downstream vendors may propose cherry-picks and are responsible for funding CI

Key changes:
- **Overview**: Updated proposal summary to describe two-tier model
- **CVE table**: Added "Under Two-Tier" column showing Tier 1 + Tier 2 scope
- **Design**: Rewritten with Tier Lifecycle, Tier 1→Tier 2 Transition Actions, Full EOL Actions, per-tier CVE Handling, and Tier 2 Branch Governance sections
- **Alternatives**: Moved old 3-release-only model to Alternative 4 (rejected), removed the two-tier model from alternatives
- **Graduation Requirements**: Updated for two-tier lifecycle (Tier 2 governance, vendor CI sponsorship)

### Special notes for your reviewer

This PR is independent of #317 (midstream fork alternative). Both are based on `main` and can be reviewed/merged separately.

PR #341, opened 2026-06-17
VEP #233: Rework to use dedicated release branch maintainer groups

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/233
**SIG label**: /sig release

### What this PR does

Reworks VEP #233 to replace the hard 3-release EOL cutoff with a dedicated release branch maintainer governance model.

Key changes from the original VEP:

- **Two explicit support levels**: The standard 3-release support window (aligned with the Kubernetes support matrix) is preserved as the primary support commitment. Beyond that, extended support is formalized as explicitly limited to security and critical fixes, dependent on dedicated maintainer availability.
- **Dedicated maintainer groups**: All release branches are owned by dedicated maintainer groups from the point they are cut, not by core `main` branch maintainers. This directly addresses the reviewer fatigue where core maintainers spend up to half their review time on backports.
- **Core maintainer oversight**: Release branch maintainer groups operate under the oversight of the project's core approvers, who define backporting policy, mentor release branch maintainers, and provide technical guidance on complex backports.
- **Maintainer-driven EOL**: Instead of a hard N-release cutoff, branches transition to EOL when no dedicated maintainer opts in to keep them alive. An EOL proposal PR is opened and must remain open for one month — anyone can object by committing to continue maintenance. This is modeled on [OpenStack's stable branch policy](https://docs.openstack.org/project-team-guide/stable-branches.html).
- **OWNERS transition at branch cut**: Automated as part of the release procedure using OWNERS_ALIASES.

This PR supersedes [PR #334](https://github.com/kubevirt/enhancements/pull/334) (two-tier model), [PR #335](https://github.com/kubevirt/enhancements/pull/335) (best-effort branches), and addresses the governance gap that motivated [PR #317](https://github.com/kubevirt/enhancements/pull/317) (midstream fork).

This is a substantial rework of the VEP. The branch lifecycle and EOL mechanics sections have been replaced entirely. The CVE motivation examples are preserved. The alternatives section has been reorganised — the original 3-release-only model and two-tier model are now listed as considered-and-rejected alternatives.

The VEP explicitly preserves the 3-release support window as the primary commitment and scopes the extended maintainer-driven model strictly to security and critical fixes. The maintainer-driven opt-in EOL process means branches without active maintainers are retired naturally rather than via an arbitrary release count, consistent with OpenStack's stable branch model.

### DIFF
--- a/veps/sig-release/233-release-branch-eol-policy/vep.md
+++ b/veps/sig-release/233-release-branch-eol-policy/vep.md
@@ -32,9 +32,20 @@ to a situation where CVE fixes, dependency updates, and other backports must be
 applied to an ever-growing number of branches, consuming significant
 contributor time and CI resources.
 
-This VEP proposes adopting a hard EOL policy modeled on the
-[Kubernetes patch release support period](https://kubernetes.io/releases/patch-releases/#support-period),
-limiting active maintenance to the **3 most recent minor releases**.
+Additionally, the same core maintainers who develop features on `main` are
+responsible for reviewing backports across all release branches. This creates
+reviewer fatigue and concentrates the maintenance burden on a small group of
+individuals — a problem that grows with each new release regardless of how many
+old branches are archived.
+
+This VEP proposes a **dedicated release branch governance model** where all
+release branches, from the point they are cut, are owned by maintainer groups
+separate from the core `main` branch maintainers. Release branches remain
+active as long as their dedicated maintainers opt in to keep them alive, and
+transition to EOL when no maintainer is willing to continue. This is modeled
+on [OpenStack's stable branch policy](https://docs.openstack.org/project-team-guide/stable-branches.html),
+which has operated at this scale (15+ project teams, dozens of repos) for over
+a decade.
 
 ## Motivation
 
@@ -45,10 +56,12 @@ only formally tracks v1.2 through v1.7 (with v1.2-v1.4 already marked EOL),
 in practice backports continue to land on branches well beyond the documented
 support window.
 
+### CVE Backport Burden
+
 Recent CVE responses illustrate how the lack of an enforced EOL compounds
 the maintenance burden:
 
-### CVE-2023-39325 / CVE-2023-44487 — HTTP/2 Rapid Reset (Oct 2023, High)
+#### CVE-2023-39325 / CVE-2023-44487 — HTTP/2 Rapid Reset (Oct 2023, High)
 
 This high-severity vulnerability in `golang.org/x/net` and
 `google.golang.org/grpc` required backports to **7 branches**: `release-0.49`,
@@ -61,11 +74,14 @@ and `release-1.2`
 [#10675](https://github.com/kubevirt/kubevirt/pull/10675),
 [#10674](https://github.com/kubevirt/kubevirt/pull/10674),
 [#10673](https://github.com/kubevirt/kubevirt/pull/10673)).
-Under a 3-release policy, only 3 branches (1.0, 1.1, 1.2) would have been
-in scope. Notably, `release-0.49` targeted Kubernetes 1.24, which had already
-been EOL for months when the fix landed.
+Under a maintainer-driven model, responsibility for these backports would
+have been distributed across dedicated release branch maintainers rather than
+falling on core `main` branch reviewers. Notably, `release-0.49` targeted
+Kubernetes 1.24, which had already been EOL for months when the fix landed —
+a branch that would likely have reached EOL earlier if maintainer opt-in were
+required to keep it alive.
 
-### CVE-2024-21626 — runc Container Escape (Jan 2024, Critical)
+#### CVE-2024-21626 — runc Container Escape (Jan 2024, Critical)
 
 This critical container escape vulnerability required backports to
 **4 branches**: `release-0.58`, `release-0.59`, `release-1.0`, and
@@ -77,7 +93,7 @@ This critical container escape vulnerability required backports to
 `release-0.58` targeted Kubernetes 1.26, which was well past its own EOL when
 the fix was merged.
 
-### CVE-2025-22869 — golang.org/x/crypto SSH Handshake Panic (2025, High)
+#### CVE-2025-22869 — golang.org/x/crypto SSH Handshake Panic (2025, High)
 
 A panic-triggering vulnerability in the Go SSH handshake was backported to
 `release-1.3` and `release-1.4`
@@ -85,25 +101,53 @@ A panic-triggering vulnerability in the Go SSH handshake was backported to
 [#14629](https://github.com/kubevirt/kubevirt/pull/14629)),
 both of which are already marked EOL in the sig-release support matrix.
 
-### CVE-2025-47913 — golang.org/x/crypto Denial of Service (2025, High)
+#### CVE-2025-47913 — golang.org/x/crypto Denial of Service (2025, High)
 
 A denial-of-service vulnerability in Go's crypto library was backported to
 `release-1.5` and `release-1.6`
 ([#16887](https://github.com/kubevirt/kubevirt/pull/16887),
 [#16870](https://github.com/kubevirt/kubevirt/pull/16870)).
 
-### CVE Backport Scope Comparison
+#### CVE Backport Summary
 
-| CVE | Date | Severity | Branches Backported | Under 3-Release Policy |
-|-----|------|----------|--------------------:|---------------------:|
-| CVE-2023-39325 | Oct 2023 | High | 7 | 3 |
-| CVE-2024-21626 | Jan 2024 | Critical | 4 | 3 |
-| CVE-2025-22869 | 2025 | High | 2 | 2 |
-| CVE-2025-47913 | 2025 | High | 2 | 2 |
+| CVE | Date | Severity | Branches Backported |
+|-----|------|----------|--------------------:|
+| CVE-2023-39325 | Oct 2023 | High | 7 |
+| CVE-2024-21626 | Jan 2024 | Critical | 4 |
+| CVE-2025-22869 | 2025 | High | 2 |
+| CVE-2025-47913 | 2025 | High | 2 |
 
-The pattern is clear: without an enforced EOL, each CVE response requires
-work across an unpredictable and growing set of branches. This creates
-several compounding problems:
+Under this VEP's governance model, the review and CI burden for these
+backports would be carried by dedicated release branch maintainers rather
+than core `main` branch reviewers. Branches without active maintainers
+would have already transitioned to EOL, naturally reducing the scope.
+
+### Reviewer Burden
+
+The unbounded branch count is only part of the problem. The same core
+maintainers who review feature work on `main` also review every backport
+cherry-pick across all release branches. This creates compounding reviewer
+fatigue:
+
+- Approximately **half of all upstream PRs are backports** — from various
+  contributors and downstream vendors — creating significant reviewer and CI
+  load
+- Individual core maintainers report spending **up to half of their review
+  time** on backports rather than feature development
+- Backports are often a source of long-running fixes where old branches are
+  discovered to be non-functional
+- Fixes for customer problems are sometimes raised upstream but never
+  backported, with missing bug tracking and no follow-up
+- Backports are sometimes raised for CVEs that don't actually affect upstream
+- Edge cases with forked dependencies (e.g.
+  [kubevirt#17692](https://github.com/kubevirt/kubevirt/pull/17692)) require
+  time-consuming manual verification by maintainers
+
+### Summary
+
+The pattern is clear: without separating release branch maintenance from core
+development and without a mechanism to EOL branches that lack active
+maintainers, the project faces:
 
 1. **Unsustainable contributor burden**: Each CVE or critical fix must be
    backported, reviewed, and merged across many branches. The HTTP/2 rapid
@@ -112,27 +156,30 @@ several compounding problems:
 2. **Unbounded CI cost**: CI lanes must be maintained and executed for every
    active release branch. Old branches frequently break due to infrastructure
    drift, consuming debugging time unrelated to the fix itself.
-3. **False sense of security**: Users on ancient releases receive sporadic CVE
+3. **Reviewer fatigue**: Core maintainers context-switch between feature
+   reviews on `main` and backport reviews across release branches, with
+   backports consuming up to half their review bandwidth.
+4. **False sense of security**: Users on ancient releases receive sporadic CVE
    fixes but no guarantee of comprehensive coverage, creating a false sense
    that the release is fully maintained.
-4. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
+5. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
    forcing function to upgrade to supported releases.
-5. **Inconsistency with stated policy**: The documentation already ties EOL to
+6. **Inconsistency with stated policy**: The documentation already ties EOL to
    Kubernetes support periods, but the lack of enforcement makes the policy
    meaningless in practice. Fixes are landing on branches targeting Kubernetes
    versions that Kubernetes itself stopped supporting long ago.
 
 ## Goals
 
-- Define a hard EOL policy for KubeVirt release branches with a fixed support
-  window
-- Align the KubeVirt release lifecycle with the Kubernetes release support
-  model
-- Reduce the number of actively maintained release branches to a bounded,
-  predictable set
+- Separate release branch maintenance from core `main` branch development by
+  establishing dedicated release branch maintainer groups
+- Define a maintainer-driven EOL policy where branches transition to EOL when
+  no dedicated maintainer opts in to keep them alive
 - Establish automated enforcement of EOL through CI and branch protection
 - Provide a clear transition plan for the existing backlog of unmaintained
   branches
+- Create a contributor onboarding path through release branch review toward
+  full SIG approver status on `main`
 
 ## Non Goals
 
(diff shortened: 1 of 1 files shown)


veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -217,28 +290,56 @@ When a release reaches EOL, the following actions are taken:
...
+offloading only old branches would leave the core of the problem untouched.
+
+Release branch maintainer group membership does not require being a core
+`main` branch approver — it is a separate group managed independently.
+However, the release branch maintainer groups operate under the oversight of
+the project's core approvers and formal maintainers, who are responsible for
+defining the backporting policy, mentoring release branch maintainers, and
+providing technical guidance on complex or uncertain backports. This ensures
+that release branch quality remains aligned with the project's standards even
+though the day-to-day review work is handled by a different group.
```
fossedihelm, line 330: Are the release branch maintainers a "static" group or will it change for every release?

fossedihelm (commented):
It looks good. Just a single question below :)

iholder101:
In theory, this direction sounds good - separating release branch maintenance from core development sounds like the right idea.

That said, I'm missing a concrete picture of who these dedicated maintainers actually are. Who do we envision stepping up to fill these groups? Do we have reason to believe people will volunteer for this role in practice?

To make sure this is realistic, should the VEP be upfront about who we expect to fill these groups?

More questions that come into mind: what's the minimum viable team per branch? and what happens if nobody steps up for the most recent branches where the bulk of backport activity happens?

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: The current implementation is per-release so it's easier to track activity and understand when a branch is eventually unmaintained. A centralised group would make this more difficult to track IMHO.

lyarwood:
> In theory, this direction sounds good - separating release branch maintenance from core development sounds like the right idea.
> 
> That said, I'm missing a concrete picture of who these dedicated maintainers actually are. Who do we envision stepping up to fill these groups? Do we have reason to believe people will volunteer for this role in practice?

I don't see this happening overnight but from my own discussions downstream I'm aware of folks who are willing to volunteer to be mentored into these roles over time yes.

I also believe that it's in the best interest of vendors to put folks forward for these roles if they have an interest in keeping branches active. So folks might end up being `voluntold` that they need to help out here but that's for vendors to work out internally and to then put these folks forward once the framework is in place.

> To make sure this is realistic, should the VEP be upfront about who we expect to fill these groups?
 
I don't think so, this is purely about getting the right framework in place and letting the community then mentor folks into these positions. I don't think it's realistic to have a list of names ahead of time to move this forward.

> More questions that come into mind: what's the minimum viable team per branch? and what happens if nobody steps up for the most recent branches where the bulk of backport activity happens?

I've assumed at least one release branch approver for now as listed in the description by the assumption that branches without this can transition to EOL:

> * **Maintainer-driven EOL**: Instead of a hard N-release cutoff, branches transition to EOL when no dedicated maintainer opts in to keep them alive. An EOL proposal PR is opened and must remain open for one month — anyone can object by committing to continue maintenance. This is modeled on [OpenStack's stable branch policy](https://docs.openstack.org/project-team-guide/stable-branches.html).

lyarwood:
Hey folks, tagging those of you who reviewed or showed interest in the original VEP PR (#234) — this PR is a substantial rework of that proposal and I'd appreciate your thoughts on the new approach.

**TL;DR of changes**: Instead of a hard 3-release EOL cutoff, the reworked VEP proposes dedicated release branch maintainer groups that own branches from the point they're cut. Branches transition to EOL when no maintainer opts in to keep them alive (with a one-month objection window), rather than hitting an arbitrary release count. This also creates a contributor onboarding path where release branch review serves as a stepping stone toward SIG approver status on `main`.

The PR description has a more detailed summary of the key changes.

/cc @aburdenthehand @dhiller @xpivarc @vladikr @fossedihelm @iholder101

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -112,27 +156,30 @@ several compounding problems:
...
-4. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
+5. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
    forcing function to upgrade to supported releases.
-5. **Inconsistency with stated policy**: The documentation already ties EOL to
+6. **Inconsistency with stated policy**: The documentation already ties EOL to
    Kubernetes support periods, but the lack of enforcement makes the policy
    meaningless in practice. Fixes are landing on branches targeting Kubernetes
    versions that Kubernetes itself stopped supporting long ago.
 
 ## Goals
```
EdDev, line 172: The listed goals read as implementation tasks rather than outcomes to achieve. For example, 
> Separate release branch maintenance from core main branch development by establishing dedicated release branch maintainer groups

describes how to implement the change, not what problem it solves. Goals should express the desired end state: e.g.,
> Define an explicit extended-support scope limited to security and critical fixes for releases beyond the standard support 
window

or
> Reduce the per-CVE review burden on core maintainers.

The mechanism (dedicated groups, OWNERS changes, opt-in EOL) belongs in the Design section.

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -112,27 +156,30 @@ several compounding problems:
...
 
-- Define a hard EOL policy for KubeVirt release branches with a fixed support
-  window
-- Align the KubeVirt release lifecycle with the Kubernetes release support
-  model
-- Reduce the number of actively maintained release branches to a bounded,
-  predictable set
+- Separate release branch maintenance from core `main` branch development by
+  establishing dedicated release branch maintainer groups
+- Define a maintainer-driven EOL policy where branches transition to EOL when
```
EdDev, line 176: The VEP replaces the hard 3-release support policy with a purely maintainer-driven model, but the two things solve different problems. The project's actual 
position — as I understand it from prior community discussions — is that the last 3 releases are the official support window (aligned with the Kubernetes 
support matrix). Beyond that, the existing practice is not "support" in any general sense: it is ad-hoc discretionary action by willing maintainers, 
specifically for CVE or critical fixes. This VEP formalizes and extends the second category, but by removing the 3-release baseline from the Goals it conflates
the two.

I'd suggest the goals explicitly preserve the 3-release window as the primary support commitment, and scope the extended maintainer-driven model strictly to security and critical fixes. Without that differentiation, the VEP risks communicating to downstream consumers that "active maintainers = full support", which is not what the project intends.

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -217,28 +290,56 @@ When a release reaches EOL, the following actions are taken:
...
+defining the backporting policy, mentoring release branch maintainers, and
+providing technical guidance on complex or uncertain backports. This ensures
+that release branch quality remains aligned with the project's standards even
+though the day-to-day review work is handled by a different group.
+
+This structure also provides a natural onboarding path for new contributors:
+reviewing backports (lower-risk changes with clear correctness criteria) under
+the mentorship of core maintainers builds codebase familiarity and review
+experience, making release branch maintainers strong candidates for future
+SIG approver status on `main`.
```
EdDev, line 336: I disagree with the characterization of release branch reviews as "lower-risk changes with clear correctness criteria." Changes to stable release branches are arguably more sensitive than changes to main: they must not alter user-visible behavior, must not introduce regressions on a platform that users depend on for production stability, and often require understanding subtle divergences between the branch and main. A release branch maintainer needs to be extra pedantic — their job is to protect stability, not advance functionality.

The logic that this makes them "strong candidates for future SIG approver status on main" is also inverted. The skill set required is different: a stable-branch maintainer who excels at conservative, risk-averse backport evaluation may not be the right match for a main-branch approver who evaluates new features. Conflating the two roles risks misaligning incentives and attracts contributors motivated by approver status rather than the maintenance-focused mindset the role actually demands. I'd suggest removing or rethinking this framing — the dedicated maintainer model stands on its own merit without needing this onboarding narrative.

veps/sig-release/233-release-branch-eol-policy/vep.md
EdDev, line 24: Reading this as a whole: what the VEP is fundamentally doing is making existing informal behavior official. In practice, KubeVirt has always allowed ad-hoc backports to older branches at the discretion of willing maintainers and approvers — the documentation even explicitly states that EOL "is currently not 
enforced". This VEP formalizes that with a structured opt-in process and adds the genuinely new element: the discretionary maintenance work can be handled by dedicated groups rather than falling on the same approvers responsible for main and the officially supported releases. Framing the Overview around what is genuinely new here (the governance separation, the opt-in EOL process) versus what is being codified from existing practice would help reviewers quickly understand the scope of the change and reduce the risk of misreading this as a broader expansion of support commitments.

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -112,27 +156,30 @@ several compounding problems:
...
    drift, consuming debugging time unrelated to the fix itself.
-3. **False sense of security**: Users on ancient releases receive sporadic CVE
+3. **Reviewer fatigue**: Core maintainers context-switch between feature
+   reviews on `main` and backport reviews across release branches, with
+   backports consuming up to half their review bandwidth.
+4. **False sense of security**: Users on ancient releases receive sporadic CVE
    fixes but no guarantee of comprehensive coverage, creating a false sense
    that the release is fully maintained.
-4. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
+5. **No upgrade incentive**: Without a hard EOL, downstream consumers have no
```
EdDev, line 165: This problem is listed as a motivation for the change, but the maintainer-driven EOL model does not resolve it. Under the proposed approach, a branch with an 
active maintainer has no defined EOL date, so downstream consumers still have no predictable forcing function to upgrade. The upgrade incentive was a direct consequence of the hard 3-release cutoff — not of who does the reviewing. If the intended scope of extended branches is limited to CVE and critical fixes only (not general support), that constraint would itself serve as the incentive: users would know that older branches are not fully maintained and would have a clear reason to upgrade. But the VEP does not state that scope explicitly. Either acknowledge that this problem is not addressed by this proposal, or clarify what "active maintenance" means for extended branches so the upgrade incentive is made explicit.

EdDev (changes requested):
I support the direction taken here.

IMO the high level observation is something like this:

## Today

The official Kubevirt support contract is for the last 3 versions, following the Kubernetes support matrix.
In parallel, unofficially, the project and its maintainer/approvers accept CVE and critical fixes beyond those releases, at their discretion.

## Proposal

Formalize the support behavior today by separating the two levels explicitly:
1. Full support for the 3 last releases.
2. Limited extended support for CVE and critical fixes for older releases, dependent on maintenance resource availability (e.g. reviewers, maintainers/approvers and CI infrastructure).

The motivation mentioned still holds, as the proposal answers the need to help maintenance burden, clarity and visibility.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: Good point. I've added a paragraph to the Overview that explicitly distinguishes what is being codified from existing informal practice versus what is genuinely new in this proposal.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: Fair point. I've softened this to acknowledge that the maintainer-driven model doesn't provide a hard forcing function, while still noting it's a genuine improvement over the current indefinite "not enforced" state where there is no EOL signal at all.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: Agreed. I've rewritten the goals to express desired outcomes rather than implementation tasks. The mechanism details (dedicated groups, OWNERS changes, opt-in EOL process) remain in the Design section where they belong.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: Agreed. I've added explicit goals to preserve the 3-release support window as the primary commitment and to scope extended support to security and critical fixes only. I've also added a clarifying paragraph to the Overview that makes this two-level distinction upfront.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: I respectfully disagree here, drawing from my own experience in OpenStack/Nova where I followed exactly this path: stable branch reviewer → stable branch approver → core (main) reviewer → core approver.

You're right that stable branch work requires a different focus — it demands extra conservatism, an understanding of what constitutes a safe backport, and careful judgement about regression risk. It is not "lower risk" in the sense of being easier to get wrong, but it is more bounded: the scope of a correct backport is generally clearer than evaluating a net-new feature design. With strong policy, clear backport criteria, and mentorship from core maintainers, that bounded scope is precisely what makes it a viable starting point for contributors who aren't yet ready for the full breadth of main branch review.

The key is that the onboarding value isn't "release branch review is easy" — it's that it provides a structured, mentored entry point with well-defined expectations. Whether those contributors eventually move to main branch approval is then a matter of demonstrated judgement over time, not an automatic progression. I'd rather have the framework in place and let the community prove it out than remove the onboarding path entirely on the grounds that the skill sets differ.

That said, I don't want this disagreement to block the effort, so I'm removing the onboarding framing from the VEP for now. It can always be revisited in a follow-up.

lyarwood:
@iholder101 @fossedihelm @EdDev - Would appreciate another round of reviews if you have time  before the deadline.

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -148,66 +208,89 @@ several compounding problems:
...
-|-------|-------------------|
-| v1.7 GA | v1.7, v1.6, v1.5 |
-| v1.8 GA | v1.8, v1.7, v1.6 — **v1.5 reaches EOL** |
-| v1.9 GA | v1.9, v1.8, v1.7 — **v1.6 reaches EOL** |
+A release branch transitions to EOL when **no dedicated maintainer opts in to
+keep it alive**. This is evaluated through an explicit opt-in process modeled
+on OpenStack's approach:
+
+1. An **EOL proposal PR** is opened against `kubevirt/sig-release` (or
+   `kubevirt/enhancements`) proposing the branch for EOL.
```
fossedihelm, line 277: Do we need to specify if this will happen at the beginning of the cycle or at the end?

fossedihelm (commented):
Just one last question, otherwise lgtm

iholder101:
Maybe it's an unpopular opinion, but TBH, I'm still not quite sure what we gain from marking old branches as EOL.

If someone wants to become an approver of an old branch - they're welcome to do so. If they participate in complex backports and prove themselves to be worthy of the position, they can nominate themselves, and I'm sure the community will bless this move if they're indeed worthy.

If someone is not an old-branch approver, or a vendor doesn't want to allocate people for that job, then it'll be harder for them to pursue a complex backport. That being said, if they want to contribute a one-time backport, willing to take the time to explain and convince the community it's beneficial, seek for reviewers and possibly wait until people have time/motivation to review - that should be possible on a best-effort basis.

Therefore I ask myself - why do we even want to change anything, maybe other than documenting that it is possible to nominate yourself as an old-branch approver?

If we get to a future situation where we have many old-branch approvers we can formalize it further and maybe create a process similar to the one you're suggesting here. But as of now, while you admit it's rather unclear who will take these positions, the practical outcome I see from this is that branches would become EOL blocking potential backports that the community might support. I'm just not sure what we gain from this move.

Doesn't it make sense to do this in the opposite direction? i.e. have people nominate themselves as old-branch approvers, and if we see a demand for it we can come up with a formalized process?

veps/sig-release/233-release-branch-eol-policy/vep.md
EdDev: I do not see a reason to bound it to any cycle.

I just hope this can be automated and managed in a way that has little friction.

What I am mainly missing is what triggers this EOL PR. How do we know that there are no longer opt-in maintainers for the branch?

veps/sig-release/233-release-branch-eol-policy/vep.md
```diff
@@ -249,115 +354,118 @@ To avoid disruption, the following transition plan is proposed:
...
 2. **Grace period**: Provide one full release cycle of notice. For example, if
    announced with v1.9 GA, the policy takes effect at v1.10 GA.
-3. **Initial EOL batch**: At the policy effective date, all releases older than
-   N-2 (where N is the current release) are marked EOL simultaneously. As of
-   v1.10 GA, this would mean v1.7 and earlier are EOL.
-4. **Archive old branches**: For branches that have been de facto dead for
+3. **Establish release branch maintainer groups**: During the grace period,
+   populate OWNERS_ALIASES with initial release branch maintainer groups and
+   onboard members.
+4. **Archive dead branches**: For branches that have been de facto dead for
```
EdDev: This requires a clarification on how it is done.

EdDev (commented):
Formalizing this makes sense, and I would go one step further and apply here the policy of reviewers and approvers being removed after a year of inactivity.

The only missing piece I see is what triggers in practice an EOL flow.
The only hint I found about it is "active maintainers", but that has not been accompanied with how to measure it.
I suggest to use the existing method for automatic removal on maintenance branches, especially for the root approvals list.
The original time of inactivity was 1 year, we can stick with it until we can agree on a more aggressive timeline.

If existing maintainers want to help support these old releases, we have no reason to stop them. If existing maintainers/approvers want to nominate someone to continue the work, they should do so through the regular process.

WDYT?

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: It shouldn't align with a main release cycle no.

I agree automation would be great but we will need to pick one or more signals to help with this. For example, no approvers or reviewers in the maintenance alias group, no PRs posted or merged for x amount of time, no CI available etc.

I'm not sure it's required for the initial implementation here and can be manual for now.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: Sure I can add references to the recent process @dhiller used

veps/sig-release/233-release-branch-eol-policy/vep.md
EdDev: The proposal will be stronger with automation or at least a policy in place.
If we combine it with the process that runs over the active approvers, this can be the trigger for now. It does align to the other points in the VEP.

lyarwood:
> Maybe it's an unpopular opinion, but TBH, I'm still not quite sure what we gain from marking old branches as EOL.

Well that's a concern as I've been trying to address an issue you yourself have highlighted many times in recent cycles, specifically approver and reviewer fatigue.

IMHO open but unmaintained release branches are a clear sign of a projects immaturity and in the current climate an excellent way to expose users to a range of threats that can be avoided by simply EOL'ing the branch in a structured way.
 
> If someone wants to become an approver of an old branch - they're welcome to do so. If they participate in complex backports and prove themselves to be worthy of the position, they can nominate themselves, and I'm sure the community will bless this move if they're indeed worthy.
>
> If someone is not an old-branch approver, or a vendor doesn't want to allocate people for that job, then it'll be harder for them to pursue a complex backport. That being said, if they want to contribute a one-time backport, willing to take the time to explain and convince the community it's beneficial, seek for reviewers and possibly wait until people have time/motivation to review - that should be possible on a best-effort basis.
> 
> Therefore I ask myself - why do we even want to change anything, maybe other than documenting that it is possible to nominate yourself as an old-branch approver?

That's essentially half of what I've proposed but as above leaving branches open and potentially unmaintained outside of this is at best naive and at worst down right negligent in the climate we find ourselves in today.
 
> If we get to a future situation where we have many old-branch approvers we can formalize it further and maybe create a process similar to the one you're suggesting here. But as of now, while you admit it's rather unclear who will take these positions, the practical outcome I see from this is that branches would become EOL blocking potential backports that the community might support. I'm just not sure what we gain from this move.
> 
> Doesn't it make sense to do this in the opposite direction? i.e. have people nominate themselves as old-branch approvers, and if we see a demand for it we can come up with a formalized process?

Thanks @iholder101, as above I really think both the nomination and EOL'ing processes are required here for the project to be taken seriously and move forward.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: ACK [updated](https://github.com/kubevirt/enhancements/compare/1d419763c4feffcb913cdde77ddd73b330997b4f..fc9196f98e558705aa22826e5594afeb14c45ac4) with a generic paragraph for now.

veps/sig-release/233-release-branch-eol-policy/vep.md
lyarwood: [Added](https://github.com/kubevirt/enhancements/compare/1d419763c4feffcb913cdde77ddd73b330997b4f..fc9196f98e558705aa22826e5594afeb14c45ac4)

EdDev (approved):
Thank you!

/lgtm

aburdenthehand:
/lgtm as well
This makes sense to me and I like that it's following an established practice by a long-term project. 
It will be interesting to see how it bears out in practice for KubeVirt. Thank you very much @lyarwood

lyarwood:
@vladikr @xpivarc would appreciate your thoughts on this as I'd like to give this a go with v1.10.0

fossedihelm:
Still lgtm from my side

Issue #342, opened 2026-06-17
VEP 342: IOThread Virtqueue Mapping for virtio-scsi


dsanatar:
**Primary contact (assignee)**:

/assign @dsanatar 

**Current Feature Stage**:

New

**Feature Gate**:
`SCSIMultiIOThread`

**Responsible SIGs**:

Primary SIG:
/sig storage

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/343

**Timeline**:

* Alpha release target (x.y): v1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/343
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/18249
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #343, opened 2026-06-17
VEP #342: IOThread Virtqueue Mapping for virtio-scsi

dsanatar:
### VEP Metadata

**Tracking issue**: 
https://github.com/kubevirt/enhancements/issues/342

**SIG label**: 
/sig storage

### What this PR does

Leverage existing `IOThreadsPolicies` API to extend the SCSI Controller to expose dedicated IOThreads for virtio-scsi devices.


veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,235 @@
...
+Cluster users running VMs that contain multiple virtio-scsi disks performing heavy I/O operations.
+
+## User Stories
+
+<!--
+List of user stories this design aims to solve
+-->
+
+* As a user I would like to apply multiple IOThreads for my virtio-scsi disks
+* As a user I would like to see a performance increase in my virtio-scsi disks that are leveraging iothread-vq-mapping
```
Acedus: nit: I think these two user stories can be combined since the overall goal is to improve performance and the means are the new iothread-vq-mapping capability.

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,235 @@
...
+
+Note: this example exposes a potential flaw in this design that may need to be discussed. Since the virtio-scsi controller will only create virtqueues equal to the number of vCPUs, we could encounter a case where we allocate more threads than the controller needs. It may be worth limiting the list of threads we provide to the controller capped at the amount of virtqueues it will be processing i.e the number of vCPUs.
+
+## Alternatives
+
+<!--
+Outline any alternative designs that have been considered)
+-->
+
+VMs that specificy `ioThreadsPolicy: supplementalPool` with virtio-scsi and blk devices will result in this supplementalPool being a shared thread pool, there could be an argument for there being a distinct pool for just the scsi controller.
```
Acedus: s/specificy/specify

Acedus (commented):
I have no issue with the approach, makes sense to me.
/cc @mhenriks @akalenyu

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,234 @@
...
+
+Disks will be assigned IOThreads like this:
+
+```
+testdisk1: 1
+testdisk2: 2
+scsi-disk: [1, 2, 3]
+```
+
+Note: this example exposes a potential flaw in this design that may need to be discussed. Since the virtio-scsi controller will only create virtqueues equal to the number of vCPUs, we could encounter a case where we allocate more threads than the controller needs. It may be worth limiting the list of threads we provide to the controller capped at the amount of virtqueues it will be processing i.e the number of vCPUs.
```
aglitke: This got me wondering how this change will impact hotplug.  Specifically, in nested cluster configurations where kubevirt provides VMs to act as hosted cluster nodes and lots of kubevirt-csi volumes are hotplugged into these VMs.  Maybe this is worth addressing in a section here?

awels:
Looks good, except you have not addressed the concern about hotplugging a bunch of disks, and getting the queues out of sync with what was there before that Adam mentions.

dsanatar:
@awels @aglitke just added a new section to address these hotplug scenarios. let me know if this covers what you were thinking.

Acedus (approved):
I agree with the approach proposed wrt hotplug considerations.

awels (commented):
/lgtm

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,257 @@
...
+
+For VMs using the `auto` policy we could consider two approaches:
+
+1) always allocate the maximum thread count to the SCSI controller (# of vCPUs)
+ - Pros: controller will still see performance gain if a large amount of disks are hotplugged
+ - Cons: creates a lot of potentially wasted threads that could stay idle for the entire VMI lifecycle
+
+2) recommend using `supplementalPool` policy for heavy-hotplug scenarios
+  - Pros: user-declared and stable, leads to less wasted threads compared to option 1
+  - Cons: could be seen as restrictive
```
akalenyu, line 187: I think this makes sense. We deliberately don't mention enforcing this, right? (going to be a pain)

Also, cc @jeniferh, thoughts will be appreciated as the author of https://developers.redhat.com/blog/2025/06/23/feature-introduction-multiple-iothreads-openshift-virtualization 🙏

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Overview
+
+<!--
+Provide a brief overview of the topic
+-->
+
+IOThread Virtqueue Mapping allows for multiple dedicated I/O threads to process an individual disk's virtqueues to improve I/O performance. Prior to libvirt 11.2.0 (QEMU 10.0), iothread-vq-mapping was exclusively supported for virtio-blk devices. This enhancement aims to extend the virtio-scsi controller to leverage the libvirt changes to enable multi I/O threads.
+
+## Motivation
```
akalenyu, line 37: quick brief for why one would use scsi instead of blk could be nice, not a must. scsi is especially useful for VMs with many disks, for example. https://www.qemu.org/2021/01/19/virtio-blk-scsi-configuration/

veps/sig-storage/iothread-vq-virtio-scsi.md
jeniferh: As far as any "recommendation" I would go with `supplementalPool` policy, but that does require the admin to set the number of threads desired (to be shared by all disks). Otherwise like you mention there are potentially lots of wasted thread overhead, and the main issue is `auto` can only assign a single thread per disk where `supplementalPool` allows disks to utilize multiple threads in parallel -- this is where we see the biggest perf gains for multi threaded applications. The downside is potentially "not enough" threads for all hotplug disks, but depending on dynamic I/O usage this can still help many cases in my mind vs being restricted to 1 thread per disk.

dsanatar: @akalenyu correct, this is just a recommendation as I don't believe enforcing this is necessary since either policy option here should still see I/O performance gains compared to if no policy was used at all.

Acedus (approved):

veps/sig-storage/iothread-vq-virtio-scsi.md
kevmw: Creating iothreads per se and having them around without doing anything isn't very expensive, so I wouldn't be worried about that. They do become expensive once you reserve a CPU for them, of course, but I understand that this isn't the default. What could happen when you spread light I/O over multiple iothreads that then each have to wake up for new requests is that you might get worse latencies than if a single thread could handle all of them at once and stay reasonably busy. I'm not sure if it's relevant, but that's what seems to be the biggest downside of just creating more iothreads than strictly needed.

But either way, `supplementalPool` with up to `num_vcpus` iothreads that are then shared by all virtio-blk disks and virtio-scsi controllers does sound like a generally better policy to me, too.

kevmw:
One general remark about the terminology in the document: virtio-scsi devices are SCSI controllers, but the document uses the term "virtio-scsi device" not for the controller, but for the SCSI disks that are attached to a virtio-scsi device. I found this a bit confusing.

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,259 @@
...
+The challenge here is that the SCSI controller can only assign threads to virtqueues (which currently is set to the number of vCPUs in a VMI) and not to individual disks like with virtio-blk. We can leverage the existing `IOThreadsPolicy` API to enable a multi-threaded SCSI controller that can process the virtqueues in parallel however, we cannot guarantee the same 1:1 thread to disk mapping which slightly changes how the policies will behave with SCSI disks.
+
+
+The following is an overview of how the SCSI controller would consume the following policies.
+
+### Policy: shared
+Each disk shares the same IOThread. With these proposed changes, we would assign this same thread to be also shared by the controller.
+
+### Policy: auto
+Each disk (excluding ones that request dedicatedIO) get assigned a thread in round robin order from the list of available iothreads. The total number of threads is equal to the amount of disks in a VMI, capped at 2 * (# of vCPUs). Since the SCSI controller cannot assign individual threads to its disks, we would instead allocate all of the "auto threads" (threads not reserved for dedicatedIO disks) to the controller.
```
kevmw: Does this essentially make it the same as `supplementalPool`, except that it uses a default number of iothreads instead of requiring the user to specify it?

It might be confusing for users that with `auto`, virtio-scsi disks get per-queue iothreads while virtio-blk disks only get a single iothread for all of their queues. This might look like inherent performance differences between virtio-scsi and virtio-blk when it's really coming from a hidden configuration difference.

dsanatar: yes and I completely agree that this could potentially be misleading for users however, I don't see a good way we can guarantee the same behavior of the `auto` policy for the two different bus types. My current plan is to just explicitly note this configuration difference in the API docs.

kevmw: If we were designing this API from scratch, the obvious way would be to use the behaviour you're proposing for virtio-scsi here for virtio-blk as well - in the common case, this should be the better model, and I assume virtio-blk only doesn't work like this because the `auto` policy was introduced before iothreads could be configured per queue.

I don't know if changing it now in virtio-blk might be considered a compatibility problem. The behaviour changes for virtio-scsi, too (even though mostly only from ignoring the policy to doing something with it; but still the same configuration will have a different result now), so at least keeping behaviour completely unchanged doesn't seem like a hard requirement. I guess it's a judgement call.

dsanatar: I think your suggestion of aligning the behavior of the `auto` policy for the two bus types is valid and probably the right move in the long term, that being said I do worry about the compatibility impact of changing the established policy. I would be happy to expand the scope of this VEP to include this new change however, it might be simpler to keep it as a followup so that we could contain it to a separate feature gate. Let me know your thoughts and I can update the VEP accordingly

kevmw: Just whatever works best for you really, I'm just sharing my thoughts on the interfaces in the hope that they are useful for you. You know how Kubevirt development works and I don't, so if you think that splitting it into two separate part is best, that's probably the way to go.

dsanatar: Ok, thank you the review, I appreciate the input.

dsanatar:
@kevmw @akalenyu after some discussion, I think it would be beneficial to include the changes to the `auto` policy in this VEP so that the behavior is consistent across the two bus types. i've updated the VEP accordingly

veps/sig-storage/iothread-vq-virtio-scsi.md
```diff
@@ -0,0 +1,262 @@
...
+## Goals
+
+<!--
+The desired outcome
+-->
+
+* Leverage existing `IOThreadsPolicies` API (shared, auto, supplementalPool), to extend the SCSI Controller to expose dedicated IOThreads for virtio-scsi devices.
+* Modify the existing `auto` policy to allocate pool of auto threads to each disk rather than a single round-robin thread. This behavior change will align with how virtio-scsi will use this policy (see `Policy: auto` section in design)
+* Add opt-in feature gate.
+* Complete performance testing to verify that virtio-scsi disks benefit from the inclusion of iothread-vq-mapping, similar to what was seen for virtio-blk.
```
EdDev, line 56: The goals read a bit more like implementation details (the how) rather than the actual objectives (the what and why).

For the Goals section, it would be great to focus purely on the user-facing outcomes we want to achieve (e.g., 'Users should be able to do X without needing Y').

dsanatar: thanks for the suggestion, I updated the section to hopefully closer align with what you were thinking

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -215,6 +215,8 @@ The feature is additive and will be behind a feature gate. On upgrade, the exist
 
 On rollback, disabling the feature gate reverts the ability to perform iothread-vq-mapping for virtio-scsi and will fallback to using a single threaded SCSI controller. This will also revert the `auto` policy behavior to only assign a single round-robin thread.
 
+Once this feature is GA and on by default, a new KubeVirt CR configuration will be added to allow users toggle off the new multi-threaded `auto` policy in favour of preserving the single thread legacy behavior.
+
```
akalenyu: I think we should introduce a toggle right from the start. So maybe 2 FGs:
- one for your original intention (virtqueue mapping for virtio-scsi)
- one for virtio behavior changes (auto)

Only the second one (virtio) also introduces a toggle. So folks during FG phase will have to also enable the config

Without starting off with both a FG and toggle, we'd have a weird graduation path where we'd have to convert folks from FG on->toggle on

akalenyu: I think this is the correct way to align while staying backwards compatible. Folks building on top of kubevirt can always opinionate their kubevirt offering to always opt in. @kevmw thank you for the insights in
https://github.com/kubevirt/enhancements/pull/343#discussion_r3691984113

akalenyu (approved):

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+      <iothread id='3'/>
+    </iothreads>
+  </driver>
+</controller>
+```
+
+The challenge here is that the SCSI controller can only assign threads to virtqueues (which currently is set to the number of vCPUs in a VMI) and not to individual disks like with virtio-blk. We can leverage the existing `IOThreadsPolicy` API to enable a multi-threaded SCSI controller that can process the virtqueues in parallel however, we cannot guarantee the same 1:1 thread to disk mapping which slightly changes how the policies will behave with SCSI disks.
+
+
+The following is an overview of how the SCSI controller would consume the following policies.
```
xpivarc, line 114: Please also explain relation to `dedicatedIOThread`

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Policy: shared
+Each disk shares the same IOThread. With these proposed changes, we would assign this same thread to be also shared by the controller.
+
+### Policy: auto
+Currently, each virtio-blk disk (excluding ones that request dedicatedIO) gets assigned a single thread in round robin order from the list of available iothreads. The total number of threads is equal to the number of disks in a VMI, capped at 2 * (# of vCPUs). Since the SCSI controller cannot assign individual threads to its disks, we would instead allocate all of the "auto threads" (threads not reserved for dedicatedIO disks) to the controller.
+
+In order to prevent this behavior from diverging for the two bus types, this implementation will also include modifications to the `auto` policy for virtio-blk so that each blk disk will now be allocated the same auto thread pool instead of a single thread. To preserve backwards compatibility, a new KubeVirt configuration will be added to act as a toggle for this virtio-blk `auto` policy behavior. To explicitly opt-in, users must have both the feature gate as well as this new toggle set.
```
xpivarc: What new KubeVirt configuration are we referencing here?

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Hotplug Considerations
+
+Since the SCSI Controller is initialized at VMI startup, the amount of threads we allocate to the controller has to be known at start time since there is currently no libvirt API to modify the thread pool of a running controller. This poses a certain limitation in hotplug scenarios when using the `auto` policy.
+
+By default KubeVirt will always create a SCSI controller to support hotplugging virtio-scsi disks into a running VM as long as the `disableHotplug` flag is not present in the spec. To enable the multi iothread performance gain for hotplug disks, we need to ensure the controller gets allocated a thread pool, even when no virtio-scsi disks exist at initial domain creation (assuming `IOThreadsPolicy` is set).
+
+However, when using the `auto` policy, the number of threads that get allocated at startup is derived from the initial number of disks in the VM. So if a large amount of scsi disks are hotplugged to a VM, the SCSI Controller could take a performance hit due to not being initialized with a large enough thread pool. This is especially relevant in a nested cluster configuration using kubevirt-csi.
```
xpivarc, line 187: Is it concern if `supplementalPool` with `ioThreads` can workaround it?

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+- [ ] Implementation supports all X use-cases
+
+It is not necessary to have all the requirements for all stages in the initial VEP.
+They can be added later as the feature progresses, and there is more clarity towards its future.
+
+Refer to https://github.com/kubevirt/community/blob/main/design-proposals/feature-lifecycle.md#releases for more details
+-->
+
+### Alpha
+- [ ] Add new feature gate to guard changes for virtio-scsi
```
xpivarc: Please name the fg

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+- [ ] Add new KubeVirt configuration to toggle new `auto` policy for virtio-blk.
```
xpivarc: Where is this described?

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,267 @@
...
+- [ ] Updates to virtio-scsi controller
+- [ ] Updates to `auto` policy behavior for virto-blk
+
+### Beta
+- [ ] Successful performance testing
+
+### GA
+- [ ] Feature gate is removed
```
xpivarc: Feature gates are only graduated

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: it's not explicitly a concern, this is just explaining the limitation of using the `auto` policy in this hotplug scenario and why we would recommend using `supplementalPool` instead.

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: updated to note that `dedicatedIOThread` is incompatible with scsi disks

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: clarified the new configuration that is being added

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: fixed

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: addressed, unless you think that the new configuration I am proposing deserves its own separate section in the design?

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: fixed

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,269 @@
...
+
+A [performance analysis](https://developers.redhat.com/blog/2025/06/23/feature-introduction-multiple-iothreads-openshift-virtualization#feature_details) was completed when multiIOThread capabilities were introduced for virtio-blk, and the study revealed that a thread pool size of 4 "can provide very significant performance improvements. In some cases up to 8 or 16 threads could provide slightly higher performance for some very fast storage environments." However, it was also found that adding more threads beyond this range did not yield better results and could potentially have a negative impact. For this reasoning, we will impose a thread pool size maximum of 8 for VMs using this auto policy. Users who wish to override this maximum can opt for the `supplementalPool` policy where an explicit pool size can be set.
```
dsanatar, line 124: @jeniferh @kevmw see new section on setting a default auto thread pool size maximum as per our discussion.

jeniferh: perfect, thank you!

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -117,14 +117,16 @@ The following is an overview of how the SCSI controller would consume the follow
...
```
akalenyu, line 124: Should we only impose it for folks that toggled MultiIOThreadAutoPolicy?

dsanatar: yes that was the idea here

akalenyu (approved):

dsanatar:
@xpivarc could you re-review this when you get the chance? I believe all outstanding comments have been addressed.

jeniferh (approved):
LGTM, thanks!

From a performance perspective, this will make "auto" policy be more effective for both bus types and should improve performance w/ good default thread count values.

fabiand:
Btw is this PR also covering virtio-blk? can ir be adapated to cover virtio.blk the same way?

dsanatar:
> Btw is this PR also covering virtio-blk? can ir be adapated to cover virtio.blk the same way?

yes, virtio-blk already supports iothread vq-mapping, the new capabilities proposed here for scsi will essentially align the two bus types for multi iothread functionality

awels:
@iholder101 @vladikr could one of you approve, it has lgtm from sig-storage.

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,269 @@
...
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
xpivarc: These can be marked

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
xpivarc: Thanks! The assumption should be tested, please add test if we don't have it

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,269 @@
...
+          bus: scsi
+        name: scsi-disk
+```
+
+Disks will be assigned IOThreads like this:
+
+```
+testdisk1: [1, 2]
+testdisk2: [1, 2]
+scsi-disk: [1, 2]
```
xpivarc, line 168: This is only with opt-in `MultiIOThreadAutoPolicy`, please have example for the other behavior as well.

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Policy: auto
+Currently, each virtio-blk disk (excluding ones that request dedicatedIO) gets assigned a single thread in round robin order from the list of available iothreads. The total number of threads is equal to the number of disks in a VMI, capped at 2 * (# of vCPUs). Since the SCSI controller cannot assign individual threads to its disks, we would instead allocate a pool of these "auto threads" (threads not reserved for dedicatedIO disks) to the controller. 
+
+In order to prevent this multiIOThread behavior from diverging for the two bus types, this implementation will also include modifications to the `auto` policy for virtio-blk so that each blk disk will now also be allocated the same auto thread pool that the SCSI controller would receive, instead of a single thread. However, to preserve backwards compatibility, a new configuration `MultiIOThreadAutoPolicy` will be added to the KubeVirt CR to act as a toggle for this new virtio-blk `auto` policy behavior. To explicitly opt-in, users must have both the feature gate as well as this new toggle enabled.
```
xpivarc, line 122: Having the `MultiIOThreadAutoPolicy` field, means we will forever support both allocation algorithms, is this intended?

xpivarc (commented):
The last major question that I have is about the need of `MultiIOThreadAutoPolicy` field

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
akalenyu: Yes - this is an implication of having to deliver a backwards compatible solution - https://github.com/kubevirt/enhancements/pull/343#discussion_r3691984113. 

We can't flip existing users off to the new allocation policy
(however tempting, there are some cases where we can't know for sure someone is better off with new allocation)

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: ack, added example of when the default single iothread auto policy is in use

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: we considered the idea of adding a separate feature gate for this new behavior however, like Alex mentioned there may be users that wish to stay with the legacy single iothread version even after we decide to GA this feature. we figured instead of tracking and graduating two FGs at different rates, a standalone toggle would suffice.

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: marked

veps/sig-storage/342-iothread-vq-virtio-scsi/vep.md
dsanatar: made a note of this in the testing section, will make sure it's also accounted for in the STP.

Issue #344, opened 2026-06-18
VEP 344: VM-Scoped Persistent ResourceClaims for DRA Device Passthrough


johnahull:
## Summary

This VEP adds a `resourceClaimTemplates` field to `VirtualMachineSpec`, enabling DRA ResourceClaims to be owned by the VM instead of the VMI pod. Device allocations persist across VM restarts, following the established `dataVolumeTemplates` pattern.

## Motivation

When using DRA with KubeVirt today, ResourceClaims are pod-scoped — restarting a VM means a new claim and potentially a different physical device. This causes data loss for NVMe drives, driver instability for GPUs, and allocation race conditions in busy clusters.

## VEP Directory

`veps/sig-compute/<this-issue-number>-vm-scoped-persistent-resource-claims/`

## Implementation PR

https://github.com/kubevirt/kubevirt/pull/17957

## Prior Art

- [GSoC 2024: Persistent Device Claims for KubeVirt](https://github.com/kubevirt/community/issues/254)

/sig compute
/kind feature

PR #345, opened 2026-06-18
VEP #344: VM-Scoped Persistent ResourceClaims for DRA Device Passthrough

johnahull:
## Summary

- Adds VEP proposing `resourceClaimTemplates` on `VirtualMachineSpec` so DRA ResourceClaims persist across VM restarts
- Follows the established `dataVolumeTemplates` pattern for consistency
- Covers API changes, controller behavior, webhook validation, alternatives, and graduation criteria

## Enhancement Issue

https://github.com/kubevirt/enhancements/issues/344

## Implementation PR

https://github.com/kubevirt/kubevirt/pull/17957

## Prior Art

- [GSoC 2024: Persistent Device Claims for KubeVirt](https://github.com/kubevirt/community/issues/254)

/sig compute
/kind feature

```release-note
VEP #344: Propose VM-scoped persistent ResourceClaims for DRA device passthrough
```

### DIFF
--- a/veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
+++ b/veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
@@ -1,0 +1,406 @@
+# VEP #344: VM-Scoped Persistent ResourceClaims for DRA Device Passthrough
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10.0
+- This VEP targets beta for version: TBD
+- This VEP targets GA for version: TBD
+
+### Release Signoff Checklist
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP introduces VM-scoped persistent ResourceClaims for Kubernetes Dynamic
+Resource Allocation (DRA). Today, KubeVirt supports DRA-provisioned devices
+(GPUs, host devices) at the VMI level, where ResourceClaims are created
+per-Pod and deleted when the Pod is removed. This means that restarting a VM
+results in a new ResourceClaim and potentially a different physical device.
+
+This proposal adds a `resourceClaimTemplates` field to `VirtualMachineSpec`
+that allows the VM controller to pre-create ResourceClaims owned by the VM
+object. These claims persist across VMI restarts and are only garbage-collected
+when the VM itself is deleted, ensuring device allocation stability.
+
+## Prior Art
+
+This VEP builds on the work identified in the [GSoC 2024: Persistent Device
+Claims for KubeVirt](https://github.com/kubevirt/community/issues/254) project
+proposal. That issue described the core problem — device allocations are lost
+when VM pods are recreated — and proposed using Kubernetes Dynamic Resource
+Allocation (DRA) ResourceClaims as the persistence mechanism. This VEP
+implements that vision by integrating persistent ResourceClaims into the VM
+controller, following the established `dataVolumeTemplates` pattern.
+
+## Motivation
+
+When using DRA with KubeVirt today, ResourceClaims are scoped to the
+virt-launcher Pod. Each time a VM is restarted (stop/start, reboot, crash
+recovery), the old Pod and its ResourceClaim are deleted, and a new claim is
+created. This causes several problems:
+
+1. **Data loss for stateful devices**: NVMe drives and other storage devices
+   provisioned via DRA contain persistent data. Losing the device allocation on
+   restart means losing access to that data.
+
+2. **Device instability**: GPU passthrough users may depend on a specific
+   physical device for driver compatibility, licensing, or workload continuity.
+   A different GPU after restart can cause driver failures or license
+   revalidation.
+
+3. **Allocation failures in busy clusters**: In the window between the old
+   claim being deleted and the new claim being created, another workload can
+   claim the device. This creates a race condition where a VM restart can fail
+   due to resource exhaustion, even though the device was available moments
+   before.
+
+4. **Inconsistency with other VM-scoped resources**: KubeVirt already supports
+   VM-scoped DataVolumes via `dataVolumeTemplates`, which persist across VMI
+   restarts. DRA devices lack an equivalent mechanism, creating an asymmetry in
+   the resource lifecycle model.
+
+## Goals
+
+- Allow users to declare ResourceClaims that are scoped to the VM lifecycle,
+  not the VMI/Pod lifecycle.
+- Ensure DRA device allocations persist across VM restarts (stop/start, reboot,
+  crash recovery).
+- Support all DRA device types (GPUs, host devices, and future DRA-managed
+  devices such as SR-IOV NICs) with a single, device-agnostic mechanism.
+- Follow established KubeVirt patterns (`dataVolumeTemplates`) for consistency
+  and maintainability.
+- Provide webhook validation to catch misconfigurations at admission time.
+
+## Non Goals
+
+- Eager allocation (creating ResourceClaims before the VM is started). The
+  current implementation creates claims at start time, matching the
+  `dataVolumeTemplates` pattern. Eager allocation with a per-entry policy
+  (e.g., `allocationMode: Eager | Lazy`) is a potential future enhancement.
+- Migration-aware ResourceClaim handling. Live migration with DRA devices is
+  out of scope for this VEP.
+- Managing ResourceClaimTemplates themselves. Users are responsible for creating
+  the `ResourceClaimTemplate` objects that this feature references.
+
+## Definition of Users
+
+- **VM administrators** who manage long-lived VirtualMachines with DRA-provisioned
+  hardware (GPUs, NVMe, host devices).
+- **Platform engineers** who configure DRA drivers and ResourceClaimTemplates
+  for their clusters.
+
+## User Stories
+
+1. As a VM administrator, I want my VM's GPU assignment to survive restarts so
+   that the guest OS driver stack remains consistent and I don't need to
+   reconfigure the device after each reboot.
+
+2. As a VM administrator, I want my VM's NVMe device to persist across restarts
+   so that data stored on the device is not lost when the VM is stopped and
+   restarted.
+
+3. As a VM administrator, I want my VM to reliably restart in a busy cluster
+   without risking allocation failure because another workload grabbed my
+   device during the restart window.
+
+4. As a platform engineer, I want a single mechanism that works for all
+   DRA-managed device types (GPUs, NVMe, SR-IOV NICs) without needing
+   device-specific configuration in KubeVirt.
+
+## Repos
+
+- `kubevirt/kubevirt` — API types, VM controller, webhook validation, RBAC
+
+## Design
+
+### API Changes
+
+A new type `ResourceClaimTemplateEntry` and a new field `resourceClaimTemplates`
+on `VirtualMachineSpec`:
+
+```go
+// ResourceClaimTemplateEntry defines a ResourceClaim that should be created
+// from a ResourceClaimTemplate and bound to this VirtualMachine's lifecycle.
+type ResourceClaimTemplateEntry struct {
+    // Name is the logical name used to match this entry to
+    // spec.template.spec.resourceClaims[].name in the VMI template.
+    Name string `json:"name"`
+    // ResourceClaimTemplateName is the name of a ResourceClaimTemplate
+    // object in the same namespace to create the ResourceClaim from.
+    ResourceClaimTemplateName string `json:"resourceClaimTemplateName"`
+}
+
+type VirtualMachineSpec struct {
+    // ... existing fields ...
+
+    // resourceClaimTemplates is a list of ResourceClaims that should be
+    // created from ResourceClaimTemplates and are tied to the
+    // VirtualMachine's lifecycle.
+    // +kubebuilder:validation:MaxItems:=256
+    // +listType=map
+    // +listMapKey=name
+    // +optional
+    ResourceClaimTemplates []ResourceClaimTemplateEntry `json:"resourceClaimTemplates,omitempty"`
+}
+```
+
+### Controller Behavior
+
+The VM controller manages ResourceClaims following the same pattern as
+`dataVolumeTemplates`:
+
+1. **Claim creation** (`handleResourceClaims`): Called during `startVMI()`,
+   before the VMI is created. For each entry in `resourceClaimTemplates`:
+   - Derives the claim name as `<vm-name>-<entry-name>`
+   - Checks if the ResourceClaim already exists in the informer cache
+   - If not, looks up the ResourceClaimTemplate, creates the ResourceClaim with
+     the VM as owner (via `OwnerReference`), and records controller expectations
+
+2. **VMI setup** (`SetupVMIFromVM`): When constructing the VMI from the VM
+   template, any `resourceClaims` entry that has a `resourceClaimTemplateName`
+   matching a `resourceClaimTemplates` entry is rewritten to use a direct
+   `resourceClaimName` pointing to the pre-created claim. This ensures the Pod
+   spec references the persistent claim rather than a template.
+
+3. **Lifecycle management**: ResourceClaim informer event handlers
+   (`add`/`update`/`delete`) resolve owner references back to the VM and
+   re-enqueue it for reconciliation, using the expectations pattern to prevent
+   duplicate operations.
+
+4. **Garbage collection**: ResourceClaims have the VM as their controller owner
+   reference. Kubernetes garbage collection automatically deletes them when the
+   VM is deleted.
+
+### Claim Lifecycle
+
+```
+VM created (Halted)     → no claims created
+VM started              → claims created, VMI created referencing them
+VM stopped              → VMI deleted, claims persist (owned by VM)
+VM restarted            → claims already exist, VMI reuses them
+VM deleted              → claims garbage-collected via owner ref
+```
+
+### Webhook Validation
+
+The validating webhook (`validateResourceClaimTemplates`) enforces:
+
+- `GPUsWithDRA` or `HostDevicesWithDRA` feature gate must be enabled
+- No duplicate `name` entries
+- Both `name` and `resourceClaimTemplateName` are required
+- Each entry must have a matching `spec.template.spec.resourceClaims[]` entry
+  by name
+
+### RBAC
+
+The virt-controller service account requires:
+- `resourceclaims`: get, list, watch, create, update, delete, patch
+- `resourceclaimtemplates`: get, list, watch
+
+## API Examples
+
+### VM with persistent GPU claim
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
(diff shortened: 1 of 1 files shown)


johnahull:
Friendly ping. This is ready for review when someone has bandwidth.

enp0s3:
Hi @johnahull! Thank you for heads up. We are running a bi-monthly VEP discussion meetings, you can share your VEP there as well.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,406 @@
...
+```
+
+### Controller Behavior
+
+The VM controller manages ResourceClaims following the same pattern as
+`dataVolumeTemplates`:
+
+1. **Claim creation** (`handleResourceClaims`): Called during `startVMI()`,
+   before the VMI is created. For each entry in `resourceClaimTemplates`:
+   - Derives the claim name as `<vm-name>-<entry-name>`
```
jean-edouard: This can easily get longer than max k8s name length, what do we do then?

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,406 @@
...
+## Update/Rollback Compatibility
+
+- The `resourceClaimTemplates` field is optional. Existing VMs without the
+  field are unaffected.
+- On rollback, orphaned ResourceClaims (created by the new controller) will
+  remain until manually deleted or until the VM is deleted (owner reference GC
+  still works regardless of the controller version).
+- The feature is gated behind the existing `GPUsWithDRA` / `HostDevicesWithDRA`
+  feature gates. No new feature gate is introduced since this is an extension
+  of existing DRA functionality.
```
jean-edouard: I don't think that's a strong-enough reason not to introduce a new FG. This VEP introduces a new API field, seems to me like we can cleanly reject the new field when some new FG is off. WDYT?

jean-edouard (commented):
I think this is very valuable to the project. My main concern is around the feature-gate topic, see in-line comment.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: Good catch. How about I add validation in the webhook to reject entries where len(vm.Name) + 1 + len(entry.Name) exceeds 63 characters, with a clear error message? This matches how KubeVirt handles DataVolume name validation. Or open to suggestions

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: I went back and forth on that one, so I'm fine either way. I can add a FG "PersistentDRAClaims"

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,413 @@
...
+- **Platform engineers** who configure DRA drivers and ResourceClaimTemplates
+  for their clusters.
+
+## User Stories
+
+1. As a VM administrator, I want my VM's GPU assignment to survive restarts so
+   that the guest OS driver stack remains consistent and I don't need to
+   reconfigure the device after each reboot.
+
+2. As a VM administrator, I want my VM's NVMe device to persist across restarts
```
enp0s3, line 106: @johnahull Besides the NVMe case which makes sense to stay owned by the VM when the VM is off, why would we need to reserve GPU or SRIOV resources for shutoff VMs? Isn't it a waste of resources that could be used by other active workloads while the VM is off?

enp0s3 (commented):
@johnahull Thank you for the contribution! Looks like the right direction.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: I'm glad you brought this up. This mirrors dataVolumeTemplates, where DataVolumes persist for the VM's lifetime regardless of stop/start. However, I meant to go back and make the default behavior to only persist on a reboot, but add a configurable `persistWhenStopped` field per entry for cases that need it. I still need to do that.

Some reasons a user might want to hold a claim when a VM is stopped:

- GPU software licensed to a specific PCI address. If you stop the VM overnight and start it in the morning, you get the same GPU. Otherwise you might get a different one and need to re-license.
- NIC with a specific MAC address that has DHCP reservations, firewall rules, or monitoring tied to it. Stop/start shouldn't change the MAC.
- Simply wanting device stability from VM creation to deletion without managing re-allocation.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,462 @@
...
+    // on stop and re-created on the next start — freeing the device for
+    // other workloads. When true, the claim persists for the entire VM
+    // lifetime, guaranteeing the same device across stop/start cycles.
+    //
+    // Claims always persist across reboots regardless of this setting,
+    // since the VMI pod is recreated without an intervening stop.
+    //
+    // +optional
+    // +kubebuilder:default=false
+    PersistWhenStopped bool `json:"persistWhenStopped,omitempty"`
```
jean-edouard: This struct looks quite different than the one in the implementation PR.
Most notably, this is a `*bool` there. Fun how the implementation PR has validation for the first 2 fields and here only the 3rd one does :)
Probably worth aligning both.

jean-edouard (approved):

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: Fixed in a1c8046:
* PersistWhenStopped is now `*bool` 
- Added `+kubebuilder:validation:MinLength=1` on `Name` and `ResourceClaimTemplateName` to match the implementation

johnahull:
Updated VEP with the `reservedFor` allocation preservation mechanism, based on hardware testing.

**Key addition:** When `persistWhenStopped: true`, the VM controller adds the VM as a consumer in the ResourceClaim's `status.reservedFor`. This is what actually prevents device deallocation — without it, the DRA resource controller deallocates the claim when the pod is deleted on stop, and the VM may get a different physical device on restart.

**How it works:**
1. Claim created and allocated by scheduler → pod added to `reservedFor`
2. Controller detects allocation → adds VM to `reservedFor`
3. VM stopped → pod removed from `reservedFor`, VM entry remains → allocation preserved
4. VM restarted → new pod uses same allocated device
5. VM deleted → controller removes VM from `reservedFor` → DRA deallocates → GC deletes claim

**RBAC:** Requires `resourceclaims/binding: patch` permission (Kubernetes uses the `/binding` subresource for `reservedFor` modifications, not `/status`).

**Tested on Dell XE9680:** 4 VMs with 4 NVMe drives, 10 stop/start + 10 reboot cycles each. Each VM held its exact physical device across all 80 cycles with zero swaps.

Implementation PR: kubevirt/kubevirt#17957

enp0s3:
> 5. VM deleted → controller removes VM from `reservedFor` → DRA deallocates → GC deletes claim

@johnahull What about The following scenarios:
* User wants to trigger manual deallocation from a non-running VM.
* User isn't satisfied from the allocation and wants to re-allocate the resource.

johnahull:
> > 5. VM deleted → controller removes VM from `reservedFor` → DRA deallocates → GC deletes claim
> 
> @johnahull What about The following scenarios:
> 
> * User wants to trigger manual deallocation from a non-running VM.
> * User isn't satisfied from the allocation and wants to re-allocate the resource.

Good question. The user can force reallocation by deleting the claim directly:

  kubectl delete resourceclaim my-vm-claim
  virtctl start my-vm

  The controller recreates missing claims on start, so the scheduler allocates fresh. I just tested and this works, doesn't require any changes

If we want to add an API option, we could add something like a **reallocateOnNextStart** field:

  ```
  resourceClaimTemplates:
  - name: nvme
    resourceClaimTemplateName: nvme-tmpl
    persistWhenStopped: true
    reallocateOnNextStart: true
```
One-shot boolean. Controller deletes the claim when it sees this on a stopped VM, clears the flag. Next start gets fresh allocation. 

Feedback?

File: veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md [hunk 5]
```diff
@@ -0,0 +1,513 @@
...
+## Overview
+
+This VEP introduces VM-scoped persistent ResourceClaims for Kubernetes Dynamic
+Resource Allocation (DRA). Today, KubeVirt supports DRA-provisioned devices
+(GPUs, host devices) at the VMI level, where ResourceClaims are created
+per-Pod and deleted when the Pod is removed. This means that restarting a VM
+results in a new ResourceClaim and potentially a different physical device.
+
+This proposal adds a `resourceClaimTemplates` field to `VirtualMachineSpec`
+that allows the VM controller to pre-create ResourceClaims owned by the VM
```
EdDev: I would suggest that in the overview, motivation and goals sections, details about the solutions will not be mentioned.
The abstract logic of preserving claims for the lifecycle of a VM object should be enough, letting the solution section to discuss the HOW.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,590 @@
...
+    // +listType=map
+    // +listMapKey=name
+    // +optional
+    ResourceClaimTemplates []ResourceClaimTemplateEntry `json:"resourceClaimTemplates,omitempty"`
+}
+```
+
+### Kubernetes DRA Background: `status.reservedFor`
+
+Kubernetes DRA ResourceClaims have a standard `status.reservedFor` field
```
alaypatel07, line 171: The status of resource claims is strictly managed my kubernetes.  Having an external component like kubevirt manage the status could lead to other issues.

If we are going this route, we need explicit buy-in from wg-device-management that the reservedFor field is flexible and will stay flexible for non-pod resources. Likely we need to also add e2e/integration/unit tests in kubernetes to make sure non-pod resources in reservedFor field does not break anything.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,590 @@
...
+
+## Repos
+
+- `kubevirt/kubevirt` — API types, VM controller, webhook validation, RBAC
+
+## Design
+
+### API Changes
+
+A new type `ResourceClaimTemplateEntry` and a new field `resourceClaimTemplates`
```
EdDev, line 129: I understand that this follows the path of `dataVolumeTemplates`, however I would like to try and challenge it.
AFAIU this pattern allows to define and express similar needs through two different API/s. This is both confusing to users and requires nontrivial logic from consumers.

If this is also something which we can relate to pods in general, a stronger motivation is given to keep this out of the VM and prefer something generic to pods as well.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
EdDev: I like this idea, it is basically a ref-count.
But I agree, it needs to be an agreed generic API, and having this appearing on the status hints it is not to be touched by other controllers.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,590 @@
...
+      resourceClaims:
+      - name: gpu-claim
+        resourceClaimTemplateName: gpu-template
+      - name: nvme-claim
+        resourceClaimTemplateName: nvme-template
+```
+
+## Alternatives
+
+### 1. Eager allocation with per-entry policy
```
EdDev, line 441: Do you mean adding a policy under `VirtualMachineInstanceResourceClaim`?

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,590 @@
...
+   DRA-managed device types (GPUs, NVMe, DRA-backed NICs) without needing
+   device-specific configuration in KubeVirt.
+
+## Repos
+
+- `kubevirt/kubevirt` — API types, VM controller, webhook validation, RBAC
+
+## Design
+
+### API Changes
```
EdDev, line 127: As the need here is for Kubevirt to manage these claims, how about having a dedicated controller that does this as an optional addition. This way, the resourceClaim lifecycle is managed as suggested here, but by explicitly creating them in parallel to the VM/VMI lifecycle and considering the type of claim requested.
The current suggestion here is to express this through a list of claims on the VM spec. I would prefer to see an external hint about the need of the claim to be managed by Kubevirt for the duration of the VM.
What I mean by a hint is: Define for which VMs (based on a label and possibly a configuration pattern) a resourceClaim should be managed by that special add-on controller.

EdDev (commented):
Thank you, an interesting topic.

I find the pattern of using the VM spec in parallel to the VMI spec challenging. I think it is problematic to express similar needs through two API/s.

It will be nice to discuss this in the DRA bi-weekly meeting.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: The idea I had in mind was to have this on the VM-level entry (e.g., ResourceClaimTemplateEntry), since it controls when the claim is created relative to the VM lifecycle

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
johnahull: Talked to the device management working group. They said that the use case we're trying to use reservedFor is supported, and to try it and let them know if there are issues

johnahull:
**Alternative design for consideration: `reservedFor`-based persistence**

cc @EdDev @alaypatel07 @enp0s3 @lyarwood @vladikr @jean-edouard

I want to put an alternative design on the table alongside the current one and get the group's read. I'm genuinely fine with either direction; the current `resourceClaimTemplates` approach works, but after validating this alternative against upstream I think it's worth weighing before we commit. Here's the shape of it.

**What's different from the current design**

Instead of a new `resourceClaimTemplates` field on `VirtualMachineSpec`, the only API change would be a single per-entry field on the existing `VirtualMachineInstanceResourceClaim` type:

```go
type VirtualMachineInstanceResourceClaim struct {
    ResourceClaimName         *string
    ResourceClaimTemplateName *string
    PersistWhenStopped        *bool  // new
}
```

The policy sits next to the reference it governs. There's no new top-level VM field and no duplicated declaration.

**Mechanism: virt-controller manages `status.reservedFor` only**

Claims are created exactly as they are today, by the scheduler/kubelet from the VMI's `spec.resourceClaims`. KubeVirt neither creates, names, owns, nor deletes claim objects. It writes exactly one field: `status.reservedFor`.

On each reconcile, virt-controller adds the VM as a consumer reference (`Resource: virtualmachines`, VM name/UID) to each allocated claim the VMI uses. The VM is added to `reservedFor` on every allocated claim, always, and that alone is what makes reboots persist. `persistWhenStopped` changes only one thing: whether the entry is removed on an explicit stop.

- **Reboot** (pod dies, VM stays): the DRA controller prunes the pod's `reservedFor` entry, but the VM entry remains, so the allocation is never cleared and the same device returns on restart. This is the default for all claims; no field required.
- **Stop:** `persistWhenStopped: true` keeps the VM entry (allocation preserved across stop/start); the default removes it (device freed).
- **Delete:** virt-controller removes the VM from `reservedFor` on all claims (guarded by a VM finalizer).

No new controller is needed for alpha; this folds into virt-controller's existing VM reconcile.

**Why it's safe (verified against k8s master, `v1.38.0-alpha.0`)**

- Deallocation is gated solely on `reservedFor` emptiness plus the builtin finalizer; ownerReferences never enter the decision (`pkg/controller/resourceclaim/controller.go:1624-1633`).
- The controller's `reservedFor` validation only prunes `pods`/`podgroups` entries and explicitly keeps any other consumer type ("we just keep it to avoid accidentally removing a reservation for an unknown consumer type"), at `controller.go:1594-1598`. So a `virtualmachines` entry is durable. This matches wg-device-management's confirmation that third-party writers to `reservedFor` are supported.

**How this would scope for alpha**

Support `persistWhenStopped` for **direct** claims (`resourceClaimName`) now, and **managed** claims (VEP-300) if/when available. The mechanism applies identically since neither is pod-owned, so `reservedFor` alone is sufficient and race-free. Both are referenced by a concrete claim name, so the persistence path is literally the same code.

Note this isn't a subset of VEP-300: persistence applies to *direct* claims a user brings, which VEP-300 never creates or manages. The two are orthogonal. VEP-300 governs claim *creation*; this VEP governs allocation *lifetime*. A managed claim can be persistent or not, and a direct claim can be persistent or not, so neither VEP subsumes the other. The natural relationship is layered: this VEP provides the `reservedFor` lifetime mechanism, and VEP-300 consumes it once it starts producing managed claims.

Under this alternative, template claims (`resourceClaimTemplateName`) would move to beta. They're the only claims that are kubelet-created and pod-owned, and they need extra machinery on two counts: the concrete claim name isn't in the spec (it must be read from `pod.status.resourceClaimStatuses`), and the generic garbage collector, which keys on ownerReferences and ignores `reservedFor`, deletes the claim when the pod dies. Persisting them requires adding the VM as a *second* ownerReference (addition, not replacement), which introduces a create-before-pod-death race.

**What I'd like your read on**

1. Between this and the current `resourceClaimTemplates` design, do you have a preference?
2. Does per-entry `persistWhenStopped` on `VirtualMachineInstanceResourceClaim` read well as the API surface?
3. If we went this way, is alpha scoped to direct and managed claims (templates in beta) acceptable?
4. Any concerns with virt-controller as a third-party writer to `reservedFor` vs. a dedicated controller? (@EdDev)

Happy to go either way; I just want to make sure we've compared them before locking in.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,587 @@
...
+- Each entry must have a matching `spec.template.spec.resourceClaims[]` entry
+  by name
+- The derived claim name (`<vm-name>-<entry-name>`) must not exceed 63
+  characters (the Kubernetes object name length limit)
+
+### RBAC
+
+The virt-controller service account requires:
+- `resourceclaims`: get, list, watch, create, delete
+- `resourceclaims/binding`: patch
```
jean-edouard, line 319:
```suggestion
- `resourceclaims/binding`: update
```
More below.

veps/sig-compute/344-vm-scoped-persistent-resource-claims/vep.md
```diff
@@ -0,0 +1,587 @@
...
+DRA ResourceClaims.
+
+**Rejected** because PVCs are a storage abstraction — they don't cover GPUs,
+DRA-backed NICs, FPGAs, or other non-storage devices. DRA ResourceClaims are the
+Kubernetes-standard API for dynamic device allocation, and they carry device
+metadata (NUMA node, PCIe topology, driver attributes) that PVCs lack. Even
+for NVMe, DRA VFIO passthrough gives direct hardware access without the
+filesystem/block driver overhead that CSI imposes.
+
+### 5. Dual declaration of resourceClaimTemplates
```
jean-edouard, line 484: Does this section belong to `## Alternatives`?

Issue #340, opened 2026-06-17
VEP 340: launchsecurity: Add InitData resource for TDX and SNP


MatiasVara:
**Primary contact (assignee)**:

/assign @MatiasVara 

**Current Feature Stage**:

**Feature Gate**: InjectInitData

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
                   - https://github.com/kubevirt/enhancements/pull/346
  * [ ] Code PRs:
                   - https://github.com/kubevirt/kubevirt/pull/18064
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #346, opened 2026-06-19
VEP #340: launchsecurity: Add InitData resource for TDX and SNP

MatiasVara:
Add InitData resource for TDX and SNP

### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/340
**SIG label**: /sig compute

### What this PR does

Add VEP about the injection of mrConfig (TDX) and hostData(SEV-SNP). These registers are used to carry a checksum of the initData used during attestation (see https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md)

### DIFF
--- a/veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
+++ b/veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
@@ -1,0 +1,364 @@
+# VEP #340: launchsecurity: add injectInitdata subresource for TDX and SNP
+
+## VEP Status Metadata
+
+### Target releases
+
+This VEP does not have graduation phases guarded by a feature gate.
+- Ship target version:
+
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
+
+Confidential VMs (CVMs) require per-VM configuration values to be injected
+before boot so the guest can use them during remote attestation. These values
+are incorporated into hardware measurement registers and become part of the
+attestation report, allowing the guest to prove that the host provided the
+expected configuration.
+
+This proposal adds two new KubeVirt subresource endpoints that allow an
+external operator to inject these values into a VMI while it is in the
+`Scheduled` phase, before QEMU starts:
+
+- `PUT virtualmachineinstances/{name}/tdx/injectInitdata` — for Intel TDX
+- `PUT virtualmachineinstances/{name}/snp/injectInitdata` — for AMD SEV-SNP
+
+The design follows the existing SEV attestation pattern where `virt-handler`
+blocks VM startup until the required attestation data is provided via a
+subresource, and the VM is explicitly unpaused by the external operator
+after injection.
+
+In this VEP, Initdata wording refers to
+[initData](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md).
+The Initdata Specification defines the key data structure and algorithms to
+inject any well-defined data from untrusted host into TEE.
+
+## Motivation
+
+Organizations deploying confidential VMs need the strongest possible
+attestation guarantees. A key use case is encrypted disk provisioning: an
+external operator injects a reference to a secret stored in a Key Broker
+Service (KBS) into the guest before boot. The guest then attests to the KBS
+and retrieves the secret only if the attestation passes. The injected values
+become part of the hardware measurement, ensuring the host cannot tamper with
+them undetected.
+
+Currently, KubeVirt supports this flow for SEV via the
+`sev/fetchCertChain` and `sev/setupSession` subresources. However, Intel TDX
+and AMD SEV-SNP have no equivalent mechanism. Users resort to hook sidecars
+that modify the libvirt domain XML directly, bypassing KubeVirt's API and
+losing auditability, RBAC control, and compatibility with future KubeVirt
+versions.
+
+This proposal fills the gap by extending the VMI API with the fields and
+subresources needed for TDX and SEV-SNP attestation-based secret delivery.
+
+## Goals
+
+- Expose TDX and SEV-SNP initdata fields in the VMI API so users can
+  configure attestation-relevant registers (`mrConfigId`, `hostData`,
+  `oemStrings`) declaratively or at runtime.
+- Provide a KubeVirt-native mechanism for external operators to inject
+  initdata into a VMI after scheduling and before boot.
+- Extend the existing SEV attestation pattern to TDX and SEV-SNP.
+
+## Non Goals
+
+- This proposal does not implement the external operator that calls the
+  subresource. The operator is a separate component outside of KubeVirt.
+- This proposal does not define how the guest uses the injected values after
+  boot (that is guest-side logic, e.g., `systemd-repart`, `systemd-cryptsetup`).
+
+## Definition of Users
+
+- Cluster Administrators: Deploy and manage the external operator that
+  automates the attestation provisioning flow. Configure RBAC permissions for
+  the `injectInitdata` subresource.
+
+- VM Users: secret delivery flow to work transparently once the cluster
+  administrator has configured the external operator.
+
+## User Stories
+
+- As a VM user, I want to configure initdata registers like `mrConfigId`,
+  `hostData`, and `oemStrings` in the VM definition so that they are
+  injected into the VM at boot time.
+- As a VM user, I want to tell the Kubernetes infrastructure to inject a
+  secret key into my confidential VM.
+- As a cluster administrator, I want to configure an external operator
+  that automates the attestation provisioning flow so that VM users do not
+  need to manage the injection manually.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+The design follows the existing SEV attestation pattern in KubeVirt. The key
+components are:
+
+### API extensions
+
+New fields added to the VMI spec:
+
+```go
+// In schema.go
+type TDX struct {
+    MRConfigId  string          `json:"mrConfigId,omitempty"`
+    Attestation *TDXAttestation `json:"attestation,omitempty"`
+}
+
+type SEVSNP struct {
+    HostData    string              `json:"hostData,omitempty"`
+    Attestation *SEVSNPAttestation  `json:"attestation,omitempty"`
+}
+
+type Firmware struct {
+    // ... existing fields ...
+    OEMStrings []string `json:"oemStrings,omitempty"`
+}
+```
+
+The `Attestation` field is a marker: when present, it signals that the VMI
+requires initdata injection before boot. Without `Attestation`, the registers
+can be set directly from the YAML. `mrConfigId`/`hostData` start empty and are
+filled via the subresource. `oemStrings` carries a reference (e.g., a KBS
+resource path) that the guest reads from SMBIOS Type 11 at boot.
+
+### Subresource endpoints
+
+Two new `PUT` endpoints registered in `virt-api`:
+
+| Endpoint | Platform | Payload |
+|---|---|---|
+| `PUT .../virtualmachineinstances/{name}/tdx/injectInitdata` | Intel TDX | `{"mrConfigId": "<base64>", "oemStrings": [...]}` |
+| `PUT .../virtualmachineinstances/{name}/snp/injectInitdata` | AMD SEV-SNP | `{"hostData": "<base64>", "oemStrings": [...]}` |
+
+Both endpoints:
+1. Verify the VMI is in `Scheduled` phase.
+2. Verify the `Attestation` field is set.
+3. Verify the initdata field (`mrConfigId`/`hostData`) is not already set.
+4. Validate the base64 value decodes to the correct length (48 bytes for TDX,
+   32 bytes for SNP).
+5. Apply a JSON patch to the VMI spec.
+
+### virt-handler blocking
+
+When `TDX.Attestation` (or `SEVSNP.Attestation`) is set, `virt-handler`
+blocks VM synchronization until `mrConfigId` (or `hostData`) is populated.
+This prevents QEMU from starting before the external operator has injected
+the values. The check mirrors the existing `shouldWaitForSEVAttestation`
+function.
+
+### Libvirt XML translation
+
+The new fields are translated to libvirt domain XML in the converter layer:
+
+- `TDX.MRConfigId` -> `<launchSecurity type="tdx"><mrConfigId>...</mrConfigId></launchSecurity>`
+- `SEVSNP.HostData` -> `<launchSecurity type="sev-snp"><hostData>...</hostData></launchSecurity>`
+- `Firmware.OEMStrings` -> `<sysinfo type="smbios"><oemStrings><entry>...</entry></oemStrings></sysinfo>`
+
+### Start strategy
+
+The VMI must be created with `startStrategy: Paused`. This ensures QEMU
+boots in paused state, giving the external operator a window to:
+1. Call the provisioner to obtain the values.
+2. Inject them via `injectInitdata`.
+3. Wait for the VMI to reach `Running` phase (QEMU started but paused).
+4. Call `unpause` to resume guest execution.
+
+This is validated at admission time: if `Attestation` is set but
+`startStrategy` is not `Paused`, the VMI is rejected.
+
+### Security Considerations
+
+- The `injectInitdata` subresource is guarded by RBAC. Only components with
+  the appropriate ClusterRole can call the endpoint. This prevents unauthorized
+  injection of attestation values.
+- The injected `mrConfigId` and `hostData` are measured by the hardware (TDX
+  or SEV-SNP respectively). Any tampering is detectable by the guest during
+  remote attestation. The host infrastructure provider cannot modify these
+  values after injection without invalidating the attestation report.
+- The `oemStrings` field carries a reference (e.g., a KBS resource path), not
+  the actual secret. The secret is only released by the Key Broker Service
+  after the guest passes attestation.
+- The design does not store secrets in Kubernetes objects. The VMI spec only
+  contains measurement IDs and resource paths.
+
+## API Examples
+
+### TDX with static mrConfigId
+
+When `mrConfigId` is known at VM creation time, it can be set directly in
+the YAML. No `attestation` field, no `startStrategy: Paused`, and no
+subresource call is needed:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  template:
+    spec:
+      domain:
+        launchSecurity:
+          tdx:
+            mrConfigId: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+        firmware:
+          bootloader:
+            efi:
+              secureBoot: false
+```
+
+### SNP with static hostData
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  template:
+    spec:
+      domain:
+        launchSecurity:
+          sevSnp:
+            hostData: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
+        firmware:
+          bootloader:
+            efi:
+              secureBoot: false
+```
+
+### TDX with dynamic injection (external operator)
+
+When the values are provisioned at runtime by an external operator, the VMI
+is created with `attestation` and `startStrategy: Paused`. The operator
+calls the subresource after the VMI reaches the `Scheduled` phase:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  template:
+    spec:
+      startStrategy: Paused
+      domain:
+        launchSecurity:
+          tdx:
+            attestation: {}
+        firmware:
+          bootloader:
+            efi:
+              secureBoot: false
+```
+
+```bash
+curl -X PUT \
+  /apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachineinstances/my-vm/tdx/injectInitdata \
+  -H "Content-Type: application/json" \
+  -d '{
+    "mrConfigId": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "oemStrings": ["kbs:///default/uuid/root"]
+  }'
+```
+
+### SNP with dynamic injection (external operator)
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+spec:
+  template:
+    spec:
+      startStrategy: Paused
+      domain:
+        launchSecurity:
+          sevSnp:
+            attestation: {}
+        firmware:
+          bootloader:
+            efi:
+              secureBoot: false
+```
+
+```bash
+curl -X PUT \
+  /apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachineinstances/my-vm/snp/injectInitdata \
+  -H "Content-Type: application/json" \
+  -d '{
+    "hostData": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
+    "oemStrings": ["kbs:///default/uuid/root"]
+  }'
+```
+
+## Alternatives
+
+### Hook sidecar
+
+The current approach used before this proposal. A hook sidecar intercepts
+the libvirt domain XML at VMI creation time and modifies it directly to inject
+`mrConfigId`/`hostData` and OEM strings.
+
+**Drawbacks:** bypasses KubeVirt's API, not auditable, no RBAC, fragile
+against KubeVirt upgrades that change the XML structure, requires a sidecar
+container per VMI.
+
+### KubeVirt-internal provisioner
+
+KubeVirt itself could contact the provisioner service instead of relying on
+an external operator.
+
+**Drawbacks:** couples KubeVirt to a specific provisioner implementation and
+contradicts the existing SEV model where KubeVirt only provides plumbing.
+
+## Scalability
+
+The subresource adds one `PUT` call per VMI creation (or two if counting
+`unpause`). This is comparable to the existing SEV attestation flow and does
+not introduce new watchers or polling loops in KubeVirt itself. The external
+operator's scalability is its own concern.
+
+## Update/Rollback Compatibility
+
+- The new fields (`mrConfigId`, `hostData`, `oemStrings`, `attestation`) are
+  optional and default to empty/nil. Existing VMI specs are unaffected.
+- The subresource endpoints are guarded by the existing feature gates
+  `WorkloadEncryptionTDX` and `WorkloadEncryptionSEVSNP`. Disabling the
+  feature gate removes access to the endpoints without affecting existing VMs.
+- Rollback to a version without this feature is safe: the new fields are
+  ignored by older versions and VMs that do not set `attestation` behave
+  identically to before.
+
+## Functional Testing Approach
+
+- Unit tests for the `TDXInjectInitdataHandler` and `SNPInjectInitdataHandler`:
+  valid injection, duplicate injection rejection, wrong phase rejection,
+  invalid base64, wrong byte length.
+- Unit tests for `shouldWaitForTDXAttestation` and
+  `shouldWaitForSNPAttestation` in `virt-handler`.
+- Unit tests for libvirt XML converter verifying `mrConfigId`, `hostData`, and
+  `oemStrings` appear in the generated XML.
+- Admission webhook tests verifying that `attestation` without
+  `startStrategy: Paused` is rejected.
+- E2E test with a sample attester: create a TDX/SNP VMI with `attestation`,
+  call the subresource, verify the VMI transitions through
+  `Scheduled` → `Running` → unpaused.
+
+## Graduation Requirements
+
+### Alpha
+
+The feature will be implemented in Alpha. We do not know if it will be possible
+to have e2e tests in Alpha due to lack of TDX hardware. We expect the feature
+to be merged without the e2e tests.
+
+### Beta
+
+We expect e2e tests in Beta. We expect the API to be stable.
+
+### GA
--- a/veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
+++ b/veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
@@ -1,0 +1,353 @@
+# VEP #340: launchsecurity: Add InitData resource for TDX and SNP
+
+## VEP Status Metadata
+
+### Target releases (InitData)
+
+- This VEP targets alpha for version: v1.11
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Confidential VMs (CVMs) require per-VM configuration values to be injected
+before boot so the guest can use them during remote attestation. These values
+are incorporated into hardware measurement registers and become part of the
+attestation report, allowing the guest to prove that the host provided the
+expected configuration.
+
+This proposal introduces a new `InitData` Custom Resource (CR) that carries
+the launch-time attestation values for a VMI.
+
+The design cleanly separates two concerns:
+
+1. **Immutable intent** in the VMI spec (`initDataRef` field).
+2. **One-shot launch-data commitment** in the `InitData` CR (write-once, referenced from VMI spec).
+
+In this VEP, InitData wording refers to
+[initData](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/initdata.md).
+The Initdata Specification defines the key data structure and algorithms to
+inject any well-defined data from untrusted host into TEE.
+
+## Motivation
+
+Organizations deploying confidential VMs need the strongest possible
+attestation guarantees. A key use case is encrypted disk provisioning: an
+external operator injects a reference to a secret stored in a Key Broker
+Service (KBS) into the guest before boot. The guest then attests to the KBS
+and retrieves the secret only if the attestation passes. The injected values
+become part of the hardware measurement, ensuring the host cannot tamper with
+them undetected.
+
+## Goals
+
+- Expose TDX and SEV-SNP InitData fields through the CR
+- Provide a declarative, Kubernetes-native mechanism for InitData commitment
+  that is compatible with GitOps workflows.
+
+## Non Goals
+
+- Implement the external operator that creates `InitData` resources.
+- Verify the cryptographic relationship between the digest and the InitData bytes.
+
+## Definition of Users
+
+- Cluster Administrators: Configure RBAC permissions for the `InitData`
+  resource.
+
+- VM Users: Secret delivery.
+
+## User Stories
+
+- As a VM user, I want to tell the infrastructure to inject InitData into my
+  confidential VM.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+The design separates immutable declarative intent (VMI spec) and one-shot
+launch-data commitment (`InitData` CR).
+
+### API extensions
+
+New fields added to the VMI spec. These are set at creation time and never
+mutated:
+
+```go
+type TDX struct {
+    InitDataRef string `json:"initDataRef,omitempty"`
+}
+
+type SEVSNP struct {
+    InitDataRef string `json:"initDataRef,omitempty"`
+}
+```
+### InitData CRD
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+them undetected.
+
+Currently, KubeVirt supports this flow for SEV via the
+`sev/fetchCertChain` and `sev/setupSession` subresources. However, Intel TDX
+and AMD SEV-SNP have no equivalent mechanism. Users resort to hook sidecars
+that modify the libvirt domain XML directly, bypassing KubeVirt's API and
+losing auditability, RBAC control, and compatibility with future KubeVirt
+versions.
+
+This proposal fills the gap by extending the VMI API with the fields and
```
0xFelix, line 59: If wonder if these fields should be in `spec` and `status` or only `status`? Are they ever intended to be written to from anywhere else than the subresource API endpoints?

Also, having them in `spec` could cause issue with GitOps workflows, that would consider changes to these fields configuration drift.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+subresources needed for TDX and SEV-SNP attestation-based secret delivery.
+
+## Goals
+
+- Expose TDX and SEV-SNP initdata fields in the VMI API so users can
+  configure attestation-relevant registers (`mrConfigId`, `hostData`,
+  `oemStrings`) declaratively or at runtime.
+- Provide a KubeVirt-native mechanism for external operators to inject
```
0xFelix, line 67: Could the external operator not PATCH the VM instead? Do you want to guard these operations with separate RBAC rules and therefore require the subresource APIs?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+The design follows the existing SEV attestation pattern in KubeVirt. The key
+components are:
+
+### API extensions
+
+New fields added to the VMI spec:
```
0xFelix, line 109: How are these fields kept in sync with the parent VM? You want the subresource to only change them in the VMI? Can these not be moved to the VMI's status instead?

0xFelix (commented):
To me it looks problematic that you suggest to add fields to `spec`, that may be overridden at runtime and could cause issues with GitOps tools.

Also, how does it behave when the VMI is owned by a VM and you modify the VMI's spec while the VM's template spec stays unchanged?

0xFelix:
VMI spec should actually be immutable, making the suggested design impossible?

MatiasVara:
> VMI spec should actually be immutable, making the suggested design impossible?

I did not go through the questions yet but bear in mind that the design is based on how SEV works now.

MatiasVara:
@alancaldelas @jean-edouard WDYT?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: The idea to have it in the spec is to support both cases: static definition in the spec and dynamically added by the operator. I do not see why we would have a env in which the GitOps workflows and the external operator would be presented. Also, this is how SEV currently works by having it in the spec.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I think by using a subresource and RBAC rules we can keep external operator limited to what it can change.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: These fields are per-VMI. They don't sync back to the parent VM template. On the first boot, the external operator injects them via the subresource. On subsequent restarts, the operator re-injects the same values into the new VMI.

MatiasVara:
> To me it looks problematic that you suggest to add fields to `spec`, that may be overridden at runtime and could cause issues with GitOps tools.
> 

The plan is not to have both the external operator and GitOps tool. 

> Also, how does it behave when the VMI is owned by a VM and you modify the VMI's spec while the VM's template spec stays unchanged?

The VMI spec is expected to diverge from the template in many cases (mutating webhooks, annotations, UUID sync, SEV session injection). Injecting mrConfigId into the VMI spec is no different — the VM template keeps tdx.attestation: {} with an empty mrConfigId, and the VMI gets the value filled at runtime. On restart, a new VMI is created from the template (empty again), and the operator re-injects.

MatiasVara:
> VMI spec should actually be immutable, making the suggested design impossible?

The VMI update admitter (vmi-update-admitter.go:103-117) rejects spec changes from external callers but explicitly allows them when the request comes from the KubeVirt service account — which is the case when a subresource handler in virt-api performs the patch. This is the same path used by sev/setupsession to inject session and dhCert into the VMI spec today.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+## Overview
+
+Confidential VMs (CVMs) require per-VM configuration values to be injected
+before boot so the guest can use them during remote attestation. These values
+are incorporated into hardware measurement registers and become part of the
+attestation report, allowing the guest to prove that the host provided the
+expected configuration.
+
+This proposal adds two new KubeVirt subresource endpoints that allow an
+external operator to inject these values into a VMI while it is in the
```
amd-kkin: What "external operator" mean exactly? Does it mean that this operator is deployed in a different environment than where the VM to be attested is running? Is the external operator supposed to be deployed and maintained by whoever is operating Trustee (the relying party)? 

My concern is that if the operator is malicious, then it could alter the initData provided.

MatiasVara: For me the external operator is the provisioner-operator which provides initData but it could be something else. The operator can be deployed in a trusted enviroment. Even if the operator is trusted, the k8s infra may be not. The operator and k8s infra shall not have confidential information. For example, if the operator injects a malicious KBS URL, the trustee client has to still certificate that the trustee server is trusted (sorry it is like a tongue twister).

alancaldelas: This seems to be a good cross over between the provisoner-operator and the attestation operator that I am proposing. Running the attestation operator in the trust side would make things much easier in terms of attestation as well.

amd-kkin: That was my initial thought as well, but actually from this document, it seems initData is meant to be passed from Trustee to the unsafe guest: https://confidentialcontainers.org/blog/2024/08/15/policing-a-sandbox/#init-data
Because it is part of the attestation, tampering should be caught. So maybe the operator can actually be deployed in the unsafe cluster? 

Also, in your design, where does the initData to be injected come from? How does the initData generated by Trustee's [provisioning plugin](https://github.com/confidential-containers/trustee/pull/1305) become available to the provisioning operator?

MatiasVara: I think you can deploy the operator in the unsafe cluster if it is not handling confidential information. The provisioner operator gets the information from the trustee provision plugin, .e.g, the path to the new resource, and generates the initData to inject it into the VM. The trustee plugin releases the secret after attestation. The plugin also check if the VM is allowed to access that resource based on the initData that it provides.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
0xFelix: > I do not see why we would have a env in which the GitOps workflows and the external operator would be presented.

They don't mutually exclude each other, no?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
0xFelix: ACK

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
0xFelix: Sounds okay, but I still wonder if this field should sit in status and not in spec.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
0xFelix: I mean, eventually that is down to the user to decide?

0xFelix:
> The VMI spec is expected to diverge from the template in many cases (mutating webhooks, annotations, UUID sync, SEV session injection). Injecting mrConfigId into the VMI spec is no different — the VM template keeps tdx.attestation: {} with an empty mrConfigId, and the VMI gets the value filled at runtime. On restart, a new VMI is created from the template (empty again), and the operator re-injects.

Fair point

> The VMI update admitter (vmi-update-admitter.go:103-117) rejects spec changes from external callers but explicitly allows them when the request comes from the KubeVirt service account — which is the case when a subresource handler in virt-api performs the patch. This is the same path used by sev/setupsession to inject session and dhCert into the VMI spec today.

Yes, got that (partially) wrong. The KubeVirt service accounts are allowed to mutate VMIs.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+  call the subresource, verify the VMI transitions through
+  `Scheduled` → `Running` → unpaused.
+
+## Graduation Requirements
+
+### Alpha
+
+The feature will be implemented in Alpha. We do not know if it will be possible
+to have e2e tests in Alpha due to lack of TDX hardware. We expect the feature
+to be merged without the e2e tests.
```
0xFelix: Can you test the chain with random values? You said it's up to guest-side logic to use these values. That way you could at least check if the values are transmitted correctly?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+### Beta
```
0xFelix, line 348: That means we should put Alpha behind a FG?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
+# VEP #340: launchsecurity: add injectInitdata subresource for TDX and SNP
+
+## VEP Status Metadata
+
+### Target releases
+
+This VEP does not have graduation phases guarded by a feature gate.
```
EdDev: Extending the API without a FG implies this is a trivial change and that you have a very strong case and relevant SIGs support.
Please add the reasoning to go straight to HA with this one.
Also, leave the format intact and specify GA target while marking the Alpha and Beta as irrelevant (strike line?).

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: Can they be behind the corresponding TDX and SEV FG?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+- Ship target version:
+
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+
+## Overview
```
EdDev, line 20: Please try to avoid adding implementation details in the overview, motivation and goals sections. These should express the need and the high-level idea of what is to be done.
I would avoid mentioning specific components (e.g. virt-handler) and fields in these sections.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: IIUC if the field is in the status then it would not be possible for the user to set it in the spec, am I right?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+}
+```
+
+The `Attestation` field is a marker: when present, it signals that the VMI
+requires initdata injection before boot. Without `Attestation`, the registers
+can be set directly from the YAML. `mrConfigId`/`hostData` start empty and are
+filled via the subresource. `oemStrings` carries a reference (e.g., a KBS
+resource path) that the guest reads from SMBIOS Type 11 at boot.
+
+### Subresource endpoints
```
EdDev: Two issues:
1. Attempting to add imperative operations to Kubevirt is not in sync with its declarative API. While the operation may work fine in some cases, we found it failing in others. IMO imperative operations should be our last result and only allowed if all other alternatives are not feasible.
2. Verification at the subresource endpoint is fragile and is not recommended. At the time the VMI (or any other object) is checked, there is no assurance this is the 1) real current status and 2) it will remain as such when the change actually is applied.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+curl -X PUT \
+  /apis/subresources.kubevirt.io/v1/namespaces/default/virtualmachineinstances/my-vm/snp/injectInitdata \
+  -H "Content-Type: application/json" \
+  -d '{
+    "hostData": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
+    "oemStrings": ["kbs:///default/uuid/root"]
+  }'
+```
+
+## Alternatives
```
EdDev, line 233: How about these options:
- Manage VMI spec mutation by an external authorized controller.
- As an alternative to a HOOK sidecar, utilize a formal (improved) plugin. I.e. a plugin that enables the domain xml mutation, injected by an external operator through the plugin API. Work has started in this direction, providing means to extend Kubevirt functionality.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
EdDev: Alpha does not require functional tests, but it is indeed preferred.
If there is no intention to add them, just do not mention them here and keep them in Beta.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
EdDev: It is required, unless everyone agrees on an exception to this.
Based on what I could read so far, I cannot see a strong reasoning for not having it.

EdDev (changes requested):

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I guess if user really wants to use ArgoCD with the external operator, it would find the way to ignore changes in mrConfigId in case attestation is used. If not attestation is used, the user can just provide mrConfig in the spec and it wont be a drift.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: > It is required, unless everyone agrees on an exception to this. Based on what I could read so far, I cannot see a strong reasoning for not having it.

do you mean I could use TDX and SEV FG?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I agree to add them. I proposed the use of the subresource because SEV was already using the same mechanism. I used to use the first option you mentioned but I though the use of a subresource for something similar to SEV would be a sort of pattern and would it be cleaner.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I see. I've followed the existing SEV pattern for attestation in which `sev/setupsession` is injected using a subresource.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: > * Manage VMI spec mutation by an external authorized controller.

I think if this is the correct approach then we need to get rid of the existing pattern for SEV.

EdDev: > I used to use the first option you mentioned

How did you overcome the validation admission controller webhook which limits the edits to the Kubevirt ServiceAccount?

> but I though the use of a subresource for something similar to SEV would be a sort of pattern and would it be cleaner.

We may be missing some context on why it was chosen at the time, but IMO the imperative commands are not a good fit in general. Only recently the disk/volume hotplug moved to a declarative approach, deprecating the subresource one.

> > * Manage VMI spec mutation by an external authorized controller.
> 
> I think if this is the correct approach then we need to get rid of the existing pattern for SEV.

It could be deprecated (but not removed, due to backward compatibility).

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
EdDev: It could use an existing FG if it makes sense. The downside is that the feature that one protects, will need to pass another stage before graduation.

MatiasVara: It has to be behind the TDX and SEV FG anyway because it does not make sense without.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: > > I used to use the first option you mentioned
> 
> How did you overcome the validation admission controller webhook which limits the edits to the Kubevirt ServiceAccount?
> 

I do not remember exactly but I remember I patched the qemu args to add support for smbios table and mrconfig, which is supported only after my PR. Maybe I was using a sidecar.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+    // ... existing fields ...
+    OEMStrings []string `json:"oemStrings,omitempty"`
+}
+```
+
+The `Attestation` field is a marker: when present, it signals that the VMI
+requires initdata injection before boot. Without `Attestation`, the registers
+can be set directly from the YAML. `mrConfigId`/`hostData` start empty and are
+filled via the subresource. `oemStrings` carries a reference (e.g., a KBS
+resource path) that the guest reads from SMBIOS Type 11 at boot.
```
fanzhangio: Moving all launch data to status would be incorrect, but mutating mixed static and dynamic configuration in the VMI spec also leaves ownership ambiguous. Keep immutable source intent in spec, make static and external modes exclusive, and report only commitment conditions in status.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+| `PUT .../virtualmachineinstances/{name}/tdx/injectInitdata` | Intel TDX | `{"mrConfigId": "<base64>", "oemStrings": [...]}` |
+| `PUT .../virtualmachineinstances/{name}/snp/injectInitdata` | AMD SEV-SNP | `{"hostData": "<base64>", "oemStrings": [...]}` |
+
+Both endpoints:
+1. Verify the VMI is in `Scheduled` phase.
+2. Verify the `Attestation` field is set.
+3. Verify the initdata field (`mrConfigId`/`hostData`) is not already set.
+4. Validate the base64 value decodes to the correct length (48 bytes for TDX,
+   32 bytes for SNP).
+5. Apply a JSON patch to the VMI spec.
```
fanzhangio: The API accepts the launch-register digest and OEM strings independently, while guest verification is left outside the design. Hardware can therefore attest digest X while the guest consumes bytes Y. Specify the exact Init-Data representation and require guest verify-before-parse plus relying-party authorization of the digest.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+This is validated at admission time: if `Attestation` is set but
+`startStrategy` is not `Paused`, the VMI is rejected.
+
+### Security Considerations
+
+- The `injectInitdata` subresource is guarded by RBAC. Only components with
+  the appropriate ClusterRole can call the endpoint. This prevents unauthorized
+  injection of attestation values.
+- The injected `mrConfigId` and `hostData` are measured by the hardware (TDX
```
fanzhangio: This currently conflates measurement with authorization. Hardware evidence proves which value was placed in `MR_CONFIG_ID` or `HOST_DATA`. it does not prove that the value is the tenant-approved configuration. A malicious operator can supply malicious Init-Data together with its matching digest and still produce internally consistent evidence.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+boots in paused state, giving the external operator a window to:
+1. Call the provisioner to obtain the values.
+2. Inject them via `injectInitdata`.
+3. Wait for the VMI to reach `Running` phase (QEMU started but paused).
+4. Call `unpause` to resume guest execution.
+
+This is validated at admission time: if `Attestation` is set but
+`startStrategy` is not `Paused`, the VMI is rejected.
+
+### Security Considerations
```
fanzhangio, line 135: Please define the complete normative chain, like this: 
exact Init-Data bytes -> canonical digest -> launch register -> guest verify-before-parse -> AS claim -> relying-party/KBS policy authorization.

CoCo explicitly requires the guest to compare the exact Init-Data hash with the launch configuration and permits KBS policy to evaluate the verified claims before releasing a resource. KubeVirt should be described as the transport and atomic commitment point, not the authority that establishes attestation success. See https://confidentialcontainers.org/docs/features/initdata/#integrity-and-attestation

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
fanzhangio: I agree with the spec/GitOps concern, but moving the launch data wholesale into status would also be incorrect: status is observed state, while these values configure QEMU and the hardware launch measurement.
Please separate three concepts:
1. Immutable intent in the VMI spec, such as source: static/external or an immutable launch-data reference.
2. The dynamically resolved, one-shot launch-data commitment, supplied through either a dedicated resource or a hardened subresource.
3. Observed status such as `WaitingForLaunchData` or `LaunchDataCommitted`.

Static and external modes should be mutually exclusive by admission. The API should not assume GitOps and the external operator cannot coexist, field ownership must be structural.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,364 @@
...
+
+The new fields are translated to libvirt domain XML in the converter layer:
+
+- `TDX.MRConfigId` -> `<launchSecurity type="tdx"><mrConfigId>...</mrConfigId></launchSecurity>`
+- `SEVSNP.HostData` -> `<launchSecurity type="sev-snp"><hostData>...</hostData></launchSecurity>`
+- `Firmware.OEMStrings` -> `<sysinfo type="smbios"><oemStrings><entry>...</entry></oemStrings></sysinfo>`
+
+### Start strategy
+
+The VMI must be created with `startStrategy: Paused`. This ensures QEMU
```
fanzhangio: Pausing is not an attestation boundary. Why `startStrategy: Paused` required? Injection is already completed while the VMI is Scheduled, before QEMU exists, so pausing QEMU does not create the injection window.
It also cannot be an attestation gate: the guest must execute before it can collect fresh evidence and communicate with the relying party. So Running -> unpause must not be interpreted as attestation success.

fanzhangio (commented):
Thanks @MatiasVara  for proposing this. I think the core problem is real, and I think moving TDX/SNP launch-data handling from hook-sidecar XML mutation into a typed KubeVirt interface is the right upstream direction. The pre-QEMU synchronization point, narrow RBAC opportunity, auditability and secretless KBS workflow are all valuable.
I left some inline comments, and overall I think the proposal currently conflates three separate operations:
1. delivery and commitment of launch data.
2. verification and appraisal of fresh attestation evidence.
3. authorization and release of a confidential resource.

KubeVirt should only own the first operation. A populated launch register or an unpaused VMI must not be represented as evidence that attestation succeeded.

I agree with @0xFelix  that the current imperative, runtime VMI-spec mutation needs stronger justification. However, moving all data into VMI status or allowing a general external controller to patch VMIs would not be sufficient. 
My preferred upstream shape is an immutable declarative intent in the VMI spec plus a typed launch-data resource/reference bound to the VMI UID. VMI status would report only operational conditions such as `WaitingForLaunchData` and `LaunchDataCommitted`.
If the subresource design is retained, it needs to be specified as a one-shot atomic compare-and-set operation with expected UID and resourceVersion, exact-retry idempotency, mutually exclusive static/external modes, and deterministic fail-closed behavior during retries, feature-gate changes and version skew.

The most important missing security invariant is the binding between the exact bytes consumed by the guest and the digest measured in `MR_CONFIG_ID` or `HOST_DATA`. Measurement shows what was supplied. the relying party still has to decide whether that value is authorized. The guest must verify the exact delivered bytes before parsing them or contacting a KBS, and KBS policy must authorize the resulting claims.

I would support the VEP after the trust model, byte-level Init-Data contract, immutable VMI identity binding, lifecycle and negative test requirements are made normative.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: > Can you test the chain with random values? You said it's up to guest-side logic to use these values. That way you could at least check if the values are transmitted correctly?

I do not get this, can you elaborate?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I will add a new FG and put all this behind.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
0xFelix: Is there a way to test the transmission of the data from VMI spec into the VM without having proper TDX hardware?

MatiasVara:
> My preferred upstream shape is an immutable declarative intent in the VMI spec plus a typed launch-data resource/reference bound to the VMI UID. VMI status would report only operational conditions such as WaitingForLaunchData and LaunchDataCommitted.

@fanzhangio Thanks for your comments, let me think about this and rewrite the VEP. TBH I only used a subresource to follow SEV pattern.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I only meant to say that RBAC allows only authorized components to inject the values. I said nothing about the values itself. I could rewrite it as:
``"RBAC restricts which components can call the endpoint. Authorization of the injected values themselves is enforced by the KBS policy at attestation time."``

MatiasVara:
@fanzhangio @0xFelix  feel free to take a look. I changed the approach in favor to using of a resource. Thanks.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
+# VEP #340: launchsecurity: Add InitData resource for TDX and SNP
+
+## VEP Status Metadata
+
+### Target releases (InitData)
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version: v1.11
```
0xFelix:
```suggestion
- This VEP targets alpha for version: v1.11
- This VEP targets beta for version:
```

Unfortunately, we are past the VEP freeze for 1.10.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
0xFelix:
```suggestion
- [x] (R) Alpha target version is explicitly mentioned and approved
```

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+type Firmware struct {
+    // ... existing fields ...
+    // Static OEM strings delivered via SMBIOS Type 11 (static mode only).
+    OEMStrings []string `json:"oemStrings,omitempty"`
+}
+```
+
+The `Attestation` field is a mode selector: when present, it signals that the
+VMI requires an `InitData` CR before boot. Without `Attestation`, the values
+are provided directly in the spec (static mode). The two modes are **mutually
```
0xFelix: Can we not make this explicit by always referring to an `InitData` CR in the VMI spec? This way it becomes obvious that an `InitData` CR is required and it is left open how that CR is created. Either statically or by an external operator.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+type InitData struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    Spec   InitDataSpec   `json:"spec"`
+    Status InitDataStatus `json:"status,omitempty"`
+}
+
+type InitDataSpec struct {
+    // Immutable reference binding this InitData to a specific VMI instance.
+    VMIRef VMIReference `json:"vmiRef"`
```
0xFelix: What does this coupling help with?

0xFelix: If required, use `corev1.ObjectReference` instead?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+}
+
+type VMIReference struct {
+    Name string    `json:"name"`
+    UID  types.UID `json:"uid"`
+}
+
+type InitDataStatus struct {
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+    Phase      InitDataPhase      `json:"phase,omitempty"`
```
0xFelix: We should express it as `Conditions` and not try to reduce them to a single `Phase` string. Drop this field.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+const (
+    InitDataPending   InitDataPhase = "Pending"
+    InitDataCommitted InitDataPhase = "Committed"
+    InitDataFailed    InitDataPhase = "Failed"
+)
+```
+
+Properties:
+- The `InitData` CR has an `ownerReference` pointing to the VMI, enabling
+  automatic garbage collection when the VMI is deleted.
```
0xFelix, line 126: What if we want to have a static `InitData` object? Can it be owned by the VM instead?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+- **Write-once**: updates to `.spec` are rejected after the `InitData` has
+  reached `Committed`.
+- **One-per-VMI**: creation of a second `InitData` are rejected for the same
+  VMI UID.
```
0xFelix: Why the inverse coupling? Let the VM/VMI point at the `InitData` CR it wants to use.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+- **UID binding**: `vmiRef.uid` must match the target VMI's UID; this prevents
```
0xFelix: How does matching of the `UID` prevent replay attacks?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+  stale or replayed `InitData` objects from being applied to a different VMI
+  instance with the same name.
+
+### Validation Rules
+
+- `Attestation` + `MRConfigId`/`HostData` present in spec: VMI - rejected (mutually exclusive modes)
```
0xFelix: I'd prefer to have a single source of truth, not two possible sources. Can we not use the InitData CR exclusively?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+  or SEV-SNP respectively). Any tampering is detectable by the guest during
+  remote attestation. The host infrastructure provider cannot modify these
+  values after commitment without invalidating the attestation report.
+- The `oemStrings` field carries a reference (e.g., a KBS resource path), not
+  the actual secret. The secret is only released by the Key Broker Service
+  after the guest passes attestation.
+- The design does not store secrets in Kubernetes objects. The `InitData` CR
+  contains measurement digests and resource paths, not cryptographic keys.
+- The write-once property and UID binding ensure that an `InitData` cannot be
+  replayed or reused for a different VMI instance.
```
0xFelix: How does it prevent replay? I can take the existing measurements and create a new CR with a different UID?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
```diff
@@ -0,0 +1,467 @@
...
+
+**Drawbacks:**
+- Mutating the VMI spec at runtime causes GitOps configuration drift.
+- Imperative one-shot PUT calls are not aligned with Kubernetes declarative
+  patterns.
+- JSON Patch with `test` operations for concurrency control is fragile.
+- Field ownership between the user (static values) and the operator (dynamic
+  values) is ambiguous when both live in the same spec field.
+
+### Hook sidecar
```
0xFelix, line 245: What about a plugin? Built on the new plugin mechanism?

MatiasVara: What is the new plugin mechanism and how could I use to implement the proposed mechanism?

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: I removed.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: yes, we can.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: Done

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: It does not. I will remove it.

veps/sig-compute/340-launchsecurity-inject-initdata/injectinitdata-proposal.md
MatiasVara: yes, you can, I remove that sentence.

Issue #309, opened 2026-05-13
VEP: CRD Field Deprecation and Removal Policy


lyarwood:
**Primary contact (assignee)**:

/assign @lyarwood

**Current Feature Stage**:

New

**Feature Gate**:

N/A — this is a process and policy VEP, not a code feature.

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs:
/sig storage
/sig network
/sig observability

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/310

**Timeline**:

* Alpha release target: v1.10
* Beta release target:
* Stable release target:

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/310
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:

KubeVirt uses `NoneConverter` for all CRDs, which means the storage version migrator can only prune fields during migration — it cannot transform them. Any field that is restructured between versions (old field replaced by a new field with a different shape) will have the old field silently dropped and the new field left empty during storage version migration. This is a data loss risk for any non-sequential upgrade path.

This VEP proposes establishing a policy for how CRD fields are deprecated and removed, ensuring that stored data is not silently lost during storage version migration. A clear upstream policy also enables downstream distributions to align their own lifecycle policies and build skip-level upgrade support (e.g. N to N+2 or N to N+3) on top of documented migration lifecycles.

A runnable demo reproducing the data loss scenario with raw etcd diffs is available at: https://github.com/lyarwood/kube-storage-version-migrator/tree/demo/demo

> [!IMPORTANT]
> Please keep this description up to date.

lyarwood:
VEP PR: https://github.com/kubevirt/enhancements/pull/310

lyarwood:
Closing this tracker in favour of a different approach.

The use case and target-release designation model that motivated this VEP are
now tracked in VEP #471 (https://github.com/kubevirt/enhancements/issues/471),
which covers the skip-level upgrade use case from N = v1.11 to N+3 = v1.14
and includes the pre-schema migration requirement and cumulative migration model.

The per-field deprecation conventions (how to mark fields deprecated, the
admission warning pattern, tombstone comments, and the review checklist) will
be contributed as a PR to the API design guidelines in kubevirt/kubevirt
(https://github.com/kubevirt/kubevirt/pull/18612) rather than as a standalone
VEP. That keeps the authoring conventions where contributors will actually
look for them.

PR #348, opened 2026-06-24
VEP #371: KubeVirt Integration Test Framework

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/371
**SIG label**: /sig control-plane
**VEP directory**: `veps/sig-control-plane/371-envtest-framework/`

### What this PR does

Proposes a KubeVirt integration test framework inspired by OpenStack
Nova's approach. The framework uses envtest (real kube-apiserver + etcd)
with real controllers running in-process and a pod simulator to exercise
multi-controller interactions in seconds without deploying a cluster.

#### Phase 1 (v1.10.0): VM and VMI controller coverage

Establish the framework with the VM controller (and all synchronizer
sub-controllers), VMI controller, and pod simulator running in-process.
The goal is to prove the approach and provide enough examples for
contributors to write integration tests for their own controller changes.

**Framework:**
- envtest with all KubeVirt CRDs
- Real VM (with all synchronizers, including instancetype) and VMI
  controllers in-process
- Pod simulator for fake scheduling
- Webhook server (optional, `WithWebhooks`)
- `make test-integration` CI target (non-voting initially)

**Sample tests (representative, not exhaustive):**
- VM lifecycle (VM → VMI → Pod → Scheduled)
- VM generation tracking and ControllerRevision lifecycle
- Regression tests (bugs #16719, #16071)
- CRD validation (structural schema, CEL rules)
- Webhook admission (mutating defaults, validating rejection)
- Instancetype ControllerRevision upgrades (migrated from functional tests)

#### Phase 2: virt-handler simulation and domain XML

- Fake libvirt gRPC server for domain XML (`WithFakeLibvirt`)
- virt-handler simulation (`WithVirtHandler`) — VMIs reach Running
- Workload-update controller (`WithWorkloadUpdateController`)
- Additional functional test migrations

#### Phase 3: Extended coverage

- virt-operator testing (`WithVirtOperator`) — install/upgrade lifecycle
- Live migration API, CDI integration, and storage API controller coverage

### Storage version migration and skip-level upgrades

The framework can also validate CRD schema evolution and detect silent
data loss from storage version migration — a key concern for skip-level
upgrades (see [VEP #309](https://github.com/kubevirt/enhancements/issues/309)).
Since envtest runs a real kube-apiserver + etcd with real schema pruning,
tests can create objects under an old CRD schema, apply the upgraded
schema, simulate the GET+PUT migration cycle, and assert that deprecated
fields are correctly converted rather than silently dropped.

A working proof of concept is available across PRs #18726–#18731 with
20 passing tests across 8 categories (VM lifecycle, generation tracking,
regressions, webhooks, CRD validation, domain XML, workload update,
instancetype upgrades). The 8 instancetype upgrade tests were migrated
from functional tests with a **13.4x speedup** (152.6s → 11.4s, no
cluster required).

The framework reuses shared matchers from `tests/framework/matcher/`
and depends on https://github.com/kubevirt/kubevirt/pull/16499
(synchronizer refactor) to avoid PatchStatus/UpdateStatus conflicts.

### DIFF
--- a/veps/sig-control-plane/371-envtest-framework/vep.md
+++ b/veps/sig-control-plane/371-envtest-framework/vep.md
@@ -1,0 +1,453 @@
+# VEP #371: KubeVirt Integration Test Framework
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Introduce an integration test framework for KubeVirt that exercises
+real controller logic end-to-end without deploying a Kubernetes cluster
+or running virtual machines. The framework uses [envtest][] to run a
+real kube-apiserver and etcd locally with KubeVirt CRDs installed and
+runs real controllers in-process against this API server.
+
+Inspired by [OpenStack Nova's functional test framework][nova-functional],
+which has used the same pattern (real services, fake external
+dependencies) for over a decade.
+
+[nova-functional]: https://docs.openstack.org/nova/latest/contributor/testing.html#functional-tests
+[envtest]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest
+
+## Motivation
+
+KubeVirt currently has two testing layers:
+
+1. **Unit tests** (`pkg/**/*_test.go`): Fast, isolated. Each controller
+   tested with GoMock — Controller A's writes are not visible to
+   Controller B.
+
+2. **Functional tests** (`tests/`): Comprehensive but slow (4-hour timeout),
+   resource-intensive, and prone to infrastructure flakiness.
+
+There is no middle ground. Bugs from multi-controller interactions,
+race conditions, or state machine transitions cannot be caught by unit
+tests and are expensive to reproduce with functional tests.
+
+A working proof of concept has been implemented in kubevirt/kubevirt
+across a series of focused PRs (#18726–#18731): 20 tests across 8
+categories, covering VM/VMI lifecycle, instancetype upgrades, webhook
+admission, CRD validation, and regression cases (see
+[Proof of Concept](./poc.md) for the per-PR breakdown, coverage matrix,
+and benchmark data). The PoC demonstrates this gap with concrete
+examples:
+
+- [Issue #16719][]: RestartRequired condition never cleared after spec
+  revert — requires real ControllerRevision lookup and multi-step
+  reconciliation that unit tests mock away.
+- [Issue #18396][]: Workload-update migrations race with virt-handler
+  DaemonSet re-roll during upgrade — a timing race between controllers
+  that functional tests catch only by luck (it's a flake).
+- Instancetype ControllerRevision upgrades — 8 functional tests that exercise
+  pure controller logic, migrated to envtest with a 13.4x speedup.
+- [CRD field deprecation and removal][issue #309] — KubeVirt uses
+  `NoneConverter` for all CRDs, so storage version migration can only
+  prune fields, not transform them. When deprecated fields are
+  replaced by structurally different fields (e.g., `preferredUseEfi`
+  bool → `preferredEfi` struct), the old field is silently dropped
+  during migration and the new field is left empty — permanent data
+  loss. Skip-level upgrades compound this: every intermediate field
+  removal is applied in a single migration. envtest can reproduce
+  this because it runs a real kube-apiserver + etcd with real schema
+  pruning, enabling tests that create objects under an old CRD schema,
+  apply the upgraded schema, simulate the GET+PUT migration cycle,
+  and assert that data is preserved or correctly converted.
+
+[issue #16719]: https://github.com/kubevirt/kubevirt/issues/16719
+[issue #18396]: https://github.com/kubevirt/kubevirt/issues/18396
+[issue #309]: https://github.com/kubevirt/enhancements/issues/309
+
+## Goals
+
+- Catch bugs arising from multi-controller interactions, race
+  conditions, and state-machine transitions that unit tests cannot
+  reach and functional tests catch only expensively or by luck.
+- Exercise controller logic against authentic API-server semantics for
+  CRD resources — OpenAPI validation, status subresource enforcement,
+  and resourceVersion behaviour.
+- Validate CRD declarative validation (structural schemas and CEL
+  rules).
+- Detect silent data loss from CRD schema evolution and storage
+  version migration, including skip-level upgrade scenarios where
+  multiple field removals compound into a single migration
+  ([VEP #309][issue #309]).
+- Enable deterministic regression tests for race conditions.
+- Run fast enough to execute on every PR in CI, each test completing
+  in seconds.
+- Provide a migration path for functional tests that don't need a running
+  guest.
+
+## Non Goals
+
+- Replacing functional tests (infrastructure integration still needs real
+  clusters) or unit tests (fastest feedback for single components).
+- Testing real hypervisor behavior (QEMU, KVM).
+
+## Definition of Users
+
+- **KubeVirt Developers**: Contributors writing or modifying
+  controller logic, CRD schemas, or webhook admission handlers.
+- **KubeVirt Reviewers**: Maintainers reviewing PRs that change
+  controller behavior or API validation rules.
+- **CI Infrastructure**: Automated pre-merge test pipelines that need
+  lightweight, hermetic test execution.
+
+## User Stories
+
+- As a developer, I want to verify the full VM lifecycle (VM → VMI →
+  pod → Scheduled) without deploying a cluster.
+- As a developer fixing a race condition, I want a regression test with
+  deterministic timing control that runs in seconds.
+- As a reviewer, I want PRs changing controller logic to include
+  integration tests demonstrating cross-controller correctness.
+- As a developer deprecating a CRD field, I want a test that creates
+  objects under the old schema, applies the new schema, simulates
+  storage version migration, and verifies no data is silently lost.
+- As a release engineer certifying a skip-level upgrade (e.g., v1.8
+  → v1.11), I want integration tests that exercise the cumulative CRD
+  schema diff across all intermediate releases and assert that
+  deprecated fields are correctly converted or explicitly removed.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Core Components
+
+**envtest**: Part of the [controller-runtime][] project and the standard
+testing substrate used by Kubebuilder, Operator SDK, and Cluster API.
+It starts a real `kube-apiserver` and `etcd` process locally (managed
+by `setup-envtest`), providing real API-server fidelity for CRD resources
+— real OpenAPI validation, status subresource enforcement, and
+resourceVersion semantics — without requiring a full Kubernetes cluster.
+Note that the kubelet, scheduler, and kube-controller-manager are absent,
+so built-in Kubernetes resources (e.g., Pods) do not behave identically to
+a real cluster; these are simulated by framework components such as the
+pod simulator. KubeVirt CRDs are loaded from the same Go definitions
+virt-operator generates in production, and tests talk to the API server
+through a real KubeVirt client — no fakes at the client layer.
+
+[controller-runtime]: https://github.com/kubernetes-sigs/controller-runtime
+
+**Real controllers**: Initialized through the same startup path the
+production virt-controller uses, wired to envtest via real shared
+informers. The VM controller and all its synchronizer sub-controllers
+run in-process alongside the VMI controller. virt-handler (which also
+acts as a controller, updating VMI status), the workload-update
+(migration) controller, and virt-operator are also supported as optional
+components (`WithVirtHandler`, `WithWorkloadUpdateController`,
+`WithVirtOperator`).
+
+**Pod simulator**: Since envtest has no kubelet, a goroutine watches for
+virt-launcher pods, binds them to a fake node, and sets their status to
+Running. Defaults to happy-path scheduling, with optional test hooks to inject
+delays or failures (e.g., unschedulable pods, OOMKilled/Evicted status). This
+drives the VMI controller through its lifecycle phases.
+
+**Webhook server**: Optional in-process HTTPS server running the real
+KubeVirt mutating and validating admission handlers.
+
+**Fake libvirt gRPC server**: An in-memory implementation of the
+virt-launcher command gRPC interface, maintained in-house alongside the
+framework in kubevirt/kubevirt. It runs the real VMI-to-domain XML
+converter and captures the generated domain XML in memory so tests can
+assert on it deterministically, without a running libvirtd or QEMU. (The
+libvirt C library itself is not a blocker — `libvirt-devel` is already
+present in the builder image.) Standing up this fake and the accompanying
+virt-handler simulation is a non-trivial effort — potentially warranting
+its own supporting doc or follow-up VEP — and is therefore scoped to Beta
+(see [Graduation Requirements](#graduation-requirements)).
+
+### Framework Options
+
+```go
+f := framework.New()                              // base: VM (inc. instancetype) + VMI controllers
+f := framework.New(framework.WithWebhooks())       // + admission webhooks
+f := framework.New(framework.WithWorkloadUpdateController()) // + workload-update controller
+f := framework.New(framework.WithFakeLibvirt())    // + domain XML conversion
+```
+
+Options compose — multiple can be combined in one `New()` call.
+
+### What envtest can and cannot test
+
+**Can test (controller logic):**
+
+- VM/VMI lifecycle, phase transitions, status updates
+- ControllerRevision creation, upgrade, conflict detection
+- Instancetype/preference application to VMI spec
+- Workload-update controller: VM live migration creation and gating logic during KubeVirt upgrades
+- Live updates (CPU, memory, and volume hotplug) — controller-level request handling and status
+  tracking, with virt-handler simulation (analogous to live migration simulation)
+- CRD structural schema and CEL validation
+- Webhook admission (mutation and validation)
+- Generation tracking, condition management, finalizer cleanup
+- Domain XML generation via the real converter — asserting the full
+  pipeline from the VM/VMI object submitted to the API server, through
+  webhook- and controller-injected defaults, down to the domain XML handed
+  to libvirt. Unit tests already cover the converter in isolation (starting
+  from an already-built domain spec); this complements them by exercising
+  that end-to-end path rather than duplicating it
+- CRD schema evolution and storage version migration — create objects
+  under one CRD version, apply an updated schema, simulate the
+  GET+PUT migration cycle, and assert field preservation
+
+**Cannot test (runtime/dataplane aspects require real infrastructure):**
+
+- Running guest OS, console access, guest agent
+- Runtime networking dataplane (CNI plugins, masquerade, bridge forwarding)
+- Live migration dataplane (actual memory transfer between nodes)
+- CDI dataplane (actual volume population and cloning)
+- Storage dataplane (snapshot I/O, restore, export, hotplug volume attachment)
+
+### Risks and Mitigations
+
+- **Controller-wiring drift**: The framework replicates the production
+  virt-controller initialization, so it must be updated when controller
+  constructors change. *Mitigation*: constructors are called directly in
+  the framework's setup, so a signature change breaks compilation and
+  produces an immediate signal rather than a silent gap (see
+  [Functional Testing Approach](#functional-testing-approach)).
+- **Fidelity gap**: There is no real hypervisor, networking, or storage,
+  and the absence of a kubelet, scheduler, and kube-controller-manager
+  means built-in resource behaviour is simulated (e.g., the pod
+  simulator) and can diverge from a real cluster. *Mitigation*: the
+  "cannot test" boundary above is explicit; the framework validates
+  controller logic and API interactions, and bugs depending on
+  infrastructure remain functional-test territory. Simulations (pod
+  simulator, fake libvirt) can also drift from real component behaviour
+  over time, so this tier complements rather than supersedes functional/e2e
+  tests and is never treated as the sole source of truth — e2e coverage
+  remains necessary even for behaviour this tier exercises.
+- **New CI dependency**: envtest requires `kube-apiserver` and `etcd`
+  binaries via `setup-envtest`, which is new for KubeVirt CI.
+  *Mitigation*: binaries are baked into the existing build container
+  image at a pinned Kubernetes version (see Functional Testing
+  Approach), so no runtime download or cluster provisioning is needed.
+- **Startup cost**: ~5-6s per envtest instance. *Mitigation*: tests
+  within an Ordered Context share a single instance to amortize the cost.
+
+## API Examples
+
+This VEP does not introduce user-facing KubeVirt API changes. The
+"API" is the Go framework used by test authors. The framework is
+intentionally thin — it wires up real controllers and lets tests use
+the same client and matcher APIs already familiar from functional tests.
+
+### Framework Setup
+
+```go
+f := framework.New(
+    framework.WithWebhooks(),
+    framework.WithWorkloadUpdateController(),
+)
+
+var _ = BeforeEach(func() { f.Start(); DeferCleanup(f.Stop) })
+```
+
+`New()` accepts composable options. `Start()` boots envtest (real
+kube-apiserver + etcd), loads all KubeVirt CRDs, creates seed data
+(namespace, a ready node), and starts the selected controllers and
+simulators. `Stop()` tears everything down.
+
+### Writing a Test
+
+Tests use `f.VirtClient()` for KubeVirt resources and
+`f.K8sClient()` for core Kubernetes resources — the same interfaces
+used in functional tests. Assertions use the existing shared matchers from
+`tests/framework/matcher/`.
+
+```go
+It("should reach Scheduled phase", func() {
+    vm := libvmi.NewVirtualMachine(
+        libvmi.New(libvmi.WithResourceMemory("128Mi")),
+        libvmi.WithRunStrategy(virtv1.RunStrategyAlways),
+    )
+    vm, err := f.VirtClient().VirtualMachine("default").Create(
+        context.Background(), vm, metav1.CreateOptions{},
+    )
+    Expect(err).ToNot(HaveOccurred())
+
+    Eventually(matcher.ThisVMIWith(
+        f.VirtClient(), vm.Namespace, vm.Name,
+    )).Should(matcher.BeInPhase(virtv1.Scheduled))
+})
+```
+
+No new helper APIs to learn — the framework provides real controllers
+and a real API server; tests interact through standard clients and
+matchers.
+
+## Alternatives
+
+### Enhanced Fake Clients (no envtest)
+
+Use `k8s.io/client-go/testing` fake clientset with wired informers.
+**Con**: no OpenAPI validation, no status subresource enforcement, no
+real resourceVersion semantics. **Decision**: envtest. The startup cost
+is acceptable for the fidelity gain, and envtest is the ecosystem
+standard (Kubebuilder, Operator SDK, Cluster API).
+
+### Mock-Based Approach (current pattern)
+
+Continue using GoMock for all controller tests. **Con**: cannot test
+multi-controller interactions; mocks verify calls made, not end state
+reached. **Decision**: keep GoMock for unit tests, add envtest for
+integration coverage.
+
+## Scalability
+
+Not applicable — test infrastructure only.
+
+## Update/Rollback Compatibility
+
+Not applicable — no production code changes.
+
+## Functional Testing Approach
+
+This VEP introduces a new **integration** test tier that sits between
+unit tests (fastest feedback for a single component) and functional/e2e
+tests (full cluster, real dataplane). The tier's scope is controller
+logic and API interactions, as bounded by the "can/cannot test" list in
+Design.
+
+The framework itself is validated by its own test suite — the PoC tests
+*are* the framework's integration tests. New framework code carries unit
+tests where the logic is isolated (e.g., pod-simulator failure hooks);
+the integration tests cover the multi-controller wiring that unit tests
+cannot. If the framework breaks (controller wiring, pod simulator, CRD
+loading), the tests fail.
+
+**Compile-time safety**: Controller constructor signatures and gRPC
+service interfaces are called directly in the framework's setup path.
+When upstream constructors or interfaces change, the framework fails to
+compile, producing an immediate signal — no separate contract test is
+needed.
+
+**CI integration**: The `make test-integration` target runs all integration
+tests on every PR alongside `make test` (unit tests). Like unit tests, the
+target is wrapped in `hack/dockerized` and runs inside the same build
+container image. The `kube-apiserver` and `etcd` binaries required by
+envtest are baked into that image at a pinned Kubernetes version — the
+latest version supported by the target KubeVirt release as defined in the
+[KubeVirt K8s Support Matrix][support-matrix]. The version can be overridden
+at build time via an `ENVTEST_K8S_VERSION` environment variable. No
+cluster provisioning or runtime binary download is needed.
+
+The suite can be run locally with `make test-integration`, which also
+uses `hack/dockerized` — no separate setup is needed beyond having the
+build container available.
+
+[support-matrix]: https://github.com/kubevirt/sig-release/blob/main/releases/k8s-support-matrix.md
+
+**Regression coverage**: Each bug fix merged with an envtest
+regression test provides ongoing validation that the framework
+correctly exercises the relevant controller paths.
+
+## Implementation History
+
+- **2026-08-17**: VEP proposed.
+- **2026-08 (PoC)**: Framework PoC posted as PRs #18726–#18731
+  (superseding PR #18238) — 20 tests across 8 categories with VM (inc.
+  instancetype), VMI, and workload-update controllers, pod simulator,
+  webhook server, and fake libvirt gRPC server. See
+  [Proof of Concept](./poc.md).
+
+## Graduation Requirements
+
+This is test-only infrastructure: it ships no user-facing runtime
+behaviour, introduces no Feature Gate, and creates no API objects or VM
+state. The standard On-By-Default Readiness and Feature-Gate-removal
+criteria therefore do not apply, and there is no runtime feature to
+enable or disable per stage. The Alpha/Beta/GA stages below combine what
+gets built with the framework's integration into the contributor and CI
+workflow — from a non-voting experiment to a required, gating pre-merge
+check. Absent a feature gate, progression through these stages is tracked
+solely by the target release versions recorded in the VEP metadata above.
+
+### Alpha (v1.10.0)
+
+Build:
+
+- envtest with all KubeVirt CRDs; real VM (with all synchronizers,
+  including instancetype) and VMI controllers in-process; pod simulator;
+  optional webhook server; the `make test-integration` CI target.
+- Representative tests: VM lifecycle, generation tracking and
+  ControllerRevision lifecycle, regression tests (#16719, #16071), CRD
+  validation, webhook admission, instancetype ControllerRevision
+  upgrades (migrated from functional tests).
+
+Criteria:
+
+- `make test-integration` runs on every PR as a **non-voting** job.
+- Flakiness is monitored via the standard KubeVirt CI dashboard.
+- Contributors are encouraged, but not required, to include integration
+  tests for controller changes.
+
+### Beta
+
+Build:
+
+- Fake libvirt gRPC server for domain XML and virt-handler simulation
+  (VMIs reach Running); workload-update controller.
+- Further functional-test migrations (instancetype application,
+  requirements, revisions).
+
+Criteria:
+
+- Guidelines published covering when integration tests are a
+  nice-to-have vs. a hard requirement for PRs touching controller logic.
+- VEP template updated to reference the integration test framework as an
+  option in the Functional Testing Approach section.
+- Suite demonstrates a low, monitored flake rate over at least one full
+  release cycle while still non-voting.
+
+### GA
+
+Build (extended coverage):
+
+- virt-operator install/upgrade lifecycle testing.
+- Live migration API: migration object lifecycle and migration
+  controller phase transitions (as OpenStack Nova tests live migration
+  through the API layer without a real backend).
+- CDI (DataVolume/DataSource) and storage (snapshot/restore) controller
+  logic against the API server.
+- Multi-node migration simulation with a stateful fake libvirt, enabling
+  the real domain manager and converter to run against in-memory domain
+  state.
+
+Criteria:
+
+- Promoted to a **voting** pre-merge check once the suite has
+  demonstrated zero flakes across multiple release cycles with a
+  representative test count (Alpha coverage as the baseline).
+- Integration tests **required** for PRs introducing new controller
+  logic or modifying existing controller behaviour, per the published
+  contribution guidelines.
+- Contributor guide for writing integration tests published in the
+  KubeVirt developer documentation.
--- a/veps/sig-control-plane/371-envtest-framework/poc.md
+++ b/veps/sig-control-plane/371-envtest-framework/poc.md
@@ -1,0 +1,49 @@
+# VEP #371 Proof of Concept
+
+Supporting material for [`vep.md`](./vep.md). This appendix records the
+per-PR breakdown, coverage matrix, and benchmark data for the working
+proof of concept; it is not required to understand the feature or its
+design.
+
+A working PoC (`tests/envtest/` in kubevirt/kubevirt) demonstrates the
+framework, posted as a series of focused PRs (superseding the original
+PR #18238):
+
+- [PR #18726][]: Framework base — envtest, controllers, pod simulator, lifecycle test
+- [PR #18727][]: Instancetype controller support — synchronizer refactors, regression tests, upgrade tests
+- [PR #18728][]: Generation tracking tests — including #18700 regression
+- [PR #18729][]: Webhook admission support
+- [PR #18730][]: CRD schema and CEL validation tests
+- [PR #18731][]: PodSimulator hooks — failure and not-ready simulation
+
+[PR #18726]: https://github.com/kubevirt/kubevirt/pull/18726
+[PR #18727]: https://github.com/kubevirt/kubevirt/pull/18727
+[PR #18728]: https://github.com/kubevirt/kubevirt/pull/18728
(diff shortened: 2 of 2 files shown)


veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,278 @@
...
+
+- Add DeploymentSimulator and DaemonSetSimulator (like PodSimulator
+  but for Deployment/DaemonSet readiness)
+- Seed install strategy ConfigMaps
+- Deterministic testing of upgrade paths, multi-round reconciliation
+  races, and configuration drift detection
+- ~29 informers, install strategy ConfigMap seeding, infrastructure
+  pod annotation matching required
+
+### Phase 4: CI integration
```
Vaivaswat2244: Should there a defined flakiness bar before this becomes a mandatory pre-merge check (e.g. N consecutive clean CI runs), or is that judged informally once Phase 1–3 land? or should we manually check on this iteratively?
Only asking since the framework's core value prop is being trustworthy enough to gate every PR on.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,278 @@
...
+integration coverage.
+
+## Trade-offs
+
+- **Controller wiring**: Framework replicates `VirtControllerApp`
+  initialization. When controller constructors change, the framework
+  must be updated.
+- **Fidelity gap**: No real hypervisor, networking, or storage. Bugs
+  depending on these remain E2E territory.
+- **envtest binaries**: `kube-apiserver` and `etcd` required via
```
Vaivaswat2244, line 271: Given Phase 4 makes this a required per-PR CI dependency, could you add a note on how kube-apiserver/etcd versions from setup-envtest will be pinned/updated relative to KubeVirt's supported Kubernetes version matrix? Mainly asking to understand whether version skew here is a one-time setup concern or something that needs ongoing maintenance attention.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,278 @@
...
+**Real controllers**: Initialized using the same pattern as
+`VirtControllerApp`, wired to envtest via real `SharedInformerFactory`.
+Currently: VM, VMI, instancetype, and workload-update controllers.
+
+**Pod simulator**: Since envtest has no kubelet, a goroutine watches
+for virt-launcher pods, binds them to a fake node, and sets their
+status to Running. This drives the VMI controller through its
+lifecycle phases.
+
+**Fake libvirt gRPC server**: CGo-free implementation of the
```
Vaivaswat2244: If `CmdServer`'s real gRPC interface changes, will there be a contract/compatibility test that fails CI, or does keeping the fake in sync rely on reviewers catching it manually during normal PR review?

Vaivaswat2244 (commented):
I've left some questions for this. Overall I think this looks pretty good. Happy to help however's most useful once discussion is finalized.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,278 @@
...
+- **Controller wiring**: Framework replicates `VirtControllerApp`
+  initialization. When controller constructors change, the framework
+  must be updated.
+- **Fidelity gap**: No real hypervisor, networking, or storage. Bugs
+  depending on these remain E2E territory.
+- **envtest binaries**: `kube-apiserver` and `etcd` required via
+  `setup-envtest`. Standard for envtest users but new for KubeVirt CI.
+- **Startup cost**: ~5-6s per envtest instance. Tests within an
+  Ordered Context should share one instance to amortize this.
+
```
Vaivaswat2244, line 275: `envtest` is already used in the `kubevirt-migration-controller` project (see [suite_test.go](https://github.com/kubevirt/kubevirt-migration-controller/blob/main/internal/controller/multinamespacestoragemig/suite_test.go)). Notably useful patterns worth adopting, registering multiple API group schemes (KubeVirt core, CDI, project-specific CRDs) into a shared `scheme.Scheme`; a `getFirstFoundEnvTestBinaryDir()` helper that lets tests run from an IDE without manually setting `KUBEBUILDER_ASSETS`; and `ErrorIfCRDPathMissing: true` to fail fast on CRD path misconfiguration. Note that this suite does not itself wire up a controller manager or informers. It's closer to the Layer 1 (envtest bootstrap) than Layer 2 (real controllers + pod simulator).

@awels  also flagged a precaution around resources coming back with missing `Kind` in some cases which came across his development. Might be worth following up with him on the exact repro before documenting the specific mechanism.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Good question. I've added a "CI gating" bullet to the Trade-offs section. The plan is to start as a non-voting job and promote to voting once it has demonstrated stability (zero flakes across multiple release cycles with a representative test count), monitored through the standard KubeVirt CI dashboard. Since envtest tests run against a local kube-apiserver+etcd with no infrastructure dependencies, the main flakiness vectors from E2E (network, node readiness, VM boot timing) don't apply — but we should still prove that empirically before gating PRs on it.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Added a note — `setup-envtest` will be pinned to the latest available version initially. If version skew with KubeVirt's supported Kubernetes versions ever causes issues, the pin can be adjusted to match the support matrix. Since the envtest binaries are just test infrastructure, this should be low-maintenance.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: The framework calls the real `CmdServer` constructors and gRPC service interfaces directly in its `Start()` method — if the interface changes, the framework won't compile. I've updated the Fake libvirt gRPC server paragraph to make this explicit. No separate contract test is needed because the framework *is* a consumer of the real interface.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Thanks for the pointer — I wasn't aware of the envtest usage there. A few notes on the patterns mentioned:

- **Multi-scheme registration**: Our framework already registers all KubeVirt schemes via the existing `AddToScheme` calls, same approach.
- **`getFirstFoundEnvTestBinaryDir`**: Nice IDE convenience. Our PoC currently relies on `KUBEBUILDER_ASSETS` (set via `setup-envtest use -p path`), but we could add something similar for IDE ergonomics.
- **`ErrorIfCRDPathMissing`**: Not directly applicable — our CRDs are generated from Go functions in `pkg/virt-operator/resource/generate/components/crds.go` rather than loaded from disk, so there's no path to misconfigure. The equivalent failure mode (CRD generation broken) would be a compile error.
- **Missing `Kind` issue from @awels**: Good to know — I'll follow up with him. In the PoC I haven't hit this, but it's worth understanding the repro scenario.

0xFelix: https://github.com/kubevirt/virt-template is making use of envtest as well. Possibly other useful bits can be found there too.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,289 @@
...
+- [Issue #18396][]: Workload-update migrations race with virt-handler
+  DaemonSet re-roll during upgrade — a timing race between controllers
+  that E2E tests catch only by luck (it's a flake).
+- Instancetype ControllerRevision upgrades — 8 E2E tests that exercise
+  pure controller logic, migrated to envtest with a 13.4x speedup.
+
+[issue #16719]: https://github.com/kubevirt/kubevirt/issues/16719
+[issue #18396]: https://github.com/kubevirt/kubevirt/issues/18396
+
+## Goals
```
0xFelix, line 77: Another goal could be to test declarative validations on our types. Those are supported by envtest too.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Definition of Users
+
+- **KubeVirt Developers**: Cheap coverage of multi-controller flows
+  and edge cases without provisioning infrastructure.
+- **KubeVirt Reviewers**: Validate that PRs do not break
+  multi-controller interactions.
+- **CI Infrastructure**: Lightweight enough to run alongside unit
+  tests on every PR — no cluster provisioning, no KubeVirt deployment,
+  no VM boot.
```
0xFelix: Reads more like user stories than a definition of users

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,289 @@
...
+multi-controller interactions; mocks verify calls made, not end state
+reached. **Decision**: keep GoMock for unit tests, add envtest for
+integration coverage.
+
+## Trade-offs
+
+- **Controller wiring**: Framework replicates `VirtControllerApp`
+  initialization. When controller constructors change, the framework
+  must be updated.
+- **Fidelity gap**: No real hypervisor, networking, or storage. Bugs
```
0xFelix, line 262: A downside of envtest is that its behavior can occasionally diverge from a real cluster. That could introduce unique, test-specific quirks that lead to false positives, requiring troubleshooting of edge cases that wouldn't actually occur on a real cluster.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Thanks for the pointer! I'll take a look at virt-template's envtest usage for any patterns worth adopting. Between this and the kubevirt-migration-controller reference from @Vaivaswat2244, it's good to see envtest already gaining traction in the KubeVirt ecosystem.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Good point — I've added this as an explicit goal: "Validate CRD declarative validation (structural schemas and CEL rules) against a real API server without deploying a cluster." The PoC already includes 2 CRD validation tests (structural schema and CEL rules), but calling it out as a first-class goal makes sense since envtest is one of the cheapest ways to exercise these against a real API server.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Fair — I've reworked the section to focus on who the users are (contributors, reviewers, CI pipelines) and moved the benefit descriptions into the User Stories section where they belong.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Valid concern. I've added an "envtest divergence" bullet to the Trade-offs section. The key distinction is that envtest runs a *real* kube-apiserver and etcd, so API semantics (validation, resourceVersion, status subresource) are authentic — it's significantly higher fidelity than fake clients. The divergence is in what envtest *doesn't* run: kubelet, scheduler, kube-controller-manager. The pod simulator fills the kubelet gap for controller testing, but tests need to be aware that they're validating controller logic and API interactions, not infrastructure integration. Anything depending on real scheduling, networking, or node behavior stays in E2E.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
+# VEP #371: Envtest Framework
```
EdDev: I am in favor introducing an integration type test in kubevirt.

I have only a small issue with this being classified as meta-VEP.
Meta VEPs, per my understanding and taking Kubernetes as the reference, are aimed at managing the VEPs themselves.

Testing fits sig-testing if we had a formal SIG for it (I think we do not).
I am unsure where else to place it.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+
+## Scalability
+
+Not applicable — test infrastructure only.
+
+## Update/Rollback Compatibility
+
+Not applicable — no production code changes.
+
+## Functional Testing Approach
```
EdDev: I think we are covering here a type of integration test, not a functional test.
We may have an issue with the VEP template, as sometime even unit tests are covered under this section.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+breaks (controller wiring, pod simulator, CRD loading), the tests
+fail.
+
+**Compile-time safety**: Controller constructor signatures and gRPC
+service interfaces (including `CmdServer`) are called directly in the
+framework's `Start()` method. When upstream constructors or interfaces
+change, the framework fails to compile, producing an immediate signal
+— no separate contract test is needed.
+
+**CI integration**: The `make test-envtest` target runs all envtest
```
EdDev: Can we use some other name? I would prefer to name it per its purpose and less per the framework it triggers.

However, if we cannot find something else, we can try to think about something else as a follow up.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Introduce an envtest-based test framework for KubeVirt that exercises
```
EdDev: Per the test coverage you described here, IMO it fits under the category if an integration test, with focus on Kubevirt controllers interaction through the API server.

The envtest part is an implementation detail which is the solution.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+
+## Motivation
+
+KubeVirt currently has two testing layers:
+
+1. **Unit tests** (`pkg/**/*_test.go`): Fast, isolated. Each controller
+   tested with GoMock — Controller A's writes are not visible to
+   Controller B.
+
+2. **E2E tests** (`tests/`): Comprehensive but slow (4-hour timeout),
```
EdDev:
```suggestion
2. **Functional tests** (`tests/`): Comprehensive but slow (4-hour timeout),
```

We tend to call them e2e, but in reality they are only functional. It is rare to see a test exercising a real e2e, especially with all the needed satellite projects.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+  and assert that data is preserved or correctly converted.
+
+[issue #16719]: https://github.com/kubevirt/kubevirt/issues/16719
+[issue #18396]: https://github.com/kubevirt/kubevirt/issues/18396
+[issue #309]: https://github.com/kubevirt/enhancements/issues/309
+
+## Goals
+
+- Exercise multi-controller interactions without a Kubernetes cluster.
+- Use envtest (real kube-apiserver + etcd) for full API fidelity.
```
EdDev: IMO this is the solution, not the goal.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+### Core Components
+
+**envtest**: CRDs loaded from the existing Go generation functions in
+`pkg/virt-operator/resource/generate/components/crds.go`. A real
+`KubevirtClient` created via `GetKubevirtClientFromRESTConfig(cfg)`.
+
+**Real controllers**: Initialized using the same pattern as
+`VirtControllerApp`, wired to envtest via real `SharedInformerFactory`.
+The VM controller and all its synchronizer sub-controllers run
+in-process alongside the VMI controller.
```
EdDev: We have also the migration controller and the virt-operator (which is also a controller).

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+
+**Pod simulator**: Since envtest has no kubelet, a goroutine watches
```
EdDev: This is an in-house logic maintained in Kubevirt or something that is external (i.e. dependency)?

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+for virt-launcher pods, binds them to a fake node, and sets their
+status to Running. Defaults to happy-path scheduling, with optional test hooks to inject delays or failures (e.g., unschedulable pods, OOMKilled/Evicted status). This drives the VMI controller through its
+lifecycle phases.
+
+**Webhook server**: Optional in-process HTTPS server with the real
+KubeVirt `ServeVMs` mutating and validating handlers.
+
+**Fake libvirt gRPC server** (beta): CGo-free implementation of the
```
EdDev: What does "beta" refers to here?

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+  schema diff across all intermediate releases and assert that
+  deprecated fields are correctly converted or explicitly removed.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Core Components
```
EdDev, line 140: In Kubevirt, virt-handler acts partially as a controller, updating the VMI object.
Without it, many scenarios will be missing crucial parts.

I see virt-launcher gRPC server mentioned, but not its client (virt-handler).

EDIT: I just noticed virt-handler mentioned towards the end of the document, as part of the phases. So it seems it was just missed from here.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+- Webhook admission (mutation and validation)
+- Generation tracking, condition management, finalizer cleanup
+- Domain XML generation via the real converter (beta)
+- CRD schema evolution and storage version migration — create objects
+  under one CRD version, apply an updated schema, simulate the
+  GET+PUT migration cycle, and assert field preservation
+
+**Cannot test (needs real infrastructure):**
+- Running guest OS, console access, guest agent
+- Real networking (CNI, masquerade, bridge)
```
EdDev: What "Real" means here?
I guess that runtime networking. But this is true for any runtime (OS) function.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+- Live migration between nodes
```
EdDev: I believe this is testable, as it involves the API layer and the runtime layer, similar to the other VM/VMI lifecycle tests. Any reason you find it not a fit?

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+- CDI integration (DataVolumes, DataSources)
+- Storage (snapshots, restore, export, hotplug volumes)
```
EdDev: The runtime aspect of these is not testable. But there is an API aspect of these. Why can't the API be tested?

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+- virt-operator deployment and upgrade (possible with extensions, see below)
+
+## API Examples
+
+This VEP does not introduce user-facing Kubernetes API changes. The
```
EdDev:
```suggestion
This VEP does not introduce user-facing Kubevirt API changes. The
```

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+### Framework Setup
+
+```go
+f := framework.New(
+    framework.WithWebhooks(),
+    framework.WithWorkloadUpdateController(),
+)
+
+var _ = BeforeEach(func() { f.Start() })
+var _ = AfterEach(func() { f.Stop() })
```
EdDev:
```suggestion
var _ = BeforeEach(func() { f.Start(); DeferCleanup(f.Stop) })
```

veps/meta-VEPs/371-envtest-framework/vep.md
EdDev: > **CI integration**

- Can we run this locally?
- Can you clarify/define if the execution is containerized?

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+in the [KubeVirt K8s Support Matrix][support-matrix], while allowing this
+version to be overridden via an `ENVTEST_K8S_VERSION` environment variable.
+
+[support-matrix]: https://github.com/kubevirt/sig-release/blob/main/releases/k8s-support-matrix.md
+
+**Regression coverage**: Each bug fix merged with an envtest
+regression test provides ongoing validation that the framework
+correctly exercises the relevant controller paths.
+
+**CI gating**: The envtest suite runs as a non-voting job initially.
```
EdDev: Please describe here the final expected solution that we are to expect.
Implementation phases should be described in the dedicated section it has.

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+| VM lifecycle | 1 | VM → VMI → Pod → Scheduled |
+| Generation tracking | 1 | ObservedGeneration/DesiredGeneration, ControllerRevision lifecycle |
+| Regression | 2 | Bug #16719 (RestartRequired stuck), bug #16071 (stale preferenceRef) |
+| Webhook admission | 2 | Mutating defaults, validating rejection |
+| CRD validation | 2 | Structural schema, CEL rules |
+| Domain XML | 1 | Real converter via gRPC |
+| Workload update | 3 | Migration creation, phase gating, deployment ID gating |
+| Instancetype upgrades | 8 | ControllerRevision upgrade from v1beta1 (migrated from E2E) |
+
+**E2E migration example** — instancetype ControllerRevision upgrades:
```
EdDev: `migration` is an overloaded term lately. Can you please elaborate to which migration you refer?

veps/meta-VEPs/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+
+| | E2E (sig-compute lane) | envtest (shared env) |
+|---|---|---|
+| 8 tests | 152.6s total, 19.1s avg | 11.4s total, 1.4s avg |
+| Speedup | — | **13.4x** |
+| Cluster required | Yes | No |
+
+## Implementation Phases
+
+### Alpha (v1.10.0): Framework with VM and VMI controllers
```
EdDev: Consider dropping the "Alpha", "Beta", etc naming and stick to the function or coverage.
In the graduation section these are used and seeing them in multiple locations is confusing. You can ref these sections from there.

I am not that clear about the difference between the two sections now.

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+
+## Implementation History
+
+- **Phase 1 (PoC)**: 20 tests across 8 categories. Framework with
+  VM (inc. instancetype), VMI, and workload-update controllers, pod
+  simulator, webhook server, fake libvirt gRPC server. Instancetype
+  ControllerRevision upgrade tests migrated from E2E (13.4x speedup).
+  Split into focused PRs: #18726–#18731 (supersedes original PR #18238).
+
+## Graduation Requirements
```
EdDev, line 381: IMO the graduation can be linked to the phases you outlines earlier, however, I would mainly focus here on when it becomes part of the regular workflow and its requirements.

I would expect here topics like:
- When is it part of the CI as non-gating and gating?
- When is it part of the VEP template as optional or required?
- When the guidelines are introduced and settled on, about when such tests are nice-to-have or a hard requirement? These will effect both VEPs and k/kubevirt contribution guidelines.

I am unsure if we want to have the coverage part related to the points above.
Technical dept is hard to complete, the linter coverage is an example.

EdDev (changes requested):
Thank you!

Looking forward to see this introduced. It will give us better tools to examine API scenarios with less cost and quicker feedback loop.

Hopefully it will also allow the functional tests to level up a bit, to check mainly user-facing results and behaviors and less API mechanics.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Agreed — moved to `veps/sig-control-plane/371-envtest-framework/` as it primarily exercises components under sig-control-plane's ownership.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Agreed — renamed throughout. The new layer is consistently called 'integration tests'; the existing `tests/` layer is now 'functional tests'. The section heading has been updated to 'Integration Testing Approach'.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Renamed to 'KubeVirt Integration Test Framework' — purpose-based rather than tied to the envtest implementation detail.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Addressed as part of the naming and terminology cleanup. The Design section now also introduces envtest as the implementation mechanism without conflating it with the purpose.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Applied.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Agreed — rewritten as 'Achieve full API fidelity by testing against a real kube-apiserver and etcd.'

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Added — the Real controllers paragraph now explicitly calls out virt-handler (noting it acts as a partial controller updating VMI status), the workload-update (migration) controller, and virt-operator as optional components with their corresponding `With*` options.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Clarified — added a note that the fake libvirt gRPC server is implemented and maintained in-house within `tests/envtest/` in kubevirt/kubevirt.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: These referred to VEP graduation phases. Removed all phase qualifiers from the Design section — Design now describes the final end-state only. Phase-specific timing is covered in Implementation Phases.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Added to Core Components — virt-handler is now explicitly called out with a note that it acts as a partial controller (updating VMI status), alongside the other optional components.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Changed to 'Runtime networking dataplane (CNI plugins, masquerade, bridge forwarding)' to be explicit about what can't be tested.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Good point — CDI DataVolume/DataSource API interactions with the VM controller have been moved to Phase 3 as future API-layer coverage, analogous to how Nova tests Cinder interactions through the API layer without a real backend.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Addressed — the Cannot test section now distinguishes runtime/dataplane (not testable) from API and controller interactions (testable in Phase 3), covering live migration API, CDI, and storage API.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Applied.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Applied.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Added two clarifying sentences: the suite runs locally with `make test-integration` (`setup-envtest` downloads binaries automatically on first run, no cluster needed), and in CI it uses the standard KubeVirt unit test container with no additional image required since there is no CGo dependency.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Agreed — the Design section now describes the final end-state only. All phase-specific timing (which options arrive in which phase) has been moved to Implementation Phases.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Clarified to 'Workload-update controller: VM live migration creation and gating logic during KubeVirt upgrades'.

veps/meta-VEPs/371-envtest-framework/vep.md
lyarwood: Renamed to Phase 1, Phase 2, Phase 3 with descriptive subtitles. The Graduation Requirements section uses Non-gating/Gating rather than Alpha/Beta/GA to avoid the confusion between graduation stages and phase names.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Simplified — Graduation Requirements now has two sections: Non-gating (make test-integration as a non-voting job, flakiness monitoring, contributor guidelines, VEP template update) and Gating (promoted to voting check once stable, integration tests required for controller PRs, contributor guide published).

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,460 @@
...
+  and assert that data is preserved or correctly converted.
+
+[issue #16719]: https://github.com/kubevirt/kubevirt/issues/16719
+[issue #18396]: https://github.com/kubevirt/kubevirt/issues/18396
+[issue #309]: https://github.com/kubevirt/enhancements/issues/309
+
+## Goals
+
+- Exercise multi-controller interactions without a Kubernetes cluster.
+- Achieve full API fidelity by testing against a real kube-apiserver and etcd.
```
0xFelix: It behaves like a real API, but crucially the kubelet is missing. So all built in resources will not work like on a real API?

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,460 @@
...
+
+**Real controllers**: Initialized using the same pattern as
+`VirtControllerApp`, wired to envtest via real `SharedInformerFactory`.
+The VM controller and all its synchronizer sub-controllers run
+in-process alongside the VMI controller. virt-handler (which also acts
+as a controller, updating VMI status), the workload-update (migration)
+controller, and virt-operator are also supported as optional components
+(`WithVirtHandler`, `WithWorkloadUpdateController`, `WithVirtOperator`).
+
+**Pod simulator**: Since envtest has no kubelet, a goroutine watches for
```
0xFelix, line 166: Mention that it is maintained as part of the framework and not as part of envtest?

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,460 @@
...
+f := framework.New(framework.WithWebhooks())       // + admission webhooks
+f := framework.New(framework.WithWorkloadUpdateController()) // + workload-update controller
+f := framework.New(framework.WithFakeLibvirt())    // + domain XML conversion
+```
+
+Options compose — multiple can be combined in one `New()` call.
+
+### What envtest can and cannot test
+
+**Can test (controller logic):**
```
0xFelix, line 199: What about live updates?

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,460 @@
...
+only `setup-envtest` binaries (kube-apiserver + etcd), no cluster
+provisioning. By default, `setup-envtest` targets the latest Kubernetes
+version supported by the target KubeVirt release as defined in the
+[KubeVirt K8s Support Matrix][support-matrix], while allowing this version
+to be overridden via an `ENVTEST_K8S_VERSION` environment variable.
+
+The suite can be run locally with `make test-integration` — `setup-envtest`
+downloads the required binaries automatically on first run. No container
+or cluster is needed; the target runs natively using the local Go toolchain,
+unlike unit tests which are wrapped in `hack/dockerized`. In CI, the
```
0xFelix: Why not (optionally) wrap it in dockerized? Isn't that a requirement for the ci?

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Addressed — the 'full API fidelity' claim in both the Goals section and the envtest Design description has been qualified to make clear it applies to CRD resources only. The absence of the kubelet, scheduler, and kube-controller-manager is now called out explicitly, with a note that built-in Kubernetes resources (e.g. Pods) are simulated by framework components such as the pod simulator.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Addressed — live updates (CPU, memory, and volume hotplug) are now listed under 'Can test', noting that controller-level request handling and status tracking are in scope with virt-handler simulation, analogous to how live migration is handled.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Good point, agreed — the CI integration section has been updated to say `make test-integration` runs via `hack/dockerized` (same as `make test`), with `kube-apiserver` and `etcd` binaries baked into the builder image at a pinned K8s version. Two PRs have been opened to land this: kubevirt/kubevirt#18841 (add envtest binaries to builder image) and kubevirt/kubevirt#18842 (default to CS10 builder).

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,289 @@
...
+- [Issue #18396][]: Workload-update migrations race with virt-handler
+  DaemonSet re-roll during upgrade — a timing race between controllers
+  that E2E tests catch only by luck (it's a flake).
+- Instancetype ControllerRevision upgrades — 8 E2E tests that exercise
+  pure controller logic, migrated to envtest with a 13.4x speedup.
+
+[issue #16719]: https://github.com/kubevirt/kubevirt/issues/16719
+[issue #18396]: https://github.com/kubevirt/kubevirt/issues/18396
+
+## Goals
```
lyarwood: This is already present in the Goals section: 'Validate CRD declarative validation (structural schemas and CEL rules) against a real API server without deploying a cluster.' Let me know if you'd like it worded differently or made more prominent.

0xFelix: Hello AI :wave: This can be resolved :)

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,468 @@
...
+- This VEP targets alpha for version: v1.10.0
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
jschintag: Nit: These two should be checked

lyarwood:
Pushed a revision pass that runs this VEP through the VEP writing guide proposed in #431, applying its section-by-section guidance:

- **Goals** de-mechanized — stated as outcomes rather than "how"/"without a cluster".
- **Trade-offs** folded into **Design** as a *Risks and Mitigations* subsection.
- Renamed *Integration Testing Approach* → **Functional Testing Approach** (matching the template heading) and clarified the unit/integration/functional test tiers.
- **Graduation Requirements** reframed as **Alpha/Beta/GA**, with an explicit note on why Feature Gate / On-By-Default Readiness criteria do not apply to test-only infrastructure; per-phase build scope folded into each stage.
- Abstracted internal file/function/type names out of **Design** (per the guide, code churns and the VEP should describe abstractions that survive refactoring).
- Moved the PoC per-PR breakdown, coverage matrix, and benchmark data into an appendix (`poc.md`) so `vep.md` stays focused on the feature.
- **Implementation History** is now a terse dated log.

@EdDev The guide in #431 was a useful checklist here — thanks for putting it together.

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -341,128 +355,84 @@ build container available.
...
+  instancetype), VMI, and workload-update controllers, pod simulator,
+  webhook server, and fake libvirt gRPC server. See
+  [Proof of Concept](./poc.md).
 
 ## Graduation Requirements
 
-The phases above describe what gets built. This section describes when
-the framework becomes part of the regular contributor and CI workflow.
+This is test-only infrastructure: it ships no user-facing runtime
+behaviour, introduces no Feature Gate, and creates no API objects or VM
```
0xFelix, line 384: If there is no Feature Gate, how do we keep track of what is Alpha, Beta, or GA? Perhaps track it targeted to versions only instead?

lyarwood: Yeah that's the intention, I can reference these versions as we go through the process here I just wanted to keep this document structure aligned with a traditional vep.

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+drives the VMI controller through its lifecycle phases.
+
+**Webhook server**: Optional in-process HTTPS server running the real
+KubeVirt mutating and validating admission handlers.
+
+**Fake libvirt gRPC server**: CGo-free implementation of the
+virt-launcher command gRPC interface, maintained in-house alongside the
+framework in kubevirt/kubevirt. Runs the real VMI-to-domain XML
+converter and stores results in memory. Tests can assert on domain XML
+without any libvirt C library dependency.
```
jschintag: How much effort would this part be? It sounds to me as if it might be it's VEP worth of effort to implement.

veps/sig-control-plane/371-envtest-framework/vep.md
```diff
@@ -0,0 +1,438 @@
...
+- VM/VMI lifecycle, phase transitions, status updates
+- ControllerRevision creation, upgrade, conflict detection
+- Instancetype/preference application to VMI spec
+- Workload-update controller: VM live migration creation and gating logic during KubeVirt upgrades
+- Live updates (CPU, memory, and volume hotplug) — controller-level request handling and status
+  tracking, with virt-handler simulation (analogous to live migration simulation)
+- CRD structural schema and CEL validation
+- Webhook admission (mutation and validation)
+- Generation tracking, condition management, finalizer cleanup
+- Domain XML generation via the real converter
```
jschintag: Isn't the Domain XML generation already covered in unit-tests?

jschintag (commented):
Thank you for the VEP @lyarwood.

I think overall it looks good, though i haven't looked at the PoCs yet.

My main hold up is that i can't estimate what the effort for the mock gRPC server would be, but reading the VEP my impression is that it is easy?

veps/sig-control-plane/371-envtest-framework/vep.md
jschintag: Also i do not think libvirt C library dependency is an issue, libvirt-devel is already installed in the builder image anyway.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: It is but this would allow us to assert the output domain XML given to libvirt starting from the VirtualMachine or VirtualMachineInstance given to the k8s API. So that would include coverage of defaults injected by our webhooks, controllers etc that were previously only something we could assert in our functional tests.

veps/sig-control-plane/371-envtest-framework/vep.md
lyarwood: Yeah it's non-trivial so I kicked it out to beta / v1.11.0 for now and it might require a follow up VEP or supporting doc.

veps/sig-control-plane/371-envtest-framework/vep.md
jschintag: Ah, that makes more sense now.

jschintag (approved):
I reviewed the first PoC and it looked great. I will take a quicker look through the others.

Overall i think this VEP is in a good place.

/lgtm

stu-gott:
Nice! This VEP is well thought out and presents a clear case.

This is not a disagreement, but something we need to stress, is that this approach is complementary to existing testing frameworks. Mocks, by nature, can hide behavior divergence, API changes at lower levels, race conditions/timing or just drift away from the real code over time. Thus we can never solely rely on this as a single source of truth. We can never fully get rid of all E2E tests in favor of this approach.

However the massive advantages laid out here cannot be understated. Well done.

jschintag:
Looking at the PoC, the updates to the Synchronizer logic could be a Bug Fix, so i do not think they would need their own VEP, but they seem to be required and are not mentioned in the VEP?

lyarwood:
> Looking at the PoC, the updates to the Synchronizer logic could be a Bug Fix, so i do not think they would need their own VEP, but they seem to be required and are not mentioned in the VEP?

@jschintag Yeah that's already broken out into a [separate bugfix PR](https://github.com/kubevirt/kubevirt/pull/16499). I'm hopeful that it doesn't require a VEP as it's an internal implementation detail that a few Synchronizers (my own included) have got wrong until now.

lyarwood:
> Nice! This VEP is well thought out and presents a clear case.
> 
> This is not a disagreement, but something we need to stress, is that this approach is complementary to existing testing frameworks. Mocks, by nature, can hide behavior divergence, API changes at lower levels, race conditions/timing or just drift away from the real code over time. Thus we can never solely rely on this as a single source of truth. We can never fully get rid of all E2E tests in favor of this approach.
> 
> However the massive advantages laid out here cannot be understated. Well done.

Thanks @stu-gott, this was suggested but I've added an explicit [callout](https://github.com/kubevirt/enhancements/compare/4230f87bec6d244f7a4bedf832ed8b6fc80de35e..e8d7374102740ed37e084eeacf9242997012b194) for this now. @jschintag @stu-gott @xpivarc hopefully that's enough but let me know if you want something more.

jschintag:
Thanks @lyarwood, from my side this is 

/lgtm

0xFelix (approved):
Excellent work!

/lgtm

lyarwood:
@xpivarc @vladikr this should also be ready to merge now ahead of the deadline

xpivarc (commented):
/approve

Issue #349, opened 2026-06-24
VEP: Block-mode backend storage for persistent VM state (EFI NVRAM, vTPM)


kvaps:
**Primary contact (assignee)**:

/assign @kvaps

**Current Feature Stage**:

New

**Feature Gate**:

VMStateBlockVolumeMode

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

VEP PR: kubevirt/enhancements#350

**Timeline**:

* Alpha release target (x.y): TBD
* Beta release target (x.y): TBD
* Stable release target (x.y): TBD

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs: kubevirt/kubevirt#18215
  * [ ] Docs PRs:

akalenyu:
@kvaps hey, checking in, we have some comments on the VEP - https://github.com/kubevirt/enhancements/pull/350#issuecomment-5416279342 (I believe this VEP predates the CDI PR closing over RWO FS being fine)

PR #350, opened 2026-06-24
VEP-349: Block-mode backend storage for persistent VM state (EFI NVRAM, vTPM)

kvaps:
### VEP Metadata

**Tracking issue**: kubevirt/enhancements#349
**SIG label**: /sig storage

### What this PR does

Proposes an **optional** `Block` volume mode for the persistent vm-state backend storage, so persistent EFI NVRAM and vTPM state can live natively on block storage (DRBD/LINSTOR, Ceph RBD) — on the same raw volumes the VM's disks already use, with no filesystem layer, and (on RWX-Block) without a per-migration state copy.

Opened at the request of @mhenriks and @jean-edouard on kubevirt/kubevirt#18215.

To be clear after the discussion there: **this is not a migration fix.** RWO-Filesystem backend storage already live-migrates via the copy-on-target mechanism (kubevirt/kubevirt#12629), so Block mode is not required for migratability — the VEP now says so explicitly. Block mode is offered as an additive, opt-in option for operators whose storage is natively block:

- **Block-native, no filesystem layer** — the small vm-state volume uses the same block storage as the VM's disks, instead of a Filesystem PVC (and, where shared/RWX state is wanted on DRBD/Ceph, without an NFS-Ganesha/CephFS layer just for it).
- **RWX-Block skips the per-migration copy** — the backend volume is shared by source and target during migration (an optimization for a tiny blob, not a correctness requirement).

vTPM on Block is implemented in the same PR (swtpm `file://` backend + libvirt `<source type='file'>`, on a second raw device); CBT stays Filesystem-only.

/sig storage
/cc @mhenriks @jean-edouard @akalenyu @awels

### DIFF
--- a/veps/sig-storage/349-block-mode-vmstate/vep.md
+++ b/veps/sig-storage/349-block-mode-vmstate/vep.md
@@ -52,8 +52,8 @@ volume lets it live-migrate on the storage the cluster already runs, with no NFS
 
 - Allow the persistent vm-state backend volume to be created in `Block` mode.
 - Store EFI NVRAM directly on the raw block device (no filesystem) and keep it live-migratable.
-- Store vTPM state on the same kind of block device via swtpm's single-file backend, keeping it
-  live-migratable.
+- Store vTPM state on a separate raw block device (one device per blob) via swtpm's single-file
+  backend, keeping it live-migratable.
 - Keep `Filesystem` the default; no behavior change for existing users; guard the new behavior behind
   a feature gate during alpha.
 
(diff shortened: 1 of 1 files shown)


lllamnyp (approved):

jean-edouard (commented):
Thanks for the thorough VEP, @kvaps!

I appreciate the effort, but I'm personally not in favor of this proposal. My reasoning:

- Every CSI driver (including LINSTOR) supports Filesystem mode. With RWO-FS, libvirt migrates the EFI/TPM state for us through the normal migration stream, so live migration works out of the box. Block mode doesn't unlock any new capability.
- This adds a second code path through all of backend-storage: creation, pod rendering, migration markers, template seeding, device sizing. Every future change has to consider both modes. Migration recovery (kubevirt/kubevirt#12629) relies on filesystem markers that don't exist on Block.
- CBT already can't work on Block. Backend-storage is a general-purpose mechanism that may grow to serve additional use cases, and Block mode would be a permanent constraint on that design space.

Interested to hear other opinions on this!

akalenyu:
> Thanks for the thorough VEP, @kvaps!
> 
> I appreciate the effort, but I'm personally not in favor of this proposal. My reasoning:
> 
> * Every CSI driver (including LINSTOR) supports Filesystem mode. With RWO-FS, libvirt migrates the EFI/TPM state for us through the normal migration stream, so live migration works out of the box. Block mode doesn't unlock any new capability.
> * This adds a second code path through all of backend-storage: creation, pod rendering, migration markers, template seeding, device sizing. Every future change has to consider both modes. Migration recovery ([kubevirt/kubevirt#12629](https://github.com/kubevirt/kubevirt/pull/12629)) relies on filesystem markers that don't exist on Block.
> * CBT already can't work on Block. Backend-storage is a general-purpose mechanism that may grow to serve additional use cases, and Block mode would be a permanent constraint on that design space.
> 
> Interested to hear other opinions on this!

I think this VEP predates https://github.com/kubevirt/containerized-data-importer/pull/4180#issuecomment-4843715988 but will let @kvaps comment on that :pray:

stu-gott:
I think @jean-edouard's point is: if every CSI driver already supports Filesystem mode and it works for migration, what's the _concrete benefit_ that outweighs permanently maintaining two parallel implementations?

@kvaps, Is there a clear value-add (performance benchmarks, specific storage backends that benefit, operational advantages)? Optional features still have maintenance costs, and those costs need to be justified by real benefits, not just architectural preference.

I'm not disagreeing that RWX-block has some value-add, I just wonder if the upside is enough to justify the extra code path.

Do we have empirical data that RWX-block is meaningfully faster than RWO-FS copy? (It's a ~1MB file, right?)

Do we have evidence that we have enough users desiring RWX-block to justify the maintenance burden?

Are there future use cases (e.g. larger state stores) that would benefit from RWX-block?

PR #354, opened 2026-06-28
VEP-353: Component Based Testing Framework for Kubevirt

Vaivaswat2244:
#### What this PR does
Introduces VEP-353: Component-Based Testing Framework.
KubeVirt currently has no testing layer between fast unit tests (which mock all API interactions) and slow e2e tests (45+ min, full cluster). This VEP proposes a reusable pkg/testframework/component/ package using envtest to test KubeVirt controllers against a real API server in-process — no cluster, no libvirt, no VMs required. The pilot target is virt-controller.

#### Special notes for your reviewer

- This VEP has no user-facing API changes and no feature gate — it is purely internal testing infrastructure
- The pilot suite targets virt-controller specifically because it is a pure Kubernetes controller whose behavior is fully observable via the API
- Alpha graduation criteria are intentionally scoped conservatively (10+ tests, <5 min CI runtime); Beta and GA criteria expand coverage to additional components
- This proposal originated as a GSoC 2026 submission — the full original proposal is included in the VEP.

### DIFF
--- a/veps/sig-compute/353-component-based-testing-framework/vep.md
+++ b/veps/sig-compute/353-component-based-testing-framework/vep.md
@@ -1,0 +1,250 @@
+# VEP #NNNN: Component-Based Testing Framework
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: TBD
+- This VEP targets beta for version: TBD
+- This VEP targets GA for version: TBD
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


lyarwood:
Hey @Vaivaswat2244, thanks for putting this together! I've been working on a very similar proposal in #348 with a working PoC at https://github.com/kubevirt/kubevirt/pull/18238. The motivation and core technology (envtest) are essentially the same.

A few key differences worth discussing: my framework runs multiple controllers together (e.g. VM + VMI) so it can catch bugs that span controller boundaries, like https://github.com/kubevirt/kubevirt/issues/16719 where the VM controller's `RestartRequired` condition was never cleared after a spec revert. It also includes a pod simulator to handle the no-kubelet gap in envtest and reuses existing KubeVirt code (libvmi builders, kubecli.KubevirtClient, CRD generation functions) rather than introducing new abstractions. 

I think there's a lot of overlap here and it would be worth collaborating on a single proposal. Happy to discuss further if you're interested!

Vaivaswat2244:
Thanks for pointing me to #348 and the PoC. Just went through it, and yeah, the overlap is real. Approach in #348  is more complete than this one: multi-controller reconciliation, catching cross-boundary bugs like #16719 is exactly the gap a single-controller test can't cover), the pod simulator solving the no-kubelet problem, and reusing existing KubeVirt code instead of new abstractions is the right call.

I'd rather consolidate into #348 than maintain two overlapping proposals. I'll plan to close this one out once we've got a direction. I'd like to contribute wherever it's most useful. Let me know how you'd like to structure this and where a second pair of hands would help most.

Closing this now

PR #357, opened 2026-06-28
VEP 356: Test Infrastructure Abstraction for Managed KubeVirt Environments

dasionov:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/356
**SIG label**: /sig compute

### What this PR does

Introduces VEP 356, proposing Go interface abstractions for test infrastructure configuration
mutations. This enables KubeVirt e2e tests to run on HCO-managed environments without maintaining
separate test forks.

Key design points:
- Focused interfaces (`FeatureGateManager`, `KubeVirtConfigManager`, `MigrationConfigManager`)
- Default implementation wraps existing direct KubeVirt CR patching
- HCO managed implementation patches HyperConverged CR instead
- Compile-time selection via `managed_hco` Go build tag
- No feature gate -- test infrastructure only, no production code changes
- Thin wrapper migration -- existing helper function signatures preserved

This VEP does not modify KubeVirt production code and does not follow the standard Alpha/Beta/GA
feature lifecycle. It uses a flat Completion Criteria checklist instead.

### DIFF
--- a/veps/sig-compute/test-infra-managed-environments.md
+++ b/veps/sig-compute/test-infra-managed-environments.md
@@ -1,0 +1,508 @@
+# VEP 356: Test Infrastructure Abstraction for Managed KubeVirt Environments
+
+## VEP Status Metadata
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Graduation criteria filled
+
+This VEP does not follow the standard Alpha/Beta/GA lifecycle because it modifies test infrastructure only,
+not KubeVirt production code. See [Completion Criteria](#completion-criteria) for graduation requirements.
+
+## Overview
+
+KubeVirt end-to-end tests frequently modify the KubeVirt Custom Resource (CR) to toggle feature gates, adjust
+migration configuration, and alter other cluster-wide settings. These modifications work in standalone KubeVirt
+deployments, but fail in environments where KubeVirt is managed by an overarching operator such as the
+HyperConverged Cluster Operator (HCO). In managed environments, the managing operator owns the KubeVirt CR and
+reconciles it, reverting any direct modifications made by tests.
+
+This VEP proposes introducing a set of small, focused Go interfaces that abstract how tests mutate KubeVirt
+configuration. A default implementation preserves today's behavior (direct KubeVirt CR patching), while alternative
+implementations can target the managing operator's API instead. All interface definitions and implementations will
+reside in `tests/libkubevirt/config/`, colocated with the existing test helpers they wrap.
+
+## Motivation
+
+KubeVirt's test suite contains serial tests that directly patch the KubeVirt CR via strategic merge patch or JSON
+patch to configure the system under test. The key functions driving this behavior are:
+
+- `UpdateKubeVirtConfigValueAndWait()` -- strategic merge patches the entire `spec.configuration` of the KubeVirt
+  CR, then waits for propagation to virt-controller, virt-api, and virt-handler. Used by ~35 direct call sites
+  across the test suite, with additional indirect usage through wrapper functions.
+- `EnableFeatureGate()` / `DisableFeatureGate()` -- manipulate the `FeatureGates` and `DisabledFeatureGates` slices
+  on the KubeVirt CR.
+- `RegisterKubevirtConfigChange()` -- applies targeted JSON patches to the KubeVirt CR using functional options
+  (`KvChangeOption`).
+- Migration-specific helpers like `SetDedicatedMigrationNetwork()` that modify migration configuration fields.
+
+This approach assumes exclusive ownership of the KubeVirt CR, which is true in a vanilla KubeVirt deployment but
+not in managed environments.
+
+In managed environments (e.g., OpenShift Virtualization), HCO acts as a meta-operator that deploys and manages
+KubeVirt, CDI, and other operands. The HyperConverged CR is the single source of truth -- HCO copies configuration
+values from it to the operand CRs. If a test directly modifies the KubeVirt CR, HCO's reconciliation loop detects
+the out-of-band change and reverts it, incrementing the `kubevirt_hco_out_of_band_modifications_total` metric. This
+makes tests that modify KubeVirt configuration unreliable or entirely non-functional in these environments.
+
+Currently, the test framework has **no awareness of managed environments** -- there are no skip conditions,
+decorators, or alternative code paths for HCO-managed clusters. This gap prevents KubeVirt tests from being reused
+on managed platforms, forcing consumers to maintain separate test forks or skip large portions of the test suite.
+
+Enabling tests to run correctly on managed environments improves the overall testing coverage of KubeVirt as users
+actually deploy it, and reduces the maintenance burden of separate test forks.
+
+## Goals
+
+- Define small, focused Go interfaces for each category of KubeVirt configuration mutation performed by tests
+  (feature gates, migration configuration, general KubeVirt configuration, etc.).
+- Provide a **default implementation** that patches the KubeVirt CR directly, preserving current behavior with
+  zero regressions.
+- Define an **HCO implementation** that patches the HyperConverged CR instead of the KubeVirt CR directly.
+- Place all interface definitions and implementations in `tests/libkubevirt/config/`, colocated with the existing
+  test helpers they wrap. All consumers of these interfaces are internal to the test suite.
+- Migrate existing test utilities (`EnableFeatureGate`, `DisableFeatureGate`, `UpdateKubeVirtConfigValueAndWait`,
+  `RegisterKubevirtConfigChange`, etc.) to use the new interfaces.
+
+## Non Goals
+
+- Modifying or extending HCO itself. The managed implementation will use HCO's existing API (patching the
+  HyperConverged CR).
+- Replacing or refactoring the KubeVirt CR API. The interfaces abstract the *test infrastructure*, not the
+  product API.
+- Automatically detecting whether the environment is managed. The implementation to use will be selected at compile
+  time via Go build tags, not auto-detection. Auto-detection may be considered as a future convenience enhancement
+  but is explicitly deferred from this proposal.
+- Covering all possible managing operators. The initial managed implementation targets HCO specifically. The
+  interface design allows additional implementations in the future.
+- Modifying non-KubeVirt-CR resources like `MigrationPolicy` CRDs, which are separate custom resources not owned
+  by HCO's reconciliation.
+
+## Definition of Users
+
+- **KubeVirt test developers**: Write and maintain KubeVirt e2e tests. They use the new interfaces transparently --
+  the default implementation preserves their current workflow.
+- **Managed platform test consumers**: Teams running KubeVirt tests on managed platforms (e.g., OpenShift
+  Virtualization on OCP). They provide a managed implementation and configure the test suite to use it.
+
+## User Stories
+
+### As a KubeVirt test developer
+I want to write tests that toggle feature gates and modify KubeVirt configuration without worrying about the
+deployment topology, so that my tests work on both standalone and managed environments.
+
+### As a managed platform test consumer
+I want to run KubeVirt e2e tests against my HCO-managed cluster by swapping in a managed implementation, so that
+I can validate KubeVirt behavior as real users experience it without maintaining a separate test fork.
+
+## Repos
+
+- `kubevirt/kubevirt` -- primary repo; interfaces, default implementation, and test migration
+- `kubevirt/hyperconverged-cluster-operator` -- reference for HCO API; no code changes required in HCO itself
+
+## Design
+
+### Interface Definitions
+
+The design introduces a set of focused interfaces, each covering a specific category of KubeVirt configuration
+mutation. All interfaces and implementations are colocated in `tests/libkubevirt/config/`, close to the existing
+test helpers they wrap.
+
+```go
+// FeatureGateManager controls KubeVirt feature gate activation.
+type FeatureGateManager interface {
+    EnableFeatureGate(feature string) error
+    DisableFeatureGate(feature string) error
+}
+
+// KubeVirtConfigManager manages KubeVirt cluster-wide configuration.
+// UpdateConfiguration uses merge semantics (not replace): only fields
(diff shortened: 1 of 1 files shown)


veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+Enabling tests to run correctly on managed environments improves the overall testing coverage of KubeVirt as users
+actually deploy it, and reduces the maintenance burden of separate test forks.
+
+## Goals
+
+- Define small, focused Go interfaces for each category of KubeVirt configuration mutation performed by tests
+  (feature gates, migration configuration, general KubeVirt configuration, etc.).
+- Provide a **default implementation** that patches the KubeVirt CR directly, preserving current behavior with
+  zero regressions.
+- Define a **managed implementation** (e.g., `ManagedKubeVirt`) that patches the managing operator's API (e.g.,
```
lyarwood: Why don't we just call this `HCO` given it's specifically for that operator and not generic to other managed operators?

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+  the HyperConverged CR) instead of the KubeVirt CR directly.
+- Place interface definitions in the `staging/` area of the repository so that external consumers can depend on
+  the contract without importing the full `kubevirt/kubevirt` module. Implementations reside in the main module
+  under `tests/` due to their dependencies on test infrastructure packages.
```
lyarwood: Wouldn't all consumers of this interface be internal? We'd support external implementations but they wouldn't depend on the interface, only implement it.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+- `kubevirt/kubevirt` -- primary repo; interfaces, default implementation, and test migration
+- `kubevirt/hyperconverged-cluster-operator` -- reference for HCO API; no code changes required in HCO itself
+
+## Design
+
+### Interface Definitions
+
+The design introduces a set of focused interfaces, each covering a specific category of KubeVirt configuration
+mutation. All interfaces reside in the staging area under a new package
+(e.g., `staging/src/kubevirt.io/client-go/testing/config/`).
```
lyarwood: As above I think these need to be defined close to the consumer of these implementations.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+  unaffected.
+- **Managed platform consumers**: Can adopt the managed implementation at their own pace.
+
+## Functional Testing Approach
+
+- **Unit tests**: Each interface implementation (direct and managed) will have unit tests using fake clients to
+  verify correct patch generation and API interactions.
+- **Standalone CI**: Runs with the default (direct) implementation. Any regression in existing test behavior
+  indicates a bug in the migration, not the design.
+- **Managed CI**: A separate CI lane builds and runs the KubeVirt test suite with `-tags managed_hco` against an
```
lyarwood: Which SIG lane(s) would this cover?

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+
+The managed implementation must:
+1. Patch the HyperConverged CR.
+2. Poll the KubeVirt CR until the expected configuration change appears (confirming HCO has reconciled).
+3. Read the KubeVirt CR's `ResourceVersion` after HCO reconciliation.
+4. Run the existing propagation wait logic using that `ResourceVersion`.
+
+The total timeout for the managed path will be longer than the direct path to account for HCO reconciliation delay.
+
+### Field Mapping: KubeVirt CR to HyperConverged CR
```
lyarwood, line 230: FWIW I don't think you need this level of implementation detail in the VEP.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+The managed implementation must handle these unmappable fields gracefully. The recommended approach is to document
+which fields are unsupported and return an error if a test attempts to set them through the managed path. Tests
+that require these fields may need to be skipped or annotated for managed environments.
+
+### Additional KubeVirt CR Mutation Patterns
+
+Beyond the helper functions listed above, several test files directly patch the KubeVirt CR bypassing all
+abstractions. These include modifications to:
+
+- `spec.customizeComponents` (canary upgrade tests, operator tests)
```
lyarwood: This has been dropped btw

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,539 @@
...
+- `spec.uninstallStrategy` (operator tests)
+- `spec.certificateRotateStrategy` (operator tests, export tests)
+- `spec.infra` (virt-API auto-scaling)
+- `spec.workloads` / `NodePlacement` (VMI lifecycle tests)
+- `metadata.annotations` (Velero backup tests)
+
+These direct-patching patterns are out of scope for the initial interface design. Most are confined to
+operator-specific tests that verify KubeVirt operator behavior itself (not user-facing functionality). The
+interfaces target the most common test pattern: modifying `spec.configuration` to set up the system under test.
```
lyarwood: Should this be merged with the section above?

lyarwood (commented):
Finally found time to look at this, definitely agree with the direction but I have a few nits and questions.

veps/sig-compute/test-infra-managed-environments.md
dasionov: good call, done

veps/sig-compute/test-infra-managed-environments.md
dasionov: agreed

veps/sig-compute/test-infra-managed-environments.md
dasionov: done

veps/sig-compute/test-infra-managed-environments.md
dasionov: starting with sig-compute, we can exapnd later to other sigs once compute is stable

veps/sig-compute/test-infra-managed-environments.md
dasionov: Trimmed the detailed bullet list down to a summary paragraph

veps/sig-compute/test-infra-managed-environments.md
dasionov: Removed, thanks for the heads up.

veps/sig-compute/test-infra-managed-environments.md
dasionov: sure

dasionov:
[Change](https://github.com/kubevirt/enhancements/compare/bbc8846b5c8331f54d4acb182f9aec4b1c24ba11..105d101c072fd40fcb4b9068183b09c0d86316b6): Address @lyarwood's Comments

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+) error {
+    // Map KubeVirt configuration fields to their HyperConverged CR
+    // equivalents under spec.virtualization.*, spec.security.*, etc.
+    // Patch the HyperConverged CR.
+    // Wait for HCO reconciliation to propagate changes to KubeVirt CR.
+    // Then wait for KubeVirt component propagation (healthz polling).
+}
+
+// ... other methods follow the same mapping pattern
+```
```
iholder101, line 205: One concern I have with the current design is that it hardcodes HCO awareness into the upstream test infrastructure (the `managed_hco` build tag, HCO-specific implementation). Ideally, Kubevirt shouldn't favor a specific deployment method. This isn't just an architectural purity point - Helm-based deployments with reconciliation loops (ArgoCD, Flux) have the exact same problem, and the current VEP wouldn't help them.

Ideally, the upper layers that deploy Kubevirt will supply their own tests adjustment in their own repo and stay completely out-of-tree for Kubevirt.

Have you considered a more generic, out-of-tree approach? For example, the test infrastructure could call an external binary for config mutations when an env var like `KUBEVIRT_CONFIG_PROVIDER=/path/to/binary` is set. The contract would be simple - "set this config field to this value" - and the provider decides how. HCO users supply a binary that patches the HyperConverged CR, Helm users supply one that updates values and triggers a sync, and the default (no env var) keeps the current direct KubeVirt CR patching.

This way upstream KubeVirt doesn't need to know about HCO types or any other deployment tool, and new deployment methods can be supported without upstream changes.

WDYT?

lyarwood: Agreed, I somehow missed that you want to carry the HCO implementation in-tree. The only implementation we should be carrying is the existing KubeVirt CR logic, everything else needs to live elsewhere.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+    DisableFeatureGate(feature string) error
+}
+
+// KubeVirtConfigManager manages KubeVirt cluster-wide configuration.
+// UpdateConfiguration uses merge semantics (not replace): only fields
+// explicitly set in the provided config are applied. Zero-valued fields
+// are ignored, matching the behavior of the existing strategic merge
+// patch approach.
+type KubeVirtConfigManager interface {
+    UpdateConfiguration(config v1.KubeVirtConfiguration) error
```
Barakmor1, line 127: have you thought about passing a context.Context as part of the API? It could be useful for handling timeouts, cancellation, and other request-scoped concerns.

Barakmor1, line 127: Hey, I was wondering about the error handling approach for this API. Do we expect the interface implementation to handle retries and only propagate an error after retrying for a reasonable amount of time?

More generally, do we expect retries here, or should failures be handled in another way? For example, should the implementation perform some recovery logic, or should it return the error back to the caller and let the caller decide how to proceed?

My assumption is that retry handling probably shouldn't be done on our side, since we don't really know how long it may take for external managers (such as HCO) to process the request.

Maybe it would be useful to document the expected behavior somewhere, so we have a clear contract around error handling and responsibilities. If there is a pattern we want to enforce, we could also consider adding a linter or another validation mechanism to help keep the behavior consistent.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+
+```go
+// ConfigManager combines all configuration management capabilities.
+type ConfigManager interface {
+    FeatureGateManager
+    KubeVirtConfigManager
+    MigrationConfigManager
+}
+```
+
```
Barakmor1, line 148: I noticed that this API exposes both overall configuration changes and specific configuration changes (for example, migration configuration updates and feature gate toggling). I think we should clarify which approach we want to take here:

1. Do we want this interface to allow general KubeVirt CR modifications?
2. Do we want to explicitly define and support only specific field modifications?

Since option 2 can be implemented using option 1, I don't see a strong reason to expose both approaches, unless I'm missing something. It would be good to decide on a single model to keep the API simpler.

Barakmor1 (commented):
Thanks @dasionov ,
I left some comments

veps/sig-compute/test-infra-managed-environments.md
Barakmor1: Alternatively, we could provide a test library that external parties can use to implement and validate their own interface implementations. However, I'm not sure whether this would be a better approach compared to providing an external binary.

0xFelix: +1 to the external binary

What do you think about providing stubs in the test framework that make gRPC calls to an interface that the external binary could implement?

That way every vendor would be free to provide their own implementation.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+The managed implementation must:
+1. Patch the HyperConverged CR.
+2. Poll the KubeVirt CR until the expected configuration change appears (confirming HCO has reconciled).
+3. Read the KubeVirt CR's `ResourceVersion` after HCO reconciliation.
+4. Run the existing propagation wait logic using that `ResourceVersion`.
+
+The total timeout for the managed path will be longer than the direct path to account for HCO reconciliation delay.
+
+### Field Mapping: KubeVirt CR to HyperConverged CR
```
0xFelix, line 230: The code responsible for this mapping should live outside KubeVirt.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+| KubeVirt Configuration Field | HCO Equivalent |
+|---|---|
+| `seccompConfiguration` | None |
+| `developerConfiguration.logVerbosity` | None |
+| `tlsConfiguration` | None |
+| `networkConfiguration` (binding plugins) | None |
+| `supportedGuestAgentVersions` (deprecated) | None |
+
+The HCO implementation must handle these unmappable fields gracefully. The recommended approach is to return an
+error if a test attempts to set them through the HCO path. Tests that require these fields may need to be skipped
```
0xFelix, line 267: Can we please Fail tests instead? Skips are easily missed.

veps/sig-compute/test-infra-managed-environments.md
```diff
@@ -0,0 +1,508 @@
...
+or annotated for managed environments.
+
+Additionally, some test files directly patch the KubeVirt CR bypassing all helper abstractions -- modifying fields
+like `spec.uninstallStrategy`, `spec.certificateRotateStrategy`, `spec.infra`, `spec.workloads`, and
+`metadata.annotations`. These patterns are out of scope for the initial interface design since they are confined
+to operator-specific tests that verify KubeVirt operator behavior itself.
+
+### Implementation Selection
+
+The active implementation is selected at compile time via Go build tags:
```
0xFelix, line 277: Or with an ENV var like suggested above. If ENV is unset fallback to in-tree implementation, otherwise spin up binary und talk gRPC to it.

0xFelix (commented):
Could we reduce the scope of this VEP to just specify the config manager abstraction, an in-tree implementation that preserves the current behavior, and a gRPC interface contract? This would allow vendors to implement the interface in a separate binary, which could easily be supplied via an environment variable during test runs. 

This approach would also completely avoid introducing dependencies between KV and HCO.

dasionov:
/hold
putting this on hold. 
the efforts to introduce such a platform isn't worth it at the moment.

PR #358, opened 2026-06-28
VEP 190: Sync design with alpha implementation

iholder101:
### VEP Metadata

### What this PR does

Syncs VEP 190 with the merged alpha implementation. Each commit addresses one area where the design diverged from code.

### DIFF
--- a/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
+++ b/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
@@ -197,7 +197,7 @@ The plugin's container will have visibility only to its plugin folder in an isol
 ### The Node Hook
 
 To support node-level customizations, hooks can also define operations triggered by the virt-handler pod.
-These are invoked via gRPC (or ttRPC, as used by NRI) calls from virt-handler to a plugin server
+These are invoked via gRPC calls from virt-handler to a plugin server
 (a DaemonSet running on each node).
 The communication is via a socket (its path is defined by the `Plugin` CR).
 The existence of a socket marks the readiness of the plugin - similarly to NRI plugins.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -350,36 +326,23 @@ kind: Plugin
...
+            }
           }
-        }
-      ]
-  # Advanced hook: Plugin sidecar
-  - type: Plugin
-    socketPath: "/var/run/kubevirt-plugin/my-cool-plugin/my-plugin-socket.sock"
+        ]
+  # Advanced hook: Sidecar
+  - sidecar:
```
Acedus, line 328: nit: I know it just shows all possible options but we explicitly allow only one of sidecar or CEL expression.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -393,24 +356,11 @@ spec:
...
       - PostVMStop
-
-  selector:
-    matchLabels:
-      use-plugin: my-plugin
-    matchFields:
-      - key: metadata.name
-        operator: Prefix
-        value: vm-
-    conditions: # CEL-based
```
Acedus, line 410: Looking at it now, I think a list of conditions may have been preferable since one may want to define several instead of one with `&&`.

Acedus (approved):
Thanks @iholder101, a nit and a thought.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: This is a bit confusing, I think that's valid, as the example shows two separate hooks in the list - one CEL and one sidecar. Each hook specifies exactly one of the two. The list itself can contain a mix of both types.

This would have been invalid:
```yaml
domainHooks:
- cel:
    expression: 'Domain{Title: "test"}'
  sidecar:
    socketPath: "/var/run/kubevirt-plugin/my-plugin/hook.sock"
```

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Yeah, good point. With many different conditions it will be cleaner that way. Although I wonder why it should be of AND semantics and not OR, or which one we should choose.

Let's revisit this in the next cycle.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
Acedus: Yeah that's fair.

iholder101:
@Acedus the CEL syntax was wrong, fixed it.

Acedus (approved):

iholder101:
@xpivarc @vladikr :pray:

vladikr:
yeah, perfect, thanks! 
/approve

Issue #359, opened 2026-06-29
VEP: Public Pod Rendering API for Standalone VM Execution (pkg/render)


vladikr:
**Primary contact (assignee)**:

/assign @vladikr

**Current Feature Stage**:

New

**Feature Gate**:

`RenderAPI` (lifecycle tracker only — always enabled, removed at GA)

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/18277
  * [ ] Docs PRs:

**Additional context**:

External tools like [kubevirt-vm-to-pod](https://github.com/vladikr/kubevirt-vm-to-pod) currently import 8 internal KubeVirt packages (including `testutils` in production code) to generate Pod specs from VM/VMI definitions. These packages have no stability guarantees and break across minor versions.

This VEP proposes a new public `kubevirt.io/kubevirt/pkg/render` package exposing `PodFromVM()` and `PodFromVMI()` functions that:
- Wrap the internal TemplateService pipeline behind a stable `Options` struct
- Require no running cluster, controllers, or API server — pure offline transformations
- Enable external tools to generate Pod YAML without fragile internal dependencies

This also benefits KubeVirt internally: **174+ test call sites** use the verbose `NewTemplateService` (14 positional args) + `NewFakeClusterConfigUsingKV` setup ceremony, which these functions can replace with a one-liner.

> [!IMPORTANT]
> Please keep this description up to date.

PR #360, opened 2026-06-29
VEP-359: Public Pod Rendering API for Standalone VM Execution

vladikr:
### VEP Metadata

  **Tracking issue**: https://github.com/kubevirt/enhancements/issues/359
  **SIG label**: /sig compute

  ### What this PR does

  Introduces VEP #359: a new public `kubevirt.io/kubevirt/pkg/render` package exposing `PodFromVM()` and `PodFromVMI()` functions for rendering Pod specs from VirtualMachine/VMI definitions without requiring a
  running cluster or controllers.

  This addresses a real pain point: external tools (e.g. kubevirt-vm-to-pod) currently import 8 internal KubeVirt packages — including `testutils` in production — to generate Pod specs. These have no stability
  guarantees and break across minor versions.

  The package also benefits KubeVirt internally: 174+ test call sites currently go through a verbose `NewTemplateService` (14 positional args) + `NewFakeClusterConfigUsingKV` setup that `PodFromVM`/`PodFromVMI`
  can replace with a one-liner.

  - The package is purely additive — no changes to existing KubeVirt code
  - Uses a `RenderAPI` feature gate as a lifecycle tracker only (always enabled, removed at GA) (NOT 100% sure if this is needed)
  - Implementation PR: [TBD]

### DIFF
--- a/veps/sig-compute/359-public-pod-render-api/vep.md
+++ b/veps/sig-compute/359-public-pod-render-api/vep.md
@@ -1,0 +1,471 @@
+# VEP #359: Public Pod Rendering API for Standalone VM Execution
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created: https://github.com/kubevirt/enhancements/issues/359
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt's internal `TemplateService.RenderLaunchManifest` is the only
+mechanism for generating a Pod spec from a VirtualMachineInstance, but it lives
+in an internal package (`pkg/virt-controller/services`) and has never been
+exposed as a stable, public API. External tools that need to produce Pod
+manifests from VM definitions must vendor KubeVirt's internal packages, creating
+fragile dependencies that break on minor version upgrades.
+
+This VEP proposes a new public rendering API, `kubevirt.io/render`, exposing
+stable `PodFromVM()` and `PodFromVMI()` functions. These wrap the existing
+rendering pipeline -- defaults, mutations, network setup, and template rendering
+-- behind a clean interface that does not leak controller internals to consumers.
+
+The package follows the same model as `kubevirt.io/api` and
+`kubevirt.io/client-go`: it lives under `staging/src/kubevirt.io/render/` as a
+standalone Go module with a minimal dependency footprint for external consumers.
+
+## Motivation
+
+Running KubeVirt VMs outside of Kubernetes clusters is an emerging use case.
+Edge deployments, developer workstations, and CI/CD pipelines all benefit from
+converting a VirtualMachine definition into a standalone Pod manifest that can be
+executed by Podman or another OCI-compatible runtime.
+
+The [kubevirt-vm-to-pod](https://github.com/vladikr/kubevirt-vm-to-pod) tool
+addresses this use case today. However, to produce a correct Pod spec it must
+import **eight internal KubeVirt packages**:
+
+| Internal package | Purpose |
+|---|---|
+| `pkg/testutils` | Creating a fake `ClusterConfig` (no public constructor exists) |
+| `pkg/virt-controller/services` | `TemplateService` and `RenderLaunchManifest` |
+| `pkg/defaults` | `SetVirtualMachineDefaults`, `SetDefaultVirtualMachineInstance` |
+| `pkg/virt-api/webhooks/mutating-webhook/mutators` | `ApplyNewVMIMutations` |
+| `pkg/network/vmispec` | `SetDefaultNetworkInterface` |
+| `pkg/util` | `SetDefaultVolumeDisk` |
+| `pkg/virt-controller/watch/vm` | `SetupVMIFromVM`, `AutoAttachInputDevice` |
+| `pkg/virt-config` | `ClusterConfig` type |
+
+None of these are public API. Any can change signature, move, or disappear
+between minor releases. The tool is forced to use `testutils` -- a package
+explicitly intended for test code -- in production, because there is no other way
+to construct a valid `ClusterConfig` without a running Kubernetes cluster.
+
+Every KubeVirt version bump has the potential to break external consumers. A
+stable, public rendering API would eliminate this fragility and enable a healthy
+ecosystem of tools that build on KubeVirt's Pod rendering logic.
+
+### Internal benefit to KubeVirt
+
+Beyond external consumers, the render module can serve as the foundation for
+simplifying KubeVirt's own VMI→Pod rendering path in `virt-controller`. Today,
+the controller assembles a `ClusterConfig` via informers, creates a
+`TemplateService` with 14 positional arguments, applies defaults and mutations
+in a specific order, then calls `RenderLaunchManifest`. The render module
+encapsulates this entire pipeline behind a single function call with an explicit
+`Options` struct. Migrating `virt-controller` to use the render module is a
+larger refactor that would follow as a separate effort after the module is
+established.
+
+Additionally, this work lays groundwork for the gradual extraction of rendering
+logic from `virt-controller/services` into `kubevirt.io/render`, which will also
+benefit future hypervisor abstraction efforts
+([VEP-97](https://github.com/kubevirt/enhancements/pull/97)).
+
+## Goals
+
+- Expose a public Go module (`kubevirt.io/render`) for rendering a Pod spec
+  from a VirtualMachine or VirtualMachineInstance definition.
+- Include the full defaults and mutations pipeline (VM defaults, VMI defaults,
+  VMI mutations, network defaults, volume-disk pairing) so that callers receive
+  a complete, ready-to-use Pod spec without reimplementing any of these steps.
+- Accept configuration through an explicit `Options` struct instead of requiring
+  callers to construct a `ClusterConfig` or interact with informer caches.
+- Work fully offline — no running Kubernetes cluster, no KubeVirt controllers,
+  no API server connection required. The functions must be pure transformations
+  from VM/VMI spec + options → Pod spec.
+- Provide a foundation for simplifying virt-controller's VMI→Pod rendering
+  path in a future refactor.
+- Version the render module in lockstep with KubeVirt releases — the rendered
+  Pod spec is specific to the KubeVirt version it was built from.
+
+## Non Goals
+
+- **Replacing the internal `TemplateService`**. The internal implementation
+  continues to exist and evolve for use by `virt-controller`. The public API
+  wraps it; it does not replace it.
+- **Supporting live-cluster features** such as migration, hotplug, or live
+  snapshot in the public API. These require controller coordination and are out
+  of scope for offline rendering.
+- **Exposing `ClusterConfig` internals**. The public API accepts feature gates
+  and configuration values as plain Go types. Consumers never import
+  `pkg/virt-config`.
+- **Cross-version compatibility**. The rendered Pod spec is specific to the
+  KubeVirt version the module was built from. A Pod rendered with one version
+  is not expected to work with a different version's virt-launcher.
+- **Providing a CLI tool**. A `virtctl render` subcommand could be built on top
+  of this package in the future, but is not part of this VEP.
+
+## Definition of Users
+
+- **External tool developers** building utilities on top of KubeVirt's VM
+  model (e.g., kubevirt-vm-to-pod, custom GitOps pipelines, templating engines).
+- **Edge and standalone VM operators** who generate Pod YAML for execution
+  outside Kubernetes via Podman or similar container runtimes.
+- **CI/CD systems** that need to validate or generate VM Pod specs offline
+  without deploying the full KubeVirt control plane.
+
+## User Stories
+
+- As an external tool developer, I want a stable Go API to generate a Pod spec
+  from a VirtualMachine definition so that my tool does not break on every
+  KubeVirt minor release.
+- As an edge operator, I want to convert VM YAML into a standalone Pod manifest
+  offline, without depending on KubeVirt's internal packages or test utilities.
+- As a CI engineer, I want to render and validate VM Pod specs in a pipeline
+  without running the KubeVirt control plane, so I can catch misconfiguration
+  before deployment.
+- As a KubeVirt contributor, I want external consumers to use a supported API
+  surface so that refactoring internal packages does not generate downstream
+  breakage reports.
+- As a KubeVirt developer, I want virt-controller's VMI→Pod rendering path to
+  use a well-defined internal API so the pipeline is easier to understand,
+  test, and refactor.
+
+## Repos
+
+- kubevirt/kubevirt
+
+## Design
+
+### Package Location
+
+The package will be placed under:
+
+**`staging/src/kubevirt.io/render`** (import path: `kubevirt.io/render`)
+
+This follows the same model as `kubevirt.io/api` and `kubevirt.io/client-go`,
+making it a stable, independently consumable module with a minimal dependency
+footprint for external consumers.
+
+### Core functions
+
+```go
+// PodFromVM renders a Pod spec from a VirtualMachine definition.
+// It applies VM defaults, extracts the VMI template, applies VMI defaults
+// and mutations, configures networking, and calls RenderLaunchManifest.
+func PodFromVM(vm *virtv1.VirtualMachine, opts Options) (*corev1.Pod, error)
+
+// PodFromVMI renders a Pod spec from a VirtualMachineInstance definition.
+// It applies VMI defaults and mutations, configures networking, and calls
+// RenderLaunchManifest.
+func PodFromVMI(vmi *virtv1.VirtualMachineInstance, opts Options) (*corev1.Pod, error)
+
+// VMIFromVM extracts a VirtualMachineInstance from a VirtualMachine definition,
+// applying VM defaults. Useful for tools that need the VMI without full Pod
+// rendering.
+func VMIFromVM(vm *virtv1.VirtualMachine, opts Options) (*virtv1.VirtualMachineInstance, error)
+```
+
+All functions are stateless, pure transformations, and safe for concurrent use.
+They do not require a running Kubernetes cluster, KubeVirt controllers, or any
+external state — all configuration is passed via the `Options` struct. This
+makes them suitable for offline tooling, CI pipelines, and KubeVirt's own tests.
+
+### Options struct
+
+```go
+type Options struct {
+    // LauncherImage is the virt-launcher container image reference.
+    // Required.
+    LauncherImage string
+
+    // FeatureGates is the list of KubeVirt feature gates to enable.
+    // Matches KubeVirtConfiguration.DeveloperConfiguration.FeatureGates.
+    // Example: []string{"ImageVolume", "HostDisk"}
+    FeatureGates []string
+
+    // DisabledFeatureGates is the list of feature gates to explicitly disable.
+    // Useful for disabling Beta feature gates that are on by default.
+    // Matches KubeVirtConfiguration.DeveloperConfiguration.DisabledFeatureGates.
+    DisabledFeatureGates []string
+
+    // ExportImage is the vm-export sidecar image. Optional; defaults to
+    // the standard KubeVirt export image if empty.
+    ExportImage string
+
+    // RunAsUser is the UID for the virt-launcher process.
+    // Default: 107 (the standard qemu user).
+    RunAsUser int64
+}
+```
+
+New fields can be added to `Options` without breaking existing callers, since Go
+struct literals with named fields are forward-compatible with new fields that
+have zero-value defaults.
+
+### Architecture and Long-term Vision
+
+The package introduces two interfaces to enable proper decoupling:
+
+- **`RenderConfig`** — a minimal interface capturing only the configuration
+  methods needed by the rendering pipeline.
+- **`ManifestRenderer`** — an interface responsible for producing the final Pod
+  manifest from a `VirtualMachineInstance`.
+
+These interfaces allow the render module to avoid direct dependencies on
+internal KubeVirt components where possible. `RenderConfig` captures the
+configuration values needed by the rendering pipeline (feature gates, machine
+types, resource defaults, network settings). `ManifestRenderer` is responsible
+for producing the final Pod manifest.
+
+The existing `*virtconfig.ClusterConfig` and `*services.TemplateService`
+satisfy these interfaces without adapters, so virt-controller's existing code
+path continues to work unchanged. For offline use, the render module provides
+its own lightweight `RenderConfig` implementation built from `Options` values
+— no informers, goroutines, or Kubernetes client required. The specific
+interface methods are defined in the implementation and may evolve as the
+module matures.
+
+#### Current State (Short-term)
+
+For the initial implementation, the default `ManifestRenderer` still relies on
+`TemplateService` internally to generate correct Pod specs for offline use. This
+results in a limited internal dependency on `pkg/virt-controller/services`
+within the default implementation only. We treat this as **temporary technical
+debt**.
+
+The public API remains simple:
+
+```go
+pod, err := render.PodFromVMI(vmi, render.Options{...})
+```
+
+#### Long-term Direction
+
+The long-term goal is to move the core rendering logic into `kubevirt.io/render`
+itself. Over time we plan to:
+
+- Extract shared, hypervisor-agnostic rendering logic (defaults, mutations,
+  network and volume handling, manifest construction) into the render module.
+- Make `virt-controller` a consumer of `kubevirt.io/render` rather than the
+  owner of the rendering implementation.
+- Use the `ManifestRenderer` interface as the primary extension point.
+
+This direction aligns well with
+[VEP-97](https://github.com/kubevirt/enhancements/pull/97) (alternative
+hypervisors). As KubeVirt
+evolves to support multiple hypervisors, having rendering logic live in a
+dedicated, reusable module with a clear interface will make it easier to
+introduce hypervisor-specific renderers without duplicating shared logic.
+
+The ultimate vision is that `kubevirt.io/render` becomes the canonical home for
+Pod rendering logic in KubeVirt.
+
+### Internal rendering pipeline
+
+Internally, `PodFromVM` and `PodFromVMI` execute the same sequence of VM/VMI
+defaults, mutations, network configuration, and manifest rendering that
+KubeVirt's admission webhooks and virt-controller perform today. The render
+module centralizes this pipeline behind the public API so that:
+
+- The pipeline stays correct as internal implementations evolve.
+- External consumers automatically pick up fixes and improvements.
+- The internal packages can be refactored freely without breaking the public
+  contract.
+
+### Version compatibility
+
+The render module version is tied 1:1 to the KubeVirt version. A rendered Pod
+spec is specific to the KubeVirt release it was built from — container images,
+volume layouts, security contexts, and feature gate behavior all evolve with
+each release. Consumers should import the `kubevirt.io/render` version that
+matches their target `virt-launcher` image.
+
+Cross-version compatibility (e.g., rendering with v1.11 but running with a
+v1.10 virt-launcher) is not a goal and not guaranteed.
+
+The Go API surface (`PodFromVM`, `PodFromVMI`, `Options`) follows standard Go
+module conventions: new `Options` fields are additive with zero-value defaults,
+so code compiled against an older version continues to compile against newer
+versions.
+
+## API Examples
+
+### Rendering a Pod from a VirtualMachine
+
+```go
+package main
+
+import (
+    "fmt"
+    "os"
+
+    "sigs.k8s.io/yaml"
+    virtv1 "kubevirt.io/api/core/v1"
+    "kubevirt.io/render"
+)
+
+func main() {
+    vm := &virtv1.VirtualMachine{
+        // ... loaded from YAML or constructed programmatically
+    }
+
+    pod, err := render.PodFromVM(vm, render.Options{
+        LauncherImage: "quay.io/kubevirt/virt-launcher:v1.10.0",
+        FeatureGates:  []string{"ImageVolume", "HostDisk"},
+    })
+    if err != nil {
+        fmt.Fprintf(os.Stderr, "error: %v\n", err)
+        os.Exit(1)
+    }
+
+    out, _ := yaml.Marshal(pod)
+    fmt.Println(string(out))
+}
+```
+
+### Rendering a Pod from a VirtualMachineInstance
+
+```go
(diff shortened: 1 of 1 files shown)


Barakmor1 (commented):
Hey Vladik, thanks for working on this enhancement!

I think this could be a great improvement, especially since we already have a real use case for it, such as the VM-to-Pod tool.

One additional thing to consider is version and configuration compatibility. The KubeVirt version deployed in a cluster (or used as part of a workload deployment) may differ from the version of the imported package, which could lead to compatibility issues. We should also consider configuration mismatches, for example, if the cluster has a different set of feature gates or other configuration options enabled than what the importer code expects. It might be worth thinking about how we can validate or guard against these scenarios.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,384 @@
...
+  a complete, ready-to-use Pod spec without reimplementing any of these steps.
+- Accept configuration through an explicit `Options` struct instead of requiring
+  callers to construct a `ClusterConfig` or interact with informer caches.
+- Work fully offline — no running Kubernetes cluster, no KubeVirt controllers,
+  no API server connection required. The functions must be pure transformations
+  from VM/VMI spec + options → Pod spec.
+- Provide a foundation for simplifying virt-controller's VMI→Pod rendering
+  path in a future refactor.
+- Maintain backward compatibility of the public API within a major version,
+  following the same compatibility guarantees as `kubevirt.io/api`.
```
Barakmor1: I think that, in order to make this logic available to external tools, it would be very beneficial to reduce the coupling with other packages as much as possible. For example, we should consider ways to decouple this package from client-go and other dependencies where possible.

Otherwise, we may force external tools to use the exact same dependency versions, which can become a significant limitation.

vladikr: yes, I agree, but that's an evolution. This PR is a start, and then we will go from here. Once this is in and virt-controller is using it, we will have a larger cleanup. @fossedihelm already pointed to some client-go decoupling-related PR. 
But I see this work as a start.

Barakmor1: LGTM for the general purpose of this VEP. However, I still think it’s worth mentioning that one of the goals is to reduce the dependency tree for the renderer package.

iholder101:
> One additional thing to consider is version and configuration compatibility. The KubeVirt version deployed in a cluster (or used as part of a workload deployment) may differ from the version of the imported package, which could lead to compatibility issues. We should also consider configuration mismatches, for example, if the cluster has a different set of feature gates or other configuration options enabled than what the importer code expects. It might be worth thinking about how we can validate or guard against these scenarios.

@Barakmor1 I agree.

I think we can follow the pattern already done with `kubevirt.io/api` and `kubevirt.io/client-go` modules (as well as Kubernetes and client-go). I'm not sure how much we should invest in validation vs. placing the responsibility on the user's side to import the right versions though, as done with the other packages I've mentioned.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,384 @@
...
+## Motivation
+
+Running KubeVirt VMs outside of Kubernetes clusters is an emerging use case.
+Edge deployments, developer workstations, and CI/CD pipelines all benefit from
+converting a VirtualMachine definition into a standalone Pod manifest that can be
+executed by Podman or another OCI-compatible runtime.
+
+The [kubevirt-vm-to-pod](https://github.com/vladikr/kubevirt-vm-to-pod) tool
+addresses this use case today. However, to produce a correct Pod spec it must
+import **eight internal KubeVirt packages**:
```
iholder101, line 47: The number of internal packages is secondary IMO - the real cost is that any import from `kubevirt.io/kubevirt` pulls in the entire dependency graph (client-go, libvirt, informers, etc.) even when you only need the rendering logic.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,384 @@
...
+
+- kubevirt/kubevirt
+
+## Design
+
+### Package location
+
+```
+kubevirt.io/kubevirt/pkg/render
+```
```
iholder101: Following my [comment here](https://github.com/kubevirt/enhancements/pull/360#issuecomment-4965922057), I think we should be placing this under `staging/src/kubevirt.io/render/` as a separate Go module, following the existing pattern for `kubevirt.io/api` and `kubevirt.io/client-go`.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,384 @@
...
+
+    // ExportImage is the vm-export sidecar image. Optional; defaults to
+    // the standard KubeVirt export image if empty.
+    ExportImage string
+
+    // RunAsUser is the UID for the virt-launcher process.
+    // Default: 107 (the standard qemu user).
+    RunAsUser int64
+}
+```
```
iholder101, line 207: `TemplateService` takes 14 args (caches, client, directory paths, timeouts, etc.). For offline use these can be stubbed/defaulted, but the VEP also lists virt-controller adoption as a goal - and virt-controller needs real values for all of them. How does `Options` serve both use cases without either growing to expose everything or blocking internal adoption?

veps/sig-compute/359-public-pod-render-api/vep.md
fossedihelm: I agree with this, but ATM it looks like the render has strict dependencies on other kv packages.
They basically are:
- pkg/defaults
- pkg/virt-api/webhooks/mutating-webhook/mutators
- pkg/network/vmispec
- pkg/util
- pkg/virt-controller/services (TemplateService)
- pkg/virt-config

I investigated a possible solution, and I was able to move almost everything in the self-contained `kubevirt.io/render`.

The main big troubles are with `pkg/virt-controller/services` and `pkg/virt-config`.
For the former, it looks like a simple function invocation, but from there, it touches sooooo many files.
For the latter, it implies a new separate external module IMO `apimachinery` maybe?

fossedihelm: For reference: this is my experiment https://github.com/fossedihelm/kubevirt/commits/feature/render-api-ff/

vladikr: Thank you very much @fossedihelm !
+1 
We had an offline discussion a few days ago. In general, I think everyone agrees that kubevirt.io/render is the best course. The question is how we get there. 

The main blocker is decoupling the template service from the clusterConfig. I'm also not sure how to do this refactoring now, but if we do nothing, we won't progress.

I can restructure the VEP. Keep pkg/render for Alpha, then decouple from TemplateService from ClusterConfig and for GA move to staging/src/kubevirt.io/render/ as standalone module.

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: Well, the `Options` will only cover the public/offline use case and will stay deliberately small.

Regarding the  `virt-controller`, we don't need to force it through the minimal `Options`. 
The render package will keep internal constructors and helpers that still let you pass real clients, caches, and directory paths, so it will provide the same level of control as `TemplateService` has today.

the point is that the internal code can stay as powerful as it needs to be. And the migration to `pkg/render` can happen gradually without blocking anything.

Does this make sense?

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: @fossedihelm I think to unblock this template serice and virt-config coupling we can just separate it to interfaces
I'm thinking that we can get away with something like
```
type RenderConfig interface {
    IsFeatureGateEnabled(gate string) bool
    GetDefaultNetworkInterface() string
    IsBridgeInterfaceOnPodNetworkEnabled() bool
...
}
``` 
and
```
type ManifestRenderer interface {
    RenderLaunchManifest(vmi *v1.VirtualMachineInstance) (*corev1.Pod, error)
}
```
so then kubevirt.io/render would only depend on kubevirt.io/api and these two interfaces
What do you think?

fossedihelm: @vladikr The two interfaces will drop most of the coupling. But I think there is an issue: 
The `renderPod` function needs to initialize the struct and not the interface, no? :)

vladikr: @fossedihelm , I took a stab at this with Claude because I was also very confused. 
My point was that renderPod itself would only receives the interfaces, so it never constructs anything. 

In [PodFromVM](https://github.com/kubevirt/kubevirt/pull/18277/changes#diff-0f4e2af2ebf3a82b1841788f7aad091a1e46fdee85a28a07c9747eb514dbd83fR84)
the `newOfflineRenderConfig` creates the struct, `newOfflineRenderer` creates the templateSerice, and `renderPod` uses the `RenderConfig` and `ManifestRenderer`                                                                                                   

so renderPod passes the `RenderConfig `to defaults/mutators and calls renderer.RenderLaunchManifest(). It never sees a concrete type.                                                     
                                                                                                                                                                                                                   
The struct initialization is in `PodFromVM` and `PodFromVMI`, where we build an offlineRenderConfig for the `RenderConfig` and a TemplateService. 
For the offline case these are lightweight without informers and goroutines. 
For virt-controller, *ClusterConfig satisfies the same interfaces, so the existing code path is unchanged.                                                     
                                                                                                                                                                                                                   
I've upidated kubevirt/kubevirt#18277 if you want to take a look!

vladikr: The main issue is the template service ...
we will need to continue extracting... it's a lot of work, but will give us a clean separation and I think it would be better for VEP97 as well.

fossedihelm: Maybe I don't get it but I am referring to `services.NewTemplateService` https://github.com/kubevirt/kubevirt/pull/18277/changes#diff-0f4e2af2ebf3a82b1841788f7aad091a1e46fdee85a28a07c9747eb514dbd83fR160. This is not something we can avoid ATM.
I think this is something not doable right now. The main big gap is caused by a missing apimachinery module.
Most of the "utils" stuff can be there, of course, API compatibility is not guaranteed.
I started something in https://github.com/kubevirt/kubevirt/pull/18472

fossedihelm: I think my message was confused. Long story short: The template service is the main component for pod rendering. This is coupled to  tons of modules. In order to make render exportable, The templateService must live externally (otherwise duplicated code)

vladikr: yea, 100%

vladikr: so, I think we can continue your work and eventually extract the generic parts into render
I worte it here: https://github.com/kubevirt/enhancements/pull/360/changes#diff-1f186e37edcd341bcf4574c0368f3423a3b03fb98617ef7a0304fb8601698a50R488

vladikr: hmm, it doesn't show, but I wrote:
```
Core rendering logic moved into `kubevirt.io/render`, making it the
  canonical home for Pod rendering. Short-term: a limited dependency on
  `TemplateService` is accepted as technical debt in the default
  `ManifestRenderer` implementation. Long-term: the rendering logic is
  fully owned by `kubevirt.io/render`, which also positions it for future
  hypervisor abstraction work
  ```
  But, that's for GA, I'm pretty convinced that we can make progress with everything else for now.

File: veps/sig-compute/359-public-pod-render-api/vep.md [hunk 5]
```diff
@@ -0,0 +1,498 @@
...
+  VMI mutations, network defaults, volume-disk pairing) so that callers receive
+  a complete, ready-to-use Pod spec without reimplementing any of these steps.
+- Accept configuration through an explicit `Options` struct instead of requiring
+  callers to construct a `ClusterConfig` or interact with informer caches.
+- Work fully offline — no running Kubernetes cluster, no KubeVirt controllers,
+  no API server connection required. The functions must be pure transformations
+  from VM/VMI spec + options → Pod spec.
+- Provide a foundation for simplifying virt-controller's VMI→Pod rendering
+  path in a future refactor.
+- Maintain backward compatibility of the public API within a major version,
```
fossedihelm: Thinking about this: Do we really want to enforce this compatibility? I mean, I'd expect that every release is targeting a specific k8s/kv, and the generated pod is specific for that kv version. I do not expect that a newer virt-launcher pod will work perfectly with older kv

vladikr: +1 I agree

vladikr:
@Barakmor1 what do you think about this direction?

iholder101:
Thank you @vladikr! The new version of this PR looks much better! :)
I like the path forward!

/lgtm

vladikr:
@xpivarc can you please take a look? Looking for approval.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,504 @@
...
+mechanism for generating a Pod spec from a VirtualMachineInstance, but it lives
+in an internal package (`pkg/virt-controller/services`) and has never been
+exposed as a stable, public API. External tools that need to produce Pod
+manifests from VM definitions must vendor KubeVirt's internal packages, creating
+fragile dependencies that break on minor version upgrades.
+
+This VEP proposes a new public rendering API, `kubevirt.io/render`, exposing
+stable `PodFromVM()` and `PodFromVMI()` functions. These wrap the existing
+rendering pipeline -- defaults, mutations, network setup, and template rendering
+-- behind a clean interface that does not leak controller internals to consumers.
```
lyarwood, line 32: @vladikr appreciate the effort is entirely aligned around pod rendering at the moment but would you be interested in exposing the VMI from VM render code from this package as well? I'm assuming you're already calling it internally for the `PodFromVM` call anyway right?

lyarwood (approved):
Quick question about covering the VMI from VM use case as well but otherwise lgtm!

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,504 @@
...
+
+The package follows the same model as `kubevirt.io/api` and
+`kubevirt.io/client-go`: it lives under `staging/src/kubevirt.io/render/` as a
```
Barakmor1, line 35: +1

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,504 @@
...
+
+```go
+type Options struct {
+    // LauncherImage is the virt-launcher container image reference.
+    // Required.
+    LauncherImage string
+
+    // FeatureGates is the list of KubeVirt feature gates to enable.
+    // Example: []string{"ImageVolume", "HostDisk"}
+    FeatureGates []string
```
Barakmor1, line 192: I wonder if this should be a `map[string]bool`, since Beta feature gates are enabled by default. Alternatively, we could include a `FeatureGatesToDisable []string` field as well, to match our API.

WDYT?

veps/sig-compute/359-public-pod-render-api/vep.md
Barakmor1: It makes sense. I was just hoping we could make the internal constructor use the newly introduced public constructor, maybe through a inner struct for generating the pod manifest, so we can avoid code duplication.

Barakmor1 (commented):
Thank you @vladikr 
The VEP looks good in general.

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: Sorry, I was focused on other work and missed your comment... 
yeah.. It makes sense to me. we already have setupVMIFromVM in render.go (in my draft PR :) ) so I guess something like `VMIFromVM(vm) *VMI` would be a nice addition

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: yeah, exactly, and we will get there. 
virt-controller would become a consumer of the render module and wont own taht duplicate logic. I mean that's exactly the plan for GA 
But we can't do it now

vladikr:
@Barakmor1 @fossedihelm @lyarwood @xpivarc  updated.

veps/sig-compute/359-public-pod-render-api/vep.md
lyarwood: Awesome, yeah `VMIFromVM` would be great as a starting point thanks. We could even break these up into target specific implementations in the future if we end up with more we want to expose here, like `podRender.FromVM` or even `domainRender.FromVMI` etc.

lyarwood (approved):

vladikr:
> Pull requests that are marked with `lgtm` should receive a review from an approver within 1 week.
> 
> After that period the bot marks them with the label `needs-approver-review`.
> 
> /label needs-approver-review

@xpivarc can you please take a look?

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+explicitly intended for test code -- in production, because there is no other way
+to construct a valid `ClusterConfig` without a running Kubernetes cluster.
+
+Every KubeVirt version bump has the potential to break external consumers. A
+stable, public rendering API would eliminate this fragility and enable a healthy
+ecosystem of tools that build on KubeVirt's Pod rendering logic.
+
+### Internal benefit to KubeVirt
+
+Beyond external consumers, `pkg/render` can serve as the foundation for
```
EdDev: Do you mean `kubevirt.io/render` ?

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+`Options` struct. Migrating `virt-controller` to use `pkg/render` is a larger
+refactor that would follow as a separate effort after the package is established.
+
+Additionally, this work lays groundwork for the gradual extraction of rendering
+logic from `virt-controller/services` into `kubevirt.io/render`, which will also
+benefit future hypervisor abstraction efforts (VEP-97).
+
+## Goals
+
+- Expose a public Go package (`kubevirt.io/kubevirt/pkg/render`) for rendering a
```
EdDev: I think you later change it to exposing it through `kubevirt.io/render` repo, by handling it like the other staging go independent modules.


In general, I would prefer to see in all the intro/overview sections a general abstracted need for the public GO API without expressing how, then place this solution in the Design section.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+
+    // RunAsUser is the UID for the virt-launcher process.
+    // Default: 107 (the standard qemu user).
+    RunAsUser int64
+}
+```
+
+New fields can be added to `Options` without breaking existing callers, since Go
+struct literals with named fields are forward-compatible with new fields that
+have zero-value defaults.
```
EdDev, line 216: I think it depends on how the client code was written [1]. I would claim this is unsafe unless harden.
To "harden", I think you can pass a list of builder functions (see `libvmi`), so the client code will not need to use the struct directly (it should not be exposed).

Another option is to generalize it into a json data, with a defined formatting, versioning, etc.

[1] Client dependent code that will need recompilation if the struct changes:
`s := MyStruct{"val1", "val2"}`

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+
+`Options` serves the public/offline use case only. For internal adoption by
+`virt-controller`, the rendering pipeline uses the `RenderConfig` and
+`ManifestRenderer` interfaces directly, providing full control over
+configuration and caching without being constrained by the simplified `Options`
+surface.
+
+### Architecture and Long-term Vision
+
+The package introduces two interfaces to enable proper decoupling:
```
EdDev, line 220: This is referring to the package wiring to the internal kubevirt packages?

I'm not sure if the interface alone can decouple this package from the backend implementation.
Specifically, there is a need for a dependency inversion.

The needed interface in Go is better set per the needed usage and not grouped per the concrete object that implements it.
In order not to get into too much details here, the general flow/format/skeleton can be given and then specifics can be left to the implementation (otherwise we will need to get back to this document and update it all the time).

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+implementation built from `Options` values, using `featuregate.IsEnabled`
+directly — no informers, goroutines, or Kubernetes client required.
+
+#### Current State (Short-term)
+
+For the initial implementation, the default `ManifestRenderer` still relies on
+`TemplateService` internally to generate correct Pod specs for offline use. This
+results in a limited internal dependency on `pkg/virt-controller/services`
+within the default implementation only. We treat this as **temporary technical
+debt**.
```
EdDev, line 247: Dependency inversion is simple enough to implement initially (we have existing examples already in place).
I am not exactly clear how this will solve the cascade of dependencies this initiative attempted to avoid in the first place. Said that, if this is indeed sufficient for your needs, fine by me.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+The long-term goal is to move the core rendering logic into `kubevirt.io/render`
+itself. Over time we plan to:
+
+- Extract shared, hypervisor-agnostic rendering logic (defaults, mutations,
+  network and volume handling, manifest construction) into the render module.
+- Make `virt-controller` a consumer of `kubevirt.io/render` rather than the
+  owner of the rendering implementation.
+- Use the `ManifestRenderer` interface as the primary extension point.
+
+This direction aligns well with VEP-97 (alternative hypervisors). As KubeVirt
```
EdDev: nit: Please add a link.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+evolves to support multiple hypervisors, having rendering logic live in a
+dedicated, reusable module with a clear interface will make it easier to
+introduce hypervisor-specific renderers without duplicating shared logic.
+
+The ultimate vision is that `kubevirt.io/render` becomes the canonical home for
+Pod rendering logic in KubeVirt.
+
+### Internal wiring
```
EdDev: IMO this section goes into too low level implementation details. I would avoid it in order to keep this VEP stable in low details during the implementation changes.

It should be enough to call each step per the functional operation, in an abstract manner.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+### Version compatibility
+
+The render module version is tied 1:1 to the KubeVirt version. A rendered Pod
+spec is specific to the KubeVirt release it was built from — container images,
+volume layouts, security contexts, and feature gate behavior all evolve with
+each release. Consumers should import the `kubevirt.io/render` version that
+matches their target `virt-launcher` image.
+
+Cross-version compatibility (e.g., rendering with v1.11 but running with a
+v1.10 virt-launcher) is not a goal and not guaranteed.
```
EdDev, line 297: Worth adding to the non-goal section.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+
+The Go API surface (`PodFromVM`, `PodFromVMI`, `Options`) follows standard Go
+module conventions: new `Options` fields are additive with zero-value defaults,
+so code compiled against an older version continues to compile against newer
+versions.
```
EdDev, line 302: This is repeated from a previous location. I had some comments about the `Options` struct.
Maybe worth not duplicating/repeating this one and just ref between them.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,515 @@
...
+  migrates from internal imports to `pkg/render`, serving as a real-world
+  validation that the public API covers the necessary use cases.
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `RenderAPI` registered (always enabled, used as lifecycle tracker)
```
EdDev: I'm unclear why a FG is needed. This is not a runtime feature.
It is closer to a code refactoring.

The new package can be commented/documented as experiments/alpha inline.

In case there is a strong reason, build time flags can be used to avoid the new code from being compiled. But I do not think it is needed.

EdDev (changes requested):
Thank you, I like it!

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: I agree. Someone asked for this. I also feel that there is no need for a feature gate.

vladikr:
@EdDev I think all is addressed. can you please take another look?

EdDev (approved):
Thank you!

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+Additionally, this work lays groundwork for the gradual extraction of rendering
+logic from `virt-controller/services` into `kubevirt.io/render`, which will also
+benefit future hypervisor abstraction efforts
+([VEP-97](https://github.com/kubevirt/enhancements/pull/97)).
+
+## Goals
+
+- Expose a public Go module (`kubevirt.io/render`) for rendering a Pod spec
+  from a VirtualMachine or VirtualMachineInstance definition.
+- Include the full defaults and mutations pipeline (VM defaults, VMI defaults,
```
xpivarc, line 90: Can you elaborate what you mean by "VM defaults, VMI defaults"?

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+  migrates from internal imports to `pkg/render`, serving as a real-world
+  validation that the public API covers the necessary use cases.
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] `PodFromVM`, `PodFromVMI`, and `VMIFromVM` functions implemented
```
xpivarc, line 440: I am all for the `PodFromVMI`, tho I see some problems here. The path is not straightforwad now, for example for Container path volume, there is additional admiter. How do we reconcile this into single package? For the `PodFromVM` that is not required if you have `VMIFromVM` imho, for `VMIFromVM` I see issue as that also include multiple components on the data pah.

xpivarc (commented):
/approve
I like this but I am concerned about the cross component interaction that now needs to collapse to one pkg

xpivarc:
/hold
Allow to close the discussion

veps/sig-compute/359-public-pod-render-api/vep.md
vladikr: Well, VM defaults are `defaults.SetVirtualMachineDefaults` (fjust ills missing fields on the VM object) and VMI defaults  are `defaults.SetDefaultVirtualMachineInstance`,  `vmispec.SetDefaultNetworkInterface` , and `util.SetDefaultVolumeDisk` too
These are the same functions the webhook/controller pipeline calls.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+benefit future hypervisor abstraction efforts
+([VEP-97](https://github.com/kubevirt/enhancements/pull/97)).
+
+## Goals
+
+- Expose a public Go module (`kubevirt.io/render`) for rendering a Pod spec
+  from a VirtualMachine or VirtualMachineInstance definition.
+- Include the full defaults and mutations pipeline (VM defaults, VMI defaults,
+  VMI mutations, network defaults, volume-disk pairing) so that callers receive
+  a complete, ready-to-use Pod spec without reimplementing any of these steps.
```
asafbennatan, line 92: Is the promise the same Pod virt-controller would launch in-cluster, with anything needed to actually run offline left to the caller?

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+
+// PodFromVMI renders a Pod spec from a VirtualMachineInstance definition.
+// It applies VMI defaults and mutations, configures networking, and calls
+// RenderLaunchManifest.
+func PodFromVMI(vmi *virtv1.VirtualMachineInstance, opts Options) (*corev1.Pod, error)
+
+// VMIFromVM extracts a VirtualMachineInstance from a VirtualMachine definition,
+// applying VM defaults. Useful for tools that need the VMI without full Pod
+// rendering.
+func VMIFromVM(vm *virtv1.VirtualMachine, opts Options) (*virtv1.VirtualMachineInstance, error)
```
asafbennatan, line 178: If yes: standalone still needs the defaulted VMI that matches that Pod (we attach it to the compute container as `STANDALONE_VMI`, since there is no API watch). Is `VMIFromVM` the same defaults/mutations/network/volume-disk path `PodFromVM` uses internally, so we can `VMIFromVM` → post-process the VMI → `PodFromVMI`? Or does `PodFromVM` hide mutations we cannot recover from the Pod alone?

The comment currently says `VMIFromVM` only applies VM defaults, which looks weaker than `PodFromVM`.

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+- **Rollback**: Rolling back KubeVirt in a cluster has no effect on the render
+  module, since it is a compile-time dependency. Consumers pin to the version
+  that matches their target runtime.
+
+## Functional Testing Approach
+
+- **Unit tests**: Exercise `PodFromVM` and `PodFromVMI` with a representative
+  set of VM configurations:
+  - ContainerDisk volumes
+  - PersistentVolumeClaim volumes (stub PVC path)
```
asafbennatan, line 415: `RenderLaunchManifest` looks up every PVC. That is required to emit the Pod, not a runtime add-on. Will the library stub claims internally, and if so with which volume mode (Block vs Filesystem changes `volumeDevices` vs `volumeMounts`)? If not, how does a caller supply them without an informer cache?

veps/sig-compute/359-public-pod-render-api/vep.md
```diff
@@ -0,0 +1,471 @@
...
+```go
+// PodFromVM renders a Pod spec from a VirtualMachine definition.
+// It applies VM defaults, extracts the VMI template, applies VMI defaults
+// and mutations, configures networking, and calls RenderLaunchManifest.
+func PodFromVM(vm *virtv1.VirtualMachine, opts Options) (*corev1.Pod, error)
+
+// PodFromVMI renders a Pod spec from a VirtualMachineInstance definition.
+// It applies VMI defaults and mutations, configures networking, and calls
+// RenderLaunchManifest.
+func PodFromVMI(vmi *virtv1.VirtualMachineInstance, opts Options) (*corev1.Pod, error)
```
asafbennatan, line 173: If we set the interface binding and `VMI.UID` on the VMI before `PodFromVMI`, will `SetDefaultNetworkInterface` and the rest of this pipeline leave those fields alone?

asafbennatan (commented):
A few questions from implementing the same VM → Pod path in an offline converter. They assume this API emits the virt-controller launch Pod and that standalone runtime is the caller’s job; if that is wrong, the rest probably does not apply.

PR #361, opened 2026-06-29
VEP #172: Clarify Alpha scope is TCG-only, defer hardware acceleration to Beta

lyarwood:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/172
**SIG label**: /sig compute

### What this PR does

Clarifies the Alpha scope of VEP #172 (Cross-Architecture Virtualization) to
make explicit that only software emulation (TCG) is targeted at Alpha. Hardware-
accelerated cross-architecture virtualization (e.g. KVM/SAE) is deferred to
Beta.

Key changes:

- **Overview**: Mark software emulation as not intended for production use;
  hardware acceleration is the target for production support at Beta
- **KVM Domain Configurator**: Replace combined TCG+KVM struct/code with
  Alpha-only TCG implementation; show `kvmSupportedGuestArchitectures` and
  `kvmSupportsGuestArch()` as Beta additions
- **CPU Model Override / Graphics Device**: Remove `crossArchHardwareAccelerated`
  flag from Alpha code; document hardware acceleration refinements as Beta scope
- **Implementation Phases**: Move libvirt capabilities parsing and cross-arch-kvm
  node labels from Phase 1 (Alpha) to Phase 3 (Beta); add hardware acceleration
  infrastructure items to Phase 3
- **Graduation Requirements**: Move libvirt capabilities parsing from Alpha to
  Beta; add `kvmSupportedGuestArchitectures` and three-tier scheduling to Beta;
  add production readiness via hardware acceleration to Beta; check off 15/17
  Alpha items as completed in kubevirt/kubevirt#17199

This was prompted by a review comment on kubevirt/kubevirt#17199 noting the
diff between the VEP design (which included `kvmSupportedGuestArchitectures`)
and the implementation (TCG only). The Alpha implementation intentionally
targets only software emulation — this PR updates the VEP to make that
explicit.

### DIFF
--- a/veps/sig-compute/172-cross-arch-virtualization/vep.md
+++ b/veps/sig-compute/172-cross-arch-virtualization/vep.md
@@ -27,14 +27,16 @@ auto-detected at runtime:
 1. **Software emulation** via QEMU's TCG (Tiny Code Generator): available on
    any host with the appropriate cross-architecture QEMU system binary
    installed. Suitable for development, testing, and CI/CD workloads (10-100x
-   performance overhead).
+   performance overhead). **Not intended for production use** — software
+   emulation is provided for development and testing scenarios only.
 
 2. **Hardware-accelerated virtualization** via KVM extensions: available when
    the host platform provides hardware support for cross-architecture
    virtualization (e.g., SAE for ARM64 guests on s390x hosts).
    Performance characteristics are not yet known and will be determined
    once the upstream stack stabilizes and hardware is available for
-   benchmarking.
+   benchmarking. **This is the target for production support**, expected at
+   Beta.
 
 The emulation mode is determined automatically based on libvirt capabilities —
 users enable the feature gate and create cross-architecture VMs without needing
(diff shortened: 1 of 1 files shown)


jschintag (commented):
The changes look good, i have 1 question though:
Should we emit a warning when Software Emulation is used? That way we can let the users know what they are doing should not be done in production.

lyarwood:
> The changes look good, i have 1 question though: Should we emit a warning when Software Emulation is used? That way we can let the users know what they are doing should not be done in production.

@jschintag In the current implementation that only becomes evident post scheduling and that's why we've ended up with the condition.

If emulation was the eventual target for this feature I'd agree that we would likely need to do more to warn folks at admission time but for now it's just an alpha test-bed so I'd rather stay with the condition approach.

jschintag (approved):
Thank you for the clarification @lyarwood .

LGTM

lyarwood:
/cc @xpivarc 

Would you mind reviewing this?

enp0s3:
/lgtm

@lyarwood Thank you!

xpivarc (commented):
/approve

Issue #362, opened 2026-06-29
VEP 362: Guest Termination State


keinsword:
**Primary contact (assignee)**:

/assign @keinsword

**Current Feature Stage**:
New

**Feature Gate**:
GuestTermination

**Responsible SIGs**:

Primary SIG: /sig compute

Additional SIGs (optional): N/A

**Enhancement link**: [VEP doc](https://github.com/keinsword/kubevirt-enhancements/blob/41c2efc598a92d48ebfd46b868f034d89f171928/veps/sig-compute/362-guest-termination-state/vep.md)

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/363
  * [x] Code PRs: https://github.com/kubevirt/kubevirt/pull/18948
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

EdDev:
@keinsword , please update the description with the relevant information, especially the link to the VEP [document](https://github.com/kubevirt/enhancements/pull/363).

keinsword:
Hi @EdDev,
Done!

PR #363, opened 2026-06-29
VEP #362: Guest Termination State

keinsword:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/362
**SIG label**: /sig compute

### What this PR does
Introducing VEP for surfacing guest termination state through VM/VMI status.

None. 

## Summary by CodeRabbit

* **Documentation**
  * Added a proposal for retaining guest termination status and reasons after the launcher pod or VMI is removed.
  * Documents supported termination reasons, lifecycle behavior, condition synchronization with the owning VM, and handling for replacement VMIs.
  * Defines related events, metrics, feature-gate behavior, API examples, compatibility considerations, and testing requirements.

### DIFF
--- a/veps/sig-compute/362-guest-termination-state/vep.md
+++ b/veps/sig-compute/362-guest-termination-state/vep.md
@@ -1,0 +1,456 @@
+# VEP #362: Guest Termination State
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: 1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt can observe several different ways in which a guest or its domain stops:
+shutdown initiated from inside the guest, KubeVirt/API requested shutdown,
+hypervisor-side stop, and guest crash/panic. Today these cases are mostly
+reflected through generic VMI phase and readiness condition. For example, a guest
+shutdown can eventually appear as `phase: Succeeded` with
+`Ready=False, reason=GuestNotRunning`, while other stop paths may surface as
+generic failure, pod termination, or Kubernetes events.
+
+This proposal introduces a normalized guest termination state in the KubeVirt API.
+The state records the observed reason, message, and timestamp for a terminated
+guest/domain lifecycle. A VM-level copy also records the source VMI UID so the
+reason remains meaningful after the VMI object is deleted or replaced. It is
+intended as a durable, machine-readable source of truth for users and
+higher-level controllers that need to understand why a VM stopped.
+
+The public API is hypervisor-agnostic. Current libvirt lifecycle events are used
+only as one implementation source for populating the normalized reasons.
+
+## Motivation
+
+Users and automation often need to distinguish expected guest behavior from
+infrastructure or platform behavior. Examples:
+
+- A user runs `shutdown now` inside the guest and expects to understand that the
+  VM stopped because the guest requested it.
+- A user or controller requests a VM stop through KubeVirt and expects that stop
+  to be distinguishable from a shutdown initiated inside the guest.
+- A guest kernel panic should be observable as a guest crash instead of being
+  conflated with ordinary shutdown.
+- Higher-level platforms built on KubeVirt need a stable status surface to decide
+  whether to recover, report, alert, bill, or leave an instance stopped.
+
+Prior discussions and features address adjacent behavior but not a structured
+termination reason API:
+
+- Guest shutdown behavior and RunStrategy decide what KubeVirt should do after a
+  guest stops, but not why the guest stopped.
+- `rebootPolicy` makes guest reboot visible to KubeVirt when requested, but does
+  not expose the reason for a terminated guest/domain lifecycle.
+- Guest panic support emits Kubernetes events ([#16666](https://github.com/kubevirt/kubevirt/pull/16666))
+  and metrics ([#17836](https://github.com/kubevirt/kubevirt/pull/17836)) for one crash class,
+  but there is no general termination state covering shutdown and stop causes.
+
+## Goals
+
+- Add a normalized guest/domain termination state to VMI status.
+- Add a VM-level last termination state copied from the latest relevant VMI so
+  the reason remains visible after the VMI is deleted.
+- Distinguish at least:
+  - guest-initiated shutdown
+  - KubeVirt/API requested shutdown
+  - host/hypervisor requested shutdown
+  - host/hypervisor observed unexpected stop/failure
+  - guest crash/panic
+- Keep the public API independent from libvirt event names.
+- Preserve an observed termination reason across later low-signal terminal
+  domain notifications in the same domain lifecycle.
+- Emit a Kubernetes event when a guest termination state is observed.
+- Expose metrics for observed guest termination reasons.
+- Guard the initial implementation behind a feature gate.
+
+## Non Goals
+
+- This proposal does not change RunStrategy semantics.
+- This proposal does not change `rebootPolicy` semantics.
+- This proposal does not introduce a recovery policy or automatic restart policy.
+- This proposal does not make `Ready` condition carry historical termination state.
+- This proposal does not guarantee attribution for node crashes where the
+  virt-launcher process and hypervisor event stream are lost before an event can
+  be observed.
+- This proposal does not expose libvirt event names as KubeVirt API values.
+
+## Definition of Users
+
+- VM owners who need to understand why their VM stopped.
+- Cluster administrators and SREs investigating VM lifecycle incidents.
+- Higher-level controllers and platforms built on KubeVirt.
+- Monitoring and UI systems that need a stable source of VM stop attribution.
+- KubeVirt developers debugging guest/domain lifecycle behavior.
+
+## User Stories
+
+- As a VM owner, I want to see that my VM stopped because the guest OS requested
+  shutdown, so I can distinguish expected in-guest behavior from a failure.
+- As a cluster administrator, I want guest crashes to be distinguishable from
+  platform-requested shutdown, so alerts and dashboards can report the right
+  cause.
+- As a higher-level controller, I want a machine-readable stop reason after the
+  VMI has disappeared, so I can make recovery decisions based on the last VMI
+  lifecycle.
+- As a KubeVirt developer, I want a normalized status field instead of relying on
+  libvirt-specific lifecycle detail names, so the API can survive future
+  hypervisor implementation changes.
+
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+### API
+
+Introduce a VMI-scoped termination state:
+
+```go
+type VirtualMachineInstanceTerminationState struct {
+    // Reason is a short, stable, machine-readable termination reason.
+    // +optional
+    Reason VirtualMachineInstanceTerminationReason `json:"reason,omitempty"`
+
+    // Message is a human-readable explanation of the observed termination.
+    // +optional
+    Message string `json:"message,omitempty"`
+
+    // Timestamp is the time at which KubeVirt observed the termination event.
+    // +optional
+    Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+type VirtualMachineInstanceTerminationReason string
+
+const (
+    VirtualMachineInstanceTerminationReasonGuestShutdown              VirtualMachineInstanceTerminationReason = "GuestShutdown"
+    VirtualMachineInstanceTerminationReasonPlatformRequestedShutdown  VirtualMachineInstanceTerminationReason = "PlatformRequestedShutdown"
+    VirtualMachineInstanceTerminationReasonHostShutdown               VirtualMachineInstanceTerminationReason = "HostShutdown"
+    VirtualMachineInstanceTerminationReasonHostStoppedFailed          VirtualMachineInstanceTerminationReason = "HostStoppedFailed"
+    VirtualMachineInstanceTerminationReasonGuestCrashed               VirtualMachineInstanceTerminationReason = "GuestCrashed"
+)
+```
+
+Add it to VMI status:
+
+```go
+type VirtualMachineInstanceStatus struct {
+    // TerminationState reports the observed termination state for this VMI
+    // lifecycle.
+    // +optional
+    TerminationState *VirtualMachineInstanceTerminationState `json:"terminationState,omitempty"`
+}
+```
+
+Add a VM-level copy that includes the VMI UID:
+
+```go
+type VirtualMachineTerminationState struct {
+    // VMIUID identifies the VMI lifecycle this termination state came from.
+    // +optional
+    VMIUID types.UID `json:"vmiUID,omitempty"`
+
+    // Reason is a short, stable, machine-readable termination reason.
+    // +optional
+    Reason VirtualMachineInstanceTerminationReason `json:"reason,omitempty"`
+
+    // Message is a human-readable explanation of the observed termination.
+    // +optional
+    Message string `json:"message,omitempty"`
+
+    // Timestamp is the time at which KubeVirt observed the termination event.
+    // +optional
+    Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+type VirtualMachineStatus struct {
+    // LastTerminationState reports the latest observed termination state copied
+    // from a VMI owned by this VM.
+    // +optional
+    LastTerminationState *VirtualMachineTerminationState `json:"lastTerminationState,omitempty"`
+}
+```
+
+The VMI field is the authoritative state for a specific VMI lifecycle. The VM
+field is a durable copy for user-facing inspection after the VMI is deleted. The
+`vmiUID` field prevents consumers from accidentally treating a previous VMI
+lifecycle as the current one.
+
+### Reason semantics
+
+`GuestShutdown` means the guest OS requested shutdown/poweroff and KubeVirt
+observed a normal guest termination.
+
+`PlatformRequestedShutdown` means KubeVirt requested domain shutdown as part of
+VM/VMI stop or deletion flow and the observed termination corresponds to that
+request.
+
+`HostShutdown` means the hypervisor host side requested shutdown or destroy, but
+KubeVirt did not associate the event with a tracked platform shutdown request.
+
+`HostStoppedFailed` means the hypervisor reported that the domain stopped
+unexpectedly or failed without a more specific guest crash signal.
+
+`GuestCrashed` means the guest reported or triggered a crash/panic signal.
+
+The exact names are open to SIG review. The important contract is that the
+reason values are normalized KubeVirt concepts, not libvirt event names.
+
+### Event collection
+
+The initial implementation can derive the normalized state from the existing
+virt-launcher to virt-handler domain status flow.
+
+virt-launcher records normalized termination events in the internal domain status
+sent to virt-handler. This internal value is not the public API. It is a
+transport detail used so virt-handler can update the VMI/VM API, emit events, and
+increment metrics.
+
+Current libvirt mapping:
+
+| Libvirt event | Libvirt detail | Normalized reason |
+|---------------|----------------|-------------------|
+| `DOMAIN_EVENT_SHUTDOWN` | `DOMAIN_EVENT_SHUTDOWN_GUEST` | `GuestShutdown`, unless matched to a pending KubeVirt shutdown intent |
+| `DOMAIN_EVENT_SHUTDOWN` | `DOMAIN_EVENT_SHUTDOWN_HOST` | `HostShutdown`, unless matched to a pending KubeVirt shutdown intent |
+| `DOMAIN_EVENT_CRASHED` | `DOMAIN_EVENT_CRASHED_PANICKED` | `GuestCrashed` |
+| `DOMAIN_EVENT_CRASHED` | `DOMAIN_EVENT_CRASHED_CRASHLOADED` | `GuestCrashed` |
+| `DOMAIN_EVENT_STOPPED` | `DOMAIN_EVENT_STOPPED_CRASHED` | `GuestCrashed` |
+| `DOMAIN_EVENT_STOPPED` | `DOMAIN_EVENT_STOPPED_FAILED` | `HostStoppedFailed` |
+| `DOMAIN_EVENT_STOPPED` | `DOMAIN_EVENT_STOPPED_DESTROYED` | no new reason |
+| `DOMAIN_EVENT_STOPPED` | `DOMAIN_EVENT_STOPPED_SHUTDOWN` | no new reason |
+| `DOMAIN_EVENT_STOPPED` | `DOMAIN_EVENT_STOPPED_MIGRATED` | no new reason |
+
+`DOMAIN_EVENT_STOPPED_CRASHED` may refine or overwrite a preceding
+`DOMAIN_EVENT_CRASHED_*` event because it confirms that the domain has stopped.
+
+`PlatformRequestedShutdown` is detected by recording a pending KubeVirt shutdown
+intent when virt-launcher issues the shutdown request on behalf of KubeVirt. The
+intent is consumed by the next matching shutdown event. The intent is considered
+fresh for the larger of a default freshness window and the VMI termination grace
+period plus a small margin. A stale intent is cleared and ignored.
+
+### Sticky internal domain termination event
+
+Some domain lifecycles produce a useful first event followed by a later terminal
+event that does not carry enough information by itself. For example:
+
+- guest shutdown: `SHUTDOWN_GUEST` followed by `STOPPED_SHUTDOWN`
+- platform-requested shutdown: `SHUTDOWN_GUEST` or `SHUTDOWN_HOST` matched to a
+  pending platform intent, followed by `STOPPED_SHUTDOWN` or `STOPPED_DESTROYED`
+- guest crash: `CRASHED_PANICKED` followed by `STOPPED_CRASHED`
+
+To avoid losing the useful reason during the final domain notification,
+virt-launcher keeps the last observed normalized termination event in its
+per-launcher metadata cache. When a later domain notification does not map to a
+new normalized reason, virt-launcher attaches the cached event to the domain
+status sent to virt-handler.
+
+The cached event is scoped to one virt-launcher/domain lifecycle. It is cleared
+when a `DOMAIN_EVENT_STARTED` event is observed. The same start event also clears
+any pending platform termination intent. This prevents a termination reason from
+leaking into a fresh domain incarnation.
+
+If a low-signal terminal event is observed without any cached normalized event,
+KubeVirt does not infer a termination reason from it. This avoids turning
+ambiguous terminal states into misleading user-visible attribution.
+
+### VM propagation
+
+When a VMI owned by a VM has `status.terminationState`, virt-controller copies it
+to `vm.status.lastTerminationState` with the source VMI UID.
+
+The VM field is intentionally named `lastTerminationState` because it is
+historical. It is not a statement that the current VMI, if one exists, is
+terminated.
+
+### Kubernetes events
+
+When KubeVirt observes a normalized termination state, it emits a Kubernetes
+event on the VMI:
+
+- Event type `Normal` for expected shutdown reasons.
+- Event type `Warning` for crash or unexpected stop reasons.
+- Event reason is the normalized termination reason.
+- Event message is the same human-readable message used in status.
+
+### Metrics
+
+Expose a counter for observed guest termination states:
+
+```text
+kubevirt_vmi_guest_os_termination_total{namespace, name, reason}
+```
+
+The metric follows the precedent of guest panic metrics and allows operators to
+track fleet-wide termination causes. Implementations should avoid adding
+high-cardinality detail labels such as raw hypervisor messages.
+
+## API Examples
+
+Guest shutdown visible on the VMI:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: fedora
+  namespace: default
+  uid: 7d5f0d4d-3df8-442f-bd64-9ee9d0f6fb8e
+status:
+  phase: Succeeded
+  terminationState:
+    reason: GuestShutdown
+    message: Guest requested shutdown of the virtual machine
+    timestamp: "2026-06-15T12:00:00Z"
+```
+
+The same reason copied to the VM after the VMI is gone:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: fedora
+  namespace: default
+status:
+  printableStatus: Stopped
+  lastTerminationState:
+    vmiUID: 7d5f0d4d-3df8-442f-bd64-9ee9d0f6fb8e
+    reason: GuestShutdown
+    message: Guest requested shutdown of the virtual machine
+    timestamp: "2026-06-15T12:00:00Z"
+```
+
+Guest crash:
+
+```yaml
+status:
+  terminationState:
+    reason: GuestCrashed
+    message: Guest crashed
+    timestamp: "2026-06-15T12:05:00Z"
+```
+
+Platform requested shutdown:
+
+```yaml
+status:
+  terminationState:
+    reason: PlatformRequestedShutdown
+    message: Platform requested shutdown of the virtual machine
+    timestamp: "2026-06-15T12:10:00Z"
+```
+
+## Alternatives
+
+### Use the Ready condition
+
+`Ready` already reports `GuestNotRunning`, `PodTerminating`, and other
+availability states. It is not a good fit for termination attribution because it
+is transient, availability-oriented, and can copy pod readiness reasons directly.
+`GuestNotRunning` also covers startup and other non-running windows, not only
+guest termination.
+
+### Add a GuestTerminated condition
+
+A condition is useful for current state, but less suitable for durable historical
+state. It also creates awkward behavior across VM restarts: a `True` condition can
+look current even when it describes a previous VMI lifecycle. A structured status
+field with source `vmiUID` is clearer for "last observed termination".
+
+### Store the reason only on the VM
+
+The VM is the best user-facing place after the VMI is deleted, but the VMI is the
+natural authoritative object for a specific domain lifecycle. Keeping both fields
+allows controllers to reason about the current VMI lifecycle while still
+preserving a durable VM-level copy.
+
+### Use only Kubernetes events or metrics
+
+Events and metrics are useful for observability, but they are not a stable API
+source of truth for controllers. Events can expire and metrics are eventually
+scraped. Status is the right API surface for current and last observed state.
+
+## Scalability
+
+The status fields are small and updated only when KubeVirt observes a guest/domain
+termination event. The metric adds one counter series per observed VMI and reason,
+matching existing KubeVirt per-VMI metric patterns. Implementations should avoid
+labels derived from raw messages, domain names, or crash payloads.
+
+## Update/Rollback Compatibility
+
+This is an additive status API and is backward compatible.
+
+During upgrade, new KubeVirt components begin populating the optional status
+fields when the feature gate is enabled. Older clients ignore the new fields.
+
+During rollback, older KubeVirt components stop updating the fields. Existing
+status values may remain until the object is updated or recreated. Consumers must
+treat the fields as optional.
+
+The initial implementation should be guarded by a feature gate, tentatively named
+`GuestTermination`.
+
+## Functional Testing Approach
+
+- Unit test mapping from hypervisor lifecycle events to normalized termination
+  reasons.
+- Unit test pending KubeVirt shutdown intent matching and stale intent cleanup.
+- Unit test that later low-signal domain notifications preserve the cached
+  normalized termination event.
+- Unit test that domain start clears stale in-launcher termination state and
+  pending platform termination intent.
+- Unit test VMI status updates for each normalized reason.
+- Unit test VM propagation from VMI `terminationState` to VM
+  `lastTerminationState`.
+- Functional test guest-initiated shutdown from inside a Linux guest.
+- Functional test KubeVirt/API requested VM stop or VMI deletion.
+- Functional test guest panic when panic devices are available.
+- Functional test unexpected hypervisor stop/failure path where practical.
+- Verify no status, event, or metric is emitted when the feature gate is disabled.
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,456 @@
...
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
```
EdDev, line 20: Please try to summarize what this VEP is all about as short as possible.
Expanding into details can be left to the motivation, goals and other sections.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,456 @@
...
+The state records the observed reason, message, and timestamp for a terminated
+guest/domain lifecycle. A VM-level copy also records the source VMI UID so the
+reason remains meaningful after the VMI object is deleted or replaced. It is
+intended as a durable, machine-readable source of truth for users and
+higher-level controllers that need to understand why a VM stopped.
+
+The public API is hypervisor-agnostic. Current libvirt lifecycle events are used
+only as one implementation source for populating the normalized reasons.
+
+## Motivation
```
EdDev, line 35: I understand that a VM consumer would be interested to know the reason and details of a VM termination.
I'm missing if this information is available today through events and if that is sufficient or not. In other words, why would it be useful through the status rather then through an event?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,456 @@
...
+
+- Guest shutdown behavior and RunStrategy decide what KubeVirt should do after a
+  guest stops, but not why the guest stopped.
+- `rebootPolicy` makes guest reboot visible to KubeVirt when requested, but does
+  not expose the reason for a terminated guest/domain lifecycle.
+- Guest panic support emits Kubernetes events ([#16666](https://github.com/kubevirt/kubevirt/pull/16666))
+  and metrics ([#17836](https://github.com/kubevirt/kubevirt/pull/17836)) for one crash class,
+  but there is no general termination state covering shutdown and stop causes.
+
+## Goals
```
EdDev, line 57: Please keep only user facing goals, agnostic as much as possible and avoid implementation details (e.g. the need for a FG).

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,456 @@
...
+
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+### API
+
+Introduce a VMI-scoped termination state:
```
EdDev: I would lean towards using a standard condition (or several, depending on the need). I would expect one to exists already which could facilitate this, but I have not checked.

I would prefer not to introduce a custom status structure for this.

EdDev (commented):
This is a partial review.

I think that for the v1.10 release, you will need a sponsor from the relevant SIG to help prioritize this.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: > I'm missing if this information is available today through events and if that is sufficient or not.

Currently, only guest OS crashes are reported through events ([#16666](https://github.com/kubevirt/kubevirt/pull/16666)), and metrics ([#17836](https://github.com/kubevirt/kubevirt/pull/17836)), kudos to @iholder101. 
None of the other possible guest shutdown reasons are surfaced. 

> why would it be useful through the status rather then through an event?

Events are useful for audit and notifications, but they're transient, and while they still may serve as a complementary notification mechanism, I believe VM status is a proper place for providing a durable, structured termination reason for users and controllers.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: In fact, in our downstream, we implemented by introducing a new `GuestTerminated` VM/VMI condition. But it's semantically misleading: condition describes current state, while termination reason is historical information about the latest VMI lifecycle.

Existing conditions such as `Ready` only describe availability and do not preserve the termination cause. A dedicated status field can explicitly carry the reason, timestamp, and originating VMI UID.

keinsword:
@EdDev appreciate your review! Please take a look at my replies.

/cc @0xFelix @iholder101 
Hi! May I ask you folks to spare a moment and review this one? Please feel free to `cc` whoever you think may be interested as well.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+  but there is no general termination state covering shutdown and stop causes.
+
+## Goals
+
+- Add a normalized guest/domain termination state to VMI status.
+- Add a VM-level last termination state copied from the latest relevant VMI so
+  the reason remains visible after the VMI is deleted.
+- Distinguish at least:
+  - guest-initiated shutdown
+  - KubeVirt/API requested shutdown
```
0xFelix: Could also distinguish between ACPI and guest-agent initiated shutdown, if that's not too granular

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+  - host/hypervisor requested shutdown
```
0xFelix: What does it constitute exactly? Can you give an example? How is it different from a KubeVirt initiated shutdown?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+  - host/hypervisor observed unexpected stop/failure
+  - guest crash/panic
+- Emit a Kubernetes event when a guest termination state is observed.
+- Expose metrics for observed guest termination reasons.
```
0xFelix: What kind of metrics? Why do we need additional metrics for it?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+
+- As a VM owner, I want to see that my VM stopped because the guest OS requested
+  shutdown, so I can distinguish expected in-guest behavior from a failure.
+- As a cluster administrator, I want guest crashes to be distinguishable from
+  platform-requested shutdown, so alerts and dashboards can report the right
+  cause.
+- As a higher-level controller, I want a machine-readable stop reason after the
+  VMI has disappeared, so I can make recovery decisions based on the last VMI
+  lifecycle.
+- As a KubeVirt developer, I want a normalized status field instead of relying on
```
0xFelix: Not sure that's a user story we definitely need

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: And how would that field be treated differently than the status condition? I.e. if you restart the VM after it was stopped for whatever reason, is the new field cleared or will it also reflect "the past"? If that's the case I'd opt for the status condition and only add it if the VM was stopped, remove it once the VM is running again.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+    Reason VirtualMachineInstanceTerminationReason `json:"reason,omitempty"`
+
+    // Message is a human-readable explanation of the observed termination.
+    // +optional
+    Message string `json:"message,omitempty"`
+
+    // Timestamp is the time at which KubeVirt observed the termination event.
+    // +optional
+    Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
```
0xFelix: It's basically the same set of fields as on a status condition

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+```
+
+Add a VM-level copy that includes the VMI UID:
+
+```go
+type VirtualMachineTerminationState struct {
+    // VMIUID identifies the VMI lifecycle this termination state came from.
+    // +optional
+    VMIUID types.UID `json:"vmiUID,omitempty"`
```
0xFelix: You could encode this into the message and re-use a status condition instead

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+    // Timestamp is the time at which KubeVirt observed the termination event.
+    // +optional
+    Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+type VirtualMachineStatus struct {
+    // LastTerminationState reports the latest observed termination state copied
+    // from a VMI owned by this VM.
+    // +optional
+    LastTerminationState *VirtualMachineTerminationState `json:"lastTerminationState,omitempty"`
```
0xFelix: Why would we keep this historical information on a running VM?

0xFelix: You say `condition describes current state`, but that's also the point of the whole `status` field, no?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+infrastructure or platform behavior. Examples:
+
+- A user runs `shutdown now` inside the guest and expects to understand that the
+  VM stopped because the guest requested it.
+- A user or controller requests a VM stop through KubeVirt and expects that stop
+  to be distinguishable from a shutdown initiated inside the guest.
+- A guest kernel panic should be observable as a guest crash instead of being
+  conflated with ordinary shutdown.
+- Higher-level platforms built on KubeVirt need a stable status surface to decide
+  whether to recover, report, alert, bill, or leave an instance stopped.
```
0xFelix: Isn't it the point of the `RunStrategy` to recover the VM in case of an unexpected failure or shutdown? Why should an external controller be concerned with that? IIRC we expose enough conditions / phases to distinguish between a running, shut down or failed VM. Isn't that enough for the mentioned purposes?

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: I'd not narrow it that much. Can hardly think of a possible use case.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: KubeVirt/API requested shutdown goes through `DomainManager.SignalShutdownVMI()` which in the end destroys the domain, e.g. `kubectl delete vmi` or `kubectl delete pod <launcher pod>`
Whereas host/hypervisor requested shutdown is when domain gets destroyed "from outside", e.g. through `virsh destroy <domain>`.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: It could be a Prometheus metric that records total number of observed guest OS termination events, partitioned by VMI and reason. It'd help to keep a history and stats for the variety of termination causes in the cluster, not only for panics/crashes as it is now.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: RunStrategy handles kubevirt’s own restart policy, but higher-level platforms may manage recovery and user-facing lifecycle independently according to for example upper-level billing logic. Existing phases distinguish broad outcomes such as Succeeded and Failed, but not the initiator or specific cause.
E.g., say, in our compute platform we'd like an instance not to be recovered if it was shutdown from inside the guest (`shutdown now`), but otherwise it should be recovered (guest crashes, qemu disconnects, etc.). And yes, there's `RunStrategy: RerunOnFailure`, but it acts on VMI phase Succeeded, while not only `shutdown now` ends up in Succeeded, if I'm not mistaken.

keinsword:
Thanks a lot for your feedback @0xFelix!
After reading your and @EdDev comments, I agree this one fits better represented as a condition, just like we already implemented it in our fork.
Do you think this is worth considering for v1.10 as a status condition? Provided that the implementation is already done and tested in our clusters.
I'm ready to reshape the VEP doc accordingly.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: Yes, I'd agree with you on this, presenting this as a condition now seems to be fitting the existing logic better, please see my last [comment](https://github.com/kubevirt/enhancements/pull/363#discussion_r3825962726) below.

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: > E.g., say, in our compute platform we'd like an instance not to be recovered if it was shutdown from inside the guest (shutdown now), but otherwise it should be recovered (guest crashes, qemu disconnects, etc.).

A combination of `RunStrategy: RerunOnFailure` and `RebootPolicy: Terminate` should allow you to exactly do that?

0xFelix:
> Do you think this is worth considering for v1.10 as a status condition? Provided that the implementation is already done and tested in our clusters.

I'm still trying to understand if this is something that is not already covered by the existing fields and options. If so, I'd agree with making this a status condition.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: AFAICS, setting `RebootPolicy: Terminate` will be leaving a VM rebooted from inside guest with VMI in `Succeeded` phase, and, subsequently, `RunStrategy: RerunOnFailure` will leave this VM stopped, i.e. unrecovered. 
This will effectively disallow a VM user to reboot their vm from isnide guest, it will simply not boot back, and this is definitely not something we'd like users to experience. So I do not see how the combination of `RunStrategy: RerunOnFailure` and `RebootPolicy: Terminate` would help here.

keinsword:
> I'm still trying to understand if this is something that is not already covered by the existing fields and options.

I wish it was, but I don't see such a field (or set of them) that would satisfy the request I'm trying to with this vep. 
Please see my [comment](https://github.com/kubevirt/enhancements/pull/363#discussion_r3830838495). 

Please let me know if you find anything else in the existing codebase that would help, would happily evaluate that.

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: I see. You are right, this combination would lead to the VM staying stopped.

What do you think about adding a field for the `Poweroff` policy? Libvirt/QEMU should be able to distinguish between a guest initiated shutdown or reboot. See https://libvirt.org/formatdomain.html#events-configuration

keinsword: Are you suggesting something like `PoweroffPolicy: Stop`, where guest poweroff suppresses recreation but crashes and guest reboots continue to follow `RunStrategy` and `RebootPolicy`?

The problem here is that `kubectl delete vmi` or `kubectl delete pod <launcher-pod>` end up in `virDomainShutdownFlags()` which in most cases tries to do graceful ACPI guest shutdown first, which makes it indistinguishable from normal ACPI shutdown from inside guest. And for both of them libvirt would apply the same configured `<on_poweroff>` action, while it would be very useful to apply different instance recovery policies for these two ways of shutting guest down, i.e:
- if that was an "intentional" shutdown from inside guest, do not recreate domain/VMI
- if the VMI (or launcher pod) was deleted from cluster, recreate it

And that's something my VEP already solves, even if we go with a status condition instead of a dedicated structure in status.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+
+KubeVirt can observe several different ways in which a guest or its domain stops:
+shutdown initiated from inside the guest, KubeVirt/API requested shutdown,
+hypervisor-side stop, and guest crash/panic. Today these cases are mostly
+reflected through generic VMI phase and readiness condition, while other stop paths
+may surface as generic failure, pod termination, or Kubernetes events.
+
+This proposal introduces a normalized guest termination state in the KubeVirt API.
+The state records the observed reason, message, and timestamp for a terminated
+guest/domain lifecycle. A VM-level copy also records the source VMI UID so the
```
stu-gott: Any time we're copying (redundant) information between resources, this introduces a race condition concern. In this case it can be particularly insidious. In general virt-handler owns and updates the VMI's records. virt-controller owns the VM's records. Which controller is then responsible for copying data from the VMI to the VM?

stu-gott: A corralary concern:

If a shutdown reason field is added, both virt-controller or virt-handler might attempt to update the VMI in a conflicting manner.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,439 @@
...
+
+- Add a normalized guest/domain termination state to VMI status.
+- Add a VM-level last termination state copied from the latest relevant VMI so
+  the reason remains visible after the VMI is deleted.
+- Distinguish at least:
+  - guest-initiated shutdown
+  - KubeVirt/API requested shutdown
+  - host/hypervisor requested shutdown
+  - host/hypervisor observed unexpected stop/failure
+  - guest crash/panic
```
stu-gott: As you mentioned above, we already capture and record events for pvpanic. We did it that way because panics immediately kill the virt-launcher and the information about the crash is otherwise lost. How can this proposal do anything different. *there's no longer a VMI to scribble a shutdown reason on*.

stu-gott (commented):
I think overall, @EdDev's question is paramount. Are the current sources of shutdown information sufficient?

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: After discussing this with @0xFelix, we've pretty much agreed it's better to represent this guest termination state through the status condition instead of a new struct in VMI/VM status. 
Again, in our downstream, this is already implemented as `GuestTerminated` condition. 

> Which controller is then responsible for copying data from the VMI to the VM?

Given that it's a condition, virt-controller handles that by syncing the VMI conditions to VM resource in `VMController.syncConditions()`. This preserves the ordering:
1. virt-launcher catches one of the "shutdown-like" libvirt events and writes it to the domain status
2. virt-handler observes the new termination event in domain status and updates the `GuestTerminated` VMI condition accordingly
3. virt-controller syncs, among the others, `GuestTerminated` condition from VMI to VM

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: > panics immediately kill the virt-launcher

If this is true, then how do those guest panic metrics introduced by Itamar even work? 🙃 

Those metrics/events work because on panics/crashes libivrt sends `VIR_DOMAIN_EVENT_CRASHED`, and crash metrics code in launcher's notify-client just tries to intercept it in the `eventCallback` [as early as possible](https://github.com/kubevirt/kubevirt/blob/de139aee8e85066225ff5c69c621c5fa5221f90f/pkg/virt-launcher/notify-client/client.go#L341).

And we can handle other libvirt's shutdown-related events the same way.

If you're interested, I'm ready open a draft implementation PR for this VEP with condition-based approach so you cold get some more insight.

 And back to the main question:
> Are the current sources of shutdown information sufficient?

No, I don't think so. So far neither of the current sources is able to help to distinguish between an "intentional" shutdown from inside guest and e.g. VMI/launcher pod deletion through `kubectl delete vmi/pod <launcher-pod>`. See my [previous](https://github.com/kubevirt/enhancements/pull/363#discussion_r3852040167) comment.

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: Unfortunate, I see.

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: > If you're interested, I'm ready open a draft implementation PR for this VEP with condition-based approach so you cold get some more insight.

If you have the code already available that would help.

keinsword: Here it goes: https://github.com/kubevirt/kubevirt/pull/18948

keinsword:
@0xFelix @EdDev @stu-gott hi!
Posted an implementation PR some time ago, would appreciate if you could take a look: https://github.com/kubevirt/kubevirt/pull/18948

keinsword:
@0xFelix @stu-gott @EdDev 
Updated the VEP doc based on the ongoing discussion in the implementation [PR](https://github.com/kubevirt/kubevirt/pull/18948).

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+controllers a value they can query directly. Mirroring it to the VM keeps the
+answer available after the terminated VMI has been deleted.
+
+This information can be used by higher-level platforms for recovery policy,
+user-facing stop reasons, alerting, and other lifecycle decisions that are not
+owned by KubeVirt's `RunStrategy`.
+
+## Goals
+
+- Report whether the guest represented by a VMI has terminated.
```
0xFelix:
```suggestion
- Report whether the guest represented by a VMI has terminated at some point.
```

IIUC the cond could still be present, even though the VMI has started back up?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+- Give a terminated guest a normalized, machine-readable reason when KubeVirt
+  observed one.
+- Keep the latest condition on the owning VM after the VMI is deleted.
```
0xFelix: This already focuses on the how, not the what (durable record of when/why guest was stopped)

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+answer available after the terminated VMI has been deleted.
+
+This information can be used by higher-level platforms for recovery policy,
+user-facing stop reasons, alerting, and other lifecycle decisions that are not
+owned by KubeVirt's `RunStrategy`.
+
+## Goals
+
+- Report whether the guest represented by a VMI has terminated.
+- Give a terminated guest a normalized, machine-readable reason when KubeVirt
```
0xFelix: Could be merged with previous goal?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+  observed one.
+- Keep the latest condition on the owning VM after the VMI is deleted.
+- Reset stale termination state when a new guest or replacement VMI starts.
```
0xFelix: So, is the cond removed when the VMI starts back up or not?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+    VirtualMachineInstanceGuestTerminated VirtualMachineInstanceConditionType = "GuestTerminated"
+    VirtualMachineGuestTerminated         VirtualMachineConditionType         = "GuestTerminated"
+)
+```
+
+For a VMI, the condition has the following meaning:
+
+| Status | Meaning |
+|--------|---------|
+| `False` | The current guest incarnation has not terminated. |
```
0xFelix: Should it be present when `False` at all?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+condition untouched while status and reason remain the same. There is no
+second, custom termination timestamp.
+
+### Reasons
+
+When the condition is `False`, it uses:
+
+| Reason | Message |
+|--------|---------|
+| `GuestNotTerminated` | `Guest has not terminated` |
```
0xFelix: Could also be removed then?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+virt-handler exposes:
+
+```text
+kubevirt_vmi_guest_os_termination_total{namespace, name, reason}
+```
+
+The counter increments when `GuestTerminated` first becomes `True` or changes to
+a different supported termination reason. Per-VMI label sets are removed during
+final VMI cleanup so the virt-handler process does not retain them indefinitely.
+No raw hypervisor messages or other unbounded details are used as labels.
```
0xFelix, line 321: Are metrics required for Alpha? Is there a real use for them?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+
+- [ ] `GuestTermination` feature gate added.
+- [ ] VMI and VM `GuestTerminated` condition types added.
+- [ ] Initial normalized reason set implemented.
+- [ ] Termination-classifying events retained before any best-effort queue.
+- [ ] Retained state scoped to a domain incarnation with a reconciliation
+      fallback when `STARTED` is missed.
+- [ ] Dedicated VM synchronization preserves the terminal condition after VMI
+      deletion and clears it for a replacement VMI.
+- [ ] Kubernetes events and Prometheus metrics implemented.
```
0xFelix: I'd defer metrics for now

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+
+- While a VMI exists and carries the condition, copy it to the VM.
+- If a replacement VMI exists but does not yet carry the condition, remove the
+  previous VM condition. This prevents a reason from the old VMI from looking
+  current.
+- Once the VMI is gone, retain the condition last copied from it.
+- A new VMI's `False`, `True`, or `Unknown` condition replaces the retained
+  value normally.
+
+VM-owned VMIs already carry `VirtualMachineControllerFinalizer`.
```
0xFelix, line 285: Can you also ensure `virt-handler` has added this condition, before `virt-controller` removes the finalizer? Both `virt-controller` and `virt-handler` reconcile the VM, which makes this part potentially brittle.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+## Design
+
+### Condition contract
+
+No new status structure is introduced. The existing VMI and VM condition types
+gain `GuestTerminated`:
+
+```go
+const (
+    VirtualMachineInstanceGuestTerminated VirtualMachineInstanceConditionType = "GuestTerminated"
```
0xFelix, line 111: Add reasons consts to API too?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+### Require exact causal attribution from the hypervisor
+
+Libvirt can remember that its shutdown API was called, but it faces the same
+correlation problem as KubeVirt: QEMU later reports only that the guest completed
+shutdown. Exact attribution would require a token propagated through the guest,
+which is not available in the current ACPI shutdown protocol.
+
+## Scalability
+
+Each VMI and VM gains at most one condition. Condition updates occur only when
```
0xFelix, line 442: I think we should avoid adding it, if not required.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+the state or reason changes.
+
+virt-launcher retains one small termination record for the current domain
+incarnation. The Prometheus counter uses one series per observed VMI and reason,
+following existing per-VMI metric patterns, and removes those label sets during
+VMI cleanup.
+
+## Update/Rollback Compatibility
```
0xFelix, line 450: Also, when upgradingg KubeVirt but not upgrading the virt-launcher, VMs will not have this condition on shutdown.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: Semantically, it'd make sense to have it in `False` until the termination event is observed, but I'd agree here: it does not carry any useful info up until the moment it is flipped to `True`, only an additional noise in status, so I'll go ahead and and redo this as `True`-only.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: Will state that explicitly, as we go to true-only cond.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: Not if we go with true-only condition. If we do, it gets cleared once replacement VMI starts.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: If we go with true-only, it is removed when the VMI starts back up.

veps/sig-compute/362-guest-termination-state/vep.md
keinsword: Just like with panic metrics implemented by Itamar, we may want to preserve some historical data for how often each termination reason takes place in the cluster. I do not expect any significant use of those, so it's totally fine to move it to Beta, and at that point we'll re-evaluate whether there's any particular need in them.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+
+This is an additive condition type and is backward compatible. Older clients
+ignore condition types they do not recognize. Consumers must treat the
+condition as optional because it may be absent before the domain is observed,
+while the feature is disabled, or during a mixed-version rollout.
+
+After rollback or disabling the gate, controllers stop publishing the condition
+and remove it when they next reconcile the affected objects. A value can remain
+temporarily on an object that is no longer being reconciled, which is why clients
+must not assume the condition is always present.
```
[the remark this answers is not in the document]
keinsword: Addressed.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+virt-launcher latches termination-classifying lifecycle data in synchronized
+per-launcher state before attempting to enqueue the generic notification. It
+then signals the notification loop independently. Queue saturation may coalesce
+notifications, but it must not discard the retained termination reason.
+
+The retained data belongs to one guest incarnation. A libvirt
+`DOMAIN_EVENT_STARTED` clears it as a fast path. Reconciliation also compares
+the active runtime identity with the identity stored alongside the retained
+event. If the identity changed, the old event is cleared even if the `STARTED`
+callback was dropped.
```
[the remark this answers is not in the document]
keinsword: Addressed.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,544 @@
...
+)
+```
+
+For a VMI, the condition has the following meaning:
+
+| Status | Meaning |
+|--------|---------|
+| `False` | The current guest incarnation has not terminated. |
+| `True` | The current guest incarnation has terminated and KubeVirt observed a supported reason. |
+| `Unknown` | The domain is no longer active, but KubeVirt could not determine a supported termination reason. |
```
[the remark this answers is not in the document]
keinsword: Addressed.

veps/sig-compute/362-guest-termination-state/vep.md
0xFelix: I meant VMIs by the way, hope that was clear :)

keinsword: 💯

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+- This VEP targets alpha for version: 1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
```
EdDev, line 16: Should be marked.

veps/sig-compute/362-guest-termination-state/vep.md
EdDev: It makes sense.
In terms of naming used, `virt-controller` is a binary, when discussing controllers it would be better to name them explicitly as vmi, vm, migration, etc and not as a combined one named "virt".

In any case, IMO this thread can be resolved.

veps/sig-compute/362-guest-termination-state/vep.md
EdDev: Thank you for handling this, I think the shape it took is good enough.
It could be improved by stating the end result without details on the "how", but this is a bit of nitpicking.

Feel free to resolve this thread.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+`Ready=False` answers whether the VMI is available. It does not answer why it
+stopped. `GuestNotRunning`, for example, is also used while a guest is starting.
+`RunStrategy` and `rebootPolicy` decide what KubeVirt should do next; neither is
+a record of what happened.
+
+Kubernetes events are useful for notification and investigation, but they are
+not enough as the only API. They expire, may be aggregated, and require a
+consumer to watch and correlate a separate stream. A condition gives users and
+controllers a value they can query directly. Mirroring it to the VM keeps the
+answer available after the terminated VMI has been deleted.
```
EdDev, line 51: The phrasing here is leaking the solution in the motivation.
It would be useful to avoid that and just express the motivation, and express in a general wording that the need is to observe the reason as part of the VM lifecycle.

veps/sig-compute/362-guest-termination-state/vep.md
EdDev: What is this thread all about? Is it a GitHub bug that the comment carried on even that the lines changed?

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+### Condition contract
+
+No new status structure is introduced. The existing VMI and VM condition types
+gain `GuestTerminated`:
```
EdDev, line 107:
```suggestion
In order to express the termination reason and details, a new condition is suggested on the VMI and VM status, named `GuestTerminated`.
```

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+```
+
+When the condition is `True`, it uses one of these normalized reasons:
+
+| Reason | Message | Meaning |
+|--------|---------|---------|
+| `GuestShutdown` | `Guest requested shutdown of the virtual machine` | The hypervisor reported a guest-completed shutdown and KubeVirt had no pending platform termination intent. |
+| `PlatformRequestedShutdown` | `Platform requested shutdown of the virtual machine` | KubeVirt had already requested termination when the shutdown was observed. |
+| `HostShutdown` | `Host requested shutdown of the virtual machine` | The hypervisor reported a host-side shutdown without a matching KubeVirt termination intent. |
+| `HostStoppedFailed` | `Host observed the virtual machine stop unexpectedly` | The hypervisor reported that the domain stopped unexpectedly or failed. |
```
EdDev, line 155: IIUC this hints that the system can observe the reason given by the hypervisor and look at the intent stated in the spec, and conclude/compute the termination reason.

I do not think this is possible unless the hypervisor exposes something like this.
There is an integrated race which is wide or narrow depending on the implementation. But it is always there.
E.g. consider the follow sequence of events:
1. Guest user asks to shutdown the guest OS.
2. VM owner/operator, asks to stop the VM a second later.

We will report as if the termination was due to the VM operation? That would be wrong.
We get into more trouble when we thing about the fact the controllers and virt-handler reconcile all the time. If the input appears now as an intentional stop of the VM, it may change the termination reason unless everything is conditioned on a state machine of some kind.

Also, another point is the usage of "Platform" and "Host". The terms are not that common, or I am not familiar with them.
"Owner", "Operator", "Command", "Hypervisor", I understand.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+period plus 30 seconds. The longer value allows guests with a long graceful
+shutdown period to finish without losing the platform intent.
+
+While an intent is pending, `PlatformRequestedShutdown` deliberately takes
+precedence over `GuestShutdown` or `HostShutdown`. Reaching
+`SignalShutdownVMI` means KubeVirt has already committed to terminating that
+VMI. In the usual deletion path the VMI has a `deletionTimestamp`, and that
+deletion cannot be cancelled. If the guest independently starts shutting down
+at the same time, the controlling lifecycle operation is still the KubeVirt
+termination.
```
EdDev, line 218: I feedback about this earlier, IMO this breaks what a resource "status" represents in Kubernetes ecosystem.
The status is suppose to express the actual state of the resource/object, it should not map between the intent and the actual and assume anything from it by itself.
In this context, that means that if a guest has gone through a graceful shutdown, that fact can be expressed as the status, nothing else. It is the owner or observer of the CR to decide how to interpret it, not the controllers that mange the CR.

veps/sig-compute/362-guest-termination-state/vep.md
```diff
@@ -0,0 +1,555 @@
...
+### Reliable delivery and lifecycle scope
+
+The condition is meant to survive the transient source event, so a supported
+termination event cannot exist only in the bounded, best-effort libvirt event
+queue.
+
+virt-launcher latches termination-classifying lifecycle data in synchronized
+per-launcher state before attempting to enqueue the generic notification. It
+then signals the notification loop independently. Queue saturation may coalesce
+notifications, but it must not discard the retained termination reason.
```
EdDev, line 234: It is hard to understand what this actually means.
Do you mean that once the termination information is populated at the virt-launcher, it is considered immutable?

I also wander if the events the hypervisor emits are reliable, i.e. that it is assured not to miss one.
Conceptually, if we had this information from a hypervisor status/state query, it would be safer.

EdDev (commented):
Partial feedback, will continue.

0xFelix:
Thanks for putting this VEP together, @keinsword!

Given there are still several open questions regarding the design, we are not going to accept this for the 1.10 development cycle. However, we will leave this VEP open so it can be further refined for a future cycle.

To move this forward, we need to address a few main technical areas:

Use cases: These are still a bit unclear. Would it not be possible for platforms building on this to adopt KubeVirt's `RunStrategy` instead? This would let users choose reboot/shutdown behavior natively (possibly combined with `spec.rebootPolicy`).

RunStrategy compatibility: It is unclear how this should function with `RunStrategy` in general. It seems it only works meaningfully with `Manual`, because as soon as a VM is restarted, the Condition would be lost. This effectively mandates `Manual` just to offload lifecycle decisions to an external controller.

Event delivery: There are unresolved technical uncertainties about the event delivery mechanism and the actual source of the event. The design currently combines an intent with other operational behaviors that might not always hold true.

Feel free to tackle these points in the next cycle. Thanks again for your work on this!

PR #370, opened 2026-07-02
VEP144: Unlimited memlock limits, remove ReservedOverhead

bgartzi:
### VEP Metadata

**SIG label**:  /sig compute

### What this PR does
This VEP proposes removing the `ReservedOverhead` field and `ReservedOverheadMemlock` feature gate (now in Alpha) and existing logic supporting SR-IOV, SEV and RT in favor of a global, default `'unlimited'` memlock limit for all virtqemud instances running in virt-launcher pods.

### DIFF
--- a/veps/sig-compute/144-memory-lock-rlimits/vep.md
+++ b/veps/sig-compute/144-memory-lock-rlimits/vep.md
@@ -4,8 +4,8 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.8
-- This VEP targets beta for version: v1.9
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
 - This VEP targets GA for version:
 
 ### Release Signoff Checklist
@@ -13,25 +13,23 @@
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
 - [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
-- [x] (R) Alpha target version is explicitly mentioned and approved
-- [x] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
 - [ ] (R) GA target version is explicitly mentioned and approved
 
 ## Overview
 
-This enhancement introduces a new field to the `VirtualMachine` spec to
-specify memory lock RLimit configuration requirements. The new field
-will be placed into `spec.domain.memory.reservedOverhead`. It will
-contain a structure holding an enum and a kubernetes
-`resource.Quantity`. The enum will indicate if the memory lock RLimit
-needs to be updated, and the `Quantity` will indicate if there's some
-extra memory apart from the VM's that could be locked.
+This enhancement proposes a new VMI field to set unlimited memlock
+limits to the virtqemud processes for those instances that request it.
+This allows running VMIs that require non-default memory lock limits
+without extending the existing list of conditions under which the limit
+should be modified, no matter the logic supporting a device or workload
+lives inside the core KubeVirt source or it is provided via a plugin.
 
-Those values can be updated by other external mechanisms, such as
-mutating admission webhooks, based on other VMI spec configurations.
-Those values will be considered by the `virt-handler` to appropriately
-configure the memlock RLimits of the `virt-launcher` pod and qemu
-process.
+Currently, the unlimited memlock limit is applied to the virtqemud
+process as long as the virtual machine is SEV, RT or is attached a VFIO
+device. In other words, this VEP does not want to change the current
+limit, but let arbitrary virtual machines to opt-in.
 
 ## Motivation
 
@@ -42,24 +40,30 @@ that a process can lock is limited by the environment's memory lock
 RLimit. Libvirt updates environment memory lock limits if they are below
 expectations (see [source code][libvirt-vdpa-memlock-ratio]). Those
 expectations depend on the guest's memory size, the number of devices
-that require locking memory, and the size of other memory regions that
-could be also locked (such as the IOMMU region).
+that require locking memory, the size of other memory regions that
+could be also locked (such as the IOMMU region) and whether iommufd is
+used or not.
 
-Currently, KubeVirt (`virt-handler`) adjusts the memory lock limits of
-the `virt-launcher` under a specific set of
+Currently, KubeVirt (`virt-handler`) sets the memory lock limits of the
+`virt-launcher` to `'unlimited'` under a specific set of
 [conditions][kubevirt-memlock-limit-conditions]: When SRIOV/VFIO devices
-are attached to, or it is a real time or SEV VM.
+are attached to, or it is a real time or SEV VM. However, these are not
+the only cases in which a virtual machine might need to lock memory.
+Moreover, implementations outside of the core KubeVirt source code, such
+as hook sidecars or network binding plugins, do not have a way to make
+sure VMs won't fail to boot due to memory lock limits not being set
+appropriately.
 
-However, the cases in which the VM might need to lock memory are not
-limited to those listed above. In other words, this mechanism might need
-to be extended in the future. Moreover, it might be that the devices
-requiring to lock VM memory are implemented outside the core kubevirt
-source code, such as network binding plugins.
+A mechanism to allow running virtual machines with non-trivial memory
+lock RLimit requirements is needed.
 
-A more general mechanism that applies memory lock RLimits of the
-`virt-launcher` based on requirements is needed. Letting users and other
-components, such as admission webhooks, to specify those requirements
-will help to make the existing mechanism adapt to other use-cases.
+Initially, the approach consisted of an API field, `ReservedOverhead`,
+that was released under a feature gate in Alpha stage in release v1.8.
+This enhancement tracked the development of such feature. This allowed
+other components, such as admission webhooks, specify those
+requirements without affecting internal KubeVirt source code. Its tight
+coupling to pod-sizing was the reason why the original design was pushed
+back in favor of this, proposed in this renewed VEP.
 
 Currently, some existing mechanisms that are related to this enhancement
 exist, but each with their own limitations:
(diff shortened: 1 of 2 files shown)


bgartzi:
Tagging people involved in the latest discussions or similar proposals for awareness: @iholder101, @lyarwood, @Barakmor1, @michalskrivanek

iholder101:
Thanks @bgartzi very much for this VEP update!

 IIUC this PR basically reverts this VEP in favor of the solution in https://github.com/kubevirt/kubevirt/pull/17857, correct? If so, maybe we can just add a section in the beginning explaining the context - something like "After discussions in X, we've decided to go with approach Y instead" - and frame this as a deprecation/revert pointing to the new direction.

This way we keep the original thoughts here for historical context, while making it clear the VEP is no longer active. And we can also mention the revert plan.

WDYT? would that be cleaner?

bgartzi:
@iholder101 the problem with https://github.com/kubevirt/kubevirt/pull/17857 is that the implementation still conserves the conditions that trigger (or not) the memlock rlimit update. This VEP update wanted to remove the `reservedOverhead` API field. Without that, we can't implement `util.RequiresLockingMemory` that https://github.com/kubevirt/kubevirt/pull/17857 relies on. We would be back at the stage in which VMIs that are not VFIO, SEV or RT do not deserve to lock memory, i.e. what originally VEP144 wanted to change.

Some parties might prefer having a field to explicitly declare VMIs that require to lock memory, as setting `unlimited` memlock rlimits for every single VMI seem to be a concern for them: https://github.com/kubevirt/kubevirt/pull/17857/changes#r3506576313. I already asked them to have a look into this VEP and share their point of view.

Some other parties think that we don't need an API field to let VMs lock memory as the last few comments in #144 suggest.

Tagging people involved in the discussions and similar efforts to try to converge: @lyarwood, @fanzhangio, @harshitgupta1337, @Barakmor1, @michalskrivanek, @vladikr

bgartzi:
It was discussed on Wednesday's SIG compute meeting that a possible agreement between both sides could be simplifying the API to a unique field that would let vmis toggle whether they require to lock memory or not.

veps/sig-compute/144-memory-lock-rlimits/vep.md
```diff
@@ -4,34 +4,41 @@
...
+virt-launcher's physical memory usage resource limits.  Memory pressure
+attacks that malicious devices or qemu implementations could exploit are
+avoided given the kubelet will enforce a maximum physical memory usage
+limit on the virt-launcher pod, which encapsulates the running
+virtqemud/qemu process. So even if unlimited virtual memory lock
+accounting will be permitted, allowing VMs with theoretically unlimited
+VFIO devices attached to it, physical memory reservation will be still
+limited by virt-launcher's memory resource limits, which are set based
+on the guest's memory size and an overhead that KubeVirt calculates
+through each hypervisor's `GetMemoryOverhead` implementation.
```
xpivarc, line 41: Not all VMIs run with memory resource limits.

xpivarc (commented):
Do we force virtqemud to lock the memory or do we only allow it to do it here?

veps/sig-compute/144-memory-lock-rlimits/vep.md
bgartzi: Ack. Then having an VMI API field to enable/disable this behaviour sounds like be a better choice, doesn't it?

bgartzi:
@xpivarc:
> Do we force virtqemud to lock the memory or do we only allow it to do it here?

We just allow it to do it.

bgartzi:
I rephrased the VEP so it does not default to `'unlimited'` for all VMIs.

Instead, it proposes to remove `spec.domain.memory.reservedOverhead` in favour of a new API field `spec.domain.memory.memLock`. This will be a `Required`/`NotRequired` enum which will act as another valid condition that triggers the mechanism setting the virt-launcher pod's memlock limits to `unlimited` as it is currently done for SR-IOV, SEV and RT machines.

I'm proposing v1.10 to release this feature as Alpha, under a feature gate.

veps/sig-compute/144-memory-lock-rlimits/vep.md
```diff
@@ -4,34 +4,41 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.8
-- This VEP targets beta for version: v1.9
-- This VEP targets GA for version:
+- This VEP targets alpha for version:
+- This VEP targets beta for version:
+- This VEP targets GA for version: v1.10
```
iholder101: According to the [comment here](https://github.com/kubevirt/enhancements/pull/370#issuecomment-5266610968), I think this needs to be modified?

iholder101:
> I rephrased the VEP so it does not default to `'unlimited'` for all VMIs.
> 
> Instead, it proposes to remove `spec.domain.memory.reservedOverhead` in favour of a new API field `spec.domain.memory.memLock`. 

Maybe I'm missing something, but I think the current state of this PR doesn't reflect that.

> This will be a `Required`/`NotRequired` enum which will act as another valid condition that triggers the mechanism setting the virt-launcher pod's memlock limits to `unlimited` as it is currently done for SR-IOV, SEV and RT machines.

I'd prefer if we could use an empty struct pointer vs. nil pointer instead of the `Required`/`NotRequired` string enum. This leaves room for further extension in the future if it's needed.

----

In addition, in today's [VEP discussion meeting](https://calendar.google.com/calendar/u/0/embed?src=kubevirt@cncf.io), @xpivarc suggested to have no API field at all and let the plugins set the unlimited `prlimit` instead.

This approach is reasonable, and perhaps we should list it as an alternative in the VEP. That being said, it would mean that the `prlimit` would need to be set by a privileged DaemonSet that's called via a NodeHook. It means the plugin would need to re-implement stuff like PID lookups, although it's not that difficult to do and can be assisted by the [Plugin SDK](https://github.com/kubevirt/plugins).

I can envision having a first thin plugin for setting up an ulimited `prlimit`, then having more plugins depend on this plugin (we want to introduce plugin dependencies in the future).

That being said, virt-handler is already doing all of that, and this small API addition will just tell it to enable the same mechanism.

Thoughts? @xpivarc @bgartzi @vladikr

bgartzi:
> > I rephrased the VEP so it does not default to `'unlimited'` for all VMIs.
> > Instead, it proposes to remove `spec.domain.memory.reservedOverhead` in favour of a new API field `spec.domain.memory.memLock`.
> 
> Maybe I'm missing something, but I think the current state of this PR doesn't reflect that.
> 

What you were missing was the set of changes that were staged but not committed on my local fork...

Sorry for that.

> > This will be a `Required`/`NotRequired` enum which will act as another valid condition that triggers the mechanism setting the virt-launcher pod's memlock limits to `unlimited` as it is currently done for SR-IOV, SEV and RT machines.
> 
> I'd prefer if we could use an empty struct pointer vs. nil pointer instead of the `Required`/`NotRequired` string enum. This leaves room for further extension in the future if it's needed.
> 
> In addition, in today's [VEP discussion meeting](https://calendar.google.com/calendar/u/0/embed?src=kubevirt@cncf.io), @xpivarc suggested to have no API field at all and let the plugins set the unlimited `prlimit` instead.

Was about the plugin setting the prlimit, or the hook-sidecar requesting memlock (via internal API) to the virt-launcher? i.e. similar to the current design but without adding a new vmi spec field. This idea is covered under the `Sidecar Hook for Memory Lock configuration` section among the proposed alternatives. Now, I think I might have missed the point of the discussion yesterday.

> 
> This approach is reasonable, and perhaps we should list it as an alternative in the VEP. That being said, it would mean that the `prlimit` would need to be set by a privileged DaemonSet that's called via a NodeHook. It means the plugin would need to re-implement stuff like PID lookups, although it's not that difficult to do and can be assisted by the [Plugin SDK](https://github.com/kubevirt/plugins).

And that implicitly whoever deploying the plugin would need enough permissions so as to put a privileged daemonset.

Would also be another thing to consider that the plugin should have some sort of mechanism to notice virt-launchers fast enough so that VMs are not started without setting the limit first? I'm specially concerned about live-migration. Can't remember completely, but in that case live migration will just fail if the destination libvirt can't start, right? Is there some sort of retry in that case that I can't remember?

> 
> I can envision having a first thin plugin for setting up an ulimited `prlimit`, then having more plugins depend on this plugin (we want to introduce plugin dependencies in the future).
> 
> That being said, virt-handler is already doing all of that, and this small API addition will just tell it to enable the same mechanism.
> 
> Thoughts? @xpivarc @bgartzi @vladikr

bgartzi:
Another push to update the required/notrequired enum to an empty struct.

I will update the alternatives once I have a better understanding of it.

michalskrivanek:
@xpivarc regarding the "just always set unlimited case" and memory locking.
- with any sort of limits the cgroup protects it
- without limits there's no cgroup limit. but that's not any different from not having another protection preventing the pod from allocating too much and cause OOM

regardless, libvirt sets the limit using the same (well, we copied it) logic.
I do not see any risk or any difference in useing `unlimited` always

bgartzi:
During today's sig-compute meeting, discussions tended back into an API-less design to tackle this problem. Main concern was making the user aware of this without really needing to and giving users a toggle they could exploit to apply memory pressure.

Different solutions have been discussed:
1- Setting `'unlimited'` memlock rlimits to ALL virtqemud processes no matter what. This could have implications on not being able to swap memory when needed to.
2- Full plug-in implementation (completely out of tree). If I understand correctly, it would be a fully independent daemonset, running on all nodes with privileges. This would replicate all the logic that virt-hanlder already implements to find virtqemud of different virt-launcher to decide when to set these limits.
3- Other ways to make the virt-handler aware of the needs. One of them being one of the alternatives covered in the VEP already (see https://github.com/bgartzi/enhancements/blob/342ac99ee8a18afc512cd9bcded3f4d5c28f03c5/veps/sig-compute/144-memory-lock-rlimits/vep.md#sidecar-hook-for-memory-lock-configuration). In really simple terms, this would involve extending the virt-launcher/virt-handler API and the sidecar hook API.  The virt-handler would request virt-launcher if locking limits is needed, and the virt-launcher would ask the sidecar hook containers (if any) if there is need to (if the API version is the one that supports it).

My points and questions on each of them:
1- The simplest, of course. We wouldn't be able to swap memory on high memory load conditions. Only, if libvirt/qemu decide that it is a good idea to lock memory. If they don't decide, which currently do, we are pretty much in the same stage we are now. KubeVirt already puts a big amount of trust onto libvirt/qemu. Is there any reason for which we would not trust this specific case? Another concern was that VMs could apply memory pressure by locking all the memory they could. But right now they could also start consuming as much memory as they could, without any need to lock it, if memory limits are not set, which as discussed, is the majority of cases.
2- I'm concerned about this approach. Specially for live migration. There was already one bug related to memlock limits not being updated in the target virt-launcher in time (see https://github.com/kubevirt/kubevirt/commit/44c41fae4bb1e79ea72c54a53666a63cf400ad44). If there is not a mechanism that blocks live migration in the virt-handler until the plugin has done its thing, then we are up to hit this again. Which makes it not fully out-of-tree for kubevirt.
3- Only affects those cases in which hook-sidecar-containers are running in the virt-launcher. Solves the problem. My question here is, would users still be able to exploit this through a hook-sidecar they implement, with that new endpoint, that defaults to setting unlimited memlock rlimits? Again as in 1, how? If virtqemud limits are the ones being updated and that is something we trust, for now.

Tagging people involved in today's discussion: @iholder101 @vladikr @EdDev @xpivarc @michalskrivanek

veps/sig-compute/144-memory-lock-rlimits/vep.md
xpivarc: Actually I think otherwise, the lack of protection make it more dangerous.

bgartzi: This paragraph was rephrased so as to state that, setting memory limits was key:

> This approach is protected against memory pressure attacks that might be
exercised by malicious device implementation or guests by the launcher
pod's memory resource limits. Even if the memory lock limit is
`'unlimited'`, it accounts virtual memory locked. In case a malicious
workload tries locking more physical memory than the one expected by
KubeVirt and set into the launcher's memory resource limits, the kubelet
would bring it down. **Setting those limits is vital to be protected
against that vector, for example, explicitly declaring
`spec.domain.resources.limits.memory`.**

Of course, this was for the design in which an vmispec API field would toggle the limits on/off based on user input. Adding an API field for this is not an idea some people like.

> Actually I think otherwise, the lack of protection make it more dangerous.

The lack of protection exist right now with or without memlock. Without enforcing those limits any virt-launcher can run an out of memory attack, yet we trust the components kubevirt ships. I think we are talking about two related but different problems here.

bgartzi: ...which makes me think: if the memory limits are a concern to set unlimited memlock limits on virtqemud, how about setting unlimited memlock limits automatically **ONLY** for the existing use cases (sriov, sev rt...) **AND** those VMIs that have resource memory limits?

Then it's up to the user to find which method to set those is more convenient. e.g. a `ResourceQuota` on the namespace.

michalskrivanek: Hi @xpivarc, can you elaborate on the risk you're seeing? Compared to the current state, what new attack vector would it open? 
Also, IIUC with VFIO/SEV/AddedOverhead+memlock:required we already set `unlimited` to let hypervisor handle the locking anyway, no?

iholder101: > Hi @xpivarc, can you elaborate on the risk you're seeing? Compared to the current state, what new attack vector would it open? Also, IIUC with VFIO/SEV/AddedOverhead+memlock:required we already set `unlimited` to let hypervisor handle the locking anyway, no?

The added attack vector I can think of is that today an admin can disallow creating VMs that lock memory (with VFIO/SEV/etc) via a validating policy/webhook, if he really wants to be safe from VMs locking memory hence avoiding swap.

But if we go with the api-less approach, a malicious person can:
- Spawn a VM with no memory limits.
- Lock all memory on the node.

And the admin wouldn't have any mechanism to reject such VMs.

iholder101:
Following my [comment here](https://github.com/kubevirt/enhancements/pull/370#discussion_r3902531563), I guess that the main question is whether we want to avoid this attack vector or not.

If we don't, then setting an unconditional unlimited `prlimit` is the way to go.

If we do, then I tend to lean towards the plugin-based solution, at least as a first attempt, partially because other plugins using it is either way the main use-case for this VEP to begin with IIUC. Plus, we can always add the API in the future if we think it's necessary.

> If there is not a mechanism that blocks live migration in the virt-handler until the plugin has done its thing, then we are up to hit this again. Which makes it not fully out-of-tree for kubevirt.

I think the mechanism exists in the pre-migrate node hooks + failure policy.

veps/sig-compute/144-memory-lock-rlimits/vep.md
bgartzi: > But if we go with the api-less approach, a malicious person can:
> 
>    Spawn a VM with no memory limits.
>    Lock all memory on the node.

Not if we only set the unlimited memlock for those VMIs whose memory limits are defined. Right?

bgartzi:
@iholder101:
> I think the mechanism exists in the pre-migrate node hooks + failure policy.

I'm sorry, I don't get it. I thought pre-migration hooks were just about xml modifications: https://github.com/kubevirt/enhancements/blob/575fea62437e0378fcd62d95da348b8a21224178/veps/sig-compute/141-target-xml-mutation-hooks/vep.md?plain=1#L96-L105

And hook servers run on the virt-launcher, which makes it hard if you want it to run with privileges. As it is the case.

If you were talking about https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/190-kubevirt-structured-plugins/vep.md#node-hook-points, more specifically, PreStart and PreMigrationTarget node hook points:

IIUC, those components that are hooked, still need to find which is the virt-launcher pod that is being started and apply whatever we want there? I believe that I'm missing some context from the vep. Right now I feel it would re-implementing a big part of the virt-handler from scratch on each of these.

> If we do, then I tend to lean towards the plugin-based solution, at least as a first attempt

But the hooks that would be required for this to be proved are not even considered in the timeline of that VEP, so I am uncertain about when those first attempts could start. At least in 8 months (2 release cycles) IIUC?

veps/sig-compute/144-memory-lock-rlimits/vep.md
michalskrivanek: @iholder101 besides what @bgartzi says which I think solves the problem, I can't imagine anyone really doing that, it sounds way too hypothetical to me. 
You can only mlock the memory you have so admin can cap on the guest request size. So if admin cares about not exceeding memory on mode they should either enforce limits or count requests as limits.

iholder101: > Not if we only set the unlimited memlock for those VMIs whose memory limits are defined. Right?

That's correct.
That being said, limits are extremely dangerous and unrecommended for many reasons (i.e. unpredictable spikes in QEMU's memory could crash the VM), and we need to be careful from encouraging admins/VM owners to set limits unless strictly necessary.

> I can't imagine anyone really doing that, it sounds way too hypothetical to me.

I don't disagree :)
I think it's extremely valid to consider to just ignore this attack vector as it's indeed pretty hypothetical.

What do you think @vladikr @xpivarc ?

> if admin cares about not exceeding memory on mode they should either enforce limits

I really don't want admins to enforce limits unless strictly necessary as said above.

iholder101: To clarify my position - I don't object for an unconditionally unlimited `prlimit`. I agree that the attack vector is fairly hypothetical and I'm not sure how strict we should be here in practice.

bgartzi:
There was a misunderstanding here. Node hooks are already available in the latest kubevirt. More specifically `PreVMStart` and `PreMigrationTarget`. I will try to solve this problem via a node hook plugin.

I'm closing this PR, at least while I make sure that the hook plugin implementation fixes it.

I'm not closing the VEP tracker #144 yet. Once I make sure the plugin works, I will need to remove the existing Alpha API.

michalskrivanek:
Why do tou need to do that instead of the api-less "set it always unlimited"? Was there any further discussion? I still do not see the theoretical "attack" as real

bgartzi:
It is an idea that showed to be hard to push into a consensus. One way, or another, apifull or apiless, there would always be some concern about it.

There is supposed to be an alternative, out-of-kubevirt-core, way of doing this via a plugin. I thought the hooks needed for that were not implemented yet, which happened not to be true. I will try that and see if it is actually possible, even if that might require rewriting some of the virt-handler's internal logic.

PR #372, opened 2026-07-06
VEP #373: Structured Operation Logging 

sradco:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/373
**SIG label**: /sig observability

### What this PR does

Introduces a new VEP under `veps/sig-observability/373-structured-operation-logging/` proposing a reusable mechanism and field taxonomy for structured, machine-queryable KubeVirt operation lifecycle logs.

VM operations are the first domain adopting the mechanism, with **migration** serving as the initial Alpha vertical slice.

Key aspects:

- **Canonical structured lifecycle records**: Alpha emits feature-gated records for migration lifecycle transitions such as `started`, `completed`, and `failed`. Existing KubeVirt log lines and messages remain unchanged.
- **Reusable operation taxonomy**: Common fields such as `kubevirt.operation.type`, `kubevirt.operation.phase`, and conditional `kubevirt.operation.duration_ms`, with domain-specific context under `kubevirt.vmi.*`, `kubevirt.vm.*`, and `kubevirt.migration.*`.
- **Explicit operation and workload identity**:
  - existing `kind` / `name` / `namespace` / `uid` identify the `VirtualMachineInstanceMigration` operation object;
  - `kubevirt.vmi.*` identifies the affected VMI;
  - `kubevirt.vm.*` identifies the owning `VirtualMachine` when one exists.
- **Best-effort lifecycle logging**: Canonical records are attempted only after the corresponding VMIM status transition is successfully persisted. Logging is not transactional with the status update and does not claim exactly-once delivery.
- **OpenTelemetry alignment where semantics match**: Existing KubeVirt object-context fields remain unchanged where no directly applicable OTel semantic convention exists, while KubeVirt-specific operation attributes use the `kubevirt.*` namespace.
- **Audit/log separation**: Kubernetes audit logs remain the authoritative source for authenticated user/API-request attribution. This VEP does not propagate usernames into KubeVirt operation logs or Kubernetes Events.

This enables downstream consumers such as Loki queries, Perses dashboards, and VM operation-history UIs to reliably query operation lifecycle information without fragile message parsing.

### Special notes for reviewers

- Alpha target: v1.11
- Beta/GA targets: TBD
- Feature gate: `StructuredOperationLogging`
- Alpha scope: migration controller only
- Alpha lifecycle phases: `started`, `completed`, `failed`; `in_progress` is deferred
- No new CRDs, API fields, username annotations, or additional API writes
- Existing KubeVirt log lines remain unchanged; the feature adds separate canonical lifecycle records when enabled
- Kubernetes Events remain unchanged
- Downstream Jira tracking: [CNV-92178](https://redhat.atlassian.net/browse/CNV-92178)
- Related upstream direction:
  - Kubernetes structured logging [KEP-1602](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging)
  - Kubernetes contextual logging [KEP-3077](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging)

### Major change from the original proposal

Based on review feedback, username/initiator enrichment was removed from the VEP.

The proposal now has a clear separation of responsibility:

- **Kubernetes audit logs** answer who made the API request.
- **KubeVirt structured operation logs** describe what happened to the asynchronous operation afterward, including lifecycle phase, outcome, duration, and operation-specific context such as migration source and target nodes.


veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,244 @@
...
+
+Log severity follows OTel LogRecord conventions:
+- K8s Event `type: Normal` → SeverityText `INFO`, SeverityNumber 9
+- K8s Event `type: Warning` → SeverityText `WARN`, SeverityNumber 13
+
+### Username Propagation
+
+1. virt-api mutating admission webhook intercepts lifecycle-changing mutations (start, stop, restart, migrate, add/remove volume, snapshot)
+2. Webhook reads `admission.Request.UserInfo.Username`
+3. Webhook stamps annotation `kubevirt.io/last-modified-by: <username>` on the resource
```
avlitman: can users manually set this annotation? I mean the webhook always overwrites?

sradco: Good question. No. The mutating admission webhook always overwrites this annotation with the authenticated username from the admission request, so any user-supplied value gets replaced on the next lifecycle operation.

I'll also add a validating webhook rule to reject direct patches to this annotation, so it can only be set by the system. Will update the VEP to clarify this.

veps/sig-observability/NNNN-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,244 @@
...
+4. virt-controller/virt-handler reads the annotation when processing the operation
+5. Username is included in the structured log entry
+6. For system-initiated operations, username is the service account (e.g., `system:serviceaccount:kubevirt:kubevirt-controller`)
+
+### Contextual Logging
+
+Use `logr.Logger.WithValues()` to establish operation context at the top of each reconcile loop:
+
+```go
+logger := ctrl.Log.WithValues(
```
avlitman: this should be `logger := log.Log.With(` like we use in kubevirt

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,244 @@
...
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Introduce structured, machine-parseable logging for VM operation lifecycle events in KubeVirt components (virt-controller, virt-handler, virt-api). Each operation event (migration, lifecycle change, storage operation, snapshot, network hotplug) will be emitted as a JSON log entry with a well-defined field taxonomy. Additionally, the username of the user who initiated the operation will be propagated via a virt-api admission webhook annotation and included in the structured log output.
```
avlitman: which specific events map to each type (migration, lifecycle change, ...) maybe worth adding

veps/sig-observability/NNNN-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,244 @@
...
+- No new API calls or watchers — only adds fields to existing log output
+- Username annotation is a single annotation write per lifecycle mutation (already happening in admission webhook path)
+- Log volume increase is negligible (adding ~200 bytes per operation log line)
+- No new components deployed
+
+## Update/Rollback Compatibility
+
+- **Update**: Old logs without new fields will simply not have the structured fields. LogQL queries using these fields will return empty results for old entries. No breaking change.
+- **Rollback**: Removing the username annotation write is safe — controllers will log `user.name: ""` (empty). Downstream queries handle empty gracefully (`=~".*"`).
+- **Feature gate**: Consider gating the annotation write behind a feature gate (e.g., `StructuredOperationLogging`) for Alpha.
```
avlitman: the vep text should be definitive, not "consider"

sradco: Fixed

veps/sig-observability/NNNN-structured-operation-logging/vep.md
sradco: Fixed

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: Fixed

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+#### Annotation Integrity
+
+Users cannot manually set or spoof the `kubevirt.io/last-modified-by` annotation:
+
+- The **mutating admission webhook always overwrites** any existing value with the authenticated username from `admission.Request.UserInfo.Username` on every lifecycle-changing mutation.
+- A **validating admission webhook** rejects any direct writes to this annotation (e.g., via `kubectl annotate` or metadata patches) unless the request originates from the virt-api service account. This ensures the annotation can only be set by the system.
+
+### Contextual Logging
+
+Use `log.Log.With()` to establish operation context at the top of each reconcile loop:
```
iholder101, line 190: > each reconcile loop

Can we make a list of which reconcile loops we want to get this into?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+| `storage` | `AddVolume`, `RemoveVolume`, `VolumeReady`, `DataVolumeCreated`, `DataVolumeReady`, `DataVolumeFailed` |
+| `snapshot` | `SnapshotStarted`, `SnapshotSucceeded`, `SnapshotFailed`, `RestoreStarted`, `RestoreSucceeded` |
+| `network` | `InterfaceHotplug`, `InterfaceHotunplug`, `NetworkReady` |
+| `scheduling` | `SchedulingFailed`, `Scheduled`, `Evicted`, `NodeDrainStarted` |
+
+### Username Propagation
+
+1. virt-api mutating admission webhook intercepts lifecycle-changing mutations (start, stop, restart, migrate, add/remove volume, snapshot)
+2. Webhook reads `admission.Request.UserInfo.Username`
+3. Webhook stamps annotation `kubevirt.io/last-modified-by: <username>` on the resource
```
iholder101: The admission webhook covers direct VM/VMI spec edits, and subresource endpoints (start/stop/migrate/etc.) are in virt-api so they can stamp the annotation too - but each handler needs explicit instrumentation rather than a single webhook.

The harder cases:

- Indirect CRs (VMIM, VirtualMachineRestore) - user posts a separate object, VMI isn't directly mutated.
- Controller-initiated (eviction, node maintenance) - no user request at all.

How is the username captured for these?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
+# VEP #373: Structured Operation Logging with Username Enrichment
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets GA for version:
```
iholder101: Where's Beta? :)
Also what's the FG name?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+| `user.name` | [User](https://opentelemetry.io/docs/specs/semconv/registry/attributes/user/) | string | yes | Who initiated the operation (short name/login) |
+
+#### KubeVirt-Specific Attributes (new, `kubevirt.` prefix)
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `kubevirt.vm.name` | string | yes | Target VM name (may differ from `k8s.object.name` when the involved object is a Migration or DataVolume) |
+| `kubevirt.operation.type` | string | yes | Category (see [Operation Type Mapping](#operation-type-mapping) below): `migration`, `lifecycle`, `storage`, `snapshot`, `network`, `scheduling` |
+| `kubevirt.operation.phase` | string | yes | Lifecycle: `started`, `in_progress`, `completed`, `failed` |
+| `kubevirt.operation.duration_ms` | int | no | Duration in ms (only for `completed`/`failed` phase) |
```
iholder101: How is `duration_ms` tracked for operations that span multiple components? E.g., a migration goes through virt-controller (orchestration) -> virt-handler source (pre-copy) -> virt-handler target (receive). Which component owns the start time, and what happens if that component restarts mid-operation?

iholder101 (commented):
Thanks @sradco! Nice direction!
See comments below

veps/sig-observability/373-structured-operation-logging/vep.md
Aseeef: What if two different flows get triggered by two different users (e.g. migrate by A, hotplug by B)?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+- As a platform engineer, I want to build a Perses LogsTable dashboard that filters VM operations by namespace, operation type, and username without writing fragile regex.
+- As a security auditor, I want to see which user initiated a VM deletion or migration.
+- As an SRE, I want to filter operations by severity (Normal/Warning) and by source component to quickly find errors.
+
+## Repos
+
+- `kubevirt/kubevirt` — virt-controller, virt-handler, virt-api, virt-operator changes
+- `kubevirt/enhancements` — this VEP
+
+## Design
```
Aseeef, line 73: Is this VEP proposing adding structured logging for specific key points in KubeVirt or talking about creating a general purpose new logging framework for KubeVirt?

Because for a general purpose logging framework, I feel some of the taxonomy is too rigid. For example its hard to fit all the places we want to instrument into an operation type/phase.

Some are more generic like say a device plugin just became unhealthy. That doesn't neccesarily have to do anything with a specific VM. Or is this structured logging intended just for VMs?  I think this is worth clarifying.

veps/sig-observability/373-structured-operation-logging/vep.md
Aseeef: Talking of the feature gate, would this require extensive changes to how logging is done today? So if you have a log today like "Migration completed successfully", do intend to replace that or augment that? How exactly do you intend to feature gate this? Would it be something like (for example):

```
if fg.isEnabled()
  log.newLog
else
  log.oldLog
```

veps/sig-observability/373-structured-operation-logging/vep.md
Aseeef: +1

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+1. **Consistent field names** across all controllers for LogQL filtering
+2. **Username tracking** to know who initiated an operation
+3. **Operation lifecycle phases** (started, completed, failed) for duration calculation and in-progress detection
+4. **Machine-parseable format** that won't break between KubeVirt versions
+
+The Kubernetes project itself has adopted structured logging ([KEP-1602](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging)) and contextual logging ([KEP-3077](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging)). This VEP brings KubeVirt in line with that direction.
+
+## Goals
+
+- Define a standard field taxonomy for VM operation log entries
```
Aseeef: I think its worth clarifying how field taxonomy is enforced at compile time (e.g. structs, consts). And how this shared taxonomy will it be shared between components (virt-handler, virt-launcher, etc). For example, will it be part of the shared kubevirt api pkg?

I think this VEP should talk more about how you will ensure someone writing logs in virt-launcher won't accidentally report itself as virt-controller or other ways how you might prevent really obvious mistakes like this. Of course building a logger once and passing it around is part of the solution but I don't think its the whole solution. I think more needs to be said here (e.g., not having to construct an entire logger from scratch everytime and reusing some sort of helper/editing an existing reference?) For example, I think if it can be implemented cheap, reflection-based intrumentation might be really power here... Or maybe typed helpers with required fields? The point I'm (trying to) making here is we can do more to reduce mistakes and boilerplate code when logging.

veps/sig-observability/373-structured-operation-logging/vep.md
Aseeef: Edit: I see the stated purpose is to define a `field taxonomy for VM operation log entries` so if you consider everything else out of scope, its worth explicitly adding as a non-goal. And if you have any plans to extend this into something more general (which I think would be really more valuable) add an additional discussion about that.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Graduation Requirements
+
+### Alpha (v1.10)
+
+- [ ] Structured logging implemented in virt-controller, virt-handler, and virt-api for all operation types
+- [ ] Username annotation stamped by virt-api for all lifecycle mutations
+- [ ] Validating webhook rejects direct user writes to the annotation
+- [ ] Unit and integration tests passing
+- [ ] Documentation: field taxonomy documented in kubevirt.io
```
Aseeef, line 331: How this taxonomy will be documented. Manually? I fear if the current structure ends up being too rigid/requires regularly extending the existing taxonomy would be tedious. How do we make sure the docs don't become stale? Basically what I'm looking for is a single source of truth that the log field contract. Maybe there is some way to take advantage of existing auto gen infra?

Furthermore, IIRC, field taxonomy is essentially going to be API right? I really think more care needs to be given then to ensuring the same level of API regression gaurentees as our normal API. For example, I know with our "normal" API we have a unit test that does a wire-test to make sure backward compatability doesn't get broken.

Aseeef: Edit: I see the `Guaranteeing log schema stability as a formal API (best-effort stable, may evolve between minor versions)` as a listed non-goal. But IMO, this is going to raise red flags with a lot of people. The principle is, if you expose, document it, and people are relying on it, it is formal API.

Aseeef (commented):
Thank you for this VEP. Clearly a valuable addition :)!

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: The intent is a general framework (shared package, typed constants, contextual loggers) that any KubeVirt component can use, but VM operation lifecycle events are the first domain instrumented in Alpha. 
The kubevirt.operation.* taxonomy isn't imposed on all logs.
Future domains (device health, operator lifecycle) would add their own field extensions using the same framework.
I'll update the VEP to clarify this two-layer design and add a "Future Work" section for non-VM domains.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: Updated the VEP.
Feature gate name: `StructuredOperationLogging`. 
Beta target: v1.11, GA: v1.12.

The feature augments existing logs.
The human-readable msg stays unchanged.
When FG is on, additional kubevirt.* / k8s.* / user.* fields are attached via contextual logging. 
No if/else at each log site, the FG is checked once at the top of each reconcile to decide whether the logger gets enriched. 
When FG off = identical to today's output.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: I propose that the annotation is stamped on the triggering object, not on a shared VMI:

- VMIM creation → annotation on the VMIM
- VirtualMachineRestore → annotation on the Restore CR
- Controller-initiated (eviction/drain) → no annotation, logged as service account identity

This also answers @Aseeef's concurrency question.
Each operation reads username from its own triggering object, so concurrent flows by different users don't conflict.

For phased delivery: 
Alpha covers direct VM/VMI mutations + VMIM.
Snapshot/Restore moves to Beta.

BTW, This follows the same "capture identity at ingress, propagate to execution" pattern used by AWS CloudTrail and GCP Audit Logs.

Updated the VEP with a per-object stamping rules table and phased rollout.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: `duration_ms` is derived from CR status timestamps (e.g., migrationState.startTimestamp / endTimestamp), not in-memory timers. 

The component emitting the completed/failed log computes end - start from the persisted status.
This is restart-safe and requires no cross-component coordination.
If a component restarts, the timestamps are still in the CR. 
For operations without persisted timestamps, the field is omitted (it's already optional).

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: Field keys will be typed constants in a shared package (pkg/log/structuredlog/), Typos become compile errors. Component identity is set once at binary startup (not per-call-site), so virt-launcher physically cannot misreport as virt-controller. 

A typed builder with required fields ensures you can't forget mandatory fields. 
I added the details in the updated "Implementation Safety" section.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: +1.
I reworked the non-goal and added a "Schema Stability" section with graduated commitments: 
Alpha allows renames with release notes, 
Beta requires deprecation notice before removal, 
GA freezes field names with the same guarantees as KubeVirt API types. 

The source of truth is the Go constants in pkg/log/structuredlog/ with a unit test that catches accidental renames (similar to API wire-format tests).

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: I added an "Instrumented Reconcile Loops (Alpha)" table to the VEP: 
VM controller, Migration controller, DataVolume controller, Snapshot controller, Network controller (all in virt-controller), VMI reconcile loop (virt-handler), and admission webhooks (virt-api for username stamping).

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: +1. I addressed this in the `Implementation Safety` section.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,346 @@
...
+| Field | OTel Source | Type | Required | Description |
+|-------|-------------|------|----------|-------------|
+| `k8s.namespace.name` | [K8s Resource](https://opentelemetry.io/docs/specs/semconv/resource/k8s) | string | yes | Target VM namespace |
+| `k8s.object.kind` | [K8s Resource](https://opentelemetry.io/docs/specs/semconv/resource/k8s) | string | yes | K8s resource kind being operated on |
+| `k8s.object.name` | [K8s Resource](https://opentelemetry.io/docs/specs/semconv/resource/k8s) | string | yes | Specific resource instance name |
+| `k8s.event.reason` | [K8s Events](https://opentelemetry.io/docs/specs/semconv/registry/attributes/k8s/) | string | yes | Event reason string (e.g., `SuccessfulMigration`) |
+| `k8s.event.reporter.name` | [K8s Events](https://opentelemetry.io/docs/specs/semconv/registry/attributes/k8s/) | string | yes | Source controller: `virt-controller`, `virt-handler` |
+| `user.name` | [User](https://opentelemetry.io/docs/specs/semconv/registry/attributes/user/) | string | yes | Who initiated the operation (short name/login) |
+
+#### KubeVirt-Specific Attributes (new, `kubevirt.` prefix)
```
enp0s3, line 119: @sradco What is the policy of updating the `operations` attribute after GA (if I have a new controller with new operation types and its own phases)? Is this VEP aims to define the general policy of field taxonomy stability and deprecation procedures?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,346 @@
...
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `kubevirt.vm.name` | string | yes | Target VM name (may differ from `k8s.object.name` when the involved object is a Migration or DataVolume) |
+| `kubevirt.operation.type` | string | yes | Category (see [Operation Type Mapping](#operation-type-mapping) below): `migration`, `lifecycle`, `storage`, `snapshot`, `network`, `scheduling` |
+| `kubevirt.operation.phase` | string | yes | Lifecycle: `started`, `in_progress`, `completed`, `failed` |
```
enp0s3, line 125: @sradco Will there be any relationship between the `phase` typed constants and the API types? I can tell that not all the CRDs have phase in their statuses.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,346 @@
...
+
+Each domain would define its own field taxonomy using the shared package and contextual logger patterns. Extension to new domains does not require a new VEP — contributors add domain-specific constants to the shared package and instrument the relevant reconcile loops.
+
+## Implementation History
+
+- 2026-07: VEP created
+
+## Graduation Requirements
+
+### Alpha (v1.10)
```
enp0s3, line 325: @sradco It seems a lot of milestones for one release. For example, in order to propagate the contextual logger via `ctx` I think you will need to refactor some pieces of the current codebase.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,346 @@
...
+Because the annotation lives on the triggering object (not on a shared VMI), concurrent operations by different users do not conflict. For example: user A creates a VMIM (migration) while user B hotplugs a volume on the same VM — each operation reads its username from its own triggering object independently.
+
+#### Annotation Integrity
+
+Users cannot manually set or spoof the `kubevirt.io/last-modified-by` annotation:
+
+- The **mutating admission webhook always overwrites** any existing value with the authenticated username from `admission.Request.UserInfo.Username` on every lifecycle-changing mutation.
+- A **validating admission webhook** rejects any direct writes to this annotation (e.g., via `kubectl annotate` or metadata patches) unless the request originates from the virt-api service account. This ensures the annotation can only be set by the system.
+
+### Implementation Safety
```
enp0s3, line 184: @sradco Maybe we can utilize the linter in order to validate that developers are using the typed Go constants from the shared package, and not string literals for the key names.

enp0s3 (commented):
@sradco Thank you for the VEP! great initiative. I have a few questions below

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: Yes, this VEP defines the general stability/deprecation policy for the taxonomy.

After GA, field names are frozen, but additive evolution stays allowed: new kubevirt.operation.type values, optional fields, and new domain prefixes (e.g. kubevirt.device.*) can be added via typed constants + release notes without a new VEP. 

Renames/removals require the Beta+/GA deprecation path. 
I added an explicit table under Schema Stability.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: No coupling to CRD `.status.phase`. 
`kubevirt.operation.phase` is a logging-domain enum (started / in_progress / completed / failed) for consumers. 
Controllers map their own status/events onto these at emit time. 
CRDs without a status phase are fine, nothing in this VEP requires adding one. 
Clarified in a new "Operation Phase vs API Phase" section.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: +1. 
Alpha was too broad.
I narrowed it to a vertical slice: shared package + FG + migration controller instrumentation + VMIM username stamping/validation.
Alpha uses `log.Log.With()` and passes the logger explicitly.
Full `context.Context` propagation is optional Beta work, not a blocker. 
Broader controller coverage moves to Beta.

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: +1. 
Added to Implementation Safety and Alpha graduation:
a `golangci-lint` analyzer / `go vet` pass that fails CI if taxonomy key string literals are used outside `pkg/log/structuredlog`.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### KubeVirt-Specific Attributes (new, `kubevirt.` prefix)
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `kubevirt.vm.name` | string | yes | Target VM name (may differ from `k8s.object.name` when the involved object is a Migration or DataVolume) |
+| `kubevirt.operation.type` | string | yes | Category (see [Operation Type Mapping](#operation-type-mapping) below): `migration`, `lifecycle`, `storage`, `snapshot`, `network`, `scheduling`. New values may be added after GA; see [Schema Stability](#schema-stability). |
+| `kubevirt.operation.phase` | string | yes | Logging-domain lifecycle: `started`, `in_progress`, `completed`, `failed`. Independent of CRD `.status.phase` (see [Operation Phase vs API Phase](#operation-phase-vs-api-phase)). |
+| `kubevirt.operation.duration_ms` | int | no | Duration in ms (only for `completed`/`failed` phase). Derived from CR status timestamps (e.g., `migrationState.startTimestamp`/`endTimestamp`), not in-memory timers — restart-safe and requires no cross-component coordination. Omitted when no persisted timestamps are available. |
```
Aseeef: I'm confused when exactly this field is filled, and what is the benefit of having this? Is the timestamp logs already have not good enough for your purpose? Why not?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
...
+
+#### Severity Mapping
+
+Log severity follows OTel LogRecord conventions:
+- K8s Event `type: Normal` → SeverityText `INFO`, SeverityNumber 9
+- K8s Event `type: Warning` → SeverityText `WARN`, SeverityNumber 13
+
+#### Operation Type Mapping
+
+Each `kubevirt.operation.type` value maps to specific KubeVirt events/operations:
```
Aseeef: `kubevirt.operation.*` currently only has VM ops. I know you expressed interest in expanding this later into a more general framework. So maybe we could modify this path a bit to be more explicit that this is for VMs? Maybe `kubevirt.vm.operation.type`?

I don't understand oTel conventions so maybe this doesn't make sense.

veps/sig-observability/373-structured-operation-logging/vep.md
Aseeef: Given we would consider this taxonomy as API, shouldn't the same rules about updates the API requiring VEPs also apply here?

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
...
+The annotation `kubevirt.io/last-modified-by` is stamped on the **triggering object** — the resource whose creation or mutation initiates the operation. This avoids conflicts when concurrent operations are triggered by different users on the same VM.
+
+#### Per-Object Stamping Rules
+
+| Trigger | Annotated Object | How | Phase |
+|---------|-----------------|-----|-------|
+| Migration | VirtualMachineInstanceMigration (VMIM) | virt-api mutating admission webhook on VMIM creation | Alpha |
+| Direct VM/VMI mutation (start, stop, restart, pause) | VM or VMI | virt-api mutating admission webhook on the subresource endpoint | Beta |
+| Volume hotplug | VM | virt-api subresource handler stamps annotation on VM | Beta |
+| Controller-initiated (eviction, node drain) | — (no annotation) | Logged as service account identity: `system:serviceaccount:kubevirt:kubevirt-controller` | Beta |
```
Aseeef: But the node drains can be user triggered. So this sounds like a way to be able to bypass the audit trail.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
...
+| Snapshot / Restore | VirtualMachineSnapshot / VirtualMachineRestore | virt-api mutating admission webhook on CR creation | Beta |
+
+#### Flow
+
+1. virt-api mutating admission webhook intercepts the request
+2. Webhook reads `admission.Request.UserInfo.Username`
+3. Webhook stamps annotation `kubevirt.io/last-modified-by: <username>` on the **triggering object**
+4. virt-controller/virt-handler reads the annotation from the triggering object when processing the operation
```
Aseeef: How would virt-controller/virt-handler determine what the triggering object is? I don't think it is always obvious. For example, in the case of migrations, if the user does something to indirectly trigger the creation of a VMIM?

The example I had in mind was triggering a node drain. But I asked Claude if there was another concrete example for this and one example it said is:
virtctl migrate / migrate subresource -> virt-api creates the VMIM -> virt-api SA, not the caller

I haven't verified this information myself.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
...
+- [ ] Shared `pkg/log/structuredlog` package with typed field/enum constants and operation logger builder
+- [ ] Feature gate `StructuredOperationLogging` checked once at reconcile/handler entry (augment, do not replace `msg`)
+- [ ] Structured fields emitted for **migration** operations in virt-controller Migration controller
+- [ ] Username annotation stamped by virt-api on VMIM creation; validating webhook rejects direct user writes
+- [ ] Static check / linter forbidding string-literal taxonomy keys outside the shared package
+- [ ] Unit + integration tests for the migration path (fields present, username propagated, FG off = no new fields)
+- [ ] Field taxonomy documented (generated from or linked to the Go constants)
+
+### Beta (v1.11)
+
```
Aseeef, line 453: Consider doing two alpha releases for this. Because IIRC, Beta is now supposed to be "probably stable" especially since Beta is enabled by default now. So moving this to beta in the same release you expand to all other KubeVirt components might be risky.

On the contrary, it can also be argued that behavioral risks are more limited since this VEP only impacts logging.

veps/sig-observability/373-structured-operation-logging/vep.md
```diff
@@ -0,0 +1,390 @@
+# VEP #373: Structured Operation Logging with Username Enrichment
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version: v1.11
```
Aseeef: I remember being told in the past to not fill beta/GA until the release before. Meaning you are only allowed to target the next release. @enp0s3 is this true or did I misunderstand?

Aseeef (commented):
Overall closer to convergence, but I still foresee challenges with the last-modified-by audit trail. The “what if someone modifies the object right after” concern is still there for paths that stamp a shared object (and any later overwrite of the annotation before controllers read it).

Also, I think more is needed justifying why username enrichment is needed / what the usecase is, especially given Kubernetes’ existing audit capabilities. The current user stories don’t sufficiently address this (e.g. you can already see which user deleted a VM from audit of the DELETE request).

veps/sig-observability/373-structured-operation-logging/vep.md
sradco: `kubevirt.operation.duration_ms` is populated only on terminal records (completed / failed) when reliable persisted operation timestamps are available.

Log timestamps reflect when records were emitted, which may be later than the actual state transition due to reconciliation/retries. The explicit duration represents the operation itself and also avoids requiring consumers to correlate two best-effort log records. If it cannot be computed reliably, the field is omitted.

sradco:
Hi @Aseeef , I addressed the issues you raised in thes VEP. I narrowed its scope.
Ill be happy if you can have another look at it. Thank you.

Issue #376, opened 2026-07-07
VEP 376: Standardize Generation Tracking for VM and VMI


EdDev:
**Primary contact (assignee)**:

/assign EdDev

**Current Feature Stage**:

New

**Feature Gate**:

VMIInSyncCondition

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

VEP PR: https://github.com/kubevirt/enhancements/pull/377

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

PR #377, opened 2026-07-07
VEP #376: Standardize Generation Tracking for VM and VMI

EdDev:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/376
**SIG label**: /sig compute

### What this PR does

Introduces standard Kubernetes generation tracking for VM and VMI resources.

VM controllers track spec changes via `metadata.generation`, but KubeVirt's
existing fields carry non-standard semantics and neither resource's conditions
expose per-condition staleness information. This VEP closes those gaps by:

- Embedding `metav1.Condition` into `VirtualMachineCondition` and
  `VirtualMachineInstanceCondition`, adding `ObservedGeneration` to every
  condition so consumers can distinguish stale from fresh conditions.
- Adding a `VMIInSync` condition on `VirtualMachine` as a reliable,
  documented signal for whether the VM controller has processed the current
  VM generation against the running VMI.

The VEP includes an Open Questions section documenting a known gap: there is
no generation-tracking signal for whether virt-handler has applied the current
VMI spec to the running domain. The structural blockers (spec-only generation
tracking, VMI status write race between the VMI controller and virt-handler)
are explained there; a follow-up VEP will address this once the VMI status
ownership model is clarified.

### DIFF
--- a/veps/sig-compute/376-generation-tracking-standardization/vep.md
+++ b/veps/sig-compute/376-generation-tracking-standardization/vep.md
@@ -1,0 +1,538 @@
+# VEP #376: Standardize Generation Tracking for VM and VMI
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR): https://github.com/kubevirt/enhancements/issues/376
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Kubernetes defines a standard pattern for tracking controller sync status:
+`metadata.generation` (incremented by the API server on every spec change) combined with
+`status.observedGeneration` (written by the controller to record which generation it last
+processed). [KEP-1623](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions)
+extends this to the condition level via `metav1.Condition.ObservedGeneration`, allowing
+consumers to know precisely which conditions are stale vs. fresh with respect to the
+current spec.
+
+KubeVirt's `VirtualMachine` (VM) and `VirtualMachineInstance` (VMI) resources deviate
+from this standard in several ways: the VM's existing `status.observedGeneration` field
+carries semantics that differ from what Kubernetes consumers expect; and neither
+resource's conditions expose an `ObservedGeneration` field. This VEP proposes closing
+these gaps in a backward-compatible manner. The absence of a virt-handler processing
+signal on VMI is a known gap addressed separately (see Open Questions).
+
+## Motivation
+
+### VM `status.observedGeneration` carries non-standard semantics
+
+The field `vm.status.observedGeneration` does **not** mean what Kubernetes consumers
+expect. Per the Kubernetes convention, `status.observedGeneration` should record the
+last generation the controller has processed. In KubeVirt, this field instead records the
+VM generation that was stamped on the VMI — reflecting "what VM generation does the
+running VMI reflect" rather than "what generation has the VM controller processed."
+
+The field with the standard meaning is `vm.status.desiredGeneration`, which is set to
+`vm.Generation` on every reconcile by the VM controller. This inversion is a correctness
+hazard: any tool that reads `status.observedGeneration` — including `kubectl wait`,
+GitOps controllers, and monitoring dashboards — will draw the wrong conclusion about
+whether the VM controller is in sync with the current spec.
+
+### virt-handler has no observable generation signal on VMI (deferred)
+
+`VirtualMachineInstanceStatus` has no `observedGeneration` field. When virt-handler
+applies an in-place spec change, there is no standard mechanism for a consumer to
+determine whether the change has been applied to the running domain. This gap is
+acknowledged but not addressed in this VEP; see Open Questions for the structural
+reasons it is deferred.
+
+### VM and VMI conditions lack per-condition staleness information
+
+Both resources use custom condition structs (`VirtualMachineCondition`,
+`VirtualMachineInstanceCondition`) that do not expose an `ObservedGeneration` field.
+Without it, tools cannot distinguish a stale `Ready=True` condition (evaluated against
+an old spec) from a fresh one. This prevents correct interoperability with the ecosystem
+convention established by KEP-1623.
+
+### Generation annotation is a fragile internal mechanism
+
+The VM controller carries the parent VM's generation across reconcile boundaries via a
+VMI annotation (`VirtualMachineGenerationAnnotation`). This is an internal
+implementation detail that has become load-bearing, evidenced by a back-fill path through
+`ControllerRevision` (`patchVmGenerationFromControllerRevision`) that was added to
+recover from cases where the annotation was missing.
+
+## Goals
+
+- Align generation tracking on VM and VMI resources with Kubernetes conventions so that
+  standard tooling can correctly observe controller sync status.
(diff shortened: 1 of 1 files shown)


lyarwood (commented):
sigh, apologies I asked claude to review this against the envtest I wrote and it posted a wall of slop.

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,485 @@
+# VEP #376: Standardize Generation Tracking for VM and VMI
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
```
lyarwood, line 7: nit - v1.10.0 ?

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,538 @@
...
+The field with the standard meaning is `vm.status.desiredGeneration`, which is set to
+`vm.Generation` on every reconcile by the VM controller. This inversion is a correctness
+hazard: any tool that reads `status.observedGeneration` — including `kubectl wait`,
+GitOps controllers, and monitoring dashboards — will draw the wrong conclusion about
+whether the VM controller is in sync with the current spec.
+
+### virt-handler has no observable generation signal on VMI (deferred)
+
+`VirtualMachineInstanceStatus` has no `observedGeneration` field. When virt-handler
+applies an in-place spec change, there is no standard mechanism for a consumer to
```
lyarwood, line 56: Shouldn't `observedGeneration` only be populated by the VMI controller on the VMI?

Reading further down this sounds like it could be another condition like `HandlerInSync` against the VMI?

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,538 @@
...
+### Functional / e2e tests
+
+- Start a VM; verify `VMIInSync=True` with `observedGeneration == vm.metadata.generation`.
+- Apply a hotpluggable change (e.g., CPU hotplug); verify that `VMIInSync` remains `True`
+  and `observedGeneration` advances to the new VM generation after the controller
+  reconciles, without requiring a full VM restart.
+- Apply a non-hotpluggable change; verify `VMIInSync=True` with
+  `observedGeneration == vm.metadata.generation` after the controller reconciles, and that
+  `RestartRequired=True` is set independently.
+- Stop a VM; verify the `VMIInSync` condition is absent.
```
lyarwood, line 496: As discussed I think envtest would work really well for these tests away from e2e.

iholder101:
Thank you very much @EdDev! I really like the direction of this VEP :+1:

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,538 @@
...
+### Condition type alignment with `metav1.Condition`
+
+`VirtualMachineCondition` and `VirtualMachineInstanceCondition` are updated to embed
+`metav1.Condition` inline, replacing the individually declared fields. The only
+KubeVirt-specific field not present in `metav1.Condition` is `LastProbeTime`, which is
+retained as a deprecated field alongside the embedded struct:
+
+```go
+type VirtualMachineCondition struct {
+    metav1.Condition `json:",inline"`
```
Acedus, line 141: Note that `metav1.Condition` carries additional OpenAPI schema validations with it, meaning that we'd need to adapt all of our existing code to comply (not a problem, just something to consider wrt scope).

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,538 @@
...
+**Using `apimeta` via an adapter.** `apimeta.SetStatusCondition` operates on
+`*[]metav1.Condition` and cannot accept `*[]VirtualMachineCondition` directly. A
+thin adapter bridges the two: it extracts the embedded `metav1.Condition` from each
+element into a `[]metav1.Condition`, calls `apimeta.SetStatusCondition` on that slice,
+then merges the result back into the original `[]VirtualMachineCondition`. Because
+`LastProbeTime` is not part of `metav1.Condition` and is not touched by `apimeta`, the
+round-trip is lossless — `LastProbeTime` is carried through unchanged from the original
+entry, or left unset for newly inserted conditions.
+
+### `VMIInSync` condition on VM
```
Acedus, line 199: I feel like `VMIInSync` implies that changes made in the VM made it the VMI, maybe something like `VMIReconciled` or `VMIConverged`?

veps/sig-compute/376-generation-tracking-standardization/vep.md
```diff
@@ -0,0 +1,538 @@
...
+states. The VM controller does not use it for `VMIInSync` because the controller always
+resolves the condition to a definitive outcome (`True` or `False`) within a single
+reconcile pass. There is no persistent in-flight state that outlives a reconcile. The
+brief window between VMI creation and the first reconcile is represented by the condition
+being absent rather than `Unknown`, consistent with the convention that absence means "not
+yet determined."
+
+The defined `reason` values are:
+
+| `status` | `reason` | Meaning |
```
Acedus, line 233: Would probably have more reasons to differentiate error types.

Acedus (commented):
lgtm on the direction. I think the scope is a bit underestimated but it has nothing to do with the design.

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: Yes, I just wanted to see the initial reaction to this idea before proceeding.

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: Yes, No, Maybe, it is complicated :)

I think that in the original design of the VMI controller and virt-handler workflow, the intention was for the VMI controller to do some initial work and then pass it to the node virt-handler. Today, per the implementation, this is not the case (anymore?). I.e. both access the VMI object, mainly updating its status.

It does make sense to assure virt-handler never changes the spec, and therefore only the VMI controller should keep this (future) field in sync. I will update the wording here to reflect this.

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: Right.
I'm not sure if it needs a special note, as it seems trivial, especially when an adapter is mentioned below (see **Using `apimeta` via an adapter.**).

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: I am fine with `VMIReconciled`.

In Kubevirt codebase, "sync" is a bit more familiar as the reconcile lool, probably because it was created in the times where no other naming agreement existed. Core Kubernetes controller also use this name and the [controller example](https://github.com/kubernetes/sample-controller/blob/3e50bfd72c521dd4b1b9d832b9f2e6254b4ff148/controller.go#L241) has it).
I think the "reconcile"  wording is coming from the newer controller-runtime tooling.

Bottom line, sync and reconciled are the same, I have no strong preference here. I think that "sync" is clearer for the ones who are used to the existing term, but it does not mean it is better.

I suggest to collect opinions before we choose.

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: Many times the `reason` field is used in a problematic manner and I wanted to avoid entering that discussion here.
The `reason` needs to be pretty wide, not too specific, because it cannot control multiple reasons.
Details usually enter the message field.

Said that, if you have a specific suggestion, please share a reason so we can evaluate if it fits.

veps/sig-compute/376-generation-tracking-standardization/vep.md
EdDev: Yes, it will fit nicely here.
I should drop the "e2e" name in order to leave it open at the implementation.

We will need to have envtest available to use it, but I prefer not to make this test dependent on it.

BTW, when we'll have it, I think we need an "integration" section for the test types.

Issue #374, opened 2026-07-06
VEP 374: vhost-net Thread CPU Pinning


jean-edouard:
**Primary contact (assignee)**:

/assign jean-edouard

**Current Feature Stage**:

New

**Feature Gate**:

VhostThreadCPUIsolation

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

VEP PR: (to be linked once opened)

**Timeline**:

* Alpha release target (x.y): 1.10.0
* Beta release target (x.y): 1.11.0
* Stable release target (x.y): 1.12.0

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/378
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/18192 (PoC)
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #378, opened 2026-07-07
VEP #374: vhost-net thread CPU pinning

jean-edouard:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/374
**SIG label**: /sig compute

### What this PR does

New VEP: pin `vhost-net` worker thread(s) to a dedicated pCPU via a `vhostThreadPolicy` field, extending the existing `isolateEmulatorThread` housekeeping-cgroup mechanism.

### DIFF
--- a/veps/sig-compute/374-vhost-cpu-pinning/vep.md
+++ b/veps/sig-compute/374-vhost-cpu-pinning/vep.md
@@ -1,0 +1,148 @@
+# VEP #374: vhost-net Thread CPU Pinning
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10.0
+- This VEP targets beta for version: v1.11.0
+- This VEP targets GA for version: v1.12.0
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements](https://github.com/kubevirt/enhancements/issues/374)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add `vhostThreadPolicy` to `spec.domain.cpu` (VMI and instance type), behind a `VhostThreadCPUIsolation` feature gate. When set, KubeVirt reserves one or more extra dedicated pCPUs and pins the VM's `vhost-net` kernel worker thread(s) to them. Requires `isolateEmulatorThread`, and reuses its housekeeping-cgroup mechanism.
+
+## Motivation
+
+With `dedicatedCpuPlacement` + `isolateEmulatorThread`, all threads that don't have their own dedicated CPU, including `vhost-net`'s kernel worker thread(s), share a single pCPU in the "housekeeping" cgroup. Whenever any of those other threads get busy, whether from a high guest packet rate or from unrelated work on that shared pCPU, `vhost-net` has to wait for its turn, adding latency and cutting into network throughput.
+
+## Goals
+
+- Dedicated, opt-in pCPU(s) for `vhost-net` kernel worker threads.
+- Reuse the existing housekeeping-cgroup mechanism.
+- An API that can grow from "one shared pCPU" to "one pCPU per worker thread" without a breaking change (see Design).
+
+## Non Goals
+
+- Distinguishing `vhost-net` from other vhost-backed devices on the same QEMU process: not possible by name anyway, since the kernel assigns every vhost worker thread the same `vhost-<pid>` name regardless of device type. In practice this means `vhost-vsock` (if VSOCK is enabled) gets isolated the same way as a side effect, which is a welcome bonus, not something worth engineering around.
+- Implementing anything beyond the `Shared` policy (see Design) in Alpha. `PerThread`/`Auto` are reserved API values for later.
+- Validating that the VM actually uses a `vhost-net`-backed interface.
+
+## Definition of Users
+
+VM owners running dedicated-CPU VMs whose workloads depend on consistent, high-performance networking.
+
+## User Stories
+
+- As a user running a network-sensitive workload on a dedicated-CPU VM, I want `vhost-net` on its own pCPU so it stops competing with everything else sharing the housekeeping pCPU.
+
+## Repos
+
+kubevirt/kubevirt
+
+## Design
+
+New `VhostThreadCPUIsolation` feature gate. New optional string enum `VhostThreadPolicy`, added to the VMI's `spec.domain.cpu` and to `CPUInstancetype`, modeled after the existing `IOThreadsPolicy` (`Shared`/`Auto`, later extended with `SupplementalPool` without an API break):
+
+- `Shared`: one dedicated pCPU for all `vhost-net` worker threads. The only value implemented in Alpha.
+- `PerThread`: one dedicated pCPU per `vhost-net` worker thread. Reserved, not implemented yet.
+- `Auto`: KubeVirt picks a reasonable number of pCPUs between `Shared` and `PerThread`. Reserved, not implemented yet.
+
+An enum, rather than a bool, because a bool can only ever mean on/off and can't later grow into "one pCPU per thread" without a second field. And rather than a raw CPU count, because users don't know (and shouldn't need to know) how many `vhost-net` worker threads their VM will end up with: that count is an internal detail, driven by multiqueue settings, that KubeVirt already computes today. See Alternatives.
+
+`vmi-create-admitter` rejects the field unless the gate is enabled and `isolateEmulatorThread` is also set, so it always builds on the existing housekeeping cgroup rather than a second cgroup-management path; for Alpha, it also rejects any value other than `Shared`. The resource renderer requests one extra CPU in the virt-launcher pod spec for `Shared`, the same way it does for the emulator thread.
+
+`virt-launcher` allocates the extra pCPU from the same pool used for vCPUs/IOThreads/emulator thread. Libvirt has no notion of "vhost thread pinning" (it's a host kernel thread, not QEMU-managed), so the CPU is recorded in domain metadata (`VhostCPUSet`) instead of the domain XML, cached across migration, and read back by virt-handler. `VhostCPUSet` is defined as a CPU-list string from the start (like `EmulatorPin`'s), so `PerThread`/`Auto` can later populate it with more than one CPU without a metadata format change.
+
+virt-handler already re-applies the housekeeping cgroup on every VMI sync when `isolateEmulatorThread` is set. On top of that, for `Shared` it now:
+
+- Creates a second child cgroup (`vhost`) pinned to `VhostCPUSet`.
+- Moves any thread named `vhost-<pid>` into it instead of `housekeeping`. This name isn't specific to `vhost-net`, so a `vhost-vsock` worker (if VSOCK is enabled) is isolated the same way, which is fine: see Non Goals.
+- If no such thread exists yet (guest still booting, NIC not up, no such interface at all), that's a no-op for this sync, not an error: nothing to fail or retry.
+- Re-runs this same sweep on every later VMI sync, so a `vhost-net` worker thread that doesn't exist yet, or spawns later (e.g. a second queue coming up after boot), gets moved into `vhost` on the first sync after it appears. There's no dedicated fast path for this: convergence speed just tracks however often the VMI already syncs.
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
+# VEP #374: vhost-net Thread CPU Pinning
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10.0
+- This VEP targets beta for version: v1.11.0
+- This VEP targets GA for version: v1.12.0
```
iholder101, line 9: I think it'd be best to remove these and add the in the relevant release.
Populating these in a certain release serves as a public announcement of the willingness to graduate.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+    resources:
+      limits:
+        cpu: 6 # 4 vCPUs + 1 emulator + 1 vhost
+        memory: 4Gi
+```
+
+## Alternatives
+
+- **Boolean field**: simplest, but a bool can only ever mean on/off. Growing to `PerThread`/`Auto` later would need a second field instead of a natural extension of this one. Rejected in favor of the enum.
+- **Raw CPU count** (e.g. `vhostCPUs: 2`): more flexible than a bool, but forces users to know/guess how many `vhost-net` worker threads their VM will have, an internal detail they shouldn't need to reason about. Rejected in favor of the enum.
```
iholder101, line 98: We've tackled some real cases lately which did demand a specific amount of extra pCPUs per-case. In one case, it turned out that 1 pCPU wasn't enough though more than 2 weren't valuable.

Another alternative is to represent this with a struct. The structs for "Shared", "PerThread" and "Auto" could have no fields, and another struct ("Fixed"?) could have a field for determining the extra pCPU count.

WDYT?

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+
+kubevirt/kubevirt
+
+## Design
+
+New `VhostThreadCPUIsolation` feature gate. New optional string enum `VhostThreadPolicy`, added to the VMI's `spec.domain.cpu` and to `CPUInstancetype`, modeled after the existing `IOThreadsPolicy` (`Shared`/`Auto`, later extended with `SupplementalPool` without an API break):
+
+- `Shared`: one dedicated pCPU for all `vhost-net` worker threads. The only value implemented in Alpha.
+- `PerThread`: one dedicated pCPU per `vhost-net` worker thread. Reserved, not implemented yet.
+- `Auto`: KubeVirt picks a reasonable number of pCPUs between `Shared` and `PerThread`. Reserved, not implemented yet.
```
iholder101, line 58: There's an ongoing debate regarding whether it is best to simply pin vhost-net processes to the same pCPUs that are used by vCPU threads in order to gain cache locality. However, other voices say it could result in "steal time" from the guest in which vCPUs are context-switched in favor of vhost-net threads. I'm honestly not sure if it's better or not, but maybe it should be considered as well.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Add `vhostThreadPolicy` to `spec.domain.cpu` (VMI and instance type), behind a `VhostThreadCPUIsolation` feature gate. When set, KubeVirt reserves one or more extra dedicated pCPUs and pins the VM's `vhost-net` kernel worker thread(s) to them. Requires `isolateEmulatorThread`, and reuses its housekeeping-cgroup mechanism.
+
+## Motivation
+
+With `dedicatedCpuPlacement` + `isolateEmulatorThread`, all threads that don't have their own dedicated CPU, including `vhost-net`'s kernel worker thread(s), share a single pCPU in the "housekeeping" cgroup. Whenever any of those other threads get busy, whether from a high guest packet rate or from unrelated work on that shared pCPU, `vhost-net` has to wait for its turn, adding latency and cutting into network throughput.
```
iholder101, line 26: Do we want to deal specifically with `vhost-net` in this VEP, or all `vhost-*` threads?

iholder101 (commented):
Thank you @jean-edouard!

In addition to the comments below, libvirt's lacking support for vhost pinning. We're trying to communicate with libvirt folks in order to investigate that as well.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Goals
```
EdDev, line 28: Per my understanding, the original motivation is the observation that the network throughput that can be achieved is not optimal.
Therefore, the main goal is to optimize & maximize the network throughput.
Specifically, focused on the most common kernel spaces processing of packets (vhost-net).

The manner in which this can be achieved is an implementation solution, which the points I see here describe. I.e. manage the pCPU pinning for the relevant threads.

If we consider the abstract need here, we may find challenges which make this option less trivial and raise other issues in parallel.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+- Implementing anything beyond the `Shared` policy (see Design) in Alpha. `PerThread`/`Auto` are reserved API values for later.
+- Validating that the VM actually uses a `vhost-net`-backed interface.
+
+## Definition of Users
+
+VM owners running dedicated-CPU VMs whose workloads depend on consistent, high-performance networking.
+
+## User Stories
+
+- As a user running a network-sensitive workload on a dedicated-CPU VM, I want `vhost-net` on its own pCPU so it stops competing with everything else sharing the housekeeping pCPU.
```
EdDev, line 46: I think this story is low-level for a VM owner. What I think the user wants, is to have a VM with a vNIC that can reach the maximal possible network throughput that the platform allows.

The story above is also problematic, as we do not provide data on how to measure what "platform allows".

> With a low level **note**, I can mention that even if the vhost-net threads are not over utilizing the pCPU, it does not mean it is optimized to its full. E.g. if the vhost-net threads are running on a pCPU from NUMA1 and the vcpu threads on a pCPU from NUMA2, it will not be ideal and will impact performance.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
EdDev: I find the attempt to expose cpu pinning optimizations to the VM owner problematic.
If we are to attempt an initial optimization, I would prefer we start with "Auto" conceptually, not exposing any knob except the FG.

And I would go with pinning vhost-net threads to the vcpu threads 1:1, mimicking the BM/in-guest flow as they first attempt, measuring the results and proceeding from there.
The problem is that we may not know how to map the right vhost-net thread to the peer vcpu-thread.

If we go with the "Shared" option, it is not clear to me how one dedicated pCPU will help much. It will improve the situation for sure, but it will not take advantage of the mutiqueue AFAIU.
Maybe we can check this with multiple numbers and see the results, something that during Alpha can be done to find a good compromise.
In general, adding extra pCPU to the virt-launcher introduces a scale issue, which may be an issue for some cluster admins.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+
+`vmi-create-admitter` rejects the field unless the gate is enabled and `isolateEmulatorThread` is also set, so it always builds on the existing housekeeping cgroup rather than a second cgroup-management path; for Alpha, it also rejects any value other than `Shared`. The resource renderer requests one extra CPU in the virt-launcher pod spec for `Shared`, the same way it does for the emulator thread.
+
+`virt-launcher` allocates the extra pCPU from the same pool used for vCPUs/IOThreads/emulator thread. Libvirt has no notion of "vhost thread pinning" (it's a host kernel thread, not QEMU-managed), so the CPU is recorded in domain metadata (`VhostCPUSet`) instead of the domain XML, cached across migration, and read back by virt-handler. `VhostCPUSet` is defined as a CPU-list string from the start (like `EmulatorPin`'s), so `PerThread`/`Auto` can later populate it with more than one CPU without a metadata format change.
```
EdDev, line 64: > the CPU is recorded in domain metadata (`VhostCPUSet`) instead of the domain XML

Where is this metadata persisted?

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Goals
+
+- Dedicated, opt-in pCPU(s) for `vhost-net` kernel worker threads.
+- Reuse the existing housekeeping-cgroup mechanism.
+- An API that can grow from "one shared pCPU" to "one pCPU per worker thread" without a breaking change (see Design).
+
+## Non Goals
+
+- Distinguishing `vhost-net` from other vhost-backed devices on the same QEMU process: not possible by name anyway, since the kernel assigns every vhost worker thread the same `vhost-<pid>` name regardless of device type. In practice this means `vhost-vsock` (if VSOCK is enabled) gets isolated the same way as a side effect, which is a welcome bonus, not something worth engineering around.
```
EdDev, line 36: It sounds like a limitation, not something I would prefer to see non-goals around.
I find comparing `vhost-vsock`  to `vhost-net` problematic in terms of their priority in the system.

Is this solvable in any way? I.e. being able to map the kernel threads to their function?

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+- A `vhost-net` thread that doesn't exist yet at a given sync (guest still booting, or a multi-queue interface bringing up a later queue) is only caught on the next sync (see Design), so nothing is permanently missed, but it can sit in `housekeeping`, contending with the emulator thread, for however long that takes. There's no dedicated fast path to shorten that window today, only whatever cadence the VMI already syncs at.
+- Until `PerThread`/`Auto` are implemented, multi-queue interfaces still put all worker threads on the one `Shared` pCPU.
+- If VSOCK is also enabled, its `vhost-vsock` worker shares the same pCPU as `vhost-net` (see Non Goals). This seems unlikely to matter in practice: VSOCK carries low-volume, control-plane-style traffic in KubeVirt, not the sustained high packet rates `vhost-net` is meant to handle.
+
+## Scalability
+
+No new controllers, watches, or API resources. One extra pCPU and child cgroup per opted-in VMI today; `PerThread`/`Auto` would scale that with the number of worker threads, when implemented.
```
EdDev, line 111: Adding one pCPU per VM at the minimum and potentially more to satisfy the throughput demand is a major scale problem IMO.
In the end, it depends on the VM density cluster admins expect compared to the HW resources (of the node). I think this one requires field data.

veps/sig-compute/374-vhost-cpu-pinning/vep.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Update/Rollback Compatibility
+
+Optional, gated field, off by default: no impact on existing VMs. Rollback/older components ignore the field and metadata and fall back to today's shared housekeeping cgroup. No version-skew requirement.
+
+## Functional Testing Approach
+
+- **Unit**: cgroup thread split and no-op-when-absent behavior; webhook rejects the field when gated off, `isolateEmulatorThread` is unset, or the value isn't `Shared`; resource renderer; instancetype conflict detection.
+- **E2E**: dedicated-CPU VMI with `isolateEmulatorThread` and `vhostThreadPolicy: Shared` set gets `cores + 2` exclusive CPUs, and the `vhost-net` thread's affinity matches `VhostCPUSet`.
```
EdDev, line 120: This one is an integration test, as it checks low level knobs of the implementation.
The e2e scenario is to check it improves performance/throughput, possibly in an extreme condition.

EdDev (commented):
Thank you.

This is an interesting and challenging subject.
Between the current situation to the needs I managed to collect from potential users, this proposal puts us closer to the needed, but it is not clear if it is enough.

There are alternatives to provide network throughput using other means, like HW (SR-IOV) and user-space (vhost-user).
Each has its own pros and cons, none is ideal.

iholder101:
@jean-edouard do you plan to defer this one?

jean-edouard:
> @jean-edouard do you plan to defer this one?

@iholder101 this topic got a lot of traction since (A)I wrote this VEP, and I may not be the right person to drive this. I do apologize to you and @EdDev for not following through with your reviews.
If libvirt is close to releasing vhost pinning support, we can just wait for that and close this VEP.
If you think this VEP still has value, let me know and I'll freshen it up.

Issue #381, opened 2026-07-08
VEP 381: KSM Zero-Pages-Only Mode and Sleep Scaling Improvement


michalskrivanek:
**Primary contact (assignee)**:

/assign @mskrivan

**Current Feature Stage**:
New

**Feature Gate**:
None (extends existing `ksmConfiguration` API; no new feature gate required).

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/382

**Timeline**:
* 
* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):
* [X] Alpha
  * [X] VEP PRs: https://github.com/kubevirt/enhancements/pull/382
  * [ ] Code PRs:
  * [ ] Docs PRs:

michalskrivanek:
https://github.com/kubevirt/enhancements/pull/382#issuecomment-5392895162

PR #382, opened 2026-07-08
VEP 381: KSM Zero-Pages-Only Mode and Sleep Scaling Improvement

michalskrivanek:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/381
**SIG label**: /sig compute

### What this PR does
This VEP extends KubeVirt's existing KSM (Kernel Same-page Merging) support with a new
`ZeroPagesOnly` mode that leverages the RHEL/CentOS Stream 9 `redhat_only_zero_pages` kernel
feature. This mode makes KSM practical and efficient for Windows guest workloads by restricting
page deduplication to zero-filled pages only, eliminating the expensive stable/unstable tree
operations that yield little benefit for Windows memory patterns.

### DIFF
--- a/veps/sig-compute/381-ksm-zero-pages-only-mode/ksm-zero-pages-only-mode.md
+++ b/veps/sig-compute/381-ksm-zero-pages-only-mode/ksm-zero-pages-only-mode.md
@@ -1,0 +1,427 @@
+# VEP #NNNN: KSM Zero-Pages-Only Mode and Sleep Scaling Improvement
+
+Owners:
+- @mskrivan
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Target version is explicitly mentioned and approved
+- [X] (R) Graduation criteria filled
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non Goals](#non-goals)
+- [Definition of Users](#definition-of-users)
+- [User Stories](#user-stories)
+- [Repos](#repos)
+- [Design](#design)
+  - [API Changes](#api-changes)
+  - [KSM Mode: ZeroPagesOnly](#ksm-mode-zeropagesonly)
+  - [Improvement: Dynamic Sleep Baseline Scaling](#improvement-dynamic-sleep-baseline-scaling)
+  - [Sysfs Management](#sysfs-management)
+  - [Fallback and Capability Detection](#fallback-and-capability-detection)
+  - [Kernel and Distribution Requirements](#kernel-and-distribution-requirements)
+- [API Examples](#api-examples)
+- [Alternatives](#alternatives)
+- [Scalability](#scalability)
+- [Update/Rollback Compatibility](#updaterollback-compatibility)
+- [Functional Testing Approach](#functional-testing-approach)
+- [Implementation History](#implementation-history)
+- [Graduation Requirements](#graduation-requirements)
+
+## Overview
+
+This VEP extends KubeVirt's existing KSM (Kernel Same-page Merging) support with a new
+`ZeroPagesOnly` mode that leverages the RHEL/CentOS Stream 9 `redhat_only_zero_pages` kernel
+feature. This mode makes KSM practical and efficient for Windows guest workloads by restricting
+page deduplication to zero-filled pages only, eliminating the expensive stable/unstable tree
+operations that yield little benefit for Windows memory patterns.
+
+This feature depends on a RHEL-only kernel patch (`redhat_only_zero_pages`) that is not
+available upstream. See [Kernel and Distribution Requirements](#kernel-and-distribution-requirements)
+for the full impact analysis.
+
+## Motivation
+
+KubeVirt has supported KSM since v1.1 via the `ksmConfiguration` field in the KubeVirt CR.
+The current implementation has two limitations that prevent effective use in production:
+
+### 1. Full KSM is expensive and ineffective for Windows guests
+
+Standard KSM deduplicates all identical pages by maintaining stable and unstable comparison
+trees. This is CPU-intensive and yields diminishing returns for Windows workloads:
+
+- **Windows aggressively zeros memory.** The Windows kernel has a dedicated zero page thread
+  (priority 0) that scrubs freed pages. `VirtualAlloc` guarantees zero-initialized memory.
+  All physical pages are zeroed on boot. As a result, Windows VMs accumulate large pools of
+  zero-filled memory.
+- **Windows pages are mostly unique across VMs.** Unlike Linux VMs that share libc, kernel
+  text, and other common libraries, Windows VMs have few non-zero identical pages across
+  instances. Full tree comparison is wasted CPU.
+
+RHEL 9.9+ kernels (CentOS Stream 9) include a `redhat_only_zero_pages` sysfs knob that
+restricts KSM to only deduplicate zero-filled pages, skipping all stable/unstable tree
+operations. This dramatically reduces CPU overhead while capturing the majority of memory
+savings for Windows workloads.
+
+### 2. The `redhat_only_zero_pages` kernel feature
+
+RHEL 9.9 / CentOS Stream 9 introduced a `redhat_only_zero_pages` sysfs knob (authored by
+Andrea Arcangeli, MR !8249). When enabled, KSM's scan loop changes fundamentally:
+
+1. For each scanned page, compute the checksum as usual.
+2. If `checksum == zero_checksum`, attempt to merge with the kernel zero page.
+3. **Return immediately** — skip stable tree search, unstable tree search, and all
+   cross-page comparison.
+
+This turns KSM from a general-purpose memory deduplication engine into a lightweight
+zero-page consolidation pass. The CPU cost drops dramatically because the expensive
+operations (memcmp, tree insert/lookup, page locking for merge) are eliminated for all
+non-zero pages.
+
+This feature is **not proposed upstream** and currently exists only in RHEL/CentOS Stream
+kernels. See [Kernel and Distribution Requirements](#kernel-and-distribution-requirements)
+for the implications.
+
+## Goals
+
+1. Add a `ZeroPagesOnly` KSM mode that uses the `redhat_only_zero_pages` kernel feature,
+   exposed as a first-class API field on `KSMConfiguration`.
+2. Keep all changes backward compatible — existing `KSMConfiguration` with no `mode` field
+   behaves identically to today.
+3. As a secondary improvement, fix the sleep baseline scaling so that effective
+   `sleep_millisecs` is consistent across host sizes without requiring per-node annotations.
+   This benefits both modes but is not the primary motivation for this VEP.
+
+## Non Goals
+
+- Upstream the `redhat_only_zero_pages` kernel patch (this is a RHEL/CentOS kernel concern).
+- Expose all KSM tuning parameters (pages boost/decay/min/max) in the API. These remain
+  available as per-node annotation overrides for power users.
+- Integrate with the VEP 190 plugin framework (no dependency on node hooks).
+- Add Prometheus metrics for KSM (valuable but orthogonal, tracked separately).
+
+## Definition of Users
+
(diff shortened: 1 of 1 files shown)


michalskrivanek:
related tracker https://redhat.atlassian.net/browse/CNV-72743

iholder101:
Thanks @michalskrivanek.

One concern with this VEP is it extends an in-tree implementation of what is essentially an OS-level node memory management. In my opinion, KSM handling by Kubevirt has always been wrong, and I've stated it a few times before [[1](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2879062046)][[2](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2953996540)]. It's node-level memory tuning that leaked into core because it was convenient, not because it belongs there. I've stated many times that ideally this should be implemented out-of-tree.

In addition, the `redhat_only_zero_pages` sysfs toggle is RHEL-specific, adding distro-specific kernel feature support directly into Kubevirt core.

Here are a few alternatives to consider:
1. Plugins - Node hooks already exist and fire on every reconciliation. KSM management could be moved to a plugin that handles sysfs tuning, including distro-specific features like zero-pages-only mode, without touching core. I do consider adding another `NodeReconcile` node hook that will always fire independently of any VM lifecycle, and this could be a great use-case.
2. External controller - A standalone out-of-tree controller that handles KSM configuration and manages sysfs directly. Simpler, ships independently of KubeVirt release cycles, and keeps distro-specific logic where it belongs.

I think that with this approach we can strive to eventually deprecate `ksmConfiguration` from core altogether, which would be great.

WDYT?

michalskrivanek:
Hi @iholder101,
thank you for your thoughts!

> Thanks @michalskrivanek.
> 
> One concern with this VEP is it extends an in-tree implementation of what is essentially an OS-level node memory management. In my opinion, KSM handling by Kubevirt has always been wrong, and I've stated it a few times before [[1](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2879062046)][[2](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2953996540)]. It's node-level memory tuning that leaked into core because it was convenient, not because it belongs there. I've stated many times that ideally this should be implemented out-of-tree.

This particular VEP is not trying to address that, just to improve the performance and usefulness of the current state.
One can argue about what "core" means I'm sure:) I can see the argument of `virt-handler` being the node level manager so KSM belongs there. Counterexample would be the node labeling with CPU models. TBH both features are ugly:)

> In addition, the `redhat_only_zero_pages` sysfs toggle is RHEL-specific, adding distro-specific kernel feature support directly into Kubevirt core.

in the current form it is indeed a RHEL-kernel specific toggle until the feature reaches mainline (and gets a final name), hence the specific prefix for it. I don't think it's that much different from e.g. `containerd` or `crio`-specific code paths. It's not ideal, but at the same time RHEL-derivatives are widely used. And as far as kubvirt surface is concerned it's really just a name of a sysfs file to write to, that will be eventually changed to the final location once it bubbles through mainline to all distros (that could take a while).


> Here are a few alternatives to consider:
> 
>     1. Plugins - Node hooks already exist and fire on every reconciliation. KSM management could be moved to a plugin that handles sysfs tuning, including distro-specific features like zero-pages-only mode, without touching core. I do consider adding another `NodeReconcile` node hook that will always fire independently of any VM lifecycle, and this could be a great use-case.

I'm not a big fan of plugins for core feature (to your earlier point I don't necessarily mean it has to be in `virt-handler`, I only believe the KSM feature behavior is core responsibility of Kubevirt, not some "external" integrator/entity). 

>     2. External controller - A standalone out-of-tree controller that handles KSM configuration and manages sysfs directly. Simpler, ships independently of KubeVirt release cycles, and keeps distro-specific logic where it belongs.

but also means a lot more overhead. Plus the price to pay to cleanly deprecate, remove from API  (and then put where?)

> I think that with this approach we can strive to eventually deprecate `ksmConfiguration` from core altogether, which would be great.
> 
> WDYT?

Personally I think it's not that helpful to try to remove it because in the end someone has to do that config. Though I understand it depends on how you see that feature in general, is it core or not, is it useful. Since the point of this VEP is to make it even *more* useful I guess it's clear where I stand:)

iholder101:
> Hi @iholder101, thank you for your thoughts!
> 
> > Thanks @michalskrivanek.
> > One concern with this VEP is it extends an in-tree implementation of what is essentially an OS-level node memory management. In my opinion, KSM handling by Kubevirt has always been wrong, and I've stated it a few times before [[1](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2879062046)][[2](https://github.com/kubevirt/kubevirt/pull/14632#issuecomment-2953996540)]. It's node-level memory tuning that leaked into core because it was convenient, not because it belongs there. I've stated many times that ideally this should be implemented out-of-tree.
> 
> This particular VEP is not trying to address that, just to improve the performance and usefulness of the current state. One can argue about what "core" means I'm sure:) I can see the argument of `virt-handler` being the node level manager so KSM belongs there. Counterexample would be the node labeling with CPU models. TBH both features are ugly:)

In Kubernetes, labels are the standard mechanism to control scheduling. It's a common practice for operators to add labels to nodes in order to steer pods to be scheduled to the right place. I do agree that we add many labels and we should aim to reduce the amount of the labels we add. That being said, I don't think this example is equivalent to Kubevirt messing with kernel-level knobs in charge of memory management. Kubevirt is not a node-tuning operator - other operators exist for that purpose.

> > In addition, the `redhat_only_zero_pages` sysfs toggle is RHEL-specific, adding distro-specific kernel feature support directly into Kubevirt core.
> 
> in the current form it is indeed a RHEL-kernel specific toggle until the feature reaches mainline (and gets a final name), hence the specific prefix for it. I don't think it's that much different from e.g. `containerd` or `crio`-specific code paths. It's not ideal, but at the same time RHEL-derivatives are widely used. And as far as kubvirt surface is concerned it's really just a name of a sysfs file to write to, that will be eventually changed to the final location once it bubbles through mainline to all distros (that could take a while).

I'm not denying its usefulness, just wonder if we should be adding an API for a sysfs that is available only on RHEL. It might land in mainline Linux, and might not, might land in a different implementation, etc. Even if I put aside my thoughts about KSM needing to be out of the core, I wonder if this is mature enough to add an official API to.

> > Here are a few alternatives to consider:
> > ```
> > 1. Plugins - Node hooks already exist and fire on every reconciliation. KSM management could be moved to a plugin that handles sysfs tuning, including distro-specific features like zero-pages-only mode, without touching core. I do consider adding another ``NodeReconcile`` node hook that will always fire independently of any VM lifecycle, and this could be a great use-case.
> > ```
> 
> I'm not a big fan of plugins for core feature (to your earlier point I don't necessarily mean it has to be in `virt-handler`, I only believe the KSM feature behavior is core responsibility of Kubevirt, not some "external" integrator/entity).

I think this is the main disagreement point.

I do think that, as the Plugin VEP says in its [user-stories](https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/190-kubevirt-structured-plugins/vep.md?rgh-link-date=2026-07-19T06%3A53%3A44.000Z#user-stories), one of the stated goals is to use them for features that do not belong to the core.

> > ```
> > 2. External controller - A standalone out-of-tree controller that handles KSM configuration and manages sysfs directly. Simpler, ships independently of KubeVirt release cycles, and keeps distro-specific logic where it belongs.
> > ```
> 
> but also means a lot more overhead. Plus the price to pay to cleanly deprecate, remove from API (and then put where?)

I agree that it would demand more work and overhead. However, this argument could be brought up as a justification for basically anything not related to the core. Keeping Kubevirt cohesive and focused only on the necessary scope, hence keeping it compact and maintainable, is not less important than the added overhead IMO.

> > I think that with this approach we can strive to eventually deprecate `ksmConfiguration` from core altogether, which would be great.
> > WDYT?
> 
> Personally I think it's not that helpful to try to remove it because in the end someone has to do that config. Though I understand it depends on how you see that feature in general, is it core or not, is it useful. Since the point of this VEP is to make it even _more_ useful I guess it's clear where I stand:)

Again, I'm not denying the usefulness of this, just wonder who should maintain it and whether it's Kubevirt's business to get into node-level memory management.

Maybe others have different opinions about this.

fabiand:
> whether it's Kubevirt's business to get into node-level memory management.

It already is, because KSM is already part of KubeVirt.

Thus to me the question thenis: How does KubeVirt get out of this business.

vladikr:
> > whether it's Kubevirt's business to get into node-level memory management.
> 
> It already is, because KSM is already part of KubeVirt.
> 
> Thus to me the question thenis: How does KubeVirt get out of this business.

maybe we should take this opportunity to separate it from KubeVirt? I also feel that this is similar to swap handling, tuned, performance-profile-operator and such ... 
Can't we really fold it into one of these or start a sub-project in the org with a new operator?

vladikr:
"redhat_only_zero_pages" :roll_eyes:

enp0s3:
@michalskrivanek @fabiand Hey folks
> It already is, because KSM is already part of KubeVirt.
> 
> Thus to me the question thenis: How does KubeVirt get out of this business.

That is correct, we don't have a formal plan how to get rid of KSM management, but I think we have a quorum on the fact that we want to discontinue its development, because we think KubeVirt should not own machine configuration.
In some case it is possible to read data from the host in order to make decisions, like we do in VEP 222. But it also should be done carefully because once again host configuration is and should be out of KubeVirt control.

Have you consideted to deploy custom Node Tune Operator configuration using virt-platform-autopilot?

michalskrivanek:
> > > whether it's Kubevirt's business to get into node-level memory management.
> > 
> > 
> > It already is, because KSM is already part of KubeVirt.
> > Thus to me the question thenis: How does KubeVirt get out of this business.
> 
> maybe we should take this opportunity to separate it from KubeVirt? I also feel that this is similar to swap handling, tuned, performance-profile-operator and such ... Can't we really fold it into one of these or start a sub-project in the org with a new operator?

define "separate from KubeVirt". KSM management is part of kubevirt because it was originally supposed to be the foundation for active ballooning. That hasn't happened, so yes it looks weird as it is. If by separation you mean a different component then that's fine, but It seems to me the problem is more fundamental, @iholder101 and @enp0s3 seems to suggest it's simply not kubevirt business at all, which then just makes me ask whose business that is. It is active (even without the per-VM balooning) so it's not as trivial as just turning something on or off via performance profile or machineconfig or such. And the purpose is to increase VM density. So I'm not sure who else than kubevirt should care. 
kubevirt doesn't exist in a vacuum, it needs kernel level features that are unique to virtualization. Yes you may separate them out to as many components you can imagine, but it only hurts the usability and maintainability of the whole, and doesn't change the fact they're all part of KubeVirt.

> "redhat_only_zero_pages" 🙄

https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/8249

iholder101:
> @iholder101 and @enp0s3 seems to suggest it's simply not kubevirt business at all, which then just makes me ask whose business that is.

By "Kubevirt" we mean the kubevirt/kubevirt repo, not the Kubevirt org.

As @vladikr suggested above, a controller living in a different repo, being maintained separately and deployed by HCO seems reasonable. It would probably need to use NTO / MCO / some other operator that's expert on tuning the node resource rather than using virt-handler which shouldn't be used for node-level tunings. Plugins might also be an option.

I think it would be beautiful to take this VEP as an opportunity to plan extracting KSM management from Kubevirt.

fabiand:
IMHO if KSM is getting extracted (which makes some sense, albeit we did the jump to include it in KubeVirt, but nvm), then a VEP is not the appropriate place, my reason being: If it's decoupled from core KubeVirt, then it's more on the kube distro side, and tightly integrates or needs to align with how vendors are configuring their nodes.

Thus in this case I'd see it moving out of kubevirt into kube vendor specific places.

michalskrivanek:
the challenge with moving it out is that it is dynamic, it's not a simple MCO-like config change. I do not see a suitable existing place in neither core nor vendor specific components. Before going forward with a new one, does anyone have any other suggestion?

michalskrivanek:
> the challenge with moving it out is that it is dynamic, it's not a simple MCO-like config change. I do not see a suitable existing place in neither core nor vendor specific components. Before going forward with a new one, does anyone have any other suggestion?


ok, i'm going to avoid kubevirt and do it separately. thanks for consideration

iholder101:
> > the challenge with moving it out is that it is dynamic, it's not a simple MCO-like config change. I do not see a suitable existing place in neither core nor vendor specific components. Before going forward with a new one, does anyone have any other suggestion?
> 
> ok, i'm going to avoid kubevirt and do it separately. thanks for consideration

Thanks for the effort here @michalskrivanek.

If you're up to it, it could be great if you'd consider leading a VEP to deprecate KSM from KubeVirt after the external controller is in place.

Issue #383, opened 2026-07-08
VEP 383: THP Memory Backing for Virtual Machines


michalskrivanek:
**Primary contact (assignee)**:

/assign @michalskrivanek

**Current Feature Stage**: New

**Feature Gate**: `THPMemoryBacking`

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/384

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y): TBD
* Stable release target (x.y): TBD

* [X] Alpha
  * [X] VEP PRs: https://github.com/kubevirt/enhancements/pull/384
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:

Allow VMs declaring hugepages to use Transparent Huge Pages (THP) instead of static pre-allocated hugepages. Guest memory is backed by regular anonymous pages that are collapsed to 2M THP at runtime via `MADV_COLLAPSE`, achieving equivalent TLB performance without requiring hugepage pre-allocation on nodes.

Key motivations:
- Eliminates the need for cluster-level hugepage reservation and NUMA-aware capacity planning
- VMs become schedulable on any node with sufficient free memory

Depends on:
- [VEP #144: Memory Lock RLimit configuration](https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/144-memory-lock-rlimits/memory-lock-rlimits.md) (for virt-handler `prlimit64` mechanism)
- Kernel with `MADV_COLLAPSE`: RHEL 9.2+ or upstream 6.1+

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

PR #384, opened 2026-07-08
VEP 383: THP Memory Backing for Virtual Machines

michalskrivanek:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/383
**SIG label**: /sig compute

### What this PR does
Allow VMs declaring hugepages to use Transparent Huge Pages (THP) instead of static pre-allocated hugepages. Guest memory is backed by regular anonymous pages that are collapsed to THP at runtime via MADV_COLLAPSE, achieving equivalent TLB performance without requiring hugepage pre-allocation on nodes.

### DIFF
--- a/veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
+++ b/veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
@@ -1,0 +1,464 @@
+# VEP: THP Memory Backing for Virtual Machines
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: 1.10
+- This VEP targets beta for version: TBD
+- This VEP targets GA for version: TBD
+
+### Release Signoff Checklist
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This enhancement allows VMs declaring hugepages to run without
+pre-allocated static hugepages on the node. Instead, guest memory is
+backed by regular anonymous pages that are collapsed into Transparent
+Huge Pages (THP) at runtime using `MADV_COLLAPSE`. Combined with `mlock`
+and immediate preallocation, this achieves equivalent TLB performance to
+static hugepages.
+
+The native THP size is architecture-dependent (2M on x86_64, 1M on
+s390x, etc. — see Architecture Considerations below). This VEP uses
+"THP" generically to mean the platform's native transparent huge page
+size.
+
+The VM retains its hugepages declaration in the spec (preserving NUMA
+topology and domain XML generation), but the pod is built with regular
+memory resources. virt-handler raises `RLIMIT_MEMLOCK` on virtqemud and
+a collapse call promotes pages to THP after preallocation completes.
+
+## Motivation
+
+Static hugepages require cluster-level pre-allocation: administrators
+must configure nodes with a fixed number of reserved hugepages at boot
+or runtime. This creates operational burdens:
+
+1. **Capacity planning** — hugepages reserved at boot are unavailable
+   for other workloads, even when no VM uses them.
+2. **Scheduling constraints** — VMs can only land on nodes with
+   sufficient pre-allocated hugepages, reducing scheduler flexibility.
+3. **NUMA fragmentation** — hugepages must be allocated on the correct
+   NUMA node; imbalanced allocation leads to scheduling failures.
+4. **Cluster heterogeneity** — different VM sizes need different
+   hugepage reservations, complicating node profiles.
+5. **Operational toil** — changing hugepage reservations requires node
+   reboots or careful runtime tuning with risk of allocation failure.
+
+Modern kernels (RHEL 9.2+, upstream 6.1+) support `MADV_COLLAPSE`,
+which synchronously collapses regular pages into THPs. Combined with
+`mlock` and immediate preallocation, this achieves identical TLB
+performance to static hugepages without any pre-allocation.
+
+### Proof of Concept
+
+An out-of-tree addon
+([kubevirt-hugepages-addon](https://github.com/michalskrivanek/kubevirt-hugepages-addon))
+demonstrates feasibility using admission webhooks and a sidecar hook. It
+requires several runtime workarounds due to operating outside KubeVirt's
+control plane. Native integration eliminates all of them.
+
+## Goals
+
+- Allow VMs declaring hugepages to run without pre-allocated hugepages
+  on the node.
+- Achieve equivalent TLB performance to static hugepages via THP
+  collapse.
+- Preserve existing NUMA topology generation and domain XML structure.
+- Use the standard KubeVirt memory overhead calculation — no special
+  buffer hacks.
+- Require no changes to libvirt or QEMU.
+- Support two operational modes: guaranteed THP coverage (VM fails if
+  collapse incomplete) and best-effort (collapse is opportunistic,
+  coverage reported in status).
+
+## Non Goals
+
+- Replacing static hugepages — both modes coexist; THP mode is opt-in.
+- Guaranteeing 100% THP coverage on fragmented nodes — the guaranteed
+  mode defines a threshold and failure path, not a physical guarantee.
+- Supporting 1G hugepages dynamically — `MADV_COLLAPSE` targets the
+  architecture's native THP size, not arbitrary sizes.
+- Overcommitting guest memory — `<locked/>` prevents swapping; this is
+  a performance feature, not an overcommit feature.
+- Modifying the kernel's THP subsystem behavior (defrag, khugepaged
+  tuning).
+
+## Definition of Users
+
+- Users running VMs that benefit from hugepages performance but want
+  simpler cluster operations.
+- Cluster administrators who want to avoid static hugepage reservation
+  and NUMA planning.
+- Cloud providers offering VM-based workloads where hugepage
+  pre-allocation limits bin-packing.
+
+## User Stories
+
+- As a user, I want my VM to get THP performance without requiring
+  hugepages to be pre-allocated on the node.
+- As a cluster admin, I want to stop managing hugepage reservations
+  across heterogeneous nodes.
+- As a user, I want my VM to be schedulable on any node with sufficient
+  free memory, not only nodes with hugepages available.
+- As a user deploying latency-sensitive workloads, I want guaranteed
+  THP backing with a clear failure signal if the node cannot provide it.
+- As a user deploying general-purpose VMs, I want best-effort THP with
+  visibility into actual coverage.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt/)
+
+## Design
+
+### API
+
+A new `mode` field is added to `spec.domain.memory.hugepages`. The
+naming uses "thp" explicitly to distinguish from the kernel's dynamic
+hugepage allocation via sysfs (`nr_hugepages`):
+
+```yaml
+spec:
+  domain:
+    memory:
+      hugepages:
+        pageSize: "2Mi"
+        mode: thp
+```
+
+`mode` accepts two values:
+- `static` (default, current behavior) — pod requests hugepages
+  resources, node must have pre-allocated hugepages.
+- `thp` — pod requests regular memory, pages are collapsed to THP at
+  runtime.
+
+Absence of the `mode` field preserves current behavior (`static`).
+
+**Strictness policy (TBD):** The feature should support two operational
+modes — guaranteed (VM fails to reach `Running` if collapse does not
+achieve sufficient coverage) and best-effort (collapse is opportunistic,
+partial coverage reported in status). The exact API for expressing this
+(sub-field, enum value, separate field) is to be determined during
+review.
+
+### Component Changes
+
+#### virt-controller (pod template generation)
+
+When THP mode is selected:
+
+1. **Do not add `hugepages-*` resources** to the compute container.
+   Instead, set `resources.requests.memory` and
+   `resources.limits.memory` to `guest_memory + GetMemoryOverhead()`
+   using the standard overhead formula (same path as non-hugepages VMs).
+
+2. **Do not create the `hugepages` emptyDir volume** with
+   `medium: HugePages`.
+
+3. **Do not mount `/dev/hugepages`** in the compute container. This
+   prevents virt-launcher from writing `hugetlbfs_mount` to
+   `qemu.conf`, which would cause virtqemud to fail initialization.
+
+4. **Set RLIMIT_MEMLOCK requirement** — triggers the VEP 144 mechanism
+   so virt-handler raises the memory lock limit on virtqemud.
+
+#### virt-handler (memory lock and THP collapse)
+
+When a VMI uses THP mode:
+
+1. **Memory lock**: Call `prlimit64` on the virt-launcher/virtqemud
+   process to set `RLIMIT_MEMLOCK = RLIM_INFINITY`. This is the same
+   mechanism already used for VFIO passthrough VMs. When libvirt later
+   processes `<locked/>` in the domain XML and calls `setrlimit`, the
+   limit is already high enough and libvirt's check
+   (`current >= limit`) passes — making it a no-op.
+
+2. **THP collapse**: After QEMU preallocates memory (detected by
+   monitoring `VmLck` stabilization in `/proc/<pid>/status`):
+   - Parse `/proc/<pid>/maps` for large anonymous RW regions.
+   - Call `process_madvise(pidfd, iov, MADV_COLLAPSE)` to synchronously
+     collapse pages into THPs.
+   - Report coverage ratio in VMI status.
+   - For guaranteed policy: if coverage < threshold, transition VMI to
+     `Failed`.
+   - For best-effort policy: Falls back to `khugepaged` for regions that
+     cannot be immediately collapsed. Log result, update status, continue.
+
+
+virt-handler already runs privileged on the host with access to process
+PIDs and `CAP_SYS_PTRACE` — no additional capabilities are needed.
+
+#### Domain converter (XML generation)
+
+When THP mode is selected, the converter generates:
+
+```xml
+<memoryBacking>
+  <source type="anonymous"/>
+  <locked/>
+  <allocation mode="immediate" threads="4"/>
+</memoryBacking>
+```
+
+Instead of the static hugepages XML:
+
+```xml
+<memoryBacking>
+  <hugepages>
+    <page size="2048" unit="KiB"/>
+  </hugepages>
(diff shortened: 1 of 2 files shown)


veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
```diff
@@ -0,0 +1,464 @@
...
+- Domain converter: generate anonymous/locked/immediate XML.
+- virt-handler: set `RLIMIT_MEMLOCK` for THP-mode VMs.
+- THP `MADV_COLLAPSE` collapse call (single pass, log and report coverage).
+- Unit tests for all components.
+
+### Beta
+
+- Guaranteed policy support (VM fails if coverage below threshold).
+- E2E tests covering startup, THP coverage, NUMA, migration, both
+  policies.
```
jean-edouard: Any reason to implement the e2e tests in beta instead of alpha? I know we've accepted that in previous VEPs because of a lack of hardware, but it shouldn't become the norm... Unless you do need special hardware that's not currently available?
Maybe, as a compromise, alpha could have basic e2e coverage, and more advanced ones for beta?

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,464 @@
...
+- As a user deploying latency-sensitive workloads, I want guaranteed
+  THP backing with a clear failure signal if the node cannot provide it.
+- As a user deploying general-purpose VMs, I want best-effort THP with
+  visibility into actual coverage.
+
+## Repos
+
+- [KubeVirt](https://github.com/kubevirt/kubevirt/)
+
+## Design
```
jean-edouard, line 78: This whole section doesn't appear to touch on migrations. Is there any consideration of when `MADV_COLLAPSE` runs, or any pre-copy vs post-copy differences?
Or do we want to make THP VMIs non-migratable in alpha and tackle migration support later as part of beta?

veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
```diff
@@ -0,0 +1,464 @@
...
+  runtime.
+
+Absence of the `mode` field preserves current behavior (`static`).
+
+**Strictness policy (TBD):** The feature should support two operational
+modes — guaranteed (VM fails to reach `Running` if collapse does not
+achieve sufficient coverage) and best-effort (collapse is opportunistic,
+partial coverage reported in status). The exact API for expressing this
+(sub-field, enum value, separate field) is to be determined during
+review.
```
jean-edouard: During which review? VEP or code? Since it's API, I believe it should be part of the VEP for the get-go.

veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
```diff
@@ -0,0 +1,464 @@
...
+
+#### Domain converter (XML generation)
+
+When THP mode is selected, the converter generates:
+
+```xml
+<memoryBacking>
+  <source type="anonymous"/>
+  <locked/>
+  <allocation mode="immediate" threads="4"/>
```
jean-edouard: Will that be a hardcoded value?

jean-edouard (commented):
Nice VEP, thanks for keeping it under 500 lines :)
I have a few questions in-line, nothing major. In general, I think this VEP makes sense.

veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
michalskrivanek: no reason. will move to alpha

veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
michalskrivanek: i believe we can omit it in alpha. Though for large VMs it's likely quite helpful I have no data currently to show.

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: I think post-copy needs more thinking. IMHOit makes sense to keep pre-copy working with `best-effort`. Just a suggestion though, if anyone have strong feelings about it

veps/sig-compute/383-thp-memory-backing/vep-thp-memory-backing.md
michalskrivanek: ok, added a suggested API with nice-to-have and guaranteed modes.

michalskrivanek:
@jean-edouard thanks for the review, I made some updates (and this made it 518 lines now:(). The implementation is simple enough so we can make compromises on the API or scope if needed

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,518 @@
...
+
+- The new API field is optional; absence preserves current behavior
+  (static hugepages). Fully backward compatible.
+- On rollback to a version without the `THPMemoryBacking` feature gate:
+  VMIs configured for THP mode will be treated as standard hugepages
+  VMs. They will fail to schedule if no static hugepages are available
+  on the node. This is safe — the VM will not start, no data loss
+  occurs.
+- Running VMs are not affected by rollback until they are restarted or
+  migrated.
```
jean-edouard: Then what happens?

jean-edouard (commented):
/lgtm
Just one last question in-line.
Nit: using longer lines would make this more readable in raw mode, and as a bonus probably take you back to sub-500 lines dreamland :)

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: it's just to say they're not touched while running. I'll shorten the whole section a bit

michalskrivanek: done

jean-edouard (approved):

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+|-------|--------|---------|------------|
+| `mode` | `static`, `thp` | `static` | always |
+| `pageSize` | e.g. `"2Mi"` | required when hugepages set | always |
+| `policy` | `best-effort`, `guaranteed` | `best-effort` | `mode: thp` only |
+
+`mode` accepts two values:
+- `static` (default, current behavior) — pod requests hugepages
+  resources, node must have pre-allocated hugepages.
+- `thp` — pod requests regular memory, pages are collapsed to THP at
+  runtime.
```
iholder101: Can we rename this to `transparentHugePages` to make this more clear?

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This enhancement allows VMs declaring hugepages to run without
+pre-allocated static hugepages on the node. Instead, guest memory is
+backed by regular anonymous pages that are collapsed into Transparent
+Huge Pages (THP) at runtime using `MADV_COLLAPSE`. Combined with `mlock`
+and immediate preallocation, this achieves equivalent TLB performance to
+static hugepages.
```
iholder101: > this achieves equivalent TLB performance to static hugepages.

I think this is almost accurate, but it's only really true with 100% converage. So, it's not guaranteed at all with the `best-effort` mode, and not even entirely with `guaranteed`. Maybe worth mentioning that.

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+
+`policy` controls collapse strictness (only valid when `mode` is `thp`):
+- `best-effort` (default) — virt-handler runs `MADV_COLLAPSE` once after
+  preallocation. Partial collapse is acceptable; uncovered regions may
+  be promoted later by `khugepaged`. Coverage is reported in status and
+  the VMI reaches `Running` regardless of coverage. Pre-copy and
+  post-copy live migration are supported. Available in alpha.
+- `guaranteed` — same collapse pass, but if coverage falls below 95%
+  the VMI transitions to `Failed` with reason `THPCollapseFailed`.
+  Available in beta. The 95% threshold is fixed in virt-handler and not
```
iholder101: Ideally, we shouldn't add core logic when graduating to Beta. This will probably need another alpha round. See https://github.com/kubevirt/enhancements/blob/main/docs/feature-lifecycle.md#graduation-phases for more info.

Additionally, I'd avoid future promises and write it as something like:

```suggestion
  Will be available in the future. The 95% threshold is fixed in virt-handler and not
```

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+architecture's native THP size applies (see Architecture Considerations).
+
+`policy` controls collapse strictness (only valid when `mode` is `thp`):
+- `best-effort` (default) — virt-handler runs `MADV_COLLAPSE` once after
+  preallocation. Partial collapse is acceptable; uncovered regions may
+  be promoted later by `khugepaged`. Coverage is reported in status and
+  the VMI reaches `Running` regardless of coverage. Pre-copy and
+  post-copy live migration are supported. Available in alpha.
+- `guaranteed` — same collapse pass, but if coverage falls below 95%
+  the VMI transitions to `Failed` with reason `THPCollapseFailed`.
```
iholder101: The 95% check runs post-scheduling and depends on node fragmentation at that point in time. The scheduler is fragmentation-blind, so a failed VMI can reschedule and fail again - crash-loop with no convergence or back-pressure.

This is concerning, since the VM owner has no real way to understand which node has enough backing huge pages, therefore the VM can crash over and over again.

That being said, we can revisit this in future cycles and leave it as an idea for now.

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+    thp:
+      coverage: "98.4%"
+      collapsedBytes: 19394658304
+      totalBytes: 19713515520
+      collapseTime: "2.3s"
+  conditions:
+    - type: THPCollapsed
+      status: "True"
+      message: "98.4% of guest memory backed by THP"
+```
```
iholder101: I think we can avoid the `thp` struct and the new condition.

WDYT about setting `phase: Failed`, a `reason: THPCollapseFailed` + a k8s event with more details?

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,512 @@
...
+hugepage allocation via sysfs (`nr_hugepages`):
+
+```yaml
+spec:
+  domain:
+    memory:
+      hugepages:
+        pageSize: "2Mi"
+        mode: thp
+        policy: best-effort
```
iholder101: I think `bestEffort` is more idiomatic

iholder101 (commented):
Thanks @michalskrivanek! Looks good!

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: the trouble is with the coverage calculation. it's done in s simplified way that covers the whole cgroup (qemu + launcher + hook pods) as getting the precise number would require walk through all VMAs to find out the guest memory only. It can be done, but I would leave that for future as I believe the approximation is close enough in practice - the guest RAM sizes are >> overhead.
A brief explanation of that tradeoff makes sense

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: well, two reasons for `thp`:
- it's under `hugepages` already
-`static` is also not `staticHugePages`, it could be called `transparent`, but to me `thp` was a bit more widely recognized. 
which of the three is clearer?

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: the "forced" collapse is addressing the fragmentation (at the price of startup latency). I think it's only relevant when we get very very close to the exact boundary.

veps/sig-compute/383-thp-memory-backing/vep.md
iholder101: Yeah, good point.

At least to me, `transparent` is clearer.

veps/sig-compute/383-thp-memory-backing/vep.md
iholder101: Agreed that the forced/synchronous collapse triggers compaction and mitigates fragmentation at the cost of startup latency. The residual concern is the case where the node genuinely lacks enough free/movable memory - compaction can't manufacture order-9 blocks that don't exist, so it still fails, and the VM owner has no signal for which node would work and do reschedule loop.

Fine to defer to a future cycle, just wanted it captured as a known limitation. Does this make sense?

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: right.
i meant it more like a decision if we need to expose it or not. I see an alternative to go through that more exact calculation path where we could settle on 100%, but I don't see it clearly yet.

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: sounds good to me

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: @iholder101 naming is hard. I would appreciate to make a call rather than just a suggestion, so that once and for all you'll be the one to blame:-D
`.memory.hugepages.mode:`  `static | thp` vs `staticHugePages | transparentHugePages` vs `static | transparentHugePages`

michalskrivanek: sorry missed the response, so `transparent` and `static`. Ok, looks good to me!

fabiand: Folks, why do we expose it on the VMI level at all? Can't we just keep it on the cluster level?
It's mostly an admin sided feature, there is no performance difference to a Vm if HPs are static ot HTP allocated.
Thus the over dominant aspect is operational simplicity: THP does not require considereations from admin, static allocations do.

What am I missing?

michalskrivanek: I think it comes down to guarantees. for "static" you explicitly consume from the preconfigured pool. The THP is not guaranteed at all in `bestEffort`, and even with the `guaranteed` is not really guaranteed when collapse fails, with the scheduler implications mentioned. 
You also can't control the sizes in any way, you get whatever the THP size is (or mTHP in near future)
I do not think it's a full substitute.

I can imagine static hugepages may become less and less used and eventually we may ditch them completely, but I wouldn't even target it before THPs prove sufficient.

fabiand: Even with THP and madv collapse we can guarantee iIUIC (aka if collapse fails, then HP failed). THus this is rather moving the time of error to a later time (from sched to runtime).

I think an opportunity to have one-off configurations would be to use an annotation liek it's used for FPR.

IOW I'm not convinced that we really need to add a new user/workload facing knob to control this behavior.

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: done

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: done
(unsure about the level of detail we can provide, I think it's fine to resolve during implementation)

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: done

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: done. it's fine to revisit in another alpha round if needed

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: that is the point of `guaranteed`, however I don't think we can really claim equivalency, not without resolving https://github.com/kubevirt/enhancements/pull/384#discussion_r3911831738 . I would leave that for beta or another alpha round. (it's also a stated non-goal right now)
I think it's perfectly reasonable to strive for it, it would indeed simplify things, I just don't see us being that close to it yet

michalskrivanek:
applied changes, 
additionally took out specific metrics. they don't necessarily need to be in core components and the suggested k8s event is a good starting point.
(...and whitespace, I apologize for changing line wrapping this late. the github's diff view is still readable though)

veps/sig-compute/383-thp-memory-backing/vep.md
fabiand: @michalskrivanek basic THP does not provide guarantees. madv collapse should provide it.

IMHO this VEp MUSt include madv collapse, because without it, there is nothing that kubevirt woul dneed to do.

Btw even the condition to report the coverage should not be required if collapse is there. because collapse tells us if its 100% or not.

veps/sig-compute/383-thp-memory-backing/vep.md
```diff
@@ -0,0 +1,328 @@
...
+- virt-controller: generate pod spec without hugepages resources/volumes for THP-mode VMs.
+- Domain converter: generate anonymous/locked/immediate XML.
+- virt-handler: set `RLIMIT_MEMLOCK`, run `MADV_COLLAPSE`, support both `bestEffort` and `guaranteed` policies.
+- Unit tests for all components.
+- E2E tests: startup without pre-allocated hugepages, >95% coverage, guaranteed-policy failure path.
+
+### Beta
+
+- E2E tests: NUMA binding, pre-copy migration.
+- Documentation.
```
iholder101, line 297: Can you please add something along the lines of:
- Revisit how `guaranteed` works, especially how to to avoid a late runtime crash.
- Revisit API (cluster-level vs. API-level)

iholder101 (commented):
just one request to document what we've discussed before I approve :)

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: sure, done

veps/sig-compute/383-thp-memory-backing/vep.md
michalskrivanek: right, `MADV_COLLAPSE` is indeed going to be always called, the only difference is in enforcement when it fails - i.e. it partially succeeded. It could be 99.9% and still fail...and that would be irrelevant to the overarching goal of improving VM performance.

jean-edouard (approved):

Issue #385, opened 2026-07-09
VEP 385: Guest Device Info Metrics


0xFelix:
**Primary contact (assignee)**:

/assign @0xFelix

**Current Feature Stage**:

New

**Feature Gate**:

None. This VEP extends [VEP #143](https://github.com/kubevirt/enhancements/tree/main/veps/sig-observability/143-monitoring-stack-separation),
defines no feature gate of its own, and enters at that pipeline's maturity.

**Responsible SIGs**:

Primary SIG:
/sig observability

Additional SIGs:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/386

**Timeline**:

* Alpha release target: n/a, enters at beta alongside VEP #143
* Beta release target (1.10):
* Stable release target:

* [ ] Beta
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/386
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

Issue #143, opened 2025-11-27
VEP-81: Monitoring Stack Separation via Hybrid Architecture


sradco:
**Primary contact (assignee)**:

/assign @machadovilaca 

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig observability

Additional SIGs (optional):
/sig $SIG

**Enhancement link**:
VEP PR: [kubevirt/enhancements#81](https://www.google.com/search?q=https://github.com/kubevirt/enhancements/pull/81)

**Timeline**:

* Alpha release target (x.y): v1.9
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
@sradco This tracker is severely outdated, and it's the second time I ping for that :)

Can you:
* Link the merged VEP document.
* Link the different VEP and Implementation PRs.
* Change the title

Thank you.

sradco:
@machadovilaca Quick question about the `GetVMStats` / `AgentData` design:

We're looking at exposing guest OS uptime as a new metric (`kubevirt_vmi_guest_uptime_seconds`) via the observability controller (context: [CNV-28799](https://redhat.atlassian.net/browse/CNV-28799) -- customer request for RHV-like uptime display and reboot events).

Today, the virt-launcher agent poller doesn't poll `/proc/uptime` from the guest agent. With VEP 143's architecture, what's the intended path for the observability controller to collect new guest agent data (like uptime) that isn't currently polled by the agent poller? Specifically:

1. Does the `AgentData` mechanism in `GetVMStats` support on-demand guest agent commands, or does it only return data that's already cached by the agent poller?
2. If only cached data, is the plan to make the agent poller configurable so the controller can specify which commands to poll -- or would adding new data sources still require a PR to kubevirt/kubevirt?

We'd like to leverage VEP 143 to add this without modifying kubevirt/kubevirt core, which we believe is the end-goal of this VEP's design.

PR #386, opened 2026-07-09
VEP 385: Guest Device Info Metrics

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/385
**SIG label**: /sig observability

### What this PR does

Adds VEP 385, which surfaces the Windows VirtIO driver information the QEMU guest agent already
reports through `guest-get-devices`:

- `kubevirt_vmi_guest_device_driver_date_seconds`, one series per device per VM, carrying driver
  name, version and device ID as labels and the driver date as the value, for fleet-wide inventory
- An optional ConfigMap in which a distributor or cluster administrator declares which driver
  versions count as current, exposed as `kubevirt_guest_device_driver_latest_version_info`
- A recording rule intersecting the two, and an `OutdatedGuestDeviceDrivers` alert over it

What counts as "current" is a support-policy decision KubeVirt cannot make, so it is supplied as
data at runtime. Upstream ships no mapping, so the alert cannot fire on a default installation.

- The VEP extends VEP #143 and defines no feature gate of its own, so it enters at that pipeline's
  maturity: beta for v1.10, with no alpha stage.
- Collection rides on VEP #143's VMStatsCollector path only. Emitting from virt-handler's
  domainstats scraper in parallel is a rejected alternative.
- It depends on one change to that path: an agent command the guest does not implement
  (`guest-get-devices` on Linux) must stop failing the VMI's entire stats response. Without it,
  requesting the category cluster-wide would drop every VMStatsCollector metric for every Linux VM.
- The driver date is the metric value rather than a label, departing from the `_info` convention to
  allow `time() - value` age queries. Discussed in Alternatives.
- The built-in alert is deliberately not age-based: a stable device's current driver can be over a
  year old, so any threshold produces false positives. Also in Alternatives.
- The guest-facing half is testable in kubevirt/kubevirt's Windows lane. The metric, the rule and
  the alert live in kubevirt-observability-controller and are covered there by unit and promtool
  tests, so the test plan is split at the repository boundary.

### DIFF
--- a/veps/sig-compute/385-guest-device-info/guest-device-info.md
+++ b/veps/sig-compute/385-guest-device-info/guest-device-info.md
@@ -1,0 +1,283 @@
+# VEP #385: Guest Device Info Metrics
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Expose device driver information reported by the QEMU guest agent's `guest-get-devices` command
+as a Prometheus metric, giving VM owners a fleet-wide overview of deployed VirtIO
+driver versions on Windows VMs. Outdated driver versions can cause performance degradation,
+compatibility issues, and may lack security fixes. This metric enables detecting such VMs via
+PromQL before problems arise.
+
+## Motivation
+
+Windows VMs running on KubeVirt use VirtIO drivers for network, storage, balloon, and other
+paravirtualized devices. These drivers require periodic updates for performance, bug fixes, and
+security patches. Currently there is no centralized way to determine which driver versions are
+installed across a cluster - VM owners must check each VM individually.
+
+The QEMU guest agent already supports a `guest-get-devices` command (since QEMU 5.2) that returns
+driver name, version, date, and PCI device IDs. KubeVirt has existing infrastructure to poll the
+guest agent and expose results as Prometheus metrics. This VEP bridges the gap by adding
+`guest-get-devices` to the metrics pipeline.
+
+## Goals
+
+- Expose per-device driver information (name, version, device ID) as a Prometheus metric with
+  the driver date as the metric value, enabling both inventory queries and time-based alerting
+- Document example PromQL queries for fleet inventory, version distribution, and age-based alerting
+
+## Non Goals
+
+- Automatically updating drivers inside guest VMs
+- Defining what constitutes the "latest" or "correct" driver version
+- Shipping built-in alert rules (users write their own with documented PromQL examples)
+- Adding device info to VMI status or the Kubernetes API
+- Supporting non-PCI device types beyond what `guest-get-devices` reports
+- Supporting Linux guests (the command is Windows-only in the QEMU guest agent)
+
+## Definition of Users
+
+- **VM owners** managing Windows VMs with VirtIO drivers who need visibility into driver versions
+  for update planning, fleet health monitoring, and compliance
+- **Compliance/audit tools** that need to verify driver versions meet organizational policies
+
+## User Stories
+
+- As a VM owner, I want to query Prometheus to see which driver versions are installed on my
+  Windows VMs so I can identify VMs that need driver updates.
+- As a VM owner, I want to create a Grafana dashboard showing driver version distribution across
+  my VM fleet so I can track update rollout progress.
+- As a VM owner, I want to write Prometheus alerts that fire when VMs are running drivers older
+  than a policy threshold or a specific version known to have bugs, so I can prioritize updates.
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+
+## Design
+
+### QEMU Guest Agent Command
+
+The `guest-get-devices` command is a Windows-only QEMU guest agent command available since
+QEMU 5.2. It returns VirtIO and QEMU PCI device driver information including name, version,
+date, and PCI device/vendor IDs. A typical Windows VM reports 6-10 devices.
+
+Example output:
+```json
+{
+  "return": [
+    {
+      "driver-date": 1721001600000000000,
+      "driver-name": "Red Hat VirtIO Ethernet Adapter",
+      "driver-version": "100.95.104.26200",
+      "id": {
+        "device-id": 4161,
+        "vendor-id": 6900,
+        "type": "pci"
+      }
+    }
+  ]
+}
+```
+
+The command is Windows-only. On Linux guests it returns `CommandNotFound`, which the existing
+agent command infrastructure handles gracefully. No metric is emitted for VMs without device data.
+
+### Metric Definition
+
+**Name:** `kubevirt_vmi_guest_device_driver_date_seconds`
+
+**Type:** Gauge (using `operatormetrics.NewGauge`; labels set via `CollectorResult` per the domainstats pattern)
+
+**Labels:**
+
+| Label | Example | Source |
+|-------|---------|--------|
+| `node` | `worker-1` | VMI status (standard) |
+| `namespace` | `default` | VMI metadata (standard) |
+| `name` | `win-vm-1` | VMI metadata (standard) |
+| `driver_name` | `Red Hat VirtIO Ethernet Adapter` | `driver-name` |
+| `driver_version` | `100.95.104.26200` | `driver-version` |
+| `device_id` | `1041` | `id.device-id` (decimal to hex, no 0x prefix) |
+
+**Value:** Driver date as Unix epoch seconds (converted from the nanosecond timestamp).
+
+Using the driver date as the metric value enables `time() - value` arithmetic in PromQL for
+age-based alerting (precedent: `kubevirt_vmi_migration_start_time_seconds`).
+
+Labels omitted by design:
+- `vendor_id`: always `1af4` or `1b36` (QEMU hard-codes the filter) - low information density
+- `device_type`: always `pci` (the only type QEMU supports) - no information
+
+### Cardinality
+
+The `guest-get-devices` command already filters to VirtIO/QEMU vendor IDs only, so cardinality is
+inherently bounded to ~6-10 devices per Windows VM. With 1000 VMs this yields ~6000-10000 time
+series - comparable to existing per-item metrics like `kubevirt_vmi_filesystem_capacity_bytes`
+(one per filesystem per VM) and `kubevirt_vmi_vnic_info` (one per vNIC per VM).
+
+### Data Path
+
+The command is added to both data pipelines. The domainstats scraper path is gated behind the
+`GuestDeviceMetrics` feature gate since it introduces a new Prometheus metric. The VMStatsCollector
+path requires only the existing `VMStatsCollector` feature gate - no additional gate is needed
+because it simply passes through raw guest agent data, consistent with how other guest agent
+commands are handled in that pipeline. Driver info is essentially static during a VM's lifetime,
+so infrequent polling is sufficient.
+
+**Domainstats scraper path** (Prometheus metric, gated by `GuestDeviceMetrics`):
+```
+guest-get-devices (QEMU GA)
+  --> virt-launcher agent poller: periodically polls, parses JSON,
+      stores in AsyncAgentStore
+  --> virt-launcher cmd-server: GetDevices() reads from store
+  --> virt-handler domainstats scraper: fetches via gRPC cmd-client
+  --> device_metrics.go: emits CollectorResult per device
+  --> Prometheus scrape
+```
+
+**VMStatsCollector path** (gRPC endpoint, gated by `VMStatsCollector`):
+```
+guest-get-devices (QEMU GA)
+  --> virt-launcher: cached in agentDataCaches with per-command TTL
+  --> GetVMStats gRPC: returns raw JSON on demand
+```
+
+## API Examples
+
+### Metric Output
+
+```
+kubevirt_vmi_guest_device_driver_date_seconds{
+  node="worker-1", namespace="default", name="win-vm-1",
+  driver_name="Red Hat VirtIO Ethernet Adapter",
+  driver_version="100.95.104.26200",
+  device_id="1041"
+} 1721001600
+```
+
+### Example PromQL Queries
+
+List all drivers for a specific VM:
+```promql
+kubevirt_vmi_guest_device_driver_date_seconds{namespace="default", name="win-vm-1"}
+```
+
+Find all VMs with a specific device:
+```promql
+kubevirt_vmi_guest_device_driver_date_seconds{device_id="1041"}
+```
+
+Find all VMs running a specific driver version:
+```promql
+kubevirt_vmi_guest_device_driver_date_seconds{driver_version="100.95.104.26200"}
+```
+
+Find VMs with drivers older than 1 year:
+```promql
+(time() - kubevirt_vmi_guest_device_driver_date_seconds) > (365 * 24 * 3600)
+```
+
+Driver version distribution across cluster:
+```promql
+count by (driver_name, driver_version) (kubevirt_vmi_guest_device_driver_date_seconds)
+```
+
+## Alternatives
+
+### Add Device Info to VMI Status
+
+Device info could be added to `VirtualMachineInstanceStatus` (like `GuestOSInfo` today).
+Rejected because it adds API surface and VMI status update overhead. The primary use case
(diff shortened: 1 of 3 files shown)


0xFelix:
@sradco @machadovilaca Could you give this a look? :)

veps/sig-compute/385-guest-device-info/guest-device-info.md
```diff
@@ -0,0 +1,278 @@
...
+- As a VM owner, I want to write Prometheus alerts that fire when VMs are running drivers older
+  than a policy threshold or a specific version known to have bugs, so I can prioritize updates.
+
+## Repos
+
+- https://github.com/kubevirt/kubevirt
+
+## Design
+
+### QEMU Guest Agent Command
```
codingben, line 76: Who is going to execute this command? I assume that this is behind KubeVirt API as usual we do, can you please demonstrate output of the API instead of JSON return from the command itself?

veps/sig-compute/385-guest-device-info/guest-device-info.md
```diff
@@ -0,0 +1,278 @@
...
+- **Compliance/audit tools** that need to verify driver versions meet organizational policies
+
+## User Stories
+
+- As a VM owner, I want to query Prometheus to see which driver versions are installed on my
+  Windows VMs so I can identify VMs that need driver updates.
+- As a VM owner, I want to create a Grafana dashboard showing driver version distribution across
+  my VM fleet so I can track update rollout progress.
+- As a VM owner, I want to write Prometheus alerts that fire when VMs are running drivers older
+  than a policy threshold or a specific version known to have bugs, so I can prioritize updates.
```
codingben, line 68: @ksimon1 Do you plan to have Tekton task as well for this? to demonstrate the usage of it as an example.

veps/sig-compute/385-guest-device-info/guest-device-info.md
```diff
@@ -0,0 +1,278 @@
...
+PromQL before problems arise.
+
+## Motivation
+
+Windows VMs running on KubeVirt use VirtIO drivers for network, storage, balloon, and other
+paravirtualized devices. These drivers require periodic updates for performance, bug fixes, and
+security patches. Currently there is no centralized way to determine which driver versions are
+installed across a cluster - VM owners must check each VM individually.
+
+The QEMU guest agent already supports a `guest-get-devices` command (since QEMU 5.2) that returns
```
codingben, line 35: Do we expose it via KubeVirt API today?

veps/sig-compute/385-guest-device-info/guest-device-info.md
```diff
@@ -0,0 +1,278 @@
...
+- All changes are guarded by a feature gate (`GuestDeviceMetrics`)
+- No changes to VMI spec or status - the metric is collected and exposed independently
+- No impact on VM runtime behavior
+
+## Functional Testing Approach
+
+- Unit tests for JSON parsing of `guest-get-devices` response
+- Unit tests for metric emission with mock device data (following the pattern in
+  `filesystem_metrics_test.go`)
+- e2e testing with a real Windows VM is not feasible in kubevirt/kubevirt CI due to Windows
```
codingben, line 257: I think that generic test of invoking this command under KubeVirt API would be enough for upstream, do we plan to do API testing here?

codingben (commented):
I'm not sure I understand why we're missing KubeVirt API usage here to abstract this command from the user. I think we usually don't do this, right?

veps/sig-compute/385-guest-device-info/guest-device-info.md
0xFelix: It will be executed and exposed through the Domainstats scraper and VMStatsCollector paths, no direct way for users to execute the command.

veps/sig-compute/385-guest-device-info/guest-device-info.md
0xFelix: A task could gather the current versions through the VMStatsCollector path and check if they're outdated. But IMO a custom Prometheus alert is more suited for that.

veps/sig-compute/385-guest-device-info/guest-device-info.md
0xFelix: No and the plan is not to. The plan is to scrape the output of this command, like other QGA commands are scraped for metrics.

veps/sig-compute/385-guest-device-info/guest-device-info.md
0xFelix: There's not explicit way to trigger this, since it is not exposed through an API. The only way to tell if it works is to check for values being reported from a running Windows guest. However, doing that is difficult because of licensing issues.

iholder101:
> @sradco @machadovilaca Could you give this a look? :)

:+1: 
Specifically, how will it play out with https://github.com/kubevirt/enhancements/issues/143?

machadovilaca:
> > @sradco @machadovilaca Could you give this a look? :)
> 
> 👍 Specifically, how will it play out with #143?

overall looks good, and I see there is a plan to add the command to VMStatsCollector as part of this work

when migrating the rest of the metrics in the future, I can take care of the changes needed in the controller repository

veps/sig-compute/385-guest-device-info/guest-device-info.md
```diff
@@ -0,0 +1,279 @@
...
+  by e2e tests for other commands. Full e2e validation can be done downstream by vendors with
+  access to Windows images.
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
+- [ ] Feature gate `GuestDeviceMetrics` guards all code changes
```
codingben: Why feature gate is required for this feature, and not enabled by default? It's about adding a new command that will expose more VM info and not affecting the VM's lifecycle or some kind of other VM behaviour that can be considered as destructive.

codingben: Is there a KubeVirt community rule or something set by maintainers that tells us to keep any new feature behind gate?

0xFelix: In order to prevent regressions a FG can be very useful. We can keep it turned off, if we encounter issues with it. Also, it signals that this feature is still in graduation and therefore may be subject to change, not stable yet.

0xFelix:
> overall looks good, and I see there is a plan to add the command to VMStatsCollector
when migrating in the future, I can take care of the changes needed in the controller

@machadovilaca My idea was to add support to VMStatsCollector right away, is it something you would still hold off from at this point?

veps/sig-compute/385-guest-device-info/guest-device-info.md
iholder101: Not to talk about this specific case, but generally speaking, it makes total sense to add an FG even though the feature is disabled by default. The reasoning is that an FG tells the user: "this is dangerous and might cause problems, beware, and enable this on your responsibility!".

iholder101:
> > overall looks good, and I see there is a plan to add the command to VMStatsCollector
> > when migrating in the future, I can take care of the changes needed in the controller
> 
> @machadovilaca My idea was to add support to VMStatsCollector right away, is it something you would still hold off from at this point?

Right away but guarded with an FG?
If the answer is yes, then I think the best path would be that by the time this goes to Beta it will be folded into the effort of https://github.com/kubevirt/enhancements/issues/143, hence living in the monitoring repo.

WDYT?

machadovilaca:
> > > overall looks good, and I see there is a plan to add the command to VMStatsCollector
> > > when migrating in the future, I can take care of the changes needed in the controller
> > 
> > 
> > @machadovilaca My idea was to add support to VMStatsCollector right away, is it something you would still hold off from at this point?
> 
> Right away but guarded with an FG? If the answer is yes, then I think the best path would be that by the time this goes to Beta it will be folded into the effort of #143, hence living in the monitoring repo.
> 
> WDYT?

agreed, that was what I meant
and VMStats is already behind a FG, so a new one isn't needed for VMStats imo

what i meant in the original message was that I, in the future, would do the changes needed in https://github.com/kubevirt/kubevirt-observability-controller

0xFelix:
Right away in a sense of during this cycle but behind a FG yes. I understood that @machadovilaca proposed to implement it in a future development cycle?

iholder101:
Thank you @0xFelix @machadovilaca, sounds good to me!

@machadovilaca we said that during the next cycle we'll revisit all of the GA commands holistically, let's just remember to also consider this one when we do so.

0xFelix:
Just one more question about VMStats: Add it to VMStatsCollector without an additional FG (requiring only the VMStats FG to use it) or also behind another FG (requiring VMStats + GuestDeviceMetrics FGs)?

machadovilaca:
> Just one more question about VMStats: Add it to VMStatsCollector without an additional FG (requiring only the VMStats FG to use it) or also behind another FG (requiring VMStats + GuestDeviceMetrics FGs)?

for VMStats I don't think another FG is needed, it's already behind a FG, and fields are specifically requested

iholder101:
> > Just one more question about VMStats: Add it to VMStatsCollector without an additional FG (requiring only the VMStats FG to use it) or also behind another FG (requiring VMStats + GuestDeviceMetrics FGs)?
> 
> for VMStats I don't think another FG is needed, it's already behind a FG, and fields are specifically requested

Makes sense, but OTOH, if we already need an FG, why not to add it everywhere where this code is presented? When the first FG becomes beta (and on by default) the second one could be still alpha. But I don't have a super strong opinion on this.

0xFelix:
Open to both but don't want to overcomplicate it. Both data paths won't share a lot of code, so IMO it should be okay to regard them as separate "mini" features? Updated the VEP doc to reflect that.

codingben (approved):
/lgtm

codingben:
> Pull requests that are marked with lgtm should receive a review from an approver within 1 week.

@vladikr @xpivarc Do you plan to review it?

0xFelix:
@machadovilaca Does it look good to you? Can you lgtm too?

machadovilaca (approved):

0xFelix:
Rebased to follow https://github.com/kubevirt/enhancements/pull/339

codingben (approved):
/lgtm

Thanks Felix! Looks good from my side.

codingben (approved):
/lgtm

0xFelix:
@sradco @machadovilaca @avlitman Please take a look at the recent changes

codingben:
What is blocking us from merging this VEP?

veps/sig-compute/385-guest-device-info/vep.md
```diff
@@ -0,0 +1,479 @@
...
+(monitoring/alerting) is better served by Prometheus metrics. VMI status can be reconsidered
+in a future VEP if API access is needed.
+
+### Add Labels to Existing `kubevirt_vmi_info`
+
+Driver info could be added as labels on `kubevirt_vmi_info`. Rejected because there are multiple
+devices per VM, which would duplicate the entire `kubevirt_vmi_info` series per device and
+massively increase cardinality of the central VMI info metric.
+
+### Use Only VMStats REST Endpoint
```
iholder101: I talked to @machadovilaca offline about this alternative. Is there a reason why it's not good enough for this use-case?

0xFelix: You are right, I talked to @machadovilaca as well and we agreed to move this mostly to VMStats and outside of kubevirt/kubevirt. I'd still like to keep this VEP as a technical documentation of the feature though. Please have a look.

veps/sig-observability/385-guest-device-info/vep.md
```diff
@@ -0,0 +1,502 @@
...
+
+### Data Path
+
+`guest-get-devices` becomes another guest agent category on the collection path VEP #143 defines,
+requested and cached like the categories already on it. The contract change is one request field and
+one response field, both named `guestGetDevices`; collection, caching, aggregation, transport and
+scrape are unchanged. `kubevirt-observability-controller` parses the payload and emits one sample
+per device.
+
+Driver information is static for the life of a VM, so the command takes the longest time-to-live the
```
michalskrivanek: why is it static? what if I update the drivers in Windows?

0xFelix: Indeed, mostly static. Not expected to change often, but yes it may change. Therefore a long interval.

veps/sig-observability/385-guest-device-info/vep.md
```diff
@@ -0,0 +1,522 @@
...
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: n/a, see below
+- This VEP targets beta for version: v1.10
+- This VEP targets GA for version:
+
+This VEP extends [VEP #143](../143-monitoring-stack-separation/vep.md), defines no feature gate of
+its own, and enters at that pipeline's maturity. VEP #143 targets beta for v1.10; so does this.
```
iholder101: I believe this is incorrect.

https://github.com/kubevirt/enhancements/blob/9d6073199aff68e34fd5d69893d25346de2bd5ee/veps/sig-observability/143-monitoring-stack-separation/vep.md#target-releases

Maybe you can just say we don't need to re-track target versions here, but look at VEP 143.

veps/sig-observability/385-guest-device-info/vep.md
```diff
@@ -0,0 +1,522 @@
...
+
+The next section, [kubevirt/kubevirt Changes](#kubevirtkubevirt-changes), is
+self-contained. A reviewer of kubevirt/kubevirt needs nothing else in this
+document.
+
+## kubevirt/kubevirt Changes
+
+Two changes, both on the stats collection path VEP #143 defines: one more guest agent category, and
+tolerance for a category the guest agent does not implement. The device payload crosses the repository
+boundary as an opaque field and is parsed in `kubevirt-observability-controller`.
```
iholder101, line 109: It's nice to have this here as an informative information. However I still think it'd be better to have a small VEP-update to VEP 143 to include it there as well, so every VEP is entirely self-contained. You can reference this VEP as a reasoning for adding it, but VEP 143 must be the source-of-truth w.r.t. the guest commands `GetVMStats()` supports.

Does this make sense?

veps/sig-compute/385-guest-device-info/vep.md
iholder101: Thank you very much! Looks much better now!

veps/sig-observability/385-guest-device-info/vep.md
0xFelix: Right, updated it to point to VEP 143 instead.

veps/sig-observability/385-guest-device-info/vep.md
0xFelix: I made a small addition to VEP143 as part of this PR in an additional commit. Please let me know if you'd like to have this as a separate PR.

iholder101: I think it's perfectly reasonable to do it in this PR :+1: 

@machadovilaca can you have a quick look and share your thoughts?

iholder101:
Looks good to me!
As [said in the past](https://github.com/kubevirt/kubevirt/pull/17550#issuecomment-4786335799), we're going to revisit the monitoring commands when VEP 143 graduates to Beta. For alpha this is perfectly fine IMO.

@machadovilaca I'd be happy if you can re-review. After you LGTM I'd be happy to approve.

machadovilaca (approved):

iholder101:
/approve

@0xFelix I'm holding just to let you an opportunity to ask more folks for reviews if that's really needed, as a lot of the VEP targets repos that are outside of kubevirt/kubevirt. Please feel free to remove the hold when you're happy.

/hold

veps/sig-observability/385-guest-device-info/vep.md
```diff
@@ -0,0 +1,519 @@
...
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved: tracked by VEP #143
+- [x] (R) Beta target version is explicitly mentioned and approved: tracked by VEP #143
+- [x] (R) GA target version is explicitly mentioned and approved: tracked by VEP #143
+
+## Overview
```
xpivarc, line 23: @vladikr @iholder101 I wonder if it wouldn't be more fit to put this under the existing folder of VEP 143 as this seems more like extension than standalone VEP

iholder101: I don't object.
I think that @0xFelix's point was that this VEP is mostly targeted to non-core repos, with a small modification to VEP 143, so it deserves a separate VEP.

I don't have a strong opinion on this TBH.

xpivarc: I guess we could have a dir structure where we group related VEPs.

0xFelix: This VEP proposes a feature built on the ground work that VEP 143 laid out. I don't think it's a direct extension of VEP 143, so a separate document makes sense to me.

Grouping might make sense, but can we think about thise some more in the next cycle?

0xFelix:
@iholder101 Since Lubo had another look and we have approval from sig-observability too, I'm going to lift the hold. :)

/hold cancel

Issue #389, opened 2026-07-14
VEP 389: Introduce Server-Side Apply


fossedihelm:
**Primary contact (assignee)**:

/assign @fossedihelm

**Current Feature Stage**:

New

**Feature Gate**:

N/A — SSA is an internal implementation detail, not a user-facing feature.

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/390

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/390
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/17435
  * [ ] Docs PRs:

**Additional context**:

Adopt Kubernetes Server-Side Apply (SSA) in KubeVirt by generating typed apply configurations for all KubeVirt API groups and wiring `Apply`/`ApplyStatus` methods into the generated clientset.

SSA is not a blanket replacement for all API writes — each interaction is evaluated individually to determine whether SSA is the right fit. The goal is to make SSA available as a first-class option and adopt it where it provides clear benefits (clean field ownership, reduced conflict retries).

POC: https://github.com/kubevirt/kubevirt/pull/17435

> [!IMPORTANT]
> Please keep this description up to date.

lyarwood:
/sig control-plane

@fossedihelm should this be under `sig-control-plane`?

fossedihelm:
> [@fossedihelm](https://github.com/fossedihelm) should this be under `sig-control-plane`?

You tell me 🙂

alaypatel07:
Sig-scale will be interested in this VEP as well.

PR #390, opened 2026-07-14
VEP-389: Introduce Server-Side Apply proposal

fossedihelm:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/389
**SIG label**: /sig compute

### What this PR does

Proposes adopting Kubernetes Server-Side Apply (SSA) in KubeVirt by generating typed apply configurations for all KubeVirt API groups and wiring `Apply`/`ApplyStatus` methods into the generated clientset.

SSA is not a blanket replacement for all API writes — each interaction is evaluated individually to determine whether SSA is the right fit. The goal is to make SSA available as a first-class option and adopt it where it provides clear benefits (clean field ownership, reduced conflict retries).

POC: https://github.com/kubevirt/kubevirt/pull/17435

### DIFF
--- a/veps/sig-compute/389-server-side-apply/vep.md
+++ b/veps/sig-compute/389-server-side-apply/vep.md
@@ -1,0 +1,466 @@
+# VEP #389: Introduce Server-Side Apply
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+Adopt Kubernetes [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/)
+(SSA) in KubeVirt by generating typed apply configurations for all
+KubeVirt API groups and wiring them into the generated clientset. This
+gives every KubeVirt controller a first-class `Apply` method, enabling
+declarative, field-level ownership. Whether a given API interaction
+should move to SSA, remain as a read-modify-write update, or use
+another pattern is evaluated case by case — SSA is not a blanket
+replacement for all update paths.
+
+## Motivation
+
+KubeVirt controllers today use the classic update loop: read the
+current object, mutate fields, write back the full object. This has
+several well-known problems:
+
+- **Conflict retries**: Two controllers updating the same resource race
+  on `resourceVersion`. The loser retries, re-reads, and re-applies —
+  wasted work that scales poorly under contention.
+- **Full-object writes are error-prone**: A controller that only cares
+  about `.status.conditions` must still send the entire object back,
+  risking accidental overwrites of fields managed by other actors.
+- **No field ownership tracking**: There is no way to know which
+  controller or user owns a given field. `kubectl apply` uses the
+  `last-applied-configuration` annotation as a client-side
+  approximation, but controllers typically bypass it entirely.
+- **Strategic merge patch limitations**: Client-side patches require
+  hand-crafting patch payloads and don't compose well when multiple
+  controllers patch the same resource.
+
+Server-Side Apply solves these by moving the merge logic into the API
+server and tracking per-field ownership via managed fields. Upstream
+Kubernetes has been using SSA for its own controllers since 1.22 and
+it has been GA since 1.18.
+
+## Goals
+
+- Generate typed apply configuration types for all KubeVirt API groups
+  (`core/v1`, `export/v1beta1`, `export/v1`, `snapshot/v1alpha1`,
+  `snapshot/v1beta1`, `instancetype/v1beta1`, `pool/v1alpha1`,
+  `pool/v1beta1`, `migrations/v1alpha1`, `clone/v1alpha1`,
+  `clone/v1beta1`, `backup/v1alpha1`)
+- Wire apply configurations into the generated clientset so every
+  resource type exposes an `Apply` and `ApplyStatus` method
+- Integrate `applyconfiguration-gen` into the existing code generation
+  pipeline (`hack/generate.sh`)
+- Enable incremental adoption of SSA across KubeVirt controllers,
+  where each API interaction is evaluated individually to determine
+  whether SSA is the right fit
+
+## Non Goals
+
+- Mandate SSA for all API interactions — each call site is evaluated
+  individually; some update patterns may remain as read-modify-write
+  or patch where SSA is not the best fit
+- Migrate all controllers in a single release — adoption is
+  incremental and controller owners decide when to switch
+- Change any user-facing API or behavior — SSA is an internal
+  implementation detail of how controllers write to the API server
+- Require SSA for external consumers of the KubeVirt client-go library
+  — the existing `Update` and `Patch` methods remain available
+- Deprecate or remove the existing update-based client methods
+
+## Definition of Users
+
+- **KubeVirt developers**: Primary consumers. SSA provides an
+  additional tool for controller logic, making field ownership explicit
+  and reducing boilerplate where appropriate.
+- **Downstream/third-party controller authors**: Consumers of
+  `kubevirt.io/client-go` who can adopt `Apply` methods at their own
+  pace.
+
+## User Stories
+
+- As a KubeVirt developer, I want to update only the fields my
+  controller owns without fetching and sending back the entire object,
+  so that my controller is simpler and doesn't conflict with other
+  controllers.
+- As a KubeVirt developer, I want the API server to track which
+  controller owns each field, so that ownership conflicts are surfaced
+  explicitly instead of silently overwritten.
+- As a downstream controller author, I want typed apply configuration
+  builders for KubeVirt resources, so that I can adopt SSA without
+  manually crafting patch payloads.
+
+## Repos
+
+- kubevirt/kubevirt
+
+## Design
+
+### Code generation
+
+`applyconfiguration-gen` and `client-gen` are integrated into the
+existing `hack/generate.sh` pipeline to produce typed apply
+configurations and `Apply`/`ApplyStatus` clientset methods for all
+KubeVirt API groups. Generated output lands under
+`staging/src/kubevirt.io/client-go/applyconfigurations/`.
+
+### Adopting SSA in controllers
+
+SSA is not a blanket replacement for every API write. Each API
+interaction must be evaluated individually to determine whether SSA
+is appropriate. Factors to consider include:
+
+- **Field ownership boundaries**: SSA works best when a controller
+  owns a well-defined, non-overlapping set of fields. When multiple
+  actors legitimately write to the same fields (e.g. spec fields
+  shared between the user and a mutating webhook), SSA ownership
+  semantics require careful consideration.
+- **Read-then-write dependencies**: If the new value depends on the
+  current state of the object (e.g. incrementing a counter, toggling
+  a condition based on current conditions), a read-modify-write with
+  conflict retry may still be the correct pattern.
+- **Atomic multi-field updates**: When several fields must be updated
+  together as a consistent snapshot, `Update` with optimistic
+  concurrency provides a stronger consistency guarantee than SSA,
+  where fields are merged independently.
+
+#### `Force: true` convention
+
+Following the
+[Kubernetes SSA documentation](https://kubernetes.io/docs/reference/using-api/server-side-apply/#using-server-side-apply-in-a-controller),
+all controller Apply calls must use `Force: true`. This is the
+recommended practice for controllers: it ensures that a controller
+can always reclaim ownership of the fields it manages, even if
+another actor (e.g. a previous version of the same controller using
+`Update`) has taken ownership. Without `Force: true`, an Apply call
+that encounters a field ownership conflict returns an error instead
+of proceeding.
+
+#### Mixed SSA/Update coexistence during incremental adoption
+
+During incremental migration, some controllers will use Apply while
+others (or older versions of the same controller) still use Update.
+This mixed state is expected and safe:
+
+- **Update → Apply transition**: When a controller switches from
+  `Update` to `Apply` with `Force: true`, it takes ownership of the
+  fields it applies. Fields it previously wrote via `Update` but no
+  longer includes in its apply configuration become unowned — this is
+  intentional and must be reviewed during migration.
+- **Rolling updates (virt-handler DaemonSet)**: During a rolling
+  DaemonSet update, old pods use `Update` and new pods use `Apply` on
+  the same VMI resources. Because `Force: true` is used, the new pod's
+  Apply succeeds regardless of prior ownership. The old pod's `Update`
+  also succeeds because `Update` does not check managed fields. The
+  risk is that an old pod's `Update` may temporarily re-take ownership
+  of fields the new pod applied. This is transient — once the rollout
+  completes, only Apply-based pods remain. Each migration PR must
+  evaluate this window and confirm that the transient ownership flip
+  does not cause incorrect behavior.
+- **Different controllers on the same resource**: Two controllers
+  writing to different fields on the same resource — one via Apply,
+  one via Update — coexist without issue. The Apply controller owns
+  its fields; the Update controller implicitly owns all fields it
+  writes.
+
+Each adoption is a separate PR with a dedicated discussion to evaluate
+the tradeoffs for that specific call site. The PR should:
+
+1. Justify why SSA is the right choice for the targeted API interaction
+2. Replace `Update`/`UpdateStatus` with `Apply`/`ApplyStatus` where
+   appropriate
+3. Set a unique `FieldManager` name identifying the controller
+4. Adjust or remove retry-on-conflict loops only where SSA makes them
+   unnecessary
+5. Add/update tests
+
+#### Initial adoption candidates
+
+The following API interactions are identified as strong candidates for
+SSA adoption. Each has a single controller owning the status fields,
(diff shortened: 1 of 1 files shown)


veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,321 @@
...
+  current state of the object (e.g. incrementing a counter, toggling
+  a condition based on current conditions), a read-modify-write with
+  conflict retry may still be the correct pattern.
+- **Atomic multi-field updates**: When several fields must be updated
+  together as a consistent snapshot, `Update` with optimistic
+  concurrency provides a stronger consistency guarantee than SSA,
+  where fields are merged independently.
+
+Each adoption is a separate PR with a dedicated discussion to evaluate
+the tradeoffs for that specific call site. The PR should:
```
iholder101, line 180: This makes sense. However, do we want to come up with a list of candidates for such PRs, even if the final discussion will be made on the PRs themselves?

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,321 @@
...
+- [ ] No regressions in E2E test suite
+- [ ] Field ownership visible and correct in managed fields metadata
+- [ ] Downstream consumers documented on how to use apply configurations
+
+#### On-By-Default Readiness
+
+SSA is an internal implementation detail, not a feature-gated user
+feature. There is no feature gate — controllers are migrated via code
+changes. "On-by-default" applies implicitly once a controller is
+migrated.
```
iholder101: I feel this justification needs strengthening.

Although it's not a user-facing feature, it is a big and fundamental change that has its risks. Are we sure we don't want to feature-gate, just to be sure we don't introduce regressions?

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,321 @@
...
+existing `hack/generate.sh` pipeline to produce typed apply
+configurations and `Apply`/`ApplyStatus` clientset methods for all
+KubeVirt API groups. Generated output lands under
+`staging/src/kubevirt.io/client-go/applyconfigurations/`.
+
+### Adopting SSA in controllers
+
+SSA is not a blanket replacement for every API write. Each API
+interaction must be evaluated individually to determine whether SSA
+is appropriate. Factors to consider include:
```
iholder101, line 124: The section should address mixed SSA/Update coexistence during incremental adoption. The [K8s SSA docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/) strongly recommend `Force: true` for controllers and warn against mixing Apply and Update on the same fields. Worth stating `Force: true` as the convention here and discussing the implications for incremental migration - especially for virt-handler's rolling DaemonSet updates where old/new versions may write to the same VMI simultaneously.

iholder101 (commented):
Thanks @fossedihelm!

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: I don't think we need to provide such details in the VEP, but I can provide something

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: A feature gate would require maintaining two code paths (Apply and Update) for every migrated call site, doubling the testing surface and negating much of the simplification SSA provides.
The risk of regressions is mitigated by the adoption model itself:

- Each migration is a small, scoped PR targeting a single API interaction with dedicated review
- Every PR must pass the full e2e suite before merge
- Rollback is a simple code revert of a contained change; no  cluster-wide flag to flip
- SSA is a stable, GA Kubernetes API mechanism (since 1.18); the risk is in the migration, not the mechanism

Last but not least, how will we handle the fg during incremental adoption? a FG per controller? one single FG, but it will not cover every controller?

Also, I really believe that every controller needs a specific discussion; Critical point on VMI resource, where the status is updated by virt-controller + virt-handler (which actually is vm+migration-source+migration-target).

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: :+1:

fossedihelm:
Thank you @iholder101 comments addressed.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,393 @@
...
+#### Initial adoption candidates
+
+The following API interactions are identified as strong candidates for
+SSA adoption. Each has a single controller owning the status fields,
+with no cross-controller write contention. Final evaluation happens in
+the individual PRs.
+
+| Controller | File | Fields | Why |
+|------------|------|--------|-----|
+| VirtualMachineInstanceReplicaSet | `pkg/virt-controller/watch/replicaset/replicaset.go` | `Status.Replicas`, `ReadyReplicas`, `LabelSelector`, `Conditions` | Pure compute-and-set pattern. Single owner, no retry loop. Simplest controller in the codebase. |
```
lyarwood, line 199: The initial candidates all list `Conditions` as an SSA-managed field. With SSA, list merge behavior depends on the OpenAPI schema having `listType=map` with `listMapKey=type` on the conditions slice. If the KubeVirt CRDs don't have the correct list merge key annotations, an Apply that sends a single condition could take ownership of (and wipe) the entire conditions list.

Could you add a note in the Design section addressing this? Specifically:
- Whether the KubeVirt CRD schemas already have the right `x-kubernetes-list-type: map` / `x-kubernetes-list-map-keys: [type]` annotations on condition slices
- If not, that adding them is a prerequisite before any controller can safely Apply individual conditions

This is probably the single biggest footgun in SSA adoption and worth calling out explicitly.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,393 @@
...
+// on conflict: retry from the top
+```
+
+### Field manager conventions
+
+Each controller uses a descriptive, unique field manager name:
+
+- `virt-controller/<subsystem>` (e.g. `virt-controller/vm-controller`)
+- `virt-handler/<subsystem>`
+- `virt-api/<subsystem>`
```
lyarwood, line 278: Worth addressing what happens when a controller is renamed or refactored. Old managed field entries linger with the stale field manager name indefinitely. Should the VEP mention a cleanup strategy (e.g. an Apply with the old name containing no fields to release ownership, or documenting that stale entries are harmless)?

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,393 @@
+# VEP #389: Introduce Server-Side Apply
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
```
lyarwood: These should be filled in — the tracking issue says alpha targets v1.10.

lyarwood (commented):
Thanks @fossedihelm, this is a well-structured VEP and the incremental adoption approach is the right call. A few comments inline.

fossedihelm:
Thank you, @lyarwood, for the detailed review and comments.
I've addressed them and also added a paragraph about the `managedFields` size.
PTAL

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+
+## Motivation
+
+KubeVirt controllers today use the classic update loop: read the
+current object, mutate fields, write back the full object. This has
+several well-known problems:
+
+- **Conflict retries**: Two controllers updating the same resource race
+  on `resourceVersion`. The loser retries, re-reads, and re-applies —
+  wasted work that scales poorly under contention.
```
alaypatel07: Here are the two specific problems where server-side apply will help. When Update is used and there is a conflict, there are two general practices used today to resolve the conflict:
1. let the reconciler retry. This leads to unnecessary update calls that could take a while to succeed.
2. reconciler recognizes the conflict and makes a "get" call against apiserver. This is not scalable as the get call uses more resources then typical informer based cached get.

from sig-scale point of view, SSA is a well defined scalable solution to both of these issues. It might be good to catch instances of these and flag it as perfect fit for replacement if other conditions allow.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+- **Full-object writes are error-prone**: A controller that only cares
+  about `.status.conditions` must still send the entire object back,
```
alaypatel07, line 46: With things like conditions, it will require API annotation changes so that SSA is able to uniquely identify the exact item being replaced. The VEP should give an example of this. For example this type: https://github.com/kubevirt/kubevirt/blob/c70ff9f26ce6106cceaaf3c3de70743127fa9f88/staging/src/kubevirt.io/api/core/v1/types.go#L823 should include, `//+listMapKey` annotation in the type field

```
    //+listMapKey=map
    //+listMapKey=type
	Type   VirtualMachineInstanceConditionType `json:"type"`
```

alaypatel07: I jumped the gun, I see the review comment later: https://github.com/kubevirt/enhancements/pull/390/changes#r3638735357

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+| VirtualMachinePool | `pkg/virt-controller/watch/pool/pool.go` | `Status.Replicas`, `ReadyReplicas`, `LabelSelector`, `Conditions` | Nearly identical pattern to ReplicaSet. Single owner. |
+| VirtualMachineClone | `pkg/virt-controller/watch/clone/clone.go` | `Status.Phase`, `SnapshotName`, `RestoreName`, `TargetName`, `Conditions` | Clean state machine with well-defined phase transitions. Single owner. |
+| VirtualMachineExport | `pkg/storage/export/export/export.go` | `Status.Phase`, `ServiceName`, `Links`, `Conditions`, `TokenSecretRef` | Observes pod/service state, computes status. Single owner. |
+| VirtualMachineRestore | `pkg/storage/snapshot/restore.go` | `Status.Complete`, `RestoreTime`, `Conditions`, `Restores` | All callers funnel through single `doUpdateStatus`. Single owner. |
+| VirtualMachineSnapshot | `pkg/storage/snapshot/snapshot.go` | `Status.Phase`, `ReadyToUse`, `Error`, `Conditions` | Derives status from VMSnapshotContent. Single owner. |
+| KubeVirt Operator | `pkg/virt-operator/kubevirt.go` | `Status.Conditions`, `OperatorVersion`, `TargetKubeVirtVersion` | Single operator, single CR per cluster. Already separates status from metadata updates. |
```
alaypatel07, line 210: Are there no improvements to be made for VirtualMachinInstance Controller?

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+cached object carries its full `managedFields`.
+
+To mitigate memory pressure on the controller side, informer caches
+can strip `managedFields` from cached objects using a transform
+function, as demonstrated in
+[kubevirt/kubevirt#14243](https://github.com/kubevirt/kubevirt/pull/14243).
+This is safe when the controller interacts with the API server via
+`Apply` or `Patch` (which don't require the cached `managedFields`)
+but requires caution if the controller uses `Update` starting from a
+cached object — the update would persist the object without
```
alaypatel07, line 410: This is a much bigger issue, what happens when a kubevirt user has an external component which turn off managedFields for its informers and makes an update call say for example to label or annotation?

alaypatel07 (commented):
This is amazing to see, thanks for the VEP, added some comments inline, PTAL

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+
+Without these annotations, the list defaults to `atomic` merge
+semantics: an Apply that sends a single condition would take ownership
+of the entire conditions list and remove all other entries.
+
+KubeVirt CRDs today do not consistently carry these annotations on all
+condition slices. Adding the correct `x-kubernetes-list-type: map` and
+`x-kubernetes-list-map-keys: [type]` markers is a prerequisite before
+any controller can safely Apply individual conditions via SSA. This
+schema update must land before or alongside the first controller
```
lyarwood, line 232: This is a critical prerequisite. Should we add an explicit checklist item under **Alpha Graduation Requirements** for auditing and adding `// +listType=map` and `// +listMapKey=type` annotations to all condition slices across KubeVirt CRDs in `kubevirt.io/api`? Having it explicitly tracked ensures we don't accidentally merge a controller Apply PR before the CRD schemas support key-based list merging.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+migration that targets a conditions field.
+
+Each adoption PR that touches a list field must verify the
+corresponding CRD schema has the correct list-type annotation. The
+code generation pipeline should be extended to enforce this — or at
+minimum, the review checklist for adoption PRs must include CRD
+schema validation.
+
+**Not recommended for initial adoption**: VMI status updates in
```
lyarwood: Agree that core VMI phase updates are too complex for initial adoption given the virt-controller/virt-handler coordination. However, are there any isolated sub-controllers or non-conflicting fields in `virt-controller` (such as finalizer management or specific status sub-fields) that might be evaluated down the road? Clarifying that this exclusion applies specifically to shared phase/status reconciliation will help address @alaypatel07's question.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+emptyApply := applycorev1.VirtualMachine(name, namespace)
+_, err := client.Apply(ctx, emptyApply, metav1.ApplyOptions{
+    FieldManager: "old-controller-name",
+})
+```
+
+This cleanup can run once during the first reconcile after upgrade
+and be removed in the following release.
+
+### Interaction with existing features
```
lyarwood, line 320: Per the [VEP Writing Guide (PR #431)](https://github.com/kubevirt/enhancements/pull/431), please consider consolidating the various failure modes discussed throughout Design into a dedicated `### Risks and Mitigations` subsection. Specifically summarizing:
1. **List merge data loss:** If a CRD lacks `listType=map` annotations (mitigated by CRD schema audits/checks).
2. **Informer cache bloat:** `managedFields` growth in memory (mitigated by cache stripping transforms).
3. **Transient ownership flips:** During rolling DaemonSet updates where old `Update` and new `Apply` pods coexist (mitigated by `Force: true`).

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+object is modest, it adds up across large informer caches — every
+cached object carries its full `managedFields`.
+
+To mitigate memory pressure on the controller side, informer caches
+can strip `managedFields` from cached objects using a transform
+function, as demonstrated in
+[kubevirt/kubevirt#14243](https://github.com/kubevirt/kubevirt/pull/14243).
+This is safe when the controller interacts with the API server via
+`Apply` or `Patch` (which don't require the cached `managedFields`)
+but requires caution if the controller uses `Update` starting from a
```
lyarwood, line 409: Building on @alaypatel07's comment: it might be worth explicitly clarifying that when any client (internal or external) performs an `Update` without `managedFields` in metadata, the Kubernetes API server resets field ownership to that updating manager for all fields included in the update. This is standard SSA behavior and does not corrupt data, but reinforces why controllers shouldn't mix cache stripping with `Update` calls.

veps/sig-compute/389-server-side-apply/vep.md
```diff
@@ -0,0 +1,466 @@
...
+  interaction with dedicated review
+- Every PR must pass the full E2E suite before merge
+- Rollback is a simple code revert of a contained change — no
+  cluster-wide flag to flip
+- SSA is a stable, GA Kubernetes API mechanism (since 1.18) — the
+  risk is in the migration, not the mechanism
+
+### GA
+
+- [ ] SSA adopted wherever it was evaluated as the right fit
```
lyarwood: Per the [VEP Writing Guide (PR #431)](https://github.com/kubevirt/enhancements/pull/431), graduation criteria should be objective and verifiable. Consider making this more concrete by stating:
- [ ] All accepted candidate controllers from the initial candidate list (ReplicaSet, Pool, Snapshot, Export, Clone, Operator) have completed migration to SSA.
- [ ] Any remaining `Update`/read-modify-write call sites across core controllers have documented rationale for why SSA was not adopted.

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: This is bad behavior and the side effects are obviously bad. But everyone should be aware that `Update` should only be used by the controller owner.

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: @lyarwood Already answered here: https://github.com/kubevirt/enhancements/pull/390#discussion_r3864543113

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: > Agree that core VMI phase updates are too complex for initial adoption given the virt-controller/virt-handler coordination. However, are there any isolated sub-controllers or non-conflicting fields in virt-controller (such as finalizer management or specific status sub-fields) that might be evaluated down the road?

I think vm handler and vm controller really deserve isolated and dedicated inspection that I'd prefer to defer to the dedicated PR. WDYT?

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: Agree

veps/sig-compute/389-server-side-apply/vep.md
fossedihelm: :+1:

fossedihelm:
@alaypatel07 @lyarwood Thanks for the review. Addressed your comments. PTAL

fossedihelm:
@lyarwood @alaypatel07 I will not have the bandwidth to do all the things listed in the alpha graduation. Deferring this to 1.11

Issue #392, opened 2026-07-16
VEP 392: Remove Bazel from KubeVirt Build System


vamsikrishna-siddu:
**Primary contact (assignee)**:

/assign @vamsikrishna-siddu 

**Current Feature Stage**:

New

**Feature Gate**:

N/A (build system change, not a user-facing feature requiring a feature gate)

**Responsible SIGs**:

Primary SIG:
/sig buildsystem

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/393

**Timeline**:

* Alpha release target (x.y): TBD
* Beta release target (x.y): TBD
* Stable release target (x.y): TBD

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/393
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/18286
  * [ ] Docs PRs:

**Additional context**:

This enhancement aims to remove Bazel from the KubeVirt build system and replace it with Containerfiles, aligning KubeVirt with the approach taken by upstream Kubernetes.

Key motivations (from https://github.com/kubevirt/kubevirt/issues/14038):
- Lack of knowledge around Bazel in the community
- s390x is not supported by upstream Bazel
-  pp64le is not supported by upstream Bazel
- Staying close to Kubernetes (which already dropped Bazel)
- Bazel blocks usage of renovatebot/dependabot for automated dependency updates
- BUILD.bazel files break standard Go tooling

The initial PR decouples base images (which manage RPM dependencies via Bazel-invoked bazeldnf) from the KubeVirt component images, and replaces Bazel for component image builds, binary compilation, and functional tests using Containerfiles and standard `go build`/`go test` instead.
> [!IMPORTANT]
> Please keep this description up to date.

chandramerla:
> * s390x is not supported by upstream Bazel

Even ppc64le also isn't.

vamsikrishna-siddu:
> > * s390x is not supported by upstream Bazel
> 
> Even ppc64le also isn't.

Done @chandramerla

pkenchap:
@vamsikrishna-siddu , Thanks for proposing this VEP
Yes, ppc64le is not supported Bazel for building images in native arch . we are dependent on cross-compile to build the images. 
We are very much interested in contributing to this task .

PR #393, opened 2026-07-16
Remove Bazel from KubeVirt Build System

vamsikrishna-siddu:
### VEP Metadata

**Tracking issue**:  https://github.com/kubevirt/enhancements/issues/392
**SIG label**: /sig buildsystem

### What this PR does
Introduces VEP-392, proposing the complete removal of Bazel from the KubeVirt build system. The VEP outlines a phased migration strategy replacing Bazel with standard Go tooling (`go build`/`go test`) and Containerfiles for image builds, aligning KubeVirt with the approach taken by upstream Kubernetes.

- The initial draft implementation PR is already in progress: https://github.com/kubevirt/kubevirt/pull/18286
- RPM dependency management: invoking bazeldnf directly as a standalone tool (without Bazel) has been explored and is working. See https://github.com/kubevirt/kubevirt/pull/18534  for details. Final approach is open for discussion.
- This VEP is related to the discussion in https://github.com/kubevirt/kubevirt/issues/14038

### DIFF
--- a/veps/sig-buildsystem/392-remove-bazel/vep.md
+++ b/veps/sig-buildsystem/392-remove-bazel/vep.md
@@ -1,0 +1,213 @@
+# VEP #392: Remove Bazel from KubeVirt Build System
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.11.0 (development during v1.10.0 via fork)
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This VEP proposes the complete removal of Bazel from the KubeVirt project. Bazel is currently used across the entire build lifecycle: compiling Go binaries, running unit tests, running functional tests, building container images, managing RPM dependencies via bazeldnf, and cross-compiling for non-amd64 architectures. The existing Makefiles already serve as the developer-facing entry point but currently delegate to Bazel commands underneath.
+
+This VEP outlines a phased migration strategy replacing Bazel with standard tooling: `go build`/`go test` for compilation and testing, Containerfiles for image builds, and bazeldnf invoked as a standalone CLI (without Bazel) for RPM dependency resolution. The initial implementation PR demonstrates the full non-Bazel build path including binary compilation, unit/functional tests, base and component image builds, and multi-arch support, all without invoking Bazel.
+
+## Motivation
+
+Bazel has been deeply integrated into KubeVirt's build system since early in the project, handling binary compilation, testing, container image builds, RPM dependency management, and cross-compilation. While Bazel provides strong caching and reproducibility guarantees, it has become a significant source of friction and tech debt:
+
+- **Lack of community knowledge**: Very few contributors understand the Bazel build setup, creating a bus-factor risk and slowing down contributions.
+- **Architecture limitations**: s390x and ppc64le (Power) are not supported by upstream Bazel, requiring workarounds like cross-building on x86 or manually building Bazel for those platforms. This also makes it difficult for new architectures (e.g., RISC-V) to join the ecosystem in the future.
+- **Kubernetes divergence**: Kubernetes has already dropped Bazel. Staying aligned with Kubernetes tooling simplifies the contributor experience.
+- **Blocks automation**: `BUILD.bazel` files break standard Go tooling and prevent the use of renovatebot/dependabot for automated dependency updates. Security updates in release branches must be created manually. For example:
+  - Tools like `go mod tidy`, `go vet`, `golangci-lint`, and IDE features expect a standard Go module layout. However, `BUILD.bazel` files add non-standard dependency declarations that conflict when dependency graphs diverge between `go.mod` and Bazel's `WORKSPACE`.
+  - Dependabot and Renovatebot can automatically bump dependencies in `go.mod`, but they cannot update `BUILD.bazel` files (which requires running `gazelle`). As a result, dependency updates must be done manually.
+  - When a CVE is found, Dependabot can automatically patch every release branch. With Bazel, someone must manually run `gazelle` and update Bazel files on each branch, which is slow and error-prone for security-critical fixes.
+- **Contributor friction**: The non-standard build setup raises the barrier to entry for new contributors.
+
+Replacing Bazel with standard Go tooling and Containerfiles solves these by leveraging tools that every Go developer already knows, that work natively on all supported architectures, and that integrate seamlessly with dependency automation. Upstream Kubernetes made the same transition and the Go ecosystem provides all the primitives needed — native cross-compilation, build caching via `GOCACHE`, and standard module management.
+
+References:
+- https://github.com/kubevirt/kubevirt/issues/14038
+- Kubernetes KEP-2420 (prior art): https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2420-reducing-kubernetes-build-maintenance
+
+## Goals
+
+- Completely remove Bazel from the KubeVirt project
+- Replace Bazel-based image builds with standard multi-stage Containerfiles
+- Replace Bazel-based Go compilation with standard `go build`
+- Replace Bazel-based testing with standard `go test` and existing test frameworks
+- Decouple base image building from the main KubeVirt build process using Containerfiles
+- Remove Bazel from RPM dependency resolution using bazeldnf as a standalone CLI
+- Replace Bazel-based cross-compilation with Go's native cross-compilation
+- Replace Bazel's build caching with Go build cache (`GOCACHE`) and container layer caching
+- Maintain support for all currently supported architectures (amd64, arm64, s390x)
+- Enable standard Go tooling and dependency management workflows
+- Align with Kubernetes build patterns and tooling
+
+## Non Goals
+
+- Removing Bazel in a single PR (the migration is phased to minimize risk)
+- Changing the developer-facing Makefile interface (Make targets remain the same except `bazel-*` targets which will be removed, and only the underlying implementation changes)
+- Modifying the CI/CD pipeline architecture (only the build steps within it change)
+- Rewriting existing test logic (tests remain the same, only the test runner changes from Bazel to `go test`)
+
+## Definition of Users
+
+- **KubeVirt developers and contributors**: Benefit from a simpler, more approachable build system using standard tooling
+- **CI/CD maintainers**: Benefit from reduced complexity in build infrastructure
+- **Release engineers**: Benefit from easier backports and automated dependency updates
+- **New contributors**: Lower barrier to entry without needing to learn Bazel
+
+## User Stories
+
+- As a KubeVirt contributor, I want to build container images using standard Containerfiles so that I don't need Bazel expertise to make build changes.
+- As a CI maintainer, I want to use standard container build tools so that the build pipeline is easier to debug and maintain.
+- As a release engineer, I want to use dependabot/renovatebot so that security patches can be automated across release branches.
+- As a developer building for s390x, I want to build and test without having to build Bazel from source as a workaround for Bazel's lack of s390x support.
+- As a new contributor, I want to understand the build system quickly so that I can contribute without a steep learning curve.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+- [kubevirt/project-infra](https://github.com/kubevirt/project-infra) (prow job definitions need updating to use non-Bazel targets)
+
+## Design
+
+The migration is phased to incrementally replace each Bazel responsibility with standard tooling.
+
+This applies to the development (main) branch only. Existing release branches will not be modified and continue to use Bazel.
+
+### Development approach
+
+During the v1.10.0 cycle, implementation happens in a fork of `kubevirt/kubevirt` so that:
+- The existing Bazel build in `kubevirt/kubevirt` is not at risk of breakage during development
+- Bespoke prow jobs can be set up in `project-infra` targeting the fork, giving full CI signal on the new build path without flooding `kubevirt/kubevirt` presubmits or breaking existing jobs
+- The two build systems are completely isolated during development rather than needing a switching mechanism within the same repo
+
+Once validated, the fork's changes are merged into main targeting v1.11.0 for Alpha. Alpha and Beta changes are upstreamed from the fork to main.
+
+### Phase 1: Replace Bazel for builds and tests (Alpha)
+
+- **Base images**: Decouple RPM-dependent base images from the main build using Containerfiles — see [kubevirt#18286](https://github.com/kubevirt/kubevirt/pull/18286).
+- **Component images**: Each KubeVirt component image gets a multi-stage Containerfile that uses pre-built base images, copies in Go-compiled binaries, and configures the runtime environment.
+- **Binary compilation**: Replace Bazel `go_binary` targets with standard `go build` commands executed inside multi-stage Containerfiles.
+- **Testing**: Replace `go_test` Bazel targets with standard `go test` invocations. Unit tests and functional tests run directly via Go tooling.
+- **Cross-compilation**: Go's built-in cross-compilation (`GOOS`/`GOARCH`) for pure Go binaries. C cross-compilation for `cmd/container-disk-v2alpha/main.c` (requires a C cross-compiler toolchain). CGO cross-compilation for `virt-launcher` (libvirt) and `pkg/virt-handler/node-labeller/kvm-caps-info-plugin_amd64.go` (KVM capability detection, amd64 only). Multi-arch container builds via `podman build --platform` or `buildx`.
+- **RPM dependency resolution**: RPM resolution uses bazeldnf invoked as a standalone CLI, without Bazel. `hack/rpm-deps.sh` runs `bazeldnf fetch` and `bazeldnf rpmtree` to resolve and pin RPM packages, while `hack/rpm-base-images/generate-rpm-tars.sh` downloads the pinned RPMs and produces rootfs tars via `bazeldnf rpm2tar`. See [kubevirt#18534](https://github.com/kubevirt/kubevirt/pull/18534).
+
+### Dependency fetch resilience
+
+Today, Bazel's remote HTTP cache (`bazel-cache.kubevirt-prow.svc.cluster.local`) provides resilience against upstream unavailability: once a Bazel rule that fetches and processes an RPM or other external dependency has run successfully, its output is cached and subsequent builds never re-fetch from upstream. The non-Bazel build must provide equivalent resilience.
+
+**Go modules**: KubeVirt vendors all Go dependencies in the `vendor/` directory (checked into git). Builds use `-mod=vendor` and `GOPROXY=off`, so zero network access is needed to resolve or download Go modules at build time. This is unchanged by the Bazel removal and already matches the hermeticity Bazel provided for Go dependencies.
+
+**RPM packages**: RPM URLs and SHA checksums are pinned in `WORKSPACE` (checked into git), providing the same determinism as Bazel's `rpm()` rules. Each RPM entry in `WORKSPACE` already includes both an upstream CentOS mirror URL and a GCS mirror URL (`storage.googleapis.com/builddeps/<sha256>`). Base images are decoupled from the main build — regular CI pulls pre-built base images from the registry and never fetches RPMs directly. Only the "rebuild base images" workflow (triggered when RPM dependencies change) needs to download RPM packages. The `generate-rpm-tars.sh` script tries the GCS mirror first and falls back to upstream mirrors, providing the same resilience that Bazel's multi-URL fetching provided.
+
+**Container base images**: Base image references in Containerfiles will be pinned by digest (`@sha256:...`) rather than mutable tags. This ensures immutability and makes builds fail deterministically rather than silently pulling a changed image. The builder image is already pinned to a specific version.
+
+### Phase 2: Build caching and Bazel removal in fork (Beta)
+
+Without caching, the non-Bazel build path is functional but slower (~2x). This phase addresses performance and prepares for full Bazel removal:
+
+- **Go build cache**: Back up and restore `GOCACHE` in CI for incremental compilation
+- **Go test cache**: GCS-backed test cache, keyed per architecture (amd64/arm64/s390x). Periodic jobs run the full test suite hourly and upload the cache to GCS. Presubmit jobs restore this shared cache before running tests, significantly reducing wall-clock time when source hasn't changed. Inspired by the approach used by Kubernetes ([test-infra#16623](https://github.com/kubernetes/test-infra/pull/16623)).
+- **Container layer caching**: Leverage registry-based or local layer caching for image builds
+- **Bazel file removal in fork**: Remove all `BUILD.bazel` files, `WORKSPACE`, `.bazelrc`, and Bazel-related tooling in the fork for testing purposes. The main repo still retains Bazel files since Bazel-based builds continue to run until GA.
+
+### Phase 3: Stability and full migration (GA)
+
+Stability and performance are demonstrated to be comparable to Bazel. All Bazel-based prow jobs are permanently migrated to non-Bazel targets. Bazel files are removed from the main repository and `bazel-*` Make targets are dropped.
+
+## API Examples
+
+N/A — This is a build system change with no user-facing API impact.
+
+## Alternatives
+
+1. **Switch to another hermetic build system (e.g., Buck2, Please)**: Even less community familiarity than Bazel and doesn't align with the Kubernetes ecosystem.
+2. **Keep Bazel but invest in documentation and training**: Doesn't address the fundamental issues — s390x support, automated dependency updates, and tooling compatibility remain broken regardless of how well Bazel is documented.
+
+## Scalability
+
+Go's native compilation is fast and parallelizes well across CPU cores. Cross-compilation allows building all architecture variants from a single machine. Build caching via container layer caching and Go build cache (`GOCACHE`) is expected to bring performance close to Bazel's incremental builds.
+
+CI parallelism remains unchanged — jobs are split by SIG/component as before.
+
+## Update/Rollback Compatibility
+
+This change does not affect the runtime behavior of KubeVirt. The produced container images are functionally identical to Bazel-built images. Existing deployment manifests, operators, and upgrade paths remain unchanged.
+
+Backporting to older release branches that still use Bazel is expected to work without issues for pure code fixes (bug fixes, security patches), since the Go source code is build-system agnostic. However, backports that touch build infrastructure (new binaries, new dependencies, new container components) will require manual adaptation to work with the Bazel-based build in those branches.
+
+## Functional Testing Approach
+
+1. **Image build verification**: Verify all KubeVirt component images build successfully for all supported architectures (amd64, arm64, s390x).
+2. **E2E test suite**: Run the full existing e2e test suite against images built with Containerfiles. Each phase must pass the complete CI suite before merging.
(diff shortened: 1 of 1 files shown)


veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+### Phase 1: Decouple base images
+
+Base images that require RPM packages are decoupled from the main KubeVirt build. These are built independently using Containerfiles. The exact approach for RPM dependency resolution is TBD candidates include using bazeldnf standalone (without Bazel) or other methods.
+
+### Phase 2: Containerfile-based component images
+
+Each KubeVirt component (virt-operator, virt-api, virt-controller, virt-handler, virt-launcher) gets a multi-stage Containerfile that:
+- Uses the pre-built base images from Phase 1
+- Copies in Go-compiled binaries
+- Configures the runtime environment
```
nestoracunablanco, line 102: +1

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+Replace `go_test` Bazel targets with standard `go test` invocations. Unit tests and functional tests run directly via Go tooling.
+
+### Phase 5: Build caching strategy
+
+Before Bazel can be fully removed, its caching capabilities must be replaced:
+
+- **Go build cache**: Back up and restore `GOCACHE` in CI for incremental compilation
+- **Container layer caching**: Leverage registry-based or local layer caching for image builds
+- **Makefile-level caching**: Use file timestamps and checksums to skip unnecessary rebuild steps
```
nestoracunablanco: This is really cool. Do we have metrics/estimation about how big could be the gains?

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Phase 6: Cross-compilation
+
+Before Bazel can be fully removed, its cross-compilation capabilities must be replaced:
+
+- Go's built-in cross-compilation (`GOOS=linux GOARCH=arm64 go build`) for binaries
+- Multi-arch container builds via `podman build --platform` or `buildx` for images
```
nestoracunablanco: What do you think about implementing the optimizations (aka build caching strategy) after the cross-compilation?

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+Once caching, cross-compilation, and RPM dependency management are all handled without Bazel, remove all `BUILD.bazel` files, `WORKSPACE`, `.bazelrc`, and Bazel-related tooling from the repository.
+
+## API Examples
+
+N/A — This is a build system change with no user-facing API impact.
+
+## Alternatives
+
+1. **Switch to another hermetic build system (e.g., Buck2, Please)**: Even less community familiarity than Bazel and doesn't align with the Kubernetes ecosystem.
+2. **Keep Bazel but invest in documentation and training**: Doesn't address the fundamental issues — s390x support, automated dependency updates, and tooling compatibility remain broken regardless of how well Bazel is documented.
```
nestoracunablanco, line 139: Another point is that it makes difficult for another architectures (f.ex. RISC) to join the ecosystem.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+This change does not affect the runtime behavior of KubeVirt. The produced container images are functionally identical to Bazel-built images. Existing deployment manifests, operators, and upgrade paths remain unchanged.
+
+Backporting to older release branches that still use Bazel is expected to work without issues for pure code fixes (bug fixes, security patches), since the Go source code is build-system agnostic. However, backports that touch build infrastructure (new binaries, new dependencies, new container components) will require manual adaptation to work with the Bazel-based build in those branches.
+
+## Functional Testing Approach
+
+1. **Image build verification**: Verify all KubeVirt component images build successfully for all supported architectures (amd64, arm64, s390x).
+2. **E2E test suite**: Run the full existing e2e test suite against images built with Containerfiles. Each phase must pass the complete CI suite before merging.
+3. **Image parity**: Compare image contents between Bazel-built and Containerfile-built images to ensure functional equivalence.
```
nestoracunablanco: Container image immutability is complex, but multi-stage builds can help. The key is finding the sweet spot when comparing images.

nestoracunablanco (commented):
Looks solid, left some comments.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Goals
+
+- Completely remove Bazel from the KubeVirt project
+- Replace Bazel-based image builds with standard multi-stage Containerfiles
+- Replace Bazel-based Go compilation with standard `go build`
+- Replace Bazel-based testing with standard `go test` and existing test frameworks
+- Decouple base image building from the main KubeVirt build process using Containerfiles (RPM dependencies are still resolved via Bazel-invoked bazeldnf initially)
+- Remove Bazel from RPM dependency resolution once a suitable alternative approach is identified (TBD e.g., bazeldnf standalone or other method)
+- Replace Bazel-based cross-compilation with Go's native cross-compilation or multi-arch container builds
```
jschintag, line 53: I think it is also worth noting we have some C cross-compilation: https://github.com/kubevirt/kubevirt/blob/main/cmd/container-disk-v2alpha/main.c
and we should also identify which components need to use CGO.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+- Replace Bazel's build caching with Go build cache (`GOCACHE`) and container layer caching
+- Maintain support for all currently supported architectures (amd64, arm64, s390x)
+- Enable standard Go tooling and dependency management workflows
+- Align with Kubernetes build patterns and tooling
+- Retain the existing Makefile interface while replacing the Bazel commands underneath
```
jschintag, line 58: I think we should slowly drop all `bazel-*` make commands here as well. Otherwise it would be confusing.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+- [kubevirt/project-infra](https://github.com/kubevirt/project-infra) (prow job definitions need updating to use non-Bazel targets)
+
+## Design
+
+The migration is phased to incrementally replace each Bazel responsibility with standard tooling while keeping the existing Makefile interface stable.
+
+### Phase 1: Decouple base images
+
+Base images that require RPM packages are decoupled from the main KubeVirt build. These are built independently using Containerfiles. The exact approach for RPM dependency resolution is TBD candidates include using bazeldnf standalone (without Bazel) or other methods.
```
jschintag: Please list all options here.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Phase 3: Replace Bazel for binary compilation
+
+Replace Bazel `go_binary` targets with standard `go build` commands invoked from the Makefile. Go's native cross-compilation (`GOOS`/`GOARCH`) replaces Bazel's cross-compilation for arm64 and s390x.
+
+### Phase 4: Replace Bazel for testing
+
+Replace `go_test` Bazel targets with standard `go test` invocations. Unit tests and functional tests run directly via Go tooling.
+
+### Phase 5: Build caching strategy
```
jschintag: I think this phase should be split into it's components and implemented in each phase that affects it.
e.g. : Container layer caching in Phase 2, go cache in phase Phase 3

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
+# VEP #392: Remove Bazel from KubeVirt Build System
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
```
jschintag, line 7: When would we want to have this as alpha?

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,184 @@
...
+
+- This VEP targets alpha for version: 
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
```
jschintag: Nit: Should be checked.

jschintag (commented):
Thank you @vamsikrishna-siddu

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Without caching, the initial draft PR (kubevirt#18286) takes roughly 2x the current build time. I still need to measure the timing after implementing the caching strategy to see how much of that gap we can close. This is what iam planning to do now.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Good point . I'll add this to the Motivation section. Bazel's lack of support for new architectures (like RISC-V) makes it a blocker for future platform expansion as well.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Agreed, a byte-for-byte comparison isn't realistic. The focus will be on functional equivalence: verifying the same packages are installed, the same binaries are present, and the full e2e test suite passes. Multi-stage builds help keep the final image contents predictable.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Good catch. I've identified the following:
cmd/container-disk-v2alpha/main.c  pure C binary, needs a C cross-compiler toolchain for multi-arch builds
pkg/virt-handler/node-labeller/kvm-caps-info-plugin_amd64.go uses CGO for KVM capability detection (amd64 only)
Added this to Phase 5 in the VEP. Will ensure cross-compilation handles these correctly.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Agreed. The bazel-* Make targets will be removed as part of Phase 6 (Remove Bazel entirely) once all phases are complete. Updated the VEP to mention this.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: The option currently being explored is bazeldnf standalone (invoked directly without Bazel). Using dnf/microdnf directly in Containerfiles was considered but rejected it breaks reproducibility since RPM versions and dependencies can change between builds. bazeldnf provides the pinning and mirroring needed to keep builds deterministic. Still exploring if there are other viable alternatives.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Good point. It makes sense to address caching within each phase rather than as a separate step. Will restructure.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Not sure yet. it depends on when the initial PR (kubevirt#18286) where community reviews and says the approach is acceptable and when the VEP is approved. Will update the target version once there's more clarity on the timeline.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Done, thanks!

vamsikrishna-siddu:
@jschintag  @nestoracunablanco  I have addressed your review comment. Can you please review it again when you have some free time?

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Motivation
+
+Bazel has been deeply integrated into KubeVirt's build system since early in the project, handling binary compilation, testing, container image builds, RPM dependency management, and cross-compilation. While Bazel provides strong caching and reproducibility guarantees, it has become a significant source of friction and tech debt:
+
+- **Lack of community knowledge**: Very few contributors understand the Bazel build setup, creating a bus-factor risk and slowing down contributions.
+- **Architecture limitations**: s390x is not supported by upstream Bazel, requiring workarounds like cross-building on x86 or manually building Bazel for s390x. This also makes it difficult for new architectures (e.g., RISC-V) to join the ecosystem in the future.
+- **Kubernetes divergence**: Kubernetes has already dropped Bazel. Staying aligned with Kubernetes tooling simplifies the contributor experience.
+- **Blocks automation**: `BUILD.bazel` files break standard Go tooling and prevent the use of renovatebot/dependabot for automated dependency updates. Security updates in release branches must be created manually.
```
chandramerla: Can you put down an example or more details here to better understand what you mean by go tooling, dependency updates and security updates.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+Reference: https://github.com/kubevirt/kubevirt/issues/14038
+
+## Goals
+
+- Completely remove Bazel from the KubeVirt project
+- Replace Bazel-based image builds with standard multi-stage Containerfiles
+- Replace Bazel-based Go compilation with standard `go build`
+- Replace Bazel-based testing with standard `go test` and existing test frameworks
+- Decouple base image building from the main KubeVirt build process using Containerfiles (RPM dependencies are still resolved via Bazel-invoked bazeldnf initially)
+- Remove Bazel from RPM dependency resolution (approach TBD — bazeldnf standalone is being explored)
```
chandramerla, line 52: For dependency management, are only RPM (bazel-dnf) and go modules packages dependency was handled by bazel? Or any other package dependencies?

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+- **Kubernetes divergence**: Kubernetes has already dropped Bazel. Staying aligned with Kubernetes tooling simplifies the contributor experience.
+- **Blocks automation**: `BUILD.bazel` files break standard Go tooling and prevent the use of renovatebot/dependabot for automated dependency updates. Security updates in release branches must be created manually.
+- **Contributor friction**: The non-standard build setup raises the barrier to entry for new contributors.
+
+Replacing Bazel with standard Go tooling and Containerfiles solves these by leveraging tools that every Go developer already knows, that work natively on all supported architectures, and that integrate seamlessly with dependency automation. Upstream Kubernetes made the same transition and the Go ecosystem provides all the primitives needed native cross-compilation, build caching via `GOCACHE`, and standard module management.
+
+Reference: https://github.com/kubevirt/kubevirt/issues/14038
+
+## Goals
```
chandramerla, line 45: IMO, as we are removing bazel, which is optimised for the job what it has been doing, due to its complexity and other reasons, need to ensure that the alternatives we will going with are in acceptable trade-off for performance and for day-2 maintenance/operations. I see performance measurement is already tracked in VEP, but good to capture any details on later esp. if there is maintenance overhead.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+
+This change does not affect the runtime behavior of KubeVirt. The produced container images are functionally identical to Bazel-built images. Existing deployment manifests, operators, and upgrade paths remain unchanged.
+
+Backporting to older release branches that still use Bazel is expected to work without issues for pure code fixes (bug fixes, security patches), since the Go source code is build-system agnostic. However, backports that touch build infrastructure (new binaries, new dependencies, new container components) will require manual adaptation to work with the Bazel-based build in those branches.
+
+## Functional Testing Approach
+
+1. **Image build verification**: Verify all KubeVirt component images build successfully for all supported architectures (amd64, arm64, s390x).
+2. **E2E test suite**: Run the full existing e2e test suite against images built with Containerfiles. Each phase must pass the complete CI suite before merging.
+3. **Image parity**: Verify functional equivalence between Bazel-built and Containerfile-built images — same packages installed, same binaries present, same runtime behavior.
```
chandramerla, line 157: It will be interesting to see how images are checked for parity. As and when we know the details, it can be added.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+- 2026-07-21: Exploring bazeldnf standalone (without Bazel) for RPM dependency management [enhancements#393](https://github.com/kubevirt/enhancements/pull/393)
+
+## Graduation Requirements
+
+### Alpha
+- [ ] Base images (RPM-dependent) decoupled and built via Containerfiles (RPM resolution approach TBD)
+- [ ] All KubeVirt component images buildable via Containerfiles
+- [ ] Bazel no longer used for binary compilation
+- [ ] Bazel no longer used for functional tests
+- [ ] Bazel no longer used for unit tests
```
chandramerla: I think bazel will still be used in CI in the alpha stage, right?
Because I see non-bazel based caching is only sorted in Beta, so without caching I think CI still relies on bazel based binary compilation, func/unit tests. So "Bazel no longer used for " isn't correct. Please clarify.

chandramerla (commented):
@vamsikrishna-siddu Well written, Thanks!

Have added some comments/questions.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,185 @@
...
+- Replace Bazel's build caching with Go build cache (`GOCACHE`) and container layer caching
+- Maintain support for all currently supported architectures (amd64, arm64, s390x)
+- Enable standard Go tooling and dependency management workflows
+- Align with Kubernetes build patterns and tooling
+- Retain the existing Makefile interface while replacing the Bazel commands underneath
+
+## Non Goals
+
+- Removing Bazel in a single PR (the migration is phased to minimize risk)
+- Changing the developer-facing Makefile interface (Make targets remain the same, only the underlying implementation changes)
```
chandramerla: To be clear, better to mention as 

Changing the developer-facing Makefile interface (Make targets remain the same except bazel-* targets removed and only the underlying implementation changes)

veps/sig-buildsystem/392-remove-bazel/vep.md
jschintag: Interesting, i would have thought that virt-launcher would need CGO as well for linking to libvirt.

vamsikrishna-siddu: You are right. Virt-launcher also needs CGO for linking to libvirt.

nestoracunablanco:
Hi @vamsikrishna-siddu , I don't have any other concerns—this looks good to me. That said, I expect you'll need to pivot a bit as you move forward. I'd be happy to approve, but I want to respect @EdDev  and @lyarwood  role as the assigned reviewers. Does that work for you?

lyarwood (commented):
Thanks for the VEP @vamsikrishna-siddu — the motivation is solid and the phased approach makes sense.

A few things from the enhancements call and my own review:

**Timeline and approach suggestion from the enhancements call:**

We'd suggest deferring this to **v1.11.0** rather than leaving the target blank. During the v1.10.0 cycle, the implementation work should happen in a **fork** of kubevirt/kubevirt rather than directly in the main repo. This has two key advantages:

1. It avoids the risk of breaking the existing Bazel-based build in kubevirt/kubevirt while the non-Bazel path is being developed and stabilized
2. It allows us to write bespoke prow jobs in project-infra targeting the fork, so we can get full CI signal on the new build path without flooding kubevirt/kubevirt presubmits or breaking existing jobs

Once the fork demonstrates a stable, passing CI signal with the non-Bazel build, the changes can be brought back into kubevirt/kubevirt proper for the v1.11.0 milestone. This also naturally addresses the transition mechanism question — during the fork phase, the two build systems are completely isolated rather than needing a switching mechanism within the same repo.

See inline comments for specific technical feedback on the VEP text.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Design
+
+The migration is phased to incrementally replace each Bazel responsibility with standard tooling while keeping the existing Makefile interface stable.
+
+This applies to the development (main) branch only. Existing release branches will not be modified and continue to use Bazel.
+
+### Phase 1: Decouple base images
+
+Base images that require RPM packages are decoupled from the main KubeVirt build. These are built independently using Containerfiles. For RPM dependency resolution, invoking bazeldnf directly as a standalone tool (without Bazel) has been explored and is working — see [enhancements#393](https://github.com/kubevirt/enhancements/pull/393) for details.
```
lyarwood, line 95: This links to enhancements#393 — which is this PR itself. Should this point to kubevirt/kubevirt#18534 (the bazeldnf standalone exploration mentioned in the PR description)?

Same issue on line 164 in the Implementation History section.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,208 @@
+# VEP #392: Remove Bazel from KubeVirt Build System
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
```
lyarwood, line 7: As discussed on the enhancements call, we'd suggest targeting **v1.11.0*. During the v1.10.0 cycle, the implementation should happen in a fork of kubevirt/kubevirt so that:

1. The existing Bazel build in kubevirt/kubevirt is not at risk of breakage during development
2. Bespoke prow jobs can be set up in project-infra targeting the fork, giving full CI signal on the new build path without flooding kubevirt/kubevirt presubmits or breaking existing jobs

This fork-based approach also naturally resolves the transition mechanism question — during development the two build systems are completely isolated rather than needing a switching mechanism within the same repo.

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,208 @@
...
+## Graduation Requirements
+
+### Alpha
+- [ ] Base images (RPM-dependent) decoupled and built via Containerfiles (RPM resolution approach TBD)
+- [ ] All KubeVirt component images buildable via Containerfiles
+- [ ] Non-Bazel path for binary compilation demonstrated and validated
+- [ ] Non-Bazel path for functional tests demonstrated and validated
+- [ ] Non-Bazel path for unit tests demonstrated and validated
+- [ ] All existing e2e tests pass with Containerfile-built images
+- [ ] Multi-arch support (amd64, arm64, s390x) verified with new build path
```
lyarwood, line 176: There's a misalignment here: alpha requires multi-arch support verified (amd64, arm64, s390x), but cross-compilation is listed as a Beta requirement (line 181) and Phase 5 in the Design section says cross-compilation must be solved "before Bazel can be fully removed."

How is multi-arch verified in alpha without cross-compilation? If the answer is native builds on each architecture (rather than cross-compilation from x86), that should be stated explicitly. Otherwise, consider moving multi-arch to beta alongside cross-compilation.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Fixed it.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Updated the VEP to target v1.11.0 and added the fork-based development approach to the Design section.

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: Multi-arch was validated using emulation  in the POC. The final implementation will use cross-compilation. Restructured the VEP, cross-compilation is now part of Alpha requirements.

stu-gott:
One concern not addressed here: resilience to upstream dependency unavailability, which Bazel provided via hermetic, hash-addressed caching of fetched dependencies (both WORKSPACE repos and bazeldnf's RPM resolution).

The Phase 2 caching work (GOCACHE, GCS-backed test cache, container layer caching) is all aimed at recovering build speed, not at caching fetched dependencies — none of it helps if a Go module host, RPM repo, or base image registry is temporarily unreachable.

RPM resolution is the most exposed piece, and it's explicitly TBD (kubevirt#18534). Go module dependencies aren't discussed at all in this regard — no vendoring, no pinned/mirrored GOPROXY, no fallback behavior.

Could we get an explicit graduation criterion covering this, rather than relying on "performance comparable to Bazel" as the only bar?

vamsikrishna-siddu:
> RPM resolution is the most exposed piece, and it's explicitly TBD (kubevirt#18534). Go module dependencies aren't discussed at all in this regard — no vendoring, no pinned/mirrored GOPROXY, no fallback behavior.

Thanks for raising this, valid concern. I've added a "Dependency fetch resilience" section to the VEP and updated the graduation criteria. Here's the summary:

Go modules are already fully hermetic and unaffected by the Bazel removal. KubeVirt vendors all Go dependencies (vendor/ checked into git), builds use GOFLAGS="-mod=vendor" and GOPROXY=off, zero network access at build time. This was just undocumented in the VEP.

RPM packages are no longer TBD. bazeldnf standalone (without Bazel) is implemented and working ([kubevirt#18534](https://github.com/kubevirt/kubevirt/pull/18534)). hack/rpm-deps.sh runs bazeldnf fetch/bazeldnf rpmtree to resolve and pin RPMs, and generate-rpm-tars.sh downloads them and produces rootfs tars via bazeldnf rpm2tar. RPM URLs and SHAs remain pinned in WORKSPACE, and each entry already includes both the upstream CentOS mirror and the existing GCS mirror (storage.googleapis.com/builddeps/<sha256>). I've updated generate-rpm-tars.sh to try the GCS mirror first and fall back to upstream, same multi-URL resilience Bazel provided. Also worth noting: regular CI never fetches RPMs at all, it pulls pre-built base images from the registry. Only the "rebuild base images" workflow (triggered when RPM deps change) downloads RPMs, and that's covered by the GCS fallback.

Container base images will be pinned by digest (@sha256:...) rather than mutable tags.

I've added an explicit Beta graduation criterion covering all three:

- Go module vendoring verified (GOFLAGS="-mod=vendor", GOPROXY=off)
- RPM downloads use existing GCS mirror with fallback to upstream
- Base images pinned by digest in all Containerfiles

veps/sig-buildsystem/392-remove-bazel/vep.md
```diff
@@ -0,0 +1,213 @@
+# VEP #392: Remove Bazel from KubeVirt Build System
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.11.0
```
lyarwood: nit - I'm not sure if we want to list `v1.10.0 (via fork)` here for now?

lyarwood (commented):
Given this is all happening in the fork for v1.10.0 I think this is okay to land as is for now, the bar will definitely be higher with v1.11.0 but hopefully something we can meet given the experiences in the fork, thanks!

/lgtm

veps/sig-buildsystem/392-remove-bazel/vep.md
vamsikrishna-siddu: done.

lyarwood (approved):

xpivarc:
@dollierp @dhiller Can you please ack on this work from CI side

dollierp:
> Can you please ack on this work from CI side

The VEP looks comprehensive and the work completed so far is consistent with this document
- cache replacement validated
- non bazel fork created

/lgtm

xpivarc:
> > Can you please ack on this work from CI side
> 
> The VEP looks comprehensive and the work completed so far is consistent with this document
> 
>     * cache replacement validated
> 
>     * non bazel fork created
> 
> 
> /lgtm

/approve 
for CI team

Issue #395, opened 2026-07-16
VEP 395: OCI Artifact Import for VirtualMachines and VirtualMachineTemplates


0xFelix:
**Primary contact (assignee)**:

/assign @0xFelix

**Current Feature Stage**:

New

**Feature Gate**:

n/a (external repo + manual or HCO deployment)

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/396

**Timeline**:

* Alpha release target: v1.10
* Beta release target:
* Stable release target:

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/396
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #396, opened 2026-07-16
VEP #395: OCI Artifact Import for VirtualMachines and VirtualMachineTemplates

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/395
**SIG label**: /sig compute

### What this PR does

Adds VEP #395, proposing OCI artifact import for VirtualMachines and VirtualMachineTemplates. This is the import counterpart to VEP #256 (OCI export), completing the round-trip workflow.

Key points:
- New `VirtualMachineImport` CRD (`import.kubevirt.io`) and controller in `kubevirt/virt-import`
- Target kind auto-detected from OCI `artifactType`
- CDI extension: `Layer` on `DataVolumeSourceRegistry` for annotation-based layer selection
- `virtimportctl` CLI at alpha, moving into `virtctl` at beta
- Installed from the published manifest bundle, and deployable through HCO; no feature gate
- Three sequential phases (CDI, virt-import, HCO); `kubevirt/kubevirt` is not modified
- Alpha target: v1.10

This VEP builds on VEP #256 (OCI export). A proof-of-concept shell script exists at `kubevirt/kubevirt/hack/oci-import.sh`.

### DIFF
--- a/veps/sig-compute/395-oci-import/vep.md
+++ b/veps/sig-compute/395-oci-import/vep.md
@@ -1,0 +1,975 @@
+# VEP #395: OCI Artifact Import for VirtualMachines and VirtualMachineTemplates
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+[VEP #256](../256-oci-export-proposal/oci-export-proposal.md) introduced
+OCI artifact export for `VirtualMachines` and
+`VirtualMachineTemplates`, but explicitly deferred import as a follow-up
+design. Its graduation criteria for Beta/GA include "Import mechanism
+implemented" and "Round-trip export/import tested."
+
+This proposal introduces a `VirtualMachineImport` CRD (API group
+`import.kubevirt.io`) and controller that imports `VirtualMachines` and
+`VirtualMachineTemplates` from OCI artifacts stored in container registries.
+The target kind is auto-detected from the OCI manifest's `artifactType`
+field. The controller lives in a new `kubevirt/virt-import` repository and
+ships two images, `virt-import-controller` and `virt-import-fetcher`. It is
+installed from the manifest bundle the repository publishes with each
+release, and will also be available through the
+hyperconverged-cluster-operator (HCO).
+
+Importing disk data is delegated to CDI via `DataVolumes`. This requires
+extending CDI's `DataVolumeSourceRegistry` with a `Layer` field to
+support selecting OCI artifact layers by annotation, and handling raw
+(non-tar) layer blobs.
+
+A proof-of-concept shell script exists at `kubevirt/kubevirt/hack/oci-import.sh`
+that validates the overall approach.
+
+### Implementation Phases
+
+The work is split into three sequential phases:
+
+1. **Phase 1 - CDI** (`kubevirt/containerized-data-importer`): Extend CDI's
+   `DataVolumeSourceRegistry` with a `Layer` field, add raw blob handling
+   for OCI artifact layers, and validate incompatibility with node pull
+   mode.
+2. **Phase 2 - virt-import** (`kubevirt/virt-import`): Implement the
+   `VirtualMachineImport` CRD, the controller, the metadata-fetch binary and
+   the `virtimportctl` CLI in a new repository, together with the manifest
+   bundle that installs them.
+3. **Phase 3 - HCO** (`kubevirt/hyperconverged-cluster-operator`): Make the
+   component deployable through HCO.
+
+Each phase depends on the previous one. All three target alpha in v1.10.
+`kubevirt/kubevirt` is not modified.
+
+## Motivation
+
+By completing the import side of the OCI artifact workflow, we get:
+
+* **Full round-trip** - export a VM or template from one cluster, push to a
+  registry, import on another cluster
+* **Declarative API** - `VirtualMachineImport` CR with status tracking,
+  conditions, and garbage collection of intermediate resources
+* **Registry-native distribution** - pull VMs and templates from existing
+  registry infrastructure
+* **Portable imports** - storage-specific details are not stored in the
+  artifact; CDI applies cluster-appropriate defaults via StorageProfiles,
+  making imports work across clusters with different storage backends
+
+## Goals
+
+* Import `VirtualMachines` from OCI artifacts in registries
+* Import `VirtualMachineTemplates` from OCI artifacts in registries
+* Auto-detect target kind from OCI `artifactType`
+* Support multi-architecture OCI artifacts via optional
+  `platform.architecture` selection
+* Extend CDI's `DataVolumeSourceRegistry` with a `Layer` field for
+  annotation-based layer selection and raw blob handling
+* Publish a manifest bundle that installs the controller standalone, and
+  make the component deployable through HCO
+* Provide a `virtimportctl` CLI
+
+## Non Goals
+
+* Node pull mode for OCI artifacts (future enhancement - see
+  [Node Pull Mode](#node-pull-mode---future-enhancement))
+* Multi-VM appliance import (single VM or template per artifact)
+* Guest-level operations (sysprep, network reconfiguration) during import
+* Import of non-KubeVirt OCI artifacts (only artifacts matching the VEP #256
+  format); non-compliant artifacts are rejected
+* Periodic or scheduled imports (could be built on top of
+  `VirtualMachineImport` as a future enhancement)
+* `virtctl` integration; the CLI commands move into `virtctl` at beta
+
+## Definition of Users
+
+* **VM owner**: imports VMs from registries for cross-cluster migration or
+  disaster recovery
+* **Template owner**: imports templates from registries for sharing reusable
+  VM blueprints across clusters
+* **Cluster admin**: manages RBAC to control who can create imports;
+  configures storage classes and StorageProfiles
+
+## User Stories
+
+* As a VM owner, I want to import a `VirtualMachine` from a registry so I can 
+  recreate it on another cluster.
+* As a template owner, I want to import a `VirtualMachineTemplate` from a
+  registry so I can share reusable templates across clusters.
+* As a VM owner, I want to round-trip my VM: export to OCI, push to a
+  registry, import on another cluster - and get an equivalent VM.
+* As a cluster admin, I want imports to use cluster-appropriate storage
+  defaults so I do not need to manually specify volume modes and access
+  modes for each disk.
+
+## Repos
+
+* (Phase 1) [kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer) -
+  `Layer` on `DataVolumeSourceRegistry` with annotation-based selection,
+  raw blob handling in importer
+* (Phase 2) [kubevirt/virt-import](https://github.com/kubevirt/virt-import)
+  (new) - import API, controller, metadata-fetch binary, CLI and manifest
+  bundle
+* (Phase 3) [kubevirt/hyperconverged-cluster-operator](https://github.com/kubevirt/hyperconverged-cluster-operator) -
+  deployment of the import controller
+
+`kubevirt/virt-import` is created through the KubeVirt org's repository
+process, with @0xFelix and @akalenyu as maintainers.
+
+## Design
+
+### OCI Artifact Format (reference to VEP #256)
+
+The import mechanism consumes artifacts in the format defined by
+[VEP #256](../256-oci-export-proposal/oci-export-proposal.md). See its
+[OCI Artifact Format Specification](../256-oci-export-proposal/oci-export-proposal.md#oci-artifact-format-specification)
+for the full definition of artifact types, media types, layer annotations,
+and config blob structure.
+
+The import controller uses the `artifactType` field on the OCI manifest to
+auto-detect the target kind (`VirtualMachine` or `VirtualMachineTemplate`).
+It reads `io.kubevirt.disk.names` and `io.kubevirt.disk.size` from layer
+annotations to correlate layers to volumes and to size DataVolumes. The
+plural `io.kubevirt.disk.names` is corrected in VEP #256 by
+[#446](https://github.com/kubevirt/enhancements/pull/446).
+
+**Not in the artifact:** `volumeMode` and `accessModes` are not stored in
+the OCI artifact. CDI applies cluster-appropriate defaults via
+StorageProfiles on the target cluster. This makes imports portable across
+clusters with different storage backends.
+
+### Phase 1: CDI - OCI Artifact Layer Support
+
+This phase extends CDI's `DataVolumeSourceRegistry` to support selecting
+individual layers from OCI artifacts by annotation and importing raw
+(non-tar) blobs. All changes are in
+[kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer).
+
+#### DataVolumeSourceRegistry Layer
+
+A new `Layer` field is added to `DataVolumeSourceRegistry`:
+
+```go
+// DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
+type DataVolumeSourceRegistry struct {
+    //URL is the url of the registry source (starting with the scheme: docker, oci-archive)
+    // +optional
+    URL *string `json:"url,omitempty"`
+    //ImageStream is the name of image stream for import
+    // +optional
+    ImageStream *string `json:"imageStream,omitempty"`
+    //PullMethod can be either "pod" (default import), or "node" (node docker cache based import)
+    // +optional
+    PullMethod *RegistryPullMethod `json:"pullMethod,omitempty"`
+    //SecretRef provides the secret reference needed to access the Registry source
+    // +optional
+    SecretRef *string `json:"secretRef,omitempty"`
+    //CertConfigMap provides a reference to the Registry certs
+    // +optional
+    CertConfigMap *string `json:"certConfigMap,omitempty"`
+    //Platform describes the minimum runtime requirements of the image
+    // +optional
+    Platform *PlatformOptions `json:"platform,omitempty"`
+    // Layer selects a specific layer from an OCI artifact manifest
+    // by matching an annotation on the manifest's layer descriptors.
+    // The importer resolves the manifest (using Platform to select
+    // from an image index or verify a plain manifest), finds the
+    // matching layer, and treats its blob as raw data.
+    // Incompatible with pullMethod: node.
+    // +optional
+    Layer *LayerDescriptor `json:"layer,omitempty"`
+}
+
+type LayerDescriptor struct {
+    // Annotation selects a layer by matching an annotation on the
+    // OCI manifest's layer descriptors.
+    Annotation LayerAnnotationSelector `json:"annotation"`
+}
+
+type LayerAnnotationSelector struct {
+    // Key is the annotation key to match on the layer descriptor
+    // (e.g. "io.kubevirt.disk.names").
+    Key string `json:"key"`
+    // Value is the expected annotation value, matched exactly
+    // (e.g. "rootdisk", or "datadisk,shareddisk" for a layer backing
+    // more than one volume).
+    Value string `json:"value"`
+}
+```
+
+This follows a similar convention to podman's
+[artifact extract](https://docs.podman.io/en/latest/markdown/podman-artifact-extract.1.html)
+annotation-based layer selection. A `Digest` field for direct
+content-addressable selection could be added to `LayerDescriptor` in
+the future if needed. CDI could also infer the storage size from the
+layer's `io.kubevirt.disk.size` annotation when resolving the manifest,
+removing the need for the controller to set it on the DataVolume.
+
+When `Layer` is set, `pullMethod` must be `pod` (or unset, defaulting
+to `pod`). Node pull is incompatible with `Layer` because the container
+runtime cannot handle OCI artifacts with custom media types. CDI's
+webhook rejects DataVolumes that combine `Layer` with
+`pullMethod: node`.
+
+CDI's registry import behavior changes as follows:
+
+| Step | Normal (container image) | With Layer (OCI artifact) |
+|------|--------------------------|---------------------------|
+| Manifest parsing | `image.FromSource()` | Fetch manifest; if image index, select by `Platform`; if plain manifest, verify `Platform` matches (fail on mismatch) |
+| Layer selection | All layers, match by tar path | Match layer by annotation key-value |
+| Blob fetch | Via layer iteration | `GetBlob()` by matched layer's digest |
+| Layer processing | `tar.NewReader()`, extract `disk/` files | Raw blob path: stream decompressed blob directly |
+| Post-import inspection | `Inspect()` for labels | Skipped |
+
+#### Node Pull Mode - Future Enhancement
+
+Node pull mode cannot work for OCI artifacts because:
+
+1. The container runtime (containerd/CRI-O) expects standard container
+   image layers and would fail to pull an artifact with
+   `application/vnd.kubevirt.disk.raw+zstd` layers
+2. The sidecar container pattern used by CDI's node pull requires a valid
+   container image as the sidecar's image reference
+3. Even if the runtime could pull the artifact, it would try to unpack
+   layers as filesystem overlays, which fails for raw disk blobs
+
+Node pull would require one of:
+
+* CRI-level OCI artifact support (upstream Kubernetes work)
+* A different credential-sharing mechanism (e.g., projected service account
+  tokens with registry auth)
+* A proxy that uses node credentials to pull blobs and serves them to the
+  importer pod
+
+This can be a follow-up enhancement in the future.
+
+### Phase 2: virt-import Controller
+
+All changes are in
+[kubevirt/virt-import](https://github.com/kubevirt/virt-import).
+
+#### Repository Structure
+
+The repository follows the same pattern as `kubevirt/virt-template`:
+
+* Kubebuilder v4 scaffolding
+* A separate `kubevirt.io/virt-import-api` module and a client generated
+  from `staging/src/kubevirt.io/virt-import-client-go`
+* A controller-runtime manager with leader election
+* A single Kustomize `default` overlay, from which `make build-installer`
+  renders the released manifest bundle
+
+Two images are built: `virt-import-controller` runs the reconciler, and
+`virt-import-fetcher` runs in the metadata-fetch Job, whose image the
+controller receives as a command line flag.
+
+The OCI media types and layer annotations the importer reads are constants in
+`kubevirt.io/virt-import-api`, matching VEP #256. The artifact format is shared
+with the exporter, the constants are a copy: `kubevirt/virt-import` does not
+depend on the `kubevirt.io/kubevirt` module. Promoting them to a shared
+package is a GA item.
+
+The controller does not access container registries directly. It spawns a
+metadata-fetch Job whose Pod runs `virt-import-fetcher` and mounts the
+referenced Secret and ConfigMap for registry authentication. The fetcher
+holds no API access, so the only Pod that mounts registry credentials
+contains an OCI client and nothing else, and the controller needs no
+cluster-wide Secret or ConfigMap read permissions.
+
+#### Required and Optional API Groups
+
+`cdi.kubevirt.io` is required; every import creates DataVolumes. An
+availability controller polls discovery and starts the import reconciler once
+the group appears. While it waits it patches pending `VirtualMachineImports`
+with `Ready=False` and `Progressing=True`, reason `Waiting`, naming the
+missing group.
+
+`template.kubevirt.io` is optional. It is only needed when an artifact's
+`artifactType` names a `VirtualMachineTemplate`, and its absence fails that
+one import rather than blocking the controller. `VirtualMachineTemplates` are
+served from the manager's cache when the group is present at startup, and
+read uncached otherwise, so a group installed later needs no restart.
+
+#### CRD Definition
+
+```go
+// VirtualMachineImportSpec defines the desired import.
+type VirtualMachineImportSpec struct {
+    // Source specifies where to import from.
+    Source VirtualMachineImportSource `json:"source"`
+    // TargetName is the name for the created VM or VMTemplate.
+    // Defaults to the VirtualMachineImport CR name.
+    // +optional
+    TargetName *string `json:"targetName,omitempty"`
+    // StorageClassName overrides the default storage class for all
+    // DataVolumes created during import.
+    // +optional
+    StorageClassName *string `json:"storageClassName,omitempty"`
+}
+```
+
+`VirtualMachineImportSpec` is immutable after creation, enforced by a CEL
+validation rule (`self == oldSelf`) on the `spec` field of the CRD schema. To
+change import parameters, delete the CR and create a new one.
+
+```go
+type VirtualMachineImportSource struct {
+    // Registry specifies an OCI artifact in a container registry.
+    // +optional
+    Registry *VirtualMachineImportRegistrySource `json:"registry,omitempty"`
+}
+
+type VirtualMachineImportRegistrySource struct {
+    // URL is the registry URL
+    // (e.g. docker://registry.example.com/vms/fedora:v1).
+    URL string `json:"url"`
+    // SecretRef is the name of a Secret containing registry credentials.
+    // +optional
+    SecretRef *string `json:"secretRef,omitempty"`
+    // CertConfigMap is the name of a ConfigMap containing registry CA
+    // certificates.
+    // +optional
+    CertConfigMap *string `json:"certConfigMap,omitempty"`
+    // Platform selects the architecture variant from a multi-arch
+    // artifact.
+    // +optional
+    Platform *PlatformOptions `json:"platform,omitempty"`
+}
+
+type PlatformOptions struct {
+    // Architecture selects the platform variant (e.g. amd64, arm64).
+    Architecture *string `json:"architecture,omitempty"`
+}
+```
+
+```go
+// VirtualMachineImportStatus reports observed import state.
+type VirtualMachineImportStatus struct {
+    // Conditions represent the latest observations of the import.
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+    // ArtifactInfo contains metadata read from the OCI artifact.
+    // +optional
+    ArtifactInfo *ArtifactInfo `json:"artifactInfo,omitempty"`
+    // DiskImports tracks the status of each disk being imported.
+    DiskImports []DiskImportStatus `json:"diskImports,omitempty"`
+    // TargetRef references the created VM or VMTemplate.
+    // +optional
+    TargetRef *corev1.ObjectReference `json:"targetRef,omitempty"`
+}
+
+type ArtifactInfo struct {
+    // ArtifactType is the OCI artifactType from the manifest.
+    ArtifactType string `json:"artifactType"`
+    // Architecture is the resolved platform architecture.
+    Architecture string `json:"architecture"`
+    // DiskCount is the number of disk layers in the artifact.
+    DiskCount int `json:"diskCount"`
+}
+
+type DiskImportStatus struct {
+    // Names are the volume names backed by this layer, from the
+    // io.kubevirt.disk.names annotation.
+    Names []string `json:"names"`
+    // DataVolumeName is the name of the created DataVolume.
+    DataVolumeName string `json:"dataVolumeName"`
+    // Phase mirrors cdiv1.DataVolumePhase values (e.g. "Succeeded",
+    // "ImportInProgress", "Failed").
+    Phase string `json:"phase"`
+}
+```
+
+**Condition types and reasons:**
+
+| Condition | Status | Reason           | Meaning |
+|-----------|--------|------------------|---------|
+| `Progressing` | `True` | `FetchingMetadata` | Metadata-fetch Job running |
+| `Progressing` | `True` | `ImportingDisks` | DataVolumes created, CDI importing |
+| `Progressing` | `True` | `CreatingTarget` | Disks done, creating VM/VMTemplate |
+| `Progressing` | `False` | `ImportComplete` | Import finished successfully |
+| `Progressing` | `False` | `ImportFailed`   | Import failed |
+| `Progressing` | `True` | `Waiting`        | The `cdi.kubevirt.io` API group is not available |
+| `Ready` | `True` | `ImportComplete` | Target resource created |
+| `Ready` | `False` | `ImportFailed`  | Import failed (see message) |
+| `Ready` | `False` | `Waiting`       | The `cdi.kubevirt.io` API group is not available |
+
+#### Import Flow
+
+```mermaid
+flowchart TD
+    U(["User creates VirtualMachineImport"]) --> CTRL["virt-import-controller"]
+    CTRL -->|"1. create metadata-fetch Job"| FETCH
+
+    subgraph FETCH ["virt-import-fetcher Job (mounts registry Secret and CA ConfigMap)"]
+        direction TB
+        F1["Fetch OCI artifact from registry<br/>(Image Index or Image Manifest)"]
+        F2["Resolve platform, select manifest"]
+        F3["Fetch manifest and config blob<br/>for the resolved platform"]
+        F4["Validate artifact"]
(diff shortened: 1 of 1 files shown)


veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,711 @@
...
+
+### Import Flow
+
+```
+User creates VirtualMachineImport CR
+  |
+  v
+Import Controller (virt-import)
+  |-- 1. Fetch OCI Image Index from registry
+  |-- 2. Resolve platform:
```
geetikakay: i was checking alpine oci image and saw arch duplicates and arch unknows,what could controller do in such cases?
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v6"
}
{
  "architecture": "arm",
  "os": "linux",
  "variant": "v7"
}
 {
  "architecture": "unknown",
  "os": "unknown"
}

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,711 @@
...
+  feature gate is enabled
+* Provide a `virtctl vmimport` CLI command
+
+## Non Goals
+
+* Node pull mode for OCI artifacts (future enhancement - see
+  [Node Pull Mode](#node-pull-mode---future-enhancement))
+* Multi-VM appliance import (single VM or template per artifact)
+* Guest-level operations (sysprep, network reconfiguration) during import
+* Import of non-KubeVirt OCI artifacts (only artifacts matching the VEP #256
```
geetikakay, line 94: how we deal with non compliant ones? are we going to reject them?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: If arch cannot be resolved, then the import will fail. If it does not match, it will be ignored. Also, I don't know what `unknown` arch/os is supposed to be, but in terms of OCI artifacts created by KubeVirt there always will be a valid combination.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Yes, they're going to be rejected.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: I see what you mean now regarding the duplicates. In KubeVirt it's unlikely that we required the `variant` subkey, since `arm` is not supported. It could be added in the future if required. If there's multiple with the same architecture, it should fail too.

0xFelix:
Reworked the proposal into three distinct phases, to make the order of required changes easier to understand.

@geetikakay Also updated sections in regard to your comments

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+    CertConfigMap *string             `json:"certConfigMap,omitempty"`
+    Platform      *PlatformOptions    `json:"platform,omitempty"`
+    // LayerSelector selects a specific layer from an OCI artifact.
+    // When set, the importer treats the matching layer as a raw
+    // blob (no tar extraction).
+    // +optional
+    LayerSelector *LayerSelector      `json:"layerSelector,omitempty"`
+}
+
+type LayerSelector struct {
```
Acedus: I wonder, if we already fetch the manifest in advance, wouldn't we be able to figure out the relevant layer digest at the controller level? If so, couldn't we replace this field with something like `LayerDigest` which would enable CDI to skip the manifest parsing and go straight to pulling the blob?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+}
+```
+
+When `LayerSelector` is set, `pullMethod` must be `pod` (or unset,
+defaulting to `pod`). Node pull is incompatible with `LayerSelector`
+because the container runtime cannot handle OCI artifacts with custom media
+types. CDI's webhook rejects DataVolumes that combine `LayerSelector`
+with `pullMethod: node`.
+
+**Platform selection in CDI:** When `Platform` is set on a
```
Acedus: nit: How is that relevant to this section?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+    // Source specifies where to import from.
+    Source VirtualMachineImportSource `json:"source"`
+    // TargetName is the name for the created VM or VMTemplate.
+    // Defaults to the VirtualMachineImport CR name.
+    // +optional
+    TargetName string `json:"targetName,omitempty"`
+    // StorageClassName overrides the default storage class for all
+    // DataVolumes created during import.
+    // +optional
+    StorageClassName *string `json:"storageClassName,omitempty"`
```
Acedus, line 283: I think more granular SC override control may be desirable for VMs with multiple disks serving different purposes.

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+* **During import** (`Ready` is not `True`): deleting the CR triggers the
+  finalizer, which deletes all intermediate resources (DataVolumes and
+  their PVCs). No VM or VMTemplate has been created yet.
+* **After successful import** (`Ready` is `True`): the controller removes
+  ownerReferences from DataVolumes and removes its finalizer after the
+  target resource is created. Deleting the `VirtualMachineImport` CR at
+  this point only deletes the CR itself - the created VM or VMTemplate
+  and its PVCs persist independently.
+
+#### RBAC
```
Acedus, line 477: Should also add required user RBAC to prevent privilege escalation (e.g., the controller fetches OCI metadata directly, so it would at least need to `get` the specified `secretRef` meaning the user should have matching permissions).

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+#### Repository Structure
+
+The repository follows the same pattern as `kubevirt/virt-template`:
+
+* Kubebuilder v4 scaffolding
+* Separate `api/` module for importable types
+* Two Kustomize overlays: `default` (cert-manager), `virt-operator`
+  (external certs managed by `virt-operator`)
+
+The controller fetches OCI metadata directly and delegates disk import to
```
Acedus: > The controller fetches OCI metadata directly

How does it do that?

Acedus (commented):
Thanks @0xFelix, looking pretty good from the storage perspective IMO! Left a few comments.

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+| `Ready` | `False` | `Failed` | Import failed (see message) |
+
+#### Import Flow
+
+```
+User creates VirtualMachineImport CR
+  |
+  v
+Import Controller (virt-import)
+  |-- 1. Fetch OCI Image Index from registry
```
codingben: What happens if it's not an image index, but just the image manifest? Instead of producing an error, IMO it should check just for an image manifest.

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+  |-- 2. Resolve platform:
+  |       - platform.architecture set -> select matching manifest
+  |         - multiple matches (e.g. same arch, different variant) -> fail
+  |         - no match -> fail
+  |       - unset, single manifest   -> auto-select
```
codingben: What does mean "auto-select", what it will do? Is there a default value? Can you decide if `platform.architecture` is required or not?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+  |       - unset, multiple manifests -> fail, ask user to specify
+  |       Manifests with unknown or empty arch/os are skipped.
```
codingben: Who can produce manifests with unknown or empty arch/os? Can you ensure it won't happen when you do the export?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+the config blob.
+
+**Error semantics:** Import is all-or-nothing. If any DataVolume fails,
+the controller sets the `Ready` condition to `False` with reason `Failed`
+and does not create the target VM or VMTemplate. Intermediate resources
+(DataVolumes, PVCs) are cleaned up when the user deletes the
+`VirtualMachineImport` CR (see garbage collection below).
+
+**Naming collisions:** If any resource the controller tries to create
+already exists, the import fails. The controller does not overwrite or
```
codingben: tries to create already exists of *what*?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: True, that would simplify the part of CDI even more. Just pull the Layer with the specified digest? But what about detecting the layers artifact / media type? Could that be helpful to CDI (in the future)? When not fetching the manifest, it would have no way to discover these attributes.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: I guess we could drop it. More of a description of (hopefully) pre-existing behavior?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: How would you envision it? A map of disk name and SC?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+`kubevirt/virt-import`. `virt-operator` manages certificate rotation for
+the controller's webhook, following the same pattern as
+`kubevirt/virt-template`.
+
+#### CLI: virtctl vmimport
+
+```shell
+# Import a VM from a registry
+$ virtctl vmimport create my-import \
+    --url=docker://registry.example.com/vms/fedora:v1 \
```
codingben, line 537: `--arch` is not required?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+**Flags on `virtctl vmimport create`:**
+
+| Flag               | Default | Description                                                          |
+|--------------------|---------|----------------------------------------------------------------------|
+| `--url`            | -       | Registry URL (required, e.g. `docker://registry.example.com/vm:v1`) |
+| `--target-name`    | -       | Name for the created VM or VMTemplate (defaults to CR name)          |
+| `--secret`         | -       | Name of a Secret containing registry credentials                     |
+| `--cert-configmap` | -       | Name of a ConfigMap containing registry CA certificates              |
+| `--storage-class`  | -       | StorageClass override for all DataVolumes                            |
+| `--arch`           | -       | Architecture variant for multi-arch artifacts (e.g. `amd64`)        |
```
codingben: `amd64` is the default arch if not required?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Talk to the registry, i.e. like the import pod does?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+    kind: VirtualMachine
+    name: fedora-vm
+```
+
+### Full round-trip example
+
+```shell
+# Cluster A: export
+$ virtctl vmexport create my-export --vm=fedora-vm
+$ virtctl vmexport download my-export --format=oci --output=fedora.oci.tar
```
codingben, line 665: Can you describe what is the structure of `fedora.oci.tar` or reference if from here to an existing info about this? I want to see if it includes more settings than just the VM's disk image.

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,790 @@
...
+## Definition of Users
+
+* **VM owner**: imports VMs from registries for cross-cluster migration or
+  disaster recovery
+* **Template owner**: imports templates from registries for sharing reusable
+  VM blueprints across clusters
+* **Cluster admin**: manages RBAC to control who can create imports;
+  configures storage classes and StorageProfiles
+
+## User Stories
```
codingben, line 108: Did you consider the user story if the VM owner wants to import the VM template periodically? If yes, can you allow it?

For example, it could be useful to ensure that the VM owner imports and creates updated VM templates in the cluster daily or weekly.

codingben (changes requested):
Nice progress Felix!

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: It could, however the format produced by the export controller is always an index. Seems like a not too expensive change though.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: i.e. if no arch was specified and there's only a single manifest, then use that

In all other cases `platform.architecture` is required

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: It does not happen on export, but of course it does not hinder someone from putting together a manifest by themselves, or handing the controller a ref to a malformed manifest

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Any resource it tries to create?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Doesn't have to be

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Could leave it open, to get the default fallback behavior you asked about above

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: See L130, section `OCI Artifact Format`

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Valid story, but I'd like to defer that to the future. The prerequisite to that is the proposed design. An "import cron" like CRD could be built on top of it.

veps/sig-compute/395-oci-import/oci-import.md
Acedus: Yeah it's a bit problematic considering you'd need to know the disk names in advance :P. I guess we could maybe defer it as a potential beta improvement if sufficient feedback is provided, then it'd just be an additive change that overrides the "global" SC.

veps/sig-compute/395-oci-import/oci-import.md
Acedus: I'm not really sure we care though, correct me if I'm wrong but right now the controller would already resolve it for us so CDI could remain unaware of it?

veps/sig-compute/395-oci-import/oci-import.md
Acedus: The main difference there would be the way we access the registry, i.e., the importer pod currently mounts the secret reference and cert CM, meaning the controller itself needs no additional RBAC. That wouldn't necessarily true for the virt-controller.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: `mediaType` might be relevant for decoding / uncompressing of a disk? The other parts could be resolved by the controller. That would mean the controller would also take care of the architecture validation.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Yes, let's defer that to a follow up

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Right, the controller would need to access the Secret/CM directly when processing the `VirtualMachineImport`. Do you consider that acceptable? If the controller was to create a Pod that mounts a certain Secret or CM it would be required to have the permission to access them anyway, if I'm not mistaken?

0xFelix:
@Acedus @codingben Updated to address your comments

veps/sig-compute/395-oci-import/oci-import.md
Acedus: It's a bit of a slippery slope, creating a Pod that mounts a named secret/cm in the same namespace doesn't require any additional RBAC for the controller that creates it. Reading a secret/cm on the other hand would require cluster-wide read which is generally ill-advised... Only way I can think of is delegating the entire operation to CDI, but that would mean we'd need to teach CDI about kv's custom media type which I'm not too comfortable with either...

/cc @akalenyu.

veps/sig-compute/395-oci-import/oci-import.md
codingben: Ack thanks.

veps/sig-compute/395-oci-import/oci-import.md
codingben: Sounds good.

veps/sig-compute/395-oci-import/oci-import.md
codingben: Yep

veps/sig-compute/395-oci-import/oci-import.md
codingben: Looks good now

veps/sig-compute/395-oci-import/oci-import.md
codingben: LGTM

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,782 @@
...
+
+`VirtualMachineImportSpec` is immutable after creation. A
+`ValidatingAdmissionPolicy` with a CEL expression
+(`oldObject.spec == object.spec`) rejects any update that modifies the
+spec. To change import parameters, delete the CR and create a new one.
+
+```go
+type VirtualMachineImportSource struct {
+    // Registry specifies an OCI artifact in a container registry.
+    Registry *VirtualMachineImportRegistrySource `json:"registry"`
```
mhenriks: optional and omitempty?

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,782 @@
...
+    // Conditions represent the latest observations of the import.
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+    // ArtifactInfo contains metadata read from the OCI artifact.
+    // +optional
+    ArtifactInfo *ArtifactInfo `json:"artifactInfo,omitempty"`
+    // DiskImports tracks the status of each disk being imported.
+    DiskImports []DiskImportStatus `json:"diskImports,omitempty"`
+    // TargetRef references the created VM or VMTemplate.
+    // +optional
+    TargetRef *corev1.TypedLocalObjectReference `json:"targetRef,omitempty"`
```
mhenriks: Not a huge deal but `TypedLocalObjectReference` is deprecated. Consider `corev1.ObjectReference`

veps/sig-compute/395-oci-import/oci-import.md
mhenriks: Not sure about this. As long as the secret isn't exposed to the user why keep them from using the secret indirectly?

veps/sig-compute/395-oci-import/oci-import.md
mhenriks: Can't give this controller cluster-scope "get secrets".

How about delegating only the OCI metadata fetch (steps 1-4: fetch manifest, resolve platform, fetch config blob, validate artifact) to a Pod that mounts the secretRef and certConfigMap. The Pod outputs the parsed config blob and layer descriptors (digest, size, name per disk) -- none of which is sensitive. The controller then takes that output and proceeds with DataVolume and VM creation using its normal RBAC

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,782 @@
...
+```
+
+**Condition types and reasons:**
+
+| Condition | Status | Reason | Meaning |
+|-----------|--------|--------|---------|
+| `Progressing` | `True` | `Initializing` | Fetching OCI manifest and config |
+| `Progressing` | `True` | `ImportingDisks` | DataVolumes created, CDI importing |
+| `Progressing` | `True` | `CreatingTarget` | Disks done, creating VM/VMTemplate |
+| `Progressing` | `False` | `Reconciled` | Terminal state reached |
```
mhenriks: Progressing=False, reason=Reconciled
Ready=True, reason=ImportComplete
Progressing=False, reason=Reconciled
Ready=False, reason=Failed

Having `reason=Reconciled` for both cases seems kind of awkward to me

mhenriks (commented):
Looks great for the most part! Cluster scope "get secrets" is a hard no though

veps/sig-compute/395-oci-import/oci-import.md
Acedus: > The Pod outputs the parsed config blob and layer descriptors (digest, size, name per disk) -- none of which is sensitive.

Outputs as a log? What would it output it into? Wouldn't it be better to treat it as a delegated worker with temporary SA + RBAC owned by the VMImport CR that'll orchestrate the entire thing as a job (including the creation of the DVs)?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: Change it to reflect the reason of the Ready condition instead?

veps/sig-compute/395-oci-import/oci-import.md
mhenriks: > Outputs as a log? What would it output it into? 

Was initially thinking `/dev/termination-log` but forgot the output is limited to 4K. Maybe that's fine? If not I think simply can write data to `stdout` and direct informational messages to stderr

I don't really like the idea of delegating everything to a worker pod. Adds a lot of complexity. Not just managing rbac but progress reporting as well

0xFelix: Using `stdout` / `stderr` should work. To be honest, I prefer that as well instead of managing extra SAs/RBAC.

So the controller would create a Job that fetches the config blog + the layer descriptors and outputs them to stdout. From there the controller reads them from the Pods logs and could continue with creating the import resources.

Sounds acceptable?

mhenriks: Sounds good to me

veps/sig-compute/395-oci-import/oci-import.md
mhenriks: Yeah either that or maybe delete the progressing when failed

0xFelix: Changed it a little

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: How do we handle that in CDI?

0xFelix: Okay, updated to drop this requirement.

Acedus (approved):
Thanks @0xFelix!

codingben (approved):
/lgtm

awels (commented):
Looks good, as noted it doesn't support `pullMethod: node` in CDI, which is unfortunate since this will stop us from providing the equivalent to DataImportCrons and distributing base images with certain distributions (read OpenShift). Where the pull secret to the central repository is only available to the nodes, and there is no user in the cluster that actually has the pull secrets in order to connect. The reason for not supporting it are clearly stated though, so hopefully in the future this will improve.

akalenyu:
/cc @iholder101 
iirc you had some experience with the k8s imagevolume kep, would it make sense to extend it to support custom content types? we think it's a whole new kep because of container runtime changes but not too sure

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,824 @@
...
+### Phase 1: CDI - OCI Artifact Layer Support
+
+This phase extends CDI's `DataVolumeSourceRegistry` to support fetching
+individual layer blobs from OCI artifacts by digest and importing raw
+(non-tar) blobs. All changes are in
+[kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer).
+
+#### DataVolumeSourceRegistry LayerDigest
+
+A new `LayerDigest` field is added to `DataVolumeSourceRegistry`:
```
akalenyu: I wonder if we're going too low level here, for example,
Why don't we match the [podman](https://docs.podman.io/en/latest/markdown/podman-artifact-extract.1.html) convention of annotation-based blob selection?
This would make the DataVolume spec somewhat self describing

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,824 @@
...
+
+```go
+// DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
+type DataVolumeSourceRegistry struct {
```
akalenyu, line 162: Would you consider something like
```go
type DataVolumeSourceRegistry struct {
    // ... existing fields ...

    // Layer selects a specific layer from an OCI artifact manifest.
    // When set, the importer fetches the identified blob directly rather
    // than processing the image as a container image.
    // Incompatible with pullMethod: node.
    // +optional
    Layer *LayerDescriptor `json:"layer,omitempty"`
}

type LayerDescriptor struct {
    // Annotation selects a layer whose annotation matches the given
    // key-value pair (e.g. key: "io.kubevirt.disk.name", value: "rootdisk").
    // CDI fetches the manifest, finds the matching layer descriptor,
    // and resolves its digest and mediaType automatically.
    // Mutually exclusive with Digest.
    // +optional
    Annotation *AnnotationSelector `json:"annotation,omitempty"`

    // Digest selects a layer by content-addressable digest
    // (e.g. "sha256:abc123..."). CDI fetches the manifest and
    // resolves the mediaType from the matching descriptor.
    // Mutually exclusive with Annotation.
    // +optional
    Digest *string `json:"digest,omitempty"`
}

type AnnotationSelector struct {
    // Key is the annotation key to match
    // (e.g. "io.kubevirt.disk.name", "org.opencontainers.image.title").
    Key   string `json:"key"`
    // Value is the expected annotation value (e.g. "rootdisk").
    Value string `json:"value"`
}
```

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,824 @@
...
+clusters with different storage backends.
+
+### Phase 1: CDI - OCI Artifact Layer Support
+
+This phase extends CDI's `DataVolumeSourceRegistry` to support fetching
+individual layer blobs from OCI artifacts by digest and importing raw
+(non-tar) blobs. All changes are in
+[kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer).
+
+#### DataVolumeSourceRegistry LayerDigest
```
akalenyu: So CDI also has an `archive` content type that just untars random (non-disk-image) content to PVCs. Should we allow layer imports from that too? I get this VEP has no motivation for it (always disk images) but probably worth a mention

veps/sig-compute/395-oci-import/oci-import.md
akalenyu:
```yaml
apiVersion: cdi.kubevirt.io/v1beta1
kind: DataVolume
metadata:
  name: fedora-vm-rootdisk
spec:
  source:
    registry:
      url: ...
      layer:
        annotation:
          key: io.kubevirt.disk.name
          value: rootdisk
```

akalenyu (commented):
Nice work! want to start some discussion over the CDI API

veps/sig-compute/395-oci-import/oci-import.md
```diff
@@ -0,0 +1,824 @@
...
+
+A new `LayerDigest` field is added to `DataVolumeSourceRegistry`:
+
+```go
+// DataVolumeSourceRegistry provides the parameters to create a Data Volume from an registry source
+type DataVolumeSourceRegistry struct {
+    //URL is the url of the registry source (starting with the scheme: docker, oci-archive)
```
akalenyu, line 163: I was surprised to learn myself that CDI support `oci-archive` :O

veps/sig-compute/395-oci-import/oci-import.md
akalenyu: @0xFelix we will leave the decision up to you, can't confidently say anyone will use this manually instead of through the import controller orchestration

0xFelix: An earlier revision of the VEP suggested going with a `LayerSelector` based on labels. I like that idea too, but it would also put the "burden" of picking the right architecture on CDI. @Acedus suggested the `LayerDigest` and it felt like the change with the least impact.

In the link you've sent I see that podman supports a `--digest` flag too. :)

I'm not firmly committed to either of them, but I'm leaning towards the digest?

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: That would allow us both options supported by Podman, right? The import controller would most likely use the `Digest` though.

WDYT about starting with `LayerDescriptor` or `LayerSelector`, but only adding support for the `Digest` field right now?

veps/sig-compute/395-oci-import/oci-import.md
akalenyu: I was suggesting we support both - https://github.com/kubevirt/enhancements/pull/396#discussion_r3699164986

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: It's why I was asking about the `mediaType` in https://github.com/kubevirt/containerized-data-importer/pull/4214, to make it more clear which content types are supported. You are right that this VEP would only be interested in disk images though.

veps/sig-compute/395-oci-import/oci-import.md
0xFelix: TIL!

veps/sig-compute/395-oci-import/oci-import.md
Acedus: I think having `LayerDescriptor` would be a good idea to allow extending the feature if sufficient feedback is received in the future.

0xFelix:
@Acedus @akalenyu I updated the VEP to include the `Layer` field / `LayerDescriptor` structure instead of the plain `LayerDigest`. I also added a few more validations to the metadata-fetch job. Can you please take a look?

0xFelix:
After offline discussion with @akalenyu, I updated the VEP to switch to annotation based selection of layers. PTAL.

@Acedus @akalenyu

Acedus (approved):

akalenyu (approved):
thank you! lgtm from storage side

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,865 @@
...
+[VEP #256](../256-oci-export-proposal/oci-export-proposal.md) introduced
+OCI artifact export for `VirtualMachines` and
+`VirtualMachineTemplates`, but explicitly deferred import as a follow-up
+design. Its graduation criteria for Beta/GA include "Import mechanism
+implemented" and "Round-trip export/import tested."
+
+This proposal introduces a `VirtualMachineImport` CRD (API group
+`import.kubevirt.io`) and controller that imports `VirtualMachines` and
+`VirtualMachineTemplates` from OCI artifacts stored in container registries.
+The target kind is auto-detected from the OCI manifest's `artifactType`
```
xpivarc, line 31: This means a controller possibly requires secret/token to query OCI artifact.

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,865 @@
...
+1. **Phase 1 - CDI** (`kubevirt/containerized-data-importer`): Extend CDI's
+   `DataVolumeSourceRegistry` with a `Layer` field, add raw blob handling
+   for OCI artifact layers, and validate incompatibility with node pull
+   mode.
+2. **Phase 2 - virt-import** (`kubevirt/virt-import`): Implement the
+   `VirtualMachineImport` CRD and controller in a new repository. The
+   controller is deployable standalone via Kustomize with cert-manager.
+3. **Phase 3 - kubevirt** (`kubevirt/kubevirt`): Add the `OCIImport` feature
+   gate, wire `virt-operator` to deploy/remove the import controller, and
+   implement the `virtctl vmimport` CLI.
```
xpivarc: I see disconnect here. On one side we are making this part of core, as  we have the `virtctl vmimport` but on the other side we are making the controller outside of core. I wonder what is the reasoning....

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,865 @@
...
+
+Each phase depends on the previous one. All three target alpha in v1.10.
```
xpivarc, line 61: This smells like setting up ourself for failure :)

veps/sig-compute/395-oci-import/vep.md
0xFelix: That's what the metadata-fetch Job is intended for. It possibly requires a secret/token to query the registry, yes. But the controller does not. This follows the pattern of CDI.

veps/sig-compute/395-oci-import/vep.md
0xFelix: Faster pace of development and avoiding to overload the main repo. It is the same approache we have taken with virt-template.

veps/sig-compute/395-oci-import/vep.md
0xFelix: It's an ambitious goal :) However, if Phase 3 fails, we can always defer it to the next release.

0xFelix:
Updated this to move it from virt-operator deployment to deployment through HCO.

PTAL @akalenyu @xpivarc

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,932 @@
...
+implemented" and "Round-trip export/import tested."
+
+This proposal introduces a `VirtualMachineImport` CRD (API group
+`import.kubevirt.io`) and controller that imports `VirtualMachines` and
+`VirtualMachineTemplates` from OCI artifacts stored in container registries.
+The target kind is auto-detected from the OCI manifest's `artifactType`
+field. The controller lives in a new `kubevirt/virt-import` repository and
+ships two images, `virt-import-controller` and `virt-import-fetcher`. It is
+installed from the manifest bundle the repository publishes with each
+release, or through the hyperconverged-cluster-operator (HCO).
```
xpivarc: `or` -> and will be available through HCO. @orenc1 can you ack on this?

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,932 @@
...
+* As a cluster admin, I want imports to use cluster-appropriate storage
+  defaults so I do not need to manually specify volume modes and access
+  modes for each disk.
+
+## Repos
+
+* (Phase 1) [kubevirt/containerized-data-importer](https://github.com/kubevirt/containerized-data-importer) -
+  `Layer` on `DataVolumeSourceRegistry` with annotation-based selection,
+  raw blob handling in importer
+* (Phase 2) [kubevirt/virt-import](https://github.com/kubevirt/virt-import)
```
xpivarc, line 129: For new repos we should also mention maintainers and notify CI team, @dollierp @dhiller

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,932 @@
...
+format uses `application/vnd.kubevirt.disk.raw+zstd` - raw zstd-compressed
+blobs, not tar archives. The kubelet and container runtime would reject
+these. Disk data must be pulled via CDI's pod-pull path.
+
+### Implement the controller in kubevirt/kubevirt
+
+The API and controller could be in-tree, at
+`staging/src/kubevirt.io/api/vmimport/v1alpha1` and
+`cmd/virt-import-controller/`, sharing the OCI artifact format constants in
+`pkg/storage/oci` with the export writer. This was rejected because import
```
xpivarc, line 772: Are you considering to still reuse the format from `kubevirt/kubevirt`?

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,932 @@
...
+
+Minimal impact. Each import creates one `VirtualMachineImport` CR, one
+metadata-fetch Job, one `ConfigMap`, and one `DataVolume` per disk. The
+import controller is a single-replica deployment with leader election and
+standard watch/reconcile mechanics. DataVolume imports use CDI's existing
+scheduling and rate limiting. Intermediate resources (Jobs, ConfigMaps,
+DataVolumes, PVCs) are owned by the import CR and garbage collected on
+deletion.
+
+## Update/Rollback Compatibility
```
xpivarc, line 819: I would mention how versioning match would work between the import controller, CDI and KubeVirt

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,932 @@
...
+
+## Implementation History
+
+* 2026-07-21: Initial proposal. PR: [#396](https://github.com/kubevirt/enhancements/pull/396).
+
+## Graduation Requirements
+
+### Alpha
+
+#### Phase 1: CDI
```
xpivarc, line 917: @Acedus @akalenyu Do you see need for feature gate?

xpivarc (commented):
/approve
/hold
This sounds as very interesting work and I don't have major objections here.
@Acedus @akalenyu Please remove hold once this is up to your satisfaction. 
I would formally support small deadline extension for polishing this (one of the reason is that the VEP targets not core repo).

veps/sig-compute/395-oci-import/vep.md
akalenyu: Nope, no reason to gate IMO

veps/sig-compute/395-oci-import/vep.md
0xFelix: I'd say during Alpha we should not make changes to `kubevirt/kubevirt` and it's API yet. But once all of this is mature we can export the consts through `kubevirt/api`.

veps/sig-compute/395-oci-import/vep.md
0xFelix: Added

veps/sig-compute/395-oci-import/vep.md
0xFelix: Added and waiting on ACK from @orenc1

0xFelix:
@xpivarc @Acedus @akalenyu Made a few adjustments, mostly based on Lubo's feedback and also on https://github.com/kubevirt/enhancements/pull/446.

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,965 @@
...
+
+type LayerDescriptor struct {
+    // Annotation selects a layer by matching an annotation on the
+    // OCI manifest's layer descriptors.
+    Annotation LayerAnnotationSelector `json:"annotation"`
+}
+
+type LayerAnnotationSelector struct {
+    // Key is the annotation key to match on the layer descriptor
+    // (e.g. "io.kubevirt.disk.names").
```
akalenyu, line 210: what's the name->names change about? usually these are singular https://docs.podman.io/en/latest/markdown/podman-artifact-extract.1.html

veps/sig-compute/395-oci-import/vep.md
orenc1: if there will be a proper `csv-generator` for this new component, we could then add it to HCO bundle, yes.

veps/sig-compute/395-oci-import/vep.md
0xFelix: See item 2 in the description of https://github.com/kubevirt/enhancements/pull/446

veps/sig-compute/395-oci-import/vep.md
0xFelix: Yes, adding a `csv-generator` is the idea.

veps/sig-compute/395-oci-import/vep.md
```diff
@@ -0,0 +1,965 @@
...
+| `Progressing` | `False` | `ImportFailed`   | Import failed |
+| `Progressing` | `True` | `Waiting`        | The `cdi.kubevirt.io` API group is not available |
+| `Ready` | `True` | `ImportComplete` | Target resource created |
+| `Ready` | `False` | `ImportFailed`  | Import failed (see message) |
+| `Ready` | `False` | `Waiting`       | The `cdi.kubevirt.io` API group is not available |
+
+#### Import Flow
+
+```
+User creates VirtualMachineImport CR
```
Acedus: nit (feel free to ignore): I think it'd be much more readable if this was mermaid graph.

Acedus (approved):
One small nit, lgtm from sig-storage perspective.

veps/sig-compute/395-oci-import/vep.md
akalenyu: okay lets go with this, I will try to convince you otherwise async, don't want to block this complete VEP.
/lgtm

PR #398, opened 2026-07-17
VEP 25 Extension: Improve Pull Mode Design

Acedus:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/25 
**SIG label**: /sig storage 

### What this PR does

This VEP extension improves upon the previous incremental-backup pull-mode design by evaluating alternatives to the virt-launcher <-> virt-exportserver tunnel used to facilitate the backup process. Details on approach evaluation are included in the VEP under the new `Pull Mode Data Path` section.

### DIFF
--- a/veps/sig-storage/25-incremental-backup/vep.md
+++ b/veps/sig-storage/25-incremental-backup/vep.md
@@ -382,20 +382,12 @@ GET | /exports/{disk}/data | Streams raw disk data. Supports `offset` and `lengt
 ```mermaid
 sequenceDiagram
     participant User
-    participant Server as virt-exportserver (Shim)
-    participant Tunnel as TLS Tunnel
-    participant Launcher as virt-launcher
+    participant Server as virt-exportserver
 
     Note over User, Server: HTTP/1.1 HTTPS
     User->>Server: GET /exports/datavolumedisk1/data?offset=0&length=65536
     
-    Note over Server, Launcher: gRPC over TLS
-    Server->>Tunnel: { "op": "read", "offset": 0, "len": 65536 }
-    Tunnel->>Launcher: Forward Command
-    
-    Launcher->>Launcher: nbdClient Read
-    Launcher-->>Tunnel: Binary Payload
-    Tunnel-->>Server: [Payload]
+    Server->>Server: libnbd Pread (bind-mounted NBD socket)
     
     Server-->>User: 206 Partial Content (Stream)
  ```
@@ -406,32 +398,19 @@ GET | /exports/{disk}/map | Returns JSON of extents (clean or dirty). Params: `o
 ```mermaid
 sequenceDiagram
     participant User
-    participant Server as virt-exportserver (Shim)
-    participant Tunnel as TLS Tunnel
-    participant Launcher as virt-launcher
+    participant Server as virt-exportserver
 
     Note over User, Server: HTTP/1.1 HTTPS (Page 1)
     User->>Server: GET /exports/datavolumedisk1/map?offset=0&length=1073741824&page_size=1
     
-    Note over Server, Launcher: gRPC over TLS tunnel
-    Server->>Tunnel: { "op": "map", "offset": 0, "length": 1073741824}
-    Tunnel->>Launcher: Forward Command
-    
-    Launcher->>Launcher: nbdClient Map
-    Launcher-->>Tunnel: JSON Payload (Chunk 1)
-    Tunnel-->>Server: [Payload]
+    Server->>Server: libnbd BlockStatus (bind-mounted NBD socket)
     
     Server-->>User: {"extents": [...], "next_offset": 1073741824}
 
     Note over User, Server: HTTP/1.1 HTTPS (Page 2 - Follow Pointer)
     User->>Server: GET /exports/datavolumedisk1/map?offset=1073741824&length=1073741824&page_size=1
     
-    Server->>Tunnel: { "op": "map", "offset": 1073741824, "length": 1073741824 }
-    Tunnel->>Launcher: Forward Command
-    
-    Launcher->>Launcher: nbdClient Map
-    Launcher-->>Tunnel: JSON Payload (Chunk 2 / EOF)
-    Tunnel-->>Server: [Payload]
+    Server->>Server: libnbd BlockStatus (bind-mounted NBD socket)
     
     Server-->>User: { "extents": [...], "next_offset": null }
  ```
@@ -496,21 +475,22 @@ The flow is as follows:
 3. Once the backup job is progressing, the backup controller automatically creates an owned `VirtualMachineExport` CR, setting its source to the `VirtualMachineBackup`.
 4. The export-controller reconciles the new `VirtualMachineExport`. It identifies the source type and launches the virt-exportserver pod.
   - Configuration: The controller injects the included volume list (from the `VirtualMachineBackupStatus`) into the pod's environment variables.
-5. Once the export service is ready, the backup controller retrieves the service's name and the internal CA certificate. It then issues yet another `Backup` RPC, this time with the `Export` cmd to the virt-launcher.
-6. The virt-launcher acts as the TLS client. It establishes a persistent, self-healing TLS tunnel to the virt-exportserver via HTTP CONNECT. This reverse-tunnel architecture ensures no network changes are required for either pod.
+5. Once the export pod and service are provisioned (i.e., the `VirtualMachineExport` status has a `ServiceName` and the exporter pod exists), the backup controller instructs virt-handler to bind-mount the virt-launcher's NBD Unix socket into the virt-exportserver pod's filesystem (see [Backup Handler Export Command](#backup-handler-export-command) for the mechanism and [Pull Mode Data Path](#pull-mode-data-path) for design rationale).
+6. The virt-exportserver connects to the bind-mounted NBD Unix socket via libnbd. Since the socket is local to the pod's filesystem, no network communication between the pods is required for data transfer.
 7. Data Serving:
   - The virt-exportserver registers dynamic HTTP handlers for every volume included in the backup under the new `Backups` status field.
   - The backup controller recognizes that the `VirtualMachineExport` is serving and updates the `ExportReady` condition of the `VirtualMachineBackupStatus` to true.
   - The `VirtualMachineBackupStatus` includes the created URLs for every volume in its `includedVolumes`'s list by populating the `mapEndpoint` and `dataEndpoint` fields respectively.
-  - When a client requests a specific endpoint, it invokes the relevant RPC over the tunnel. 
-  - The virt-launcher receives the command and uses an NBD client to query the local Unix socket, it then either:
-    1. Streams the raw binary data back in the case of a `ReadRequest` or;
-    2. Streams back the bitmap in JSON format.
+  - When a client requests a specific endpoint, the virt-exportserver uses libnbd to read directly from the bind-mounted NBD socket and either:
+    1. Streams the raw binary data back in the case of a data request or;
+    2. Returns the bitmap extents in JSON format.
 
 This architecture was chosen because:
 - It preserves pull mode semantics as the data flow remains consumer-driven.
-- The virt-launcher initiates the connection outbound to the export server; no new inbound ports are opened on the VM pod, maintaining the existing security posture.
-- Leverages the existing export controller for lifecycle management of the export pod, while the backup controller manages the data connection.
+- The data path is direct (QEMU, NBD Unix socket, libnbd, HTTP response) with no network I/O between the pods (see [Pull Mode Data Path](#pull-mode-data-path) for benchmarks).
+- No new inbound ports are opened on the VM pod, maintaining the existing security posture.
+- Leverages the existing export controller for lifecycle management of the export pod.
+- The bind-mount mechanism is equivalent to existing filesystem volume hotplug logic via virt-handler.
 - Provisions only a percentage of the total virtual disk size, rather than the full amount, because it relies on scratch space and transfers the backup data over the network.
 
 In contrast:
@@ -528,6 +508,7 @@ sequenceDiagram
     end
 
     box "Data Plane (Pods)"
+        participant VH as Virt-Handler
         participant Launcher as Virt-Launcher
         participant ES as Virt-Exportserver
     end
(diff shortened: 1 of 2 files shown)


veps/sig-storage/incremental-backup.md
```diff
@@ -649,6 +624,58 @@ A:
...
+
+```
+QEMU | NBD Unix socket | libnbd Pread | HTTP response writer
+```
+
+This eliminates the need for protobuf serialization, gRPC buffer pool management, double HTTP/2 framing, and inter-pod mTLS encryption that would be required by a gRPC-based tunnel.
+
+**Implementation details:**
+- The virt-exportserver pod is scheduled on the same node as the VMI, enabling the bind-mount.
+- The export pod declares an emptyDir volume at `/var/run/kubevirt/nbd` with `MountPropagationHostToContainer`, allowing the bind-mount performed by virt-handler on the host to propagate into the container.
```
alromeros, line 643: I assume all this performance increase is just compatible with online incremental backups, as it relies on bind-mounting the virt-launcher's NBD socket. I'm still wrapping my head around if this would work with the offline backup. If it doesn't, we just need to ensure the old path remains viable on the export server side for the offline scenario. I think it's a fair trade-off either way.

alromeros (commented):
Looking great!

veps/sig-storage/incremental-backup.md
Acedus: It should work with offline backup, we just wouldn't need to bind mount anything.

veps/sig-storage/incremental-backup.md
```diff
@@ -496,21 +475,22 @@ The flow is as follows:
...
+    2. Returns the bitmap extents in JSON format.
 
 This architecture was chosen because:
 - It preserves pull mode semantics as the data flow remains consumer-driven.
-- The virt-launcher initiates the connection outbound to the export server; no new inbound ports are opened on the VM pod, maintaining the existing security posture.
-- Leverages the existing export controller for lifecycle management of the export pod, while the backup controller manages the data connection.
+- The data path is direct (QEMU, NBD Unix socket, libnbd, HTTP response) with no network I/O between the pods (see [Pull Mode Data Path](#pull-mode-data-path) for benchmarks).
+- No new inbound ports are opened on the VM pod, maintaining the existing security posture.
+- Leverages the existing export controller for lifecycle management of the export pod.
+- The bind-mount mechanism is analogous to existing volume hotplug logic via virt-handler.
```
akalenyu: Do you mean equivalent instead of analogous? When reasoning about the security model, this would make folks feel much better. I feel like this is in fact equivalent to hotplug, not analogous.

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +624,58 @@ A:
...
+QEMU | NBD Unix socket | libnbd Pread | HTTP response writer
+```
+
+This eliminates the need for protobuf serialization, gRPC buffer pool management, double HTTP/2 framing, and inter-pod mTLS encryption that would be required by a gRPC-based tunnel.
+
+**Implementation details:**
+- The virt-exportserver pod is scheduled on the same node as the VMI, enabling the bind-mount.
+- The export pod declares an emptyDir volume at `/var/run/kubevirt/nbd` with `MountPropagationHostToContainer`, allowing the bind-mount performed by virt-handler on the host to propagate into the container.
+- The virt-exportserver watches for the socket file to appear and connects to it via `libnbd.ConnectUnix`. Reads are performed using `Pread` in 1 MiB chunks and streamed directly into the HTTP response.
+- Cleanup of the bind-mount occurs when the `VirtualMachineBackup` resource is deleted.
```
akalenyu, line 662: I think hotplug path has some additional logic to protect against abnormal deletions, might benefit from the (painfully accumulated :laughing:) experience

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +624,58 @@ A:
...
+```
+QEMU | NBD Unix socket | libnbd Pread | HTTP response writer
+```
+
+This eliminates the need for protobuf serialization, gRPC buffer pool management, double HTTP/2 framing, and inter-pod mTLS encryption that would be required by a gRPC-based tunnel.
+
+**Implementation details:**
+- The virt-exportserver pod is scheduled on the same node as the VMI, enabling the bind-mount.
+- The export pod declares an emptyDir volume at `/var/run/kubevirt/nbd` with `MountPropagationHostToContainer`, allowing the bind-mount performed by virt-handler on the host to propagate into the container.
+- The virt-exportserver watches for the socket file to appear and connects to it via `libnbd.ConnectUnix`. Reads are performed using `Pread` in 1 MiB chunks and streamed directly into the HTTP response.
```
akalenyu, line 660: what happens to requests coming in before socket file appears? maybe we could hold off on export readiness over this? no traffic arrives until the bind-mount is complete, no polling needed?

veps/sig-storage/incremental-backup.md
```diff
@@ -496,21 +475,22 @@ The flow is as follows:
...
 4. The export-controller reconciles the new `VirtualMachineExport`. It identifies the source type and launches the virt-exportserver pod.
   - Configuration: The controller injects the included volume list (from the `VirtualMachineBackupStatus`) into the pod's environment variables.
-5. Once the export service is ready, the backup controller retrieves the service's name and the internal CA certificate. It then issues yet another `Backup` RPC, this time with the `Export` cmd to the virt-launcher.
-6. The virt-launcher acts as the TLS client. It establishes a persistent, self-healing TLS tunnel to the virt-exportserver via HTTP CONNECT. This reverse-tunnel architecture ensures no network changes are required for either pod.
+5. Once the export service is ready, the backup controller instructs virt-handler to bind-mount the virt-launcher's NBD Unix socket into the virt-exportserver pod's filesystem. This mechanism is analogous to the existing volume hotplug flow (see [Pull Mode Data Path](#pull-mode-data-path) for design rationale).
```
akalenyu: should we elaborate how backup controller would instruct the handler to perform the bind mount? would it involve an API similar to hotplug?

akalenyu (commented):
Great stuff!

I think this change would be much easier to reason about (especially with non sig-storage approvers) if we create a short section about equivalence to hotplug mechanism from security perspective

veps/sig-storage/incremental-backup.md
awels: Good question, I guess the virt-handler needs to identify this should take place based on the state of the VMI (which is the only thing it knows about)

veps/sig-storage/incremental-backup.md
awels: I would say equivalent to the filesystem bind mount logic. block is a whole different can of worms

veps/sig-storage/incremental-backup.md
```diff
@@ -633,14 +608,14 @@ func (s *exportServer) getBackupHandlerMap(bi export.BackupInfo) map[string]http
...
 
 Q: How are failures handled?
 
 A:
-- virt-exportserver failure: If the virt-exportserver pod is terminated, the export controller detects the state change and recreates the pod. Since the virt-launcher connects via a k8s service (which maintains a stable ClusterIP), the launcher's connection manager simply detects the TCP connection drop and enters a retry loop. Once the new server pod is ready and the endpoints update, the tunnel is re-established automatically without interrupting the backup job logic.
+- virt-exportserver failure: If the virt-exportserver pod is terminated, the export controller detects the state change and recreates the pod. The backup controller then instructs virt-handler to re-establish the bind-mount of the NBD socket into the new pod. Once the new server pod is ready and the socket is mounted, data serving resumes without interrupting the backup job logic.
```
awels: If the virt-exportserver pod is down, then the backup job would either receive a 502 or 503 if using the proxy and simply lose the connection if connecting through the service. I would argue that the client needs to be able to handle that scenario and we should call that out.

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +624,58 @@ A:
...
+Rather than tunneling backup data through an intermediary protocol, virt-handler bind-mounts the virt-launcher's NBD Unix socket directly into the virt-exportserver pod's filesystem. The virt-exportserver then connects to the QEMU NBD server via libnbd locally, resulting in a minimal data path:
+
+```
+QEMU | NBD Unix socket | libnbd Pread | HTTP response writer
+```
+
+This eliminates the need for protobuf serialization, gRPC buffer pool management, double HTTP/2 framing, and inter-pod mTLS encryption that would be required by a gRPC-based tunnel.
+
+**Implementation details:**
+- The virt-exportserver pod is scheduled on the same node as the VMI, enabling the bind-mount.
```
awels, line 658: We implemented an affinity for this exact scenario, but that affinity is preferred, not required. So in theory we could end up on a different node, should we fail the backup in that case? Or delete the export server pod and try again?

veps/sig-storage/incremental-backup.md
```diff
@@ -649,6 +624,58 @@ A:
...
+
+| Approach | Throughput | vs. Baseline |
+|---|---|---|
+| gRPC tunnel, sync Pread, 256 KiB chunks (baseline) | 89.4 MB/s | - |
+| gRPC tunnel, libnbd AIO (16 in-flight), 256 KiB chunks | 103 MB/s | +15% |
+| gRPC tunnel, libnbd AIO (16 in-flight), 1 MiB chunks | 117 MB/s | +31% |
+| **NBD socket bind-mount, sync Pread, 1 MiB chunks** | **132 MB/s** | **+48%** |
+
+> [!NOTE]
+> The bind-mount mechanism relies on virt-handler, analogous to the existing volume hotplug logic. This introduces a dependency on virt-handler for establishing and re-establishing the socket mount, which follows the same delegation pattern used by disk hotplug.
```
awels: Same analogous vs equivalence comment, and same mention filesystem here since block is a different beast.

veps/sig-storage/25-incremental-backup/vep.md
Acedus: I think we should just fix the affinity to be required, do you see any issue with that?

awels: In the PR I had it required, and I was asked to change it to preferred. I can definitely make it required.

Acedus: (I guess it depends on whether we choose to proceed with the proposed approach :P)

alromeros: Yep preferred doesn't make sense if the whole mechanism depends on co-location, we need to document the fallback here in case affinity remains unchanged

alromeros: But better to just make it required

alromeros:
We have plenty of time for others to chime in and maybe raise more comments, but it looks pretty good to me
/lgtm

awels (approved):

alromeros:
If the plan is to graduate both VEP 90 (utility volumes) and this VEP to beta, maybe we should use this PR to update the document and mark it as beta.

Acedus:
> If the plan is to graduate both VEP 90 (utility volumes) and this VEP to beta, maybe we should use this PR to update the document and mark it as beta.

https://github.com/kubevirt/enhancements/pull/433

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +640,59 @@ A:
...
+- Cleanup of the bind-mount occurs when the `VirtualMachineBackup` resource is deleted.
+
+**Throughput benchmarks** (full 20 GiB disk pull, external network storage):
+
+| Approach | Throughput | vs. Baseline |
+|---|---|---|
+| gRPC tunnel, sync Pread, 256 KiB chunks (baseline) | 89.4 MB/s | - |
+| gRPC tunnel, libnbd AIO (16 in-flight), 256 KiB chunks | 103 MB/s | +15% |
+| gRPC tunnel, libnbd AIO (16 in-flight), 1 MiB chunks | 117 MB/s | +31% |
+| **NBD socket bind-mount, sync Pread, 1 MiB chunks** | **132 MB/s** | **+48%** |
```
iholder101, line 671: nit: The `+48%` is vs the unoptimized tunnel. Vs the optimized tunnel (i.e. `AIO + 1MiB chunks: 117 MB/s) the real gain is ~13%. Maybe worth adding another line.

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +640,59 @@ A:
...
+
+> [!NOTE]
+> The bind-mount mechanism relies on virt-handler, equivalent to the existing filesystem volume hotplug logic. This introduces a dependency on virt-handler for establishing and re-establishing the socket mount, which follows the same delegation pattern.
+> Additionally, the virt-exportserver pod must be scheduled on the same node as the virt-launcher. This constraint is consistent with the hotplug model and benefits data locality, but it does restrict pod placement.
```
iholder101, line 675: So this means that under a very busy or a tainted node it wouldn't be necessarily possible to perform CBT backups. Would this be problematic in your opinion? was it discussed?

veps/sig-storage/25-incremental-backup/vep.md
```diff
@@ -649,6 +640,59 @@ A:
...
+
+#### Alternative: gRPC Tunnel
+
+In this alternative, the virt-launcher establishes a gRPC-over-HTTP/2-CONNECT tunnel to the virt-exportserver. Each read request traverses the following path:
```
iholder101, line 679: Interesting.
If we need to support backups across nodes (i.e. if the node is tainted) this could be a solution in the future. But since the underlying technology is not user-facing anyhow, this could be left for a future extension if needed.

iholder101 (commented):
Thank you @Acedus!

Looks like a great improvement to me, makes absolute sense!
Left some small questions below.

TBH I think I'm happy with the current state. Adding a hold just to give you a chance to reply and change something if needed, but feel free to remove the hold if you're happy as well.

/approve
/hold

veps/sig-storage/25-incremental-backup/vep.md
Acedus: I don't mind adding a line, but I think it's already pretty clear from the table.

iholder101: it's a nit, feel free to ignore :+1:

veps/sig-storage/25-incremental-backup/vep.md
Acedus: Good question, more so for the tainted node aspect since we currently don't add tolerations to the virt-exportserver pod. Having said that, I believe it's a reasonable tradeoff since it's a similar one to hotplugs.

veps/sig-storage/25-incremental-backup/vep.md
Acedus: That's the current solution and it's pretty over-engineered IMO (hence this VEP). I would very much like to examine "infra" ingress for virt-launchers that won't undergo DNAT exactly for such usecases, though.

Acedus:
Thanks for the review @iholder101, please unhold if you believe my responses to your comments are satisfactory.

iholder101:
> Thanks for the review @iholder101, please unhold if you believe my responses to your comments are satisfactory.

Thanks a lot @Acedus!
/unhold

Issue #399, opened 2026-07-19
VEP 399: Multi-checkpoint Incremental Backup


Acedus:
**Primary contact (assignee)**: @Acedus 

/assign @Acedus 

**Current Feature Stage**:

New

**Feature Gate**:

`MultiCheckpointIncrementalBackup`

**Responsible SIGs**: 

Primary SIG:
/sig storage

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey @Acedus. Can you please change the title to "VEP 399"? You can maybe change it to something like "VEP 399: Multi-checkpoint Incremental Backup (follow-up to VEP 25)".

This avoids confusion and is easier to parse by the vep-agent and other bots. Same for https://github.com/kubevirt/enhancements/issues/416.

Thank you!

PR #400, opened 2026-07-19
VEP 399: Multi-checkpoint Incremental Backup

Acedus:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/399
**SIG label**: /sig-storage

### What this PR does
This is an extension of https://github.com/kubevirt/enhancements/issues/25 that introduces multi-checkpoint retention, enabling backup vendors to retry incremental backups from older checkpoints when upstream failures occur with built-in checkpoint pruning.


veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
```diff
@@ -0,0 +1,259 @@
...
+2. If `forceFullBackup` was set, all previous checkpoints are purged before the new one is recorded.
+
+#### 2. Checkpoint Retention and Pruning
+The `retainCheckpoints` field controls the maximum checkpoint count. When the checkpoint list exceeds this value, the controller prunes excess checkpoints (oldest first):
+1. The controller verifies no active (non-terminal) backups reference the tracker. If any exist, pruning is deferred.
+2. For each excess checkpoint, a `DeleteCheckpoint` RPC is sent to virt-launcher, which removes the bitmap from the qcow2 file via libvirt.
+3. If a `DeleteCheckpoint` RPC fails mid-pruning, partial progress is persisted (already-pruned entries are removed from the list) and the controller requeues.
+4. If for some reason a checkpoint doesn't exist but a bitmap does, a checkpoint redefinition followed by deletion will be attempted to allow libvirt to handle the cleanup.
+
+#### 3. Checkpoint Redefinition (after VM restart)
```
alromeros, line 87: I was just thinking about how this is going to interact with the offline cbt vep I'm working on. We probably need to agree on the exact ordering of operations when a VM boots. If my vep has deferred bitmaps to inject via QMP, and this VEP needs to redefine and prune existing checkpoints, what goes first? I assume we should materialize its pending bitmaps via QMP first, so that your redefinition logic can actually see them on disk and restore the libvirt metadata properly.

alromeros (commented):
Looks really solid!

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
Acedus: > I assume we should materialize its pending bitmaps via QMP first

I don't think deferred materialization at boot can work. Bitmap creation needs to be tied directly to a successful backup, otherwise we risk inconsistency. There's also the open question of how you'd serialize pending bitmaps in a way that survives between shutdown and start.

I think the offline backup process should write bitmaps to the overlay as part of its own completion flow (probably atomically with backup success), and update the VMBT status with the new checkpoint. That way, by the time the VM boots, the bitmaps are already on disk.

As for the boot-time ordering for both modes, I think it should be:
1. Redefine all checkpoints across all associated VMBTs. 
2. Checkpoint pruning runs both at VM start (as required to set CBT status to Enabled) AND as part of VMBT reconciliation. 

If an offline backup pushes checkpoints past `retainCheckpoints` while the VM is off, pruning will be deferred until boot (I believe we can tolerate this as we'd never reach a state where too many offline backups prevent the VM from starting due to bitmap inflation since at some point we'd fail bitmap creation due to insufficient storage space anyway (and that space would be reclaimed as soon as the VM starts again). Having said that, if we'd want to support indefinite offline backups then that wouldn't work (unless the user extends the PVC).

alromeros: Makes sense, I was trying to avoid mounting the vmstate PVC as RW in the export pod, but yeah... probably deferred materialization adds way more risk and anti-patterns than simply mounting the PVC RW while the VM is stopped. I'll update my VEP accordingly.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
```diff
@@ -0,0 +1,259 @@
...
+### Checkpoint Lifecycle
+
+#### 1. Checkpoint Creation (on backup completion)
+When a `VirtualMachineBackup` completes successfully and its source is a `VirtualMachineBackupTracker`:
+1. The backup controller appends a `BackupCheckpoint` to `status.checkpoints` and updates `status.latestCheckpoint` to point to it.
+2. If `forceFullBackup` was set, all previous checkpoints are purged before the new one is recorded.
+
+#### 2. Checkpoint Retention and Pruning
+The `retainCheckpoints` field controls the maximum checkpoint count. When the checkpoint list exceeds this value, the controller prunes excess checkpoints (oldest first):
+1. The controller verifies no active (non-terminal) backups reference the tracker. If any exist, pruning is deferred.
```
awels: What happens if we have an active backup and it completes, does the pruning happen at that point in time?

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
```diff
@@ -0,0 +1,259 @@
...
+1. virt-handler detects trackers with existing checkpoints during CBT initialization (transition from `Initializing` -> `Enabled`) and sets `status.checkpointRedefinitionRequired = true`.
+2. virt-controller walks the checkpoint chain sequentially (oldest -> newest), calling `RedefineCheckpoint` RPC for each. Each checkpoint is redefined with its predecessor as the parent, preserving the QEMU bitmap parent -> child chain.
+3. Error handling:
+	1. If the checkpoint is invalid, the chain is truncated at the failing checkpoint. All checkpoints from the failing one onward are deleted in reverse order (leaf -> root) via `DeleteCheckpoint`, which uses QMP `block-dirty-bitmap-remove` for any inconsistent bitmap encountered during cleanup. The next backup will be full.
+	2. Requeue on transient errors with no truncation.
+4. After processing, `checkpointRedefinitionRequired` is cleared.
+
+#### 4. Checkpoint Deletion (on tracker deletion)
+When a VMBT is deleted, its finalizer ensures bitmap cleanup:
+1. If active (non-terminal) backups reference the tracker, deletion is blocked.
```
awels: Same question, what happens after that active backup completes, does the process run at that point or is there some other trigger?

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
```diff
@@ -0,0 +1,259 @@
...
+2. If the VMI is running with CBT enabled, all checkpoints are deleted in reverse order (newest -> oldest) via `DeleteCheckpoint` RPC. Partial failures leave remaining checkpoints for retry.
+3. If the VM is stopped, bitmap cleanup is deferred until the VM starts.
+4. If the VM has been permanently deleted (both VM and VMI gone), checkpoints are cleared immediately (no bitmap cleanup needed since the disks are gone).
```
awels: Is this true? What if I have a VM that uses a disk that it doesn't own, and we delete the VM. The disk will remain. I would argue the bitmaps should be cleaned up, but if the backend state PVC gets deleted automatically that problem should go away.

awels (commented):
Looks good a few minor questions, and it appears your mermaid graphs are not rendering properly in GH.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
Acedus: I was contemplating that, yes, I believe it should. I'll update the doc to reflect that.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
Acedus: > I would argue the bitmaps should be cleaned up, but if the backend state PVC gets deleted automatically that problem should go away.

That was the intent of that bullet. The checkpoints (and their respective bitmaps) are effectively gone.

alromeros: The vmstate PVC is now deleted automatically but with the new API https://github.com/kubevirt/enhancements/pull/313 a referenced PVC will be able to outlive the VM lifecycle. So the bullet point should be specifically applied for that case.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
```diff
@@ -0,0 +1,259 @@
...
+Multi-checkpoint support is additive to the existing single-checkpoint architecture. The `VirtualMachineBackupTracker` is extended from tracking a single `latestCheckpoint` to maintaining an ordered list of retained checkpoints, with count-based pruning.
+
+```mermaid
+graph TD
+    VMBT["VirtualMachineBackupTracker<br/>spec.retainCheckpoints: 3<br/>status.checkpoints: [cp-1, cp-2, cp-3]<br/>status.latestCheckpoint: cp-3"]
+    VM["VirtualMachine<br/>qcow2 bitmaps: cp-1, cp-2, cp-3"]
+    VMBT -- "references" --> VM
+```
+
+Each backup creates a new checkpoint and appends it to the tracker's checkpoint list. Excess checkpoints beyond `retainCheckpoints` are pruned automatically (see Checkpoint Retention and Pruning below).
```
alromeros: How much growth in state PVC usage do we expect with the increased use of retained checkpoints? Even with `retainCheckpoints` defaulting to 1, should we  consider increasing the default PVC size again for this use case? Not a must for the vep, just something to consider.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
Acedus: Indeed, but as stated in #313, CBT state is wiped on adoption so this isn't a concern.

veps/sig-storage/25-incremental-backup/25-1-multi-checkpoint/multi-checkpoint.md
Acedus: The default PVC size was actually calculated according to an expected increase in checkpoints, unrelated specifically to multi-checkpoint, rather the ability to have $n$ VMBTs (see https://github.com/kubevirt/kubevirt/blob/f179b42968c3b4bb80811d8e05b5b337a5376575/pkg/storage/cbt/cbt.go#L41-L57).

alromeros:
Looks pretty good already. I'll wait until you address @awels's pruning concern, but it's already lgtm-worthy imo.

awels:
> Looks pretty good already. I'll wait until you address @awels's pruning concern, but it's already lgtm-worthy imo.

Agreed once you address that you will have my lgtm

Acedus:
/cc @awels @alromeros @akalenyu @mhenriks 
 
 I completely rewrote the VEP according to VEP 25 beta state.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+
+Pruning is deferred while a non-terminal `VirtualMachineBackup` references the same `VirtualMachineBackupTracker`, since removing a bitmap from under a running job is unsafe. The deferral is scoped to that `VirtualMachineBackupTracker` specifically, meaning a backup on one does not hold up pruning on another, because the bitmaps are disjoint. `status.checkpoints` is therefore a steady-state bound rather than an instantaneous one, which matters mostly in pull mode where a backup lives until its client deletes it or its TTL expires.
+
+### Backing up from an older checkpoint
+
+`VirtualMachineBackup` gains `spec.fromCheckpoint`, naming the checkpoint to use as the incremental base, unset means `latestCheckpoint`.
+
+Resolution is per disk, a disk is incremental when the named checkpoint's bitmap is present and usable on it (as established in [VEP 25](../vep.md)), and full otherwise. Each checkpoint is judged on its own bitmap, so a disk missing `b(C1)` but still holding `b(C2)` is full for a `C1`-based backup and incremental for a `C2`-based one (i.e., partially covered chain degrades disk by disk instead of failing the backup).
+
+### Redefinition after restart
```
akalenyu: Waiting for a rewrite following offline discussion

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+
+`VirtualMachineBackupTracker` gains `spec.retainCheckpoints` and an ordered `status.checkpoints` list.
+
+`spec.retainCheckpoints` is the desired chain length, and the controller reconciles toward it on every sync rather than only when a backup completes. A completed backup appends its checkpoint, and anything past the count is pruned oldest first.
+
+Pruning is deferred while a non-terminal `VirtualMachineBackup` references the same `VirtualMachineBackupTracker`, since removing a bitmap from under a running job is unsafe. The deferral is scoped to that `VirtualMachineBackupTracker` specifically, meaning a backup on one does not hold up pruning on another, because the bitmaps are disjoint. `status.checkpoints` is therefore a steady-state bound rather than an instantaneous one, which matters mostly in pull mode where a backup lives until its client deletes it or its TTL expires.
+
+### Backing up from an older checkpoint
+
+`VirtualMachineBackup` gains `spec.fromCheckpoint`, naming the checkpoint to use as the incremental base, unset means `latestCheckpoint`.
```
akalenyu, line 111: What about validation? would a vendor want to know their chain broke? maybe we should alert?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+      type: Full
+```
+
+## Alternatives
+
+A separate checkpoint CRD: a checkpoint has no lifecycle independent of its chain, ordering would have to be reconstructed from object metadata, and it multiplies API objects per VM per backup without adding expressiveness.
+
+Falling back to the nearest newer usable base instead of full: a disk missing the requested base but holding a later one could produce a smaller delta, but the vendor asked for the base whose data they hold. Applying a later delta to it drops the writes in between, and the result looks valid.
+
+## Scalability
```
akalenyu, line 225: I think auto grow PVCs are starting to make a lot more sense following this VEP (new unbounded variable per VM). Should we mention things like https://docs.portworx.com/portworx-enterprise/operations/scale-portworx-cluster/autopilot/use-cases/autogrow-pvc in the VEP?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+
+This independence is what keeps vendors out of each other's way, each `VirtualMachineBackupTracker` already owns its chain outright and pruning in one cannot invalidate another's bases.
+
+### Checkpoint retention
+
+`VirtualMachineBackupTracker` gains `spec.retainCheckpoints` and an ordered `status.checkpoints` list.
+
+`spec.retainCheckpoints` is the desired chain length, and the controller reconciles toward it on every sync rather than only when a backup completes. A completed backup appends its checkpoint, and anything past the count is pruned oldest first.
+
+Pruning is deferred while a non-terminal `VirtualMachineBackup` references the same `VirtualMachineBackupTracker`, since removing a bitmap from under a running job is unsafe. The deferral is scoped to that `VirtualMachineBackupTracker` specifically, meaning a backup on one does not hold up pruning on another, because the bitmaps are disjoint. `status.checkpoints` is therefore a steady-state bound rather than an instantaneous one, which matters mostly in pull mode where a backup lives until its client deletes it or its TTL expires.
```
akalenyu: Should we define some hard limit to avoid TTL being the only backstop? `min(hardstop, ttl)`
Technically folks could set TTL to indefinite

veps/sig-storage/25-incremental-backup/incremental-backup.md
akalenyu, line 1: Maybe keep this file renamed for future reference?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+    - name: my-vm-tracker-2026-03-01_12-00-00
+      creationTime: "2026-03-01T12:00:00Z"
+    - name: my-vm-tracker-2026-03-02_12-00-00
+      creationTime: "2026-03-02T12:00:00Z"
+    - name: my-vm-tracker-2026-03-03_12-00-00
+      creationTime: "2026-03-03T12:00:00Z"
+  latestCheckpoint:
+    name: my-vm-tracker-2026-03-03_12-00-00
+    creationTime: "2026-03-03T12:00:00Z"
+  lastTrackedPodUID: 7d3f1c1e-0b7a-4a1e-9a2b-5f0c1d2e3a4b
```
akalenyu: hm?

akalenyu (commented):
solid! mainly waiting for convergence on redefinition after restart

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: Was a replacement for checkpoint redefinition logic, will be gone as part of offline discussion.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: IIRC VMExport doesn't have a hardstop either, don't see why we should limit it (it's the consumer's responsibility).

veps/sig-storage/25-incremental-backup/incremental-backup.md
Acedus: I believe we have the previous commits for that, I don't see a compelling reason to keep it.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: So with mixed disks the way of knowing if a chain broke (partially or fully) is to examine the disks' backup type on the VirtualMachineBackup status, that way a consumer can deem for themselves whether they're fine with proceeding (might be that from checkpoint C1 all disks are now full) or not by either cancelling (push mode semantics) or signaling completion and performing a backup again from a different checkpoint.
We'll clean up phantom checkpoints (i.e., ones that have no corresponding bitmaps at all) on VMI restart.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
akalenyu: I'm talking about pruning deferral hard stop

Acedus: I'd rather avoid time-based expiration.

awels (approved):

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,271 @@
...
+metadata:
+  name: my-vm-tracker
+  namespace: default
+  generation: 2
+spec:
+  source:
+    apiGroup: kubevirt.io
+    kind: VirtualMachine
+    name: my-vm
+  retainCheckpoints: 3
```
iholder101: Using primitives as field types is sometimes problematic as it's unextendible. In the future, maybe we'd want to add more configurations here (e.g. define a TTL)?

Maybe something like this is better? WDYT?
```yaml
retainCheckpoints:
  checkpointsToRetain: 3
```

alromeros (approved):

iholder101 (commented):
Thanks @Acedus! Overall looks good.

I've glanced upon the details, letting sig-storage LGTM the more storage-related aspects, but general direction looks good to me.

Left one suggestion regarding the API.

iholder101:
@akalenyu are you happy?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: I think while relevant it falls out of scope.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: I highly doubt we would due to the established cleanup semantics, but I see no harm in allowing extension.

Acedus:
/hold

Waiting for @alromeros and @akalenyu to give their lgtms as well.

alromeros (approved):

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Update/Rollback Compatibility
+
+All new fields are optional with backward-compatible defaults, and the feature is guarded by a new `MultiCheckpointIncrementalBackup` feature gate.
+
+**Upgrade**: existing `VirtualMachineBackupTracker`s keep working. `status.checkpoints` is populated from `latestCheckpoint` on first reconciliation and `historyLimit` defaults to 1, so an upgraded `VirtualMachineBackupTracker` behaves exactly as before until a vendor opts into a larger value. Raising `historyLimit` takes effect from the next backup onward and cannot recover checkpoints already trimmed.
+
+**Rollback**: an older controller ignores `status.checkpoints` and uses `latestCheckpoint`, and ignores `fromCheckpoint` on in-flight backups. Deltas from `latestCheckpoint` stay correct, because that bitmap is self-contained and does not depend on the checkpoints the old controller stopped tracking. The other retained checkpoints' bitmaps remain in the QCOW2 overlays, and an older controller has no `Reclaim` to remove them, so the cost of a rollback is wasted space rather than a broken chain.
```
mhenriks, line 239: I don't lose much sleep over rollback but consider the following:

1. You still have backup C1
2. A disk block changes between C1 and C2
3. You lose backup C2, so you request a new incremental backup from C1
4. Before that request starts, the controller is downgraded. The older controller ignores fromCheckpoint: C1 and uses the latest checkpoint, C2
5. Its delta excludes that block if it hasn’t changed again since C2. Applying the delta to C1 leaves the block with its old contents

Not sure if we can somehow force full backups after downgrade. But should maybe document that

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+
+For the reference geometry each retained checkpoint costs roughly ~1.25 MiB of overlay per disk, `ceil(V / G / 8 / S) x S` of bitmap data plus one cluster for the bitmap table, so about ~6.25 MiB per checkpoint across 5 disks. QEMU additionally holds each bitmap in memory at `V / G / 8`, about 1 MiB per bitmap per disk, and every bitmap transfers with its disk during live migration.
+
+## Update/Rollback Compatibility
+
+All new fields are optional with backward-compatible defaults, and the feature is guarded by a new `MultiCheckpointIncrementalBackup` feature gate.
+
+**Upgrade**: existing `VirtualMachineBackupTracker`s keep working. `status.checkpoints` is populated from `latestCheckpoint` on first reconciliation and `historyLimit` defaults to 1, so an upgraded `VirtualMachineBackupTracker` behaves exactly as before until a vendor opts into a larger value. Raising `historyLimit` takes effect from the next backup onward and cannot recover checkpoints already trimmed.
```
mhenriks, line 237: Do we have to ensure every tracker has populated status.checkpoints before the first Reclaim runs?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+
+Reconciling a set rather than replaying individual deletions means ordering carries no correctness weight. A crash between trimming `status.checkpoints` and deleting the bitmap leaves an orphan, which is by definition outside the next desired set. A deleted `VirtualMachineBackupTracker` needs no cleanup RPC for the same reason, its checkpoints simply leave the union.
+
+`forceFullBackup` on a `VirtualMachineBackupTracker` source reduces `status.checkpoints`, on completion, to just the checkpoint that backup created, and the retained bases are then collected like anything else outside the set. It serves as an escape hatch for a chain a vendor no longer trusts, so it discards the retained bases rather than appending to them. A forced backup that fails before creating its checkpoint leaves `status.checkpoints` unchanged, so the chain it was meant to replace survives and the vendor can retry.
+
+The controller builds the desired set from synced informers and skips the `Reclaim` call, not the backup, if any are not. A tracker missing from that union looks exactly like a tracker with no checkpoints, and its bitmaps would be deleted while its owner is still holding data against them.
```
mhenriks, line 146: What if tracker A finishes a backup and tracker B starts another before A’s update reaches the cache? Could cleanup could miss A’s new checkpoint and delete its bitmap?

mhenriks (commented):
Looks really great @Acedus! Just a couple spots where additional detail may be nice

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: So in the naive case no, because it'll simply be the unified sets and an empty set just means that tracker has no checkpoints worth keeping. Having said that, I'll change it to also include `latestCheckpoint` as that may be populated before `status.checkpoints`.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: Since backups can't run concurrently, and a tracker patch precedes a volume detach, I'd say that caches have a reasonably long enough time (at least longer by several orders of magnitude than the volume detach process) to sync before backup A becomes terminal and backup B can start. The price of a cache miss _is_ full backup which is unfortunate, but it's just highly unlikely and doesn't result in corruption. WDYT?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: Interesting scenario, I'd say the fix would be to cancel `fromCheckpoint` backups before rollback but it's such a niche case I'd rather just avoid it entirely at the potential cost of a backup going haywire. WDYT?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
iholder101: Agreed the window is narrow and there's no corruption risk. But "just a full backup" is the exact cost this VEP exists to eliminate - on a multi-TB VM it's the failure we're spending a whole feature to avoid.

It also makes correctness rest on volume detach being slow, which nothing enforces and no test would catch if it stopped being true.

We can maybe consider marking on first sighting outside the desired set and delete on the next Reclaim if still outside? It'd mean one extra cycle of wasted space, and the window closes by construction rather than by timing. Just a thought.

iholder101 (commented):
Left a non-blocking comment which we can revise in the future if necessary.

Thank you @Acedus! Looks great!

/approve

Acedus:
> Agreed the window is narrow and there's no corruption risk. But "just a full backup" is the exact cost this VEP exists to eliminate - on a multi-TB VM it's the failure we're spending a whole feature to avoid.

> It also makes correctness rest on volume detach being slow, which nothing enforces and no test would catch if it stopped being true.

> We can maybe consider marking on first sighting outside the desired set and delete on the next Reclaim if still outside? It'd mean one extra cycle of wasted space, and the window closes by construction rather than by timing. Just a thought.

I agree that we should value correctness over storage space reclamation here... I'm not sure how I feel about marking _and then_ deleting, though. I'd like to circle back to this in beta if possible.

akalenyu:
/lgtm
/hold cancel

thanks @Acedus!

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+2. Deleting a checkpoint deletes its own bitmaps and nothing else.
+3. The parent relationship of checkpoints is metadata only (nothing in the backup path reads it).
+
+This independence is what keeps vendors out of each other's way, each `VirtualMachineBackupTracker` already owns its chain outright and trimming one cannot invalidate another's bases.
+
+### Checkpoint retention
+
+`VirtualMachineBackupTracker` gains `spec.checkpointRetention` and an ordered `status.checkpoints` list.
+
+`spec.checkpointRetention.historyLimit` is the desired chain length. A completed backup appends its checkpoint and drops anything past the count, oldest first, and it does both in the same status write in order to avoid coordination issues.
```
hao-zhou-rubrik, line 103: 1. From the CR section, "Default 1". If I set this value too large, what will happen when the chain grow beyond what `CBTBackendStateOverhead` can handle? 
  2. Any performance concern if the chain length grow?
  3. This will be a major limitation for some new features we are planning. We want the capability to query delta for old snapshots without chain length limitation like what vSphere CBT offers.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+
+The append is already ordered ahead of everything else on the completion path, it precedes the teardown that releases the VMI for the next backup, so a checkpoint is never in existence-but-unnamed state while another backup is able to start.
+
+That leaves two writes to `status.checkpoints` that no backup drives: lowering `spec.checkpointRetention.historyLimit`, and dropping a checkpoint [cleanup](#cleanup) reported as absent. Both are ordinary reconciles, and both are a status write and nothing else.
+
+### Backing up from an older checkpoint
+
+`VirtualMachineBackup` gains `spec.fromCheckpoint`, naming the checkpoint to use as the incremental base, unset means `latestCheckpoint`.
+
+Resolution is per disk, a disk is incremental when the named checkpoint's bitmap is present and usable on it (as established in [VEP 25](../vep.md)), and full otherwise. Each checkpoint is judged on its own bitmap, so a disk missing `b(C1)` but still holding `b(C2)` is full for a `C1`-based backup and incremental for a `C2`-based one (i.e., partially covered chain degrades disk by disk instead of failing the backup).
```
hao-zhou-rubrik, line 113: Regarding "partially covered chain degrades disk by disk instead of failing the backup"  
  1. Can we fail the backup in this case or give vendor a parameter to fail the backup in this case? Considering this is a unfulfillable situation.  
  2. Any status/field indicates that a `full` is caused by missing bitmap or some other reasons?
  3. Follow up on point 2, what happens if there's a new disk added to the VM after checkpoint C1, and I take a new backup against C1? Will the new disk be `full`, will there be an indicate that this `full` is because it's a new disk?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
```diff
@@ -0,0 +1,291 @@
...
+
+Neither gates what triggered it, `Reclaim` is best-effort and a failure emits an event and proceeds. Nothing periodic runs in between, so a VM that stops taking backups keeps its orphans until it takes another or reboots.
+
+Reconciling a set rather than replaying individual deletions means ordering carries no correctness weight. A crash between trimming `status.checkpoints` and deleting the bitmap leaves an orphan, which is by definition outside the next desired set. A deleted `VirtualMachineBackupTracker` needs no cleanup RPC for the same reason, its checkpoints simply leave the union.
+
+`forceFullBackup` on a `VirtualMachineBackupTracker` source reduces `status.checkpoints`, on completion, to just the checkpoint that backup created, and the retained bases are then collected like anything else outside the set. It serves as an escape hatch for a chain a vendor no longer trusts, so it discards the retained bases rather than appending to them. A forced backup that fails before creating its checkpoint leaves `status.checkpoints` unchanged, so the chain it was meant to replace survives and the vendor can retry.
```
hao-zhou-rubrik, line 144: Regarding "It serves as an escape hatch for a chain a vendor no longer trusts".
Not necessarily, we may want to force a full when we don't trust the last checkpoint or the last N checkpoints. Can we skill keep the chain, or have a parameter to keep them?

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: 1. So `historyLimit` is bounded by the backend storage PVC, which is sized by the flat constant you mentioned. Bitmaps live in memory while the VM runs, so exceeding that budget has no effect on a running VM or on backups (they'll keep succeeding so long as the virt-launcher isn't memory-limited), but the next time QEMU commits bitmap state to disk (which happens during migration switchover and VM shutdown), commit failure would A. abort the migration in the case of a migration and B. fail silently and every bitmap in the affected overlays will be marked inconsistent on shutdown. When bitmaps are marked as inconsistent, they'll be rendered unusable, meaning the following backup will fallback to full on those disks and the chain will be effectively reset. It's a known limitation and we will be providing a way to control the size of the backend state PVC as part of a separate VEP. 
2. Should be negligible besides the obvious memory overhead incurred by holding bitmaps while the VM is online.
3. Full disclosure, unbounded depth isn't something we can reach by raising a limit right now. KubeVirt's change tracking is built on top of QEMU dirty bitmaps, and each retained point costs a full bitmap per disk, so roughly `disk_size / granularity / 8` in both memory and backend storage. The cost is linear in how far back you can query, which is why a limit exists at all. vSphere CBT has per-block sequence numbers so querying against a point a hundred backups back costs the same as querying the most recent one, it's an encoding difference, and there's no value we can set that can close it.

cc @kevmw - there's real demand here, I wonder if we could examine a different CBT strategy in QEMU for the future.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: 1. Yes, as explained, you may delete the VMBackup CR to signal cancel/completion and attempt from the same checkpoint again.
2. We haven't added a reason, only the type of the disk, would you be interested in having a reason for why a disk is full?
3. The new disk will be full, as per question 2's answer, there is currently no indicator as to why it's full.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
Acedus: So right now it does clear the chain, but I think we can change the design so it doesn't clear the chain and the way to clear it is just via `historyLimit: 1`.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
hao-zhou-rubrik: Yes, it would be great if we could have a reason for why a disk is full.

veps/sig-storage/25-incremental-backup/399-multi-checkpoint/vep.md
hao-zhou-rubrik: > I think we can change the design so it doesn't clear the chain

That's great, we as a vendor will choose to clean up or not based on our use case.

Issue #401, opened 2026-07-20
VEP 401: Offline Incremental Backup


alromeros:
**Primary contact (assignee)**:

/assign @alromeros 

**Current Feature Stage**: `

**Feature Gate**: `OfflineIncrementalBackup`

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y): v1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #402, opened 2026-07-20
VEP 401: Offline incremental backup

alromeros:
### VEP Metadata

**Tracking issue**: #401 
**SIG label**: /sig storage

### What this PR does

This proposal extends KubeVirt's Changed Block Tracking backup capabilities to support stopped Virtual Machines, using persisted QCOW2 dirty bitmaps exposed via `qemu-nbd`.


Acedus (commented):
From a quick pass I like the general direction, mainly due to its implied reuse of the backup source export logic, but there's quite a bit of missing information that's crucial to the implementation:
1. Assuming online backup A followed by VM shutdown and offline backup B, the offline backup would be incremental and would therefore need to adhere to the same semantics as libvirt, that is to say, verify that all bitmaps (from which the checkpoints would be redefined) are valid and consistent, and create new bitmaps that will comprise backup B's checkpoint (which is also important for maintaining the corresponding backup chain in track). Furthermore, as opposed to libvirt's use of QMP where bitmap addition is an atomic single operation for multiple disks, something like `qemu-img bitmap add` would need to be done per disk, how would we handle crash consistency? The entirety of this proposition also conflicts with what the VEP currently states (i.e., read-only mount of the relevant PVCs).
2. How would pruning work (both in the context of garbage collection and checkpoint retention in the case of https://github.com/kubevirt/enhancements/pull/400)? Do we leave cleanup for the checkpoint redefinition period when the VM is started after it underwent offline backups? Do we let the virt-exportserver perform the cleanup?

alromeros:
> From a quick pass I like the general direction, mainly due to its implied reuse of the backup source export logic, but there's quite a bit of missing information that's crucial to the implementation:
> 
>     1. Assuming online backup A followed by VM shutdown and offline backup B, the offline backup would be incremental and would therefore need to adhere to the same semantics as libvirt, that is to say, verify that all bitmaps (from which the checkpoints would be redefined) are valid and consistent, and create new bitmaps that will comprise backup B's checkpoint (which is also important for maintaining the corresponding backup chain in track). Furthermore, as opposed to libvirt's use of QMP where bitmap addition is an atomic single operation for multiple disks, something like `qemu-img bitmap add` would need to be done per disk, how would we handle crash consistency? The entirety of this proposition also conflicts with what the VEP currently states (i.e., read-only mount of the relevant PVCs).
> 
>     2. How would pruning work (both in the context of garbage collection and checkpoint retention in the case of #400)? Do we leave cleanup for the checkpoint redefinition period when the VM is started after it underwent offline backups? Do we let the virt-exportserver perform the cleanup?

Good catch on the read-only conflict and the atomicity issue. I took the hint from your last question and moved all the bitmap management to the VM startup phase. I've updated the VEP with this approach. Now the export pod stays strictly read-only and only validates the existing bitmaps. Once the VM starts, the VMI controller handles the QMP transaction right before domain creation resumes.

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+Once validated, the backup controller creates a VirtualMachineExport. The export controller spins up a pod that mounts the state PVC and all data volumes read-only, discovers the QCOW2 overlay chain, validates existing bitmaps, starts `qemu-nbd` per disk, and serves the dirty extent map and raw data over HTTP using the existing Go NBD client. While the backup is in progress, the VMI controller blocks VM start.
+
+On completion, the intent to create the new checkpoint bitmap and remove the old one is recorded in the backup CR, and the backup transitions to `Completed` immediately, unblocking the VM. The backup controller deletes the export pod as soon as the backup reaches a terminal state (`Succeeded` or `Failed`) to release PVC mounts promptly. When the VM eventually starts, the VMI controller materializes these pending bitmaps atomically via libvirt QMP before the guest resumes writing.
```
Acedus: How is
> the intent to create the new checkpoint bitmap and remove the old one is recorded in the backup CR

recorded exactly? The way to signal backup completion is to delete the backup CR so persisting intent on it wouldn't be possible.

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Overview
+
+When an offline backup is requested, the backup controller first verifies the VM is stopped and that no stale virt-launcher pods remain. Staleness is determined by listing pods with the `kubevirt.io/domain` label matching the source VM and checking for any pod not in a terminal phase (`Succeeded` or `Failed`); if such a pod exists, the backup is failed immediately. If another backup is already in progress for this VM, the controller also fails the new backup immediately with a descriptive reason, preventing indefinite hangs.
+
+Once validated, the backup controller creates a VirtualMachineExport. The export controller spins up a pod that mounts the state PVC and all data volumes read-only, discovers the QCOW2 overlay chain, validates existing bitmaps, starts `qemu-nbd` per disk, and serves the dirty extent map and raw data over HTTP using the existing Go NBD client. While the backup is in progress, the VMI controller blocks VM start.
```
awels: How are you planning on blocking the VM from starting? We have similar issues with guestfs for instance, and I don't think we every properly resolved how to block the VM from starting.

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+Checkpoint retention policy, such as maintaining multiple checkpoints for branching backup chains, is outside the scope of this VEP and is addressed by VEP 25.1.
+
+### Known limitations
+
+Only one offline incremental backup can be performed between VM starts. Since the VM is stopped, no guest writes occur after an offline backup completes. A second offline incremental backup would produce an empty delta and serve no purpose. The backup controller rejects such requests while a pending intent exists. This also avoids the technical constraint that the deferred bitmap has not been materialized on disk yet, so there is no on-disk bitmap for a subsequent backup to read dirty extents from. If a full (non-incremental) backup is needed while the VM remains stopped, it can proceed independently of pending intents.
```
awels: would it make sense to keep a 'last known working' intent around. Basically if someone attempts to do an incremental backup after it succeeded. Right now the second attempt would fail since we are now marking the intent on something that doesn't exist.

But if the backup failed in the client side, we could use the last known working to do the backup again from the working intent.

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Storage access modes
+
+State PVCs may use either RWO or RWX access modes, and the backup design must handle both. On RWO storage, Kubernetes enforces single-node attachment and QEMU's default file locking provides a secondary defense; note that the export pod must schedule on the same node where the RWO PVC was last attached, which means it will be unschedulable if that node is unavailable. On RWX storage, QEMU file locking is unreliable on NFS and distributed filesystems, and KubeVirt does not explicitly configure it. Since the export pod mounts all volumes read-only, a race that allows a virt-launcher to start concurrently shouldn't result in QCOW2 corruption from the export pod, the worst case is a read of stale bitmap data. The bidirectional controller checks (VMI controller blocks start during backup, backup controller fails the backup if a VMI appears) mitigate this window. For Beta, we can look into using an admission webhook to block the VM start reque
```
awels: RWO does not mean the export pod must be scheduled on the same node where the RWO PVC was last attached. Once the RWO volume is detached, it can be scheduled on any node that allows it to be attached.

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+st entirely.
```
awels: stray newline?

veps/sig-storage/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+Offline mode is detected automatically, if no VMI exists for the backup's source VM (or the VMI is in a terminal phase), the backup proceeds in offline mode.
+
+### VM start prevention & Deletion handling
+
+Before creating a virt-launcher pod, the VMI controller performs a lookup of the VMBackup/VMBackupTracker CRs in the namespace. If any CR references the same VM, has offline status and is in progress, the controller sets a `BackupInProgress` condition on the VMI and returns without creating the pod. The VMI stays `Pending` until the backup CR leaves `InProgress`. This follows the same pattern used for migration gating via `ActiveMigrationExistsForVMI()`.
+
+If during a short window of time a VMI appears for the source VM while the offline backup is `InProgress`, the backup controller yields priority to the VM start, it deletes the export pod and marks the backup `Failed` with reason `VMStarted`.
```
awels: Deleting the VMExport pod can take a little bit of time, during that window, the VM would be forced onto the node the VMExport is running on. Its a short window I think but it is something to think about.

awels (commented):
Biggest open question for me is the raciness between the VMExport and a VM starting, or how to prevent the VM from starting. There are other operations with similar issues (guest fs for instance) and I am not sure if we ever properly solved that issue.

veps/sig-storage/offline-incremental-backup.md
alromeros: In my current POC I'm assuming that the cr persists after completion and that a intent field in .status survives until the VM starts and processes it. We may need a better alternative in case it gets deleted earlier and record the intent in the VM. I have to think a bit more about the exact approach and be more specific here in the vep. Thanks for the comment!

veps/sig-storage/offline-incremental-backup.md
alromeros: That's a good point. We can consider implementing some kind of fallback to a 'last known working intent' to allow offline retries, but I think it significantly increases the complexity of the alpha implementation. Maybe it's acceptable for now to document this constraint (consecutive offline incrementals fail until the VM boots) and say the fallback for the client-side failure is to take a full backup? We could leave the retry logic in a future iteration.

veps/sig-storage/offline-incremental-backup.md
alromeros: yep you are right, good catch! I'll rewrite this section.

veps/sig-storage/offline-incremental-backup.md
alromeros: You are right, but I'm not sure if this is something we should actively avoid or just document as a possible limitation. I'll add it to the vep, thanks!

veps/sig-storage/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,211 @@
...
+
+### VM finalizer to prevent deletion
+
+Add a finalizer to the VM object to block deletion during backup. We rejected this because VirtualMachineExport and online VirtualMachineBackup both handle source deletion via graceful abort. A finalizer risks permanently stuck VMs if the backup controller becomes unavailable.
+
+## Update/Rollback Compatibility
+
+On upgrade, offline backup becomes available immediately behind a feature gate. On rollback, in-progress offline backups will fail, but no persistent state (Leases, VM finalizers) is left behind that could block VM operations. Checkpoints created by online CBT are compatible with offline CBT because dirty bitmaps are QCOW2 format-level metadata, not a libvirt artifact, so `qemu-nbd` reads them directly from the image.
+
+## Functional Testing Approach
```
ema-aka-young, line 185: The VEP defines some edge cases that are not reflected in this section. Some examples are: concurrent backups, stale virt-launcher (L60), mid-backup VMI fails with VMStarted (L72), corrupted bitmap fallback (L99). There may be others I haven't spotted.

veps/sig-storage/offline-incremental-backup.md
alromeros: ack following your suggestion in https://github.com/kubevirt/enhancements/pull/400

awels (approved):

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Overview
+
+When an offline backup is requested, the backup controller verifies the VM is stopped and that no stale virt-launcher pods remain. A pod is stale if it carries the `kubevirt.io/domain` label matching the source VM and is not in a terminal phase (`Succeeded` or `Failed`). If a stale pod exists, or another backup is already in progress for the same VM, the backup is failed immediately.
+
+Once validated, the backup controller creates a VirtualMachineExport. The export controller starts a pod that mounts the state PVC and all data volumes read-write, discovers the QCOW2 overlay chain, validates existing bitmaps, creates the new checkpoint bitmap on each disk, starts `qemu-nbd` per disk, and serves dirty extents and raw data over HTTP.
```
Acedus: What about scenarios where the VM includes a disk that was never included in a backup before?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+Bitmap creation on disk and tracker advancement are independent operations that occur at different points in the lifecycle. The full sequence is described in "Checkpoint and bitmap management." The backup controller deletes the export pod as soon as the backup reaches a terminal state to release PVC mounts.
+
+Offline mode is detected automatically: if no VMI exists for the backup's source VM, the backup proceeds in offline mode.
+
+### VM start prevention & Deletion handling
+
+Before creating a virt-launcher pod, the VMI controller looks up VMBackup/VMBackupTracker CRs in the namespace. If any CR references the same VM, indicates offline mode, and has the `Progressing` condition set, the controller sets a `BackupInProgress` condition on the VMI and returns without creating the pod. The VMI remains `Pending` until the backup CR is no longer progressing. This follows the same pattern used for migration gating.
+
+If a VMI appears for the source VM while the offline backup is progressing, the behavior depends on the backup mode:
+
```
Acedus, line 72: I'm not sure we want a split behavior here, backup mode shouldn't influence the failure mode. I'd rather either both fail or the VMI in question is delayed in startup.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+- **Push mode**: the backup controller fails the backup with reason `VMStarted`, deletes the export pod, and does not advance the tracker. A partial QCOW2 write on the target PVC is not a usable backup file, so completing would advance the tracker past data that was never fully written.
+
+If the source VM is deleted during the backup, the export pod is deleted and the backup is marked `Failed` with reason `VMDeleted`.
+
+### Backup deadline
+
+The backup controller uses `spec.ttlDuration` as the maximum allowed duration for an offline backup:
+
+- **Push mode**: TTL expiry before the write completes is a failure. The controller deletes the export pod and sets `Failed` with reason `DeadlineExceeded`. The tracker is not advanced.
+- **Pull mode**: TTL expiry is a completion event, consistent with online pull mode where TTL-triggered abort is treated as success. The controller advances the tracker, deletes the export pod, and sets `Complete`. Incomplete pulls are the user's responsibility (see "Known limitations").
```
Acedus: Truthfully I feel like this is a design oversight rather than an intended behavior... We should probably make pull mode fail if the TTL is expired (it would indicate that only partial progress was made and therefore we should not consider the backup and its associated checkpoint valid for subsequent ones).

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Checkpoint and bitmap management
+
+Offline incremental backups follow the same checkpoint semantics as the online (libvirt) path.
+
+On startup, the export pod validates the bitmaps referenced by the `VirtualMachineBackupTracker` on each disk overlay using `qemu-img info --output=json`. If any bitmap has the `in-use` flag set or is marked inconsistent (e.g., from a prior VM crash), the backup falls back to full mode for that disk and the tracker is updated accordingly.
+
+The bitmap lifecycle is split between the export pod and the next VM start:
+
+1. **Create new checkpoint bitmap (startup):** After validation, the export pod runs `qemu-img bitmap --add` on each overlay. This applies to every backup type (Full and Incremental), because each backup must create the tracking anchor for the next one. The bitmap is created before `qemu-nbd` starts, since `qemu-nbd --read-only` locks the overlay against modification. The new bitmap starts empty and does not advance the checkpoint; that is the tracker's role.
```
Acedus: I think the terminology is a bit inconsistent here. Checkpoints are a libvirt-specific abstraction, bitmaps are completely separate; it just so happens that libvirt builds on top qemu bitmaps to facilitate backup checkpoints.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+2. **Serve data:** `qemu-nbd --read-only` starts per disk, serving the previous checkpoint's dirty extents and data over HTTP.
+3. **Advance tracker (completion):** The backup controller updates the `VirtualMachineBackupTracker` only when the backup completes (see "Pull mode completion semantics" and "Backup deadline"). If the backup fails (VM deletion, pod crash), the tracker is not advanced and the chain is preserved for retry.
+4. **Remove old bitmap (deferred to VM start):** Verification of the new bitmap and removal of the old checkpoint bitmap are handled during checkpoint redefinition when the VM next starts.
+
+Since `qemu-img bitmap` operates per-disk with no multi-disk transaction support, a pod crash during bitmap creation can leave partial state (bitmap added on some disks but not others). The next backup's startup validation detects this by checking each disk's bitmaps against the tracker and falls back to full mode for any inconsistent disk.
```
Acedus: Pod crash as in the virt-exportserver? Wouldn't it start again since the VMExport is still up at which point we can simply validate whether A. we've already created a bitmap that matches in name to the ongoing backup's checkpoint name and B. if it does have a bitmap but it's inconsistent we can remove it and try creating it again?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+The bitmap management phase runs while the VM is stopped and after the backup controller has verified no virt-launcher pods exist.
+
+### Pruning and cleanup
+
+During checkpoint redefinition at VM start, only the specific bitmap being replaced is removed; other bitmaps retained by the tracker (e.g., for multi-checkpoint retention per VEP 25.1) are left untouched.
```
Acedus: > During checkpoint redefinition at VM start, only the specific bitmap being replaced is removed
Not sure I understand, which bitmap would be replaced?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+Orphaned bitmaps from prior crashes or interrupted operations are also detected during redefinition. A bitmap is orphaned only if no entry in the tracker's checkpoint list references it. Bitmaps belonging to older retained checkpoints are not removed.
+
+Checkpoint retention policy is outside the scope of this VEP and is addressed by VEP 25.1.
+
+### Boot-time checkpoint ordering
+
+When the VM starts after offline backups, the VMI controller processes checkpoints in order:
+
+1. **Redefine checkpoints** across all associated `VirtualMachineBackupTracker` CRs. Redefinition recreates bitmap tracking in the running VM's backend. Offline checkpoints (QCOW2 bitmaps written directly to disk by the export pod) have no backend equivalent to redefine. Redefinition failure for offline checkpoints must not invalidate the checkpoint. Offline checkpoints are identified by a distinguishing marker in the checkpoint name; the redefinition process skips them and clears only the "redefinition needed" flag.
```
Acedus: Checkpoints are simply libvirt metadata used to show which disks were "backed up" at a certain point in time associated with a specific domain (as well as their individual metadata, e.g., backup type per disk). I don't really see a reason to treat online checkpoints differently from offline since ultimately we control what goes into a checkpoint and what doesn't, it's simply a convenience.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+2. **Checkpoint pruning** removes bitmaps exceeding `retainCheckpoints`, including any accumulated while the VM was stopped.
```
Acedus, line 118: While true I feel like this is bleeding in scope exposure to VEP 25.1.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Known limitations
+
+Only one offline incremental backup can be performed between VM starts. No guest writes occur while the VM is stopped, so a second offline incremental would produce an empty delta. The backup controller rejects such requests when the latest checkpoint bitmap on disk is empty. A full (non-incremental) backup while the VM remains stopped proceeds normally.
```
Acedus: I think it would be better to distinguish the type at the disk level rather than the backup as a whole. Disks may vary in whether the next backup will have them as full or incremental due to a plethora of reasons (e.g., disk hotplug).

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+
+**Pull mode trust boundary.** Any event that ends the backup window (TTL expiry, CR deletion, or VMI appearance) advances the tracker regardless of whether the user pulled all data. Since the VM is stopped, this does not compromise on-disk data integrity, but an incomplete pull creates a gap in the user's external backup chain that KubeVirt does not detect. Accidental CR deletion has the same effect, even if no data was pulled.
```
Acedus: Which is why true failure (i.e., TTL expiry) should constitute as such instead of a successful operation.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,259 @@
...
+- Feature gate enabled by default.
+- No data corruption incidents reported during Alpha.
+- E2E tests passing in CI for at least one release cycle.
+
+### GA (TBD)
+
+- Feature gate removal.
+- Stable in Beta for at least one release cycle.
+
+## References
```
Acedus, line 240: Probably worth including VEP #25.1 as well seeing as it's referenced in the doc.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: Makes sense. I wanted to prioritize parity with online mode here, should this also be addressed in the online pull flow? Maybe it's worth adding a `completion semantics` or something similar in your pull mode improvements PR.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: The issue here is that completion semantics for pull mode are a bit brittle, and if we remove ttl expiry and VM startup as completion signals we'd only have vmbackup deletion left as the completion signal. That would make me prefer delaying startup for push mode so we wait for completion and leave more completion signals available for pull mode. wdyt?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: Okay you are right this might add unnecessary complexity. My assumption was that redefinition wouldn't work for bitmaps libvirt didn't create as they lacked some checkpoint metadata. I'll look into removing this.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: okay that's a good point, falling back to full can be avoided in most cases.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: okay I'll update the wording to clarify that full vs incremental is a per-disk decision.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: gotcha I'll clean up the checkpoint vs bitmap terminology

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: You're right, I mixed redefinition and replacement up. I'll clean this up.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,254 @@
...
+
+The current CBT implementation (VEP #25) supports incremental backups only for running VMs. When a VM stops, the VMI is deleted, the libvirt domain no longer exists, and the only backup option is a full copy.
+
+oVirt solves this using direct `qemu-nbd --bitmap` manipulation, but its design assumes bare-metal with root access. This VEP adapts that approach to run in non-privileged pods.
+
+## Goals
+
+- Enable incremental backups for stopped VMs using persisted QCOW2 dirty bitmaps.
+- Prevent data corruption from VM start/delete/migration during backup.
+- The export pod runs using the same non-privileged SecurityContext as existing VirtualMachineExport pods.
```
Acedus: I don't really think this qualifies as a goal resulting from the motivation?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
Acedus: I don't think it's brittle, it's just a set of conditions for "completion". If the VM has a backup associated with it that isn't terminal, then its startup should be gated, am I missing something?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,254 @@
...
+
+A VMI appearing for the source VM while the offline backup is progressing triggers the start prevention gate, delaying virt-launcher creation until the backup reaches a terminal state. In pull mode, the backup controller treats VMI appearance as a completion trigger: it advances the tracker, deletes the export pod, and marks the backup `Done`. In push mode, the backup continues until the export pod finishes writing; the gate clears once the backup completes normally via pod phase.
+
+Deletion of the source VM during the backup causes the export pod to be deleted and the backup to be marked `Failed` with reason `VMDeleted`.
+
+### Backup deadline
+
+As with online backup, `spec.ttlDuration` applies only to pull mode. TTL expiry deletes the export pod and sets `Failed` with reason `DeadlineExceeded`. The tracker is not advanced.
```
Acedus: Not for long. I think that TTL expiry semantics should apply to push mode the same way they do to pull mode, I'll ensure that's the case for online as well. For alpha, I think we can omit explicit backup TTL semantics for offline and they _should_ implicitly mimic the online path's.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,254 @@
...
+
+  The output format, file naming, and directory layout match online push mode (see VEP #25), so existing restore tooling works unchanged.
+
+  The backup completes when all disks have been written and the export pod exits with phase `Succeeded`. The backup controller detects completion via the pod's phase, not via VMI status (there is no VMI). If the pod fails or is killed, the backup is marked `Failed` and the tracker is not advanced. See "Completion semantics" for the full distinction between push and pull.
+
+### Completion semantics
+
+Offline backup has two distinct completion models depending on the mode.
+
+**Pull mode** has no explicit vendor completion signal. Any event that ends the backup window advances the tracker and marks the backup `Done`:
```
Acedus: This contradicts the statement that TTL expiration will render the backup as failed.

Acedus (commented):
Looking good, few rather minor comments but aside from them I think we're good to go.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: ouch yep that's a remnant from the previous iteration, I'll change it

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: Agree, ack

Acedus (approved):
Thanks @alromeros!

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: so pull mode has no deterministic completion signal. Once the backup is created it stays progressing, and nothing tells us when the external consumer has actually finished pulling. So if we gate VM start on the backup being terminal, startup would stay blocked even after the vendor already has all the data, waiting on a state that may never resolve on its own (just ttl expiry, but we decided to consider that a failure). If we make CR deletion the authoritative "pull complete" notification, then gating on terminal is somewhat fine. Until then, I don't think VMI startup should be blocked in pull mode.

Acedus: I meant that we should gate startup according to two conditions:
1. Is there even a backup to begin with? (this handles both modes);
2. If there is a backup, is it terminal? (this handles push mode).

CR deletion has always been the authoritative "pull complete" signal as far as we (and libvirt for matter) are concerned. Having said that, we're always level-triggered (we don't gate pending notification, we simply compare the active state of the world (whether there's a backup and whether it's terminating) against the desired (VMI start).

awels (approved):

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,245 @@
...
+
+The absence of a VMI for the source VM triggers offline mode automatically.
+
+### VM start prevention & Deletion handling
+
+Before creating a virt-launcher pod, the VMI controller looks up VMBackup/VMBackupTracker CRs in the namespace. When a CR references the same VM, indicates offline mode, and has the `Progressing` condition set, the controller sets a `BackupInProgress` condition on the VMI and returns without creating the pod. The VMI remains `Pending` until the backup CR is no longer progressing. This is the same gating mechanism used for migration.
+
+VMI creation during offline backup is progressing is gated the same way in both modes. Virt-launcher creation is delayed until the backup reaches a terminal state. In push mode the gate clears once the export pod finishes writing. In pull mode it clears once the vendor deletes the backup CR.
+
+Deletion of the source VM during the backup causes the export pod to be deleted and the backup to be marked `Failed` with reason `VMDeleted`.
```
iholder101, line 73: So IIUC, whenever a backup is ongoing:
* During VM start - the creation will be delayed.
* During VM deletion - the deletion will **not** be blocked, but the backup will be cancelled.

Isn't this inconsistent? Should we cancel the update during VM start in order to avoid blocking it, in the same way that we handle VM stops?

In addition what about stopping (not deleting) a VM while a backup is ongoing?

veps/sig-storage/401-offline-incremental-backup/offline-incremental-backup.md
iholder101: @ema-aka-young is this one addressed?

veps/sig-storage/offline-incremental-backup.md
iholder101: @awels Are you ok with deferring the retry logic and just documenting the constraint for alpha? Should solving this be a requirement for beta?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,245 @@
...
+
+1. **Redefine checkpoints** across all associated `VirtualMachineBackupTracker` CRs. Redefinition recreates bitmap tracking in the running VM's backend. Offline checkpoints are redefined the same way as online checkpoints; the bitmap already exists on disk and `--redefine` adopts it.
+2. **Checkpoint pruning** removes bitmaps exceeding `retainCheckpoints`, including any accumulated while the VM was stopped.
+
+### Known limitations
+
+Only one offline incremental backup can be performed between VM starts. No guest writes occur while the VM is stopped, so a second offline incremental would produce an empty delta. The export pod checks the latest checkpoint bitmap on startup and reports an empty bitmap via status; the backup controller fails the request based on that status. A full (non-incremental) backup while the VM remains stopped proceeds normally.
+
+**Pull mode trust boundary.** CR deletion advances the tracker regardless of whether the user pulled all data. On-disk data remains intact (the VM is stopped and nothing has changed on the volume), but an incomplete pull creates a gap in the user's external backup chain that KubeVirt does not detect.
```
iholder101, line 124: Could we close this instead of documenting it? Track which ranges are fetched via `/data`, advance the tracker only if coverage is complete, else fail. Similar to the excluded "explicit completion signal" alternative - but this is data-integrity, not just UX.

WDYT?

iholder101:
@awels was [this comment](https://github.com/kubevirt/enhancements/pull/402#pullrequestreview-4822822087) addressed?

> Biggest open question for me is the raciness between the VMExport and a VM starting, or how to prevent the VM from starting. There are other operations with similar issues (guest fs for instance) and I am not sure if we ever properly solved that issue.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: Good question. That asymmetry was left on purpose since blocking a start is drastically different from cancelling a deletion. There are completion signals for both push and pull mode that easily allow delaying the VM start. We just need to prevent virt-launcher pod creation until the signal (backup CR deletion or VMI appearance for pull, backup completion for push). It's a cheap and easily manageable operation. Blocking deletion would require a finalizer which would prevent deletion indefinitely if there are issues with the export or backup controllers. Also, the most important thing during this design was consistency with online (regular) backup. Online backup doesn't block deletion, so it doesn't make sense to block deletion just for offline backup and it's something that should in any case be included in VEP 25.

> In addition what about stopping (not deleting) a VM while a backup is ongoing?

The VEP is about offline backup, so the VM is already stopped. The opposite case should be handled by online backup. It's something worth considering, but IMO not something that should be addressed in this VEP.

veps/sig-storage/offline-incremental-backup.md
alromeros: This was addressed, the comment is referring to an earlier version. The export pod now creates the bitmap directly on startup so there is no retry logic left to defer.

alromeros:
> @awels was [this comment](https://github.com/kubevirt/enhancements/pull/402#pullrequestreview-4822822087) addressed?
> 
> > Biggest open question for me is the raciness between the VMExport and a VM starting, or how to prevent the VM from starting. There are other operations with similar issues (guest fs for instance) and I am not sure if we ever properly solved that issue.

@iholder101 this one was addressed in the `VM start prevention & Deletion handling` section.

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: Yep that's a good point. So the export server can track which ranges the vendor actually fetched and gate completion on full coverage so CR deletion advances the tracker only if coverage is complete. If it's incomplete the backup fails and the tracker stays the same.

alromeros: Okay I'll update the vep to address this

veps/sig-storage/401-offline-incremental-backup/offline-incremental-backup.md
ema-aka-young: Yes, it is. Thanks @alromeros

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
alromeros: ack

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
iholder101: > There are completion signals for both push and pull mode that easily allow delaying the VM start. 

Still I wonder if that's the right action in this case.
Wouldn't it be better to unblock the VM start but cancel the backup?

iholder101:
Left a small comment above, but looks good to me overall.
Thanks a lot!

/approve

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,243 @@
...
+
+  The output format, file naming, and directory layout match online push mode (see VEP #25), so existing restore tooling works unchanged.
+
+  The backup completes when all disks have been written and the export pod exits with phase `Succeeded`. The backup controller detects completion via the pod's phase, not via VMI status (there is no VMI). If the pod fails or is killed, the backup is marked `Failed` and the tracker is not advanced. See "Completion semantics" for the full distinction between push and pull.
+
+### Completion semantics
+
+Offline backup has two distinct completion models depending on the mode.
+
+**Pull mode** completes on CR deletion, but only once the export server has served the full dirty-extent set for every disk. The server tracks fetched ranges against the bitmap and advances the tracker only on full coverage. An incomplete pull or TTL expiry fails the backup and does not advance the tracker, so the same extents are offered again on the next run.
```
Acedus: I don't think we should involve ourselves with what the consumer does with what we expose. IMO it should simply be CR deletion and nothing more.

alromeros: @iholder101 @Acedus we have opposing views here. I can see how advancing the tracker towards full coverage might be a sensible choice and something we may end up adopting, but to be fair, my intent with this VEP is just to design the offline CBT feature and match the online behavior wherever possible.

Since the online behavior is agnostic to what the consumer does, I'm inclined to drop my latest commit and leave these possible improvements as a future enhancement. wdyt?

awels: I agree with @Acedus here. If the user deletes the VMBackup resource, it means we are done, regardless of if they actually pulled the data.

awels:
I am good with giving it an lgtm, once you remove that last commit, the rest is nitpicking.

alromeros:
Removed the latest commit, holding in case @iholder101 wants to step in 
/hold

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
iholder101: Sure, I don't feel too strongly about this. Let's go with @Acedus approach.

iholder101:
> Removed the latest commit, holding in case @iholder101 wants to step in /hold

You can remove the hold from my POV.
@Acedus are you happy?

veps/sig-storage/25-incremental-backup/401-offline-incremental-backup/offline-incremental-backup.md
```diff
@@ -0,0 +1,245 @@
...
+### Boot-time checkpoint ordering
+
+When the VM starts after offline backups, the VMI controller processes checkpoints in order:
+
+1. **Redefine checkpoints** across all associated `VirtualMachineBackupTracker` CRs. Redefinition recreates bitmap tracking in the running VM's backend. Offline checkpoints are redefined the same way as online checkpoints; the bitmap already exists on disk and `--redefine` adopts it.
+2. **Checkpoint pruning** removes bitmaps exceeding `retainCheckpoints`, including any accumulated while the VM was stopped.
+
+### Known limitations
+
+Only one offline incremental backup can be performed between VM starts. No guest writes occur while the VM is stopped, so a second offline incremental would produce an empty delta. The export pod checks the latest checkpoint bitmap on startup and reports an empty bitmap via status; the backup controller fails the request based on that status. A full (non-incremental) backup while the VM remains stopped proceeds normally.
```
hao-zhou-rubrik, line 122: Is there's a named reason for this failure? For policy-based periodic backups, it would be helpful if we can get a named reason, or a reliable precheck.

Issue #403, opened 2026-07-21
VEP 403 : Add a VirtualMachineInstance NMI injection subresource


kmatsuo3225:
**Primary contact (assignee)**: 

/assign @kmatsuo3225

**Current Feature Stage**:

**Feature Gate**:

None

**Responsible SIGs**: 

Primary SIG:
/sig compute

Additional SIGs (optional):

**Enhancement link**:

#404

**Timeline**:

* Beta release target (x.y): 1.10
* Stable release target (x.y): 1.11

* [ ] Beta
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:
* [ ] GA
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #404, opened 2026-07-21
VEP 403 Add a VirtualMachineInstance NMI injection subresource

kmatsuo3225:
### VEP Metadata

**Tracking issue**: #403 
**SIG label**: /sig compute

### What this PR does

This proposal adds a new runtime operation for VirtualMachineInstance (VMI) to inject a Non-Maskable Interrupt (NMI) into a running guest.
It is exposed through a new virtctl inject-nmi command and a corresponding VMI subresource.

### DIFF
--- a/veps/sig-compute/403-inject-nmi/vep.md
+++ b/veps/sig-compute/403-inject-nmi/vep.md
@@ -1,0 +1,182 @@
+# VEP 403: Add a VirtualMachineInstance NMI injection subresource
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets Alpha for version: v1.10
+- This VEP targets beta for version: v1.11
+- This VEP targets GA for version: v1.12
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+To support kernel and operating system hang analysis, this proposal introduces a Non-Maskable Interrupt (NMI) injection feature for VirtualMachineInstance. When a guest becomes unresponsive, users can inject an NMI to trigger the guest OS's crash dump mechanism (for example, Linux kdump), preserving diagnostic information for post-mortem analysis before performing reset or power-off operations.
+
+The operation is exposed as a new VMI subresource under the KubeVirt subresource API. The `virtctl inject-nmi` command is provided as a convenient client to call this API.
+
+The primary goal of this feature is only to provide a way to inject a NMI.
+It does not guarantee that a crash dump is produced. Dump generation, storage location, reboot behavior, and post-processing are controlled by the guest OS configuration, such as Linux kdump/NMI settings or Windows crash dump settings.
+If a guest OS is properly configured, it may generate a crash dump in response to the injected NMI.
+
+The design follows the same architectural pattern currently used by KubeVirt runtime subresources such as `pause`, `unpause`, `freeze`, `unfreeze`, `reset`, and `softreboot`
+
+## Motivation
+
+KubeVirt already provides a VM memory dump feature, but the earlier KubeVirt design proposal for this feature explicitly frames the solution around later inspection with Volatility3. [\[github.com\]](https://github.com/kubevirt/community/blob/main/design-proposals/vm-memory-dump.md)
+That workflow is useful for generic memory inspection and forensic analysis, but it does not address the main operational requirement for Linux kernel hang analysis. In practice, kernel hang analysis is typically performed on kdump-generated `vmcore` files using kernel-oriented tooling such as `crash` together with matching kernel debuginfo. Red Hat documentation for `crash` explains that it operates on kdump-generated dump files, and Linux kdump documentation explicitly lists NMI as one of the events that can trigger crash dump capture. [\[docs.redhat.com\]](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/analyzing-a-core-dump_assembly_managing-kernel-command-line-parameters-with-uki#running-and-exiting-the-crash-utility_analyzing-a-core-dump), [\[kernel.org\]](https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html)
+
+For mission-critical workloads, the inability to trigger a guest crash dump before resorting to reset or power-off is a significant operational gap. Red Hat’s kdump guide explicitly emphasizes that crash dump data is especially important in business-critical environments. Operational guidance from multiple vendors similarly recommends sending an NMI to a hung VM in order to force dump generation and preserve state for later root-cause analysis before taking more destructive recovery actions. This need is not limited to Linux: Microsoft’s guidance likewise underscores the importance of collecting crash dumps in Windows environments. [\[docs.redhat.com\]](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/installing-kdump_assembly_managing-kernel-command-line-parameters-with-uki#what-is-kdumpinstalling-kdump), [\[learn.microsoft.com\]](https://learn.microsoft.com/en-us/troubleshoot/windows-client/performance/generate-a-kernel-or-complete-crash-dump), [\[community.ibm.com\]](https://community.ibm.com/community/user/blogs/sachin-bappalige/2025/07/02/vmcore-crash-analysis-in-enterprise-power-system)
+
+KubeVirt currently has a gap here compared to mainstream virtualization platforms. Hyper-V exposes `Debug-VM -InjectNonMaskableInterrupt`; VMware vSphere/ESXi exposes `Send_NMI_To_Guest` and `vmdumper ... nmi`; libvirt exposes `virsh inject-nmi`. Users migrating VM operational practices into KubeVirt reasonably expect an equivalent platform-native function instead of requiring direct access to launcher internals or host-side tools. [\[learn.microsoft.com\]](https://learn.microsoft.com/en-us/powershell/module/hyper-v/debug-vm?view=windowsserver2025-ps), [\[knowledge.broadcom.com\]](https://knowledge.broadcom.com/external/article/301246/how-to-send-nmi-to-guest-os-on-esxi-6x-o.html), [\[docs.redhat.com\]](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-generic_commands-injecting_nmi)
+
+## Goals
+
+Provide a KubeVirt-native API and CLI mechanism to inject an NMI into a running VMI.
+
+## Non Goals
+
+- This proposal does not define guest OS configuration for crash dump generation.
+- This proposal does not guarantee that every guest OS will react identically to an injected NMI.
+- This proposal does not replace or redesign the existing `memory-dump` feature.
+- This proposal does not standardize guest-side dump collection or post-processing tools.
+
+## Definition of Users
+
+* VM owner.
+
+### User Story 1: Hung guest operating system
+
+As a VM owner, when a guest becomes stuck and unresponsive but the VMI remains alive, I want to inject an NMI so that the guest operating system can generate a crash dump for later kernel analysis.
+(Prerequisites such as enabling kdump on Linux are OS-specific and are typically configured by users following instructions from their OS vendor.)
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### High-level design
+
+This proposal introduces a new VMI subresource:
+
+`PUT /apis/subresources.kubevirt.io/<version>/namespaces/{namespace}/virtualmachineinstances/{name}/injectnmi`
+
+The subresource is the primary API surface for NMI injection. Client tooling such as virtctl may invoke this API:
+
+`virtctl inject-nmi <vmi> [-n <namespace>]`
+
+The command targets VirtualMachineInstance only.
+A VM-level alias is intentionally out of scope; it can be added later if the community wants parity with pause/unpause VM resolution behavior.
+
+The subresource request is handled through the same architecture already used by runtime VMI subresources:
+
+1. `virt-api` authenticates and authorizes the request and forwards the request to `virt-handler`.
+2. `virt-handler` resolves the target VMI launcher socket and forwards the request over the existing cmd/gRPC channel.
+3. `virt-launcher` delegates the operation to the domain manager.
+4. The domain manager invokes the hypervisor backend to inject an NMI.
+
+The KubeVirt client-go library exposes a method for invoking this subresource. The `virtctl inject-nmi` command parses the command arguments and uses this method to invoke the API.
+
+### API and authorization
+
+The operation should be modeled as a new PUT subresource under `virtualmachineinstances`, following the same authorization model used by current subresources. KubeVirt’s subresource architecture already performs per-subresource authorization checks and exposes runtime VMI operations through the aggregated subresource API.
+
+The expected authorization model is identical to the model used by softreboot:
+
+* resource: `virtualmachineinstances/injectnmi`
+* verb: `update`
+
+### Prerequisite
+
+The operation should require:
+
+* the VMI exists
+* the VMI is currently running
+
+### Observability
+
+- On successful backend delivery, emit Normal event reason "NMIInjected".
+- On validation or backend delivery failure, emit Warning event reason "NMIInjectionFailed" with a sanitized error message.
+
+### Backend implementation
+
+The target backend implementation is libvirt.
+
+The `DomainManager` interface in `virt-launcher/virtwrap/manager.go` already defines methods such as `PauseVMI`, `ResetVMI`, and `SoftRebootVMI`.
+An `InjectNMIVMI` method will be added following the same implementation pattern.
+
+When virt-launcher receives the request, it invokes `virDomainInjectNMI()` through libvirt.
+
+[libvirt virDomainInjectNMI documentation](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainInjectNMI)
+
+The libvirt API simply injects an NMI and returns. The NMI serves as a trigger for the guest OS's crash dump mechanism, and the subsequent dump collection process is handled entirely by the OS.
+
+## API Examples
+
+### HTTP API
+
+```http
+PUT /apis/subresources.kubevirt.io/v1/namespaces/test/virtualmachineinstances/my-vmi/injectnmi
+```
+
+### CLI
+
+```bash
+virtctl inject-nmi my-vmi -n test
+```
+
+### Expected responses
+
+* `202 Accepted` when the request is accepted and dispatched
+* `403 Forbidden` the user is not authorized to perform this operation on the VMI
+* `404 Not Found` if the VMI does not exist
+
+## Scalability
+
+This is a per-VMI, stateless control-plane request that follows the same request/forwarding path as existing runtime subresources. It is not expected to introduce new scalability characteristics beyond those already present for VMI runtime operations. Existing runtime subresource architecture already supports similar requests like `pause`, `freeze`, `reset`, and `softreboot`.
+
+## Update/Rollback Compatibility
+
+Clients invoking the new injectnmi subresource against older clusters should receive a clear server-side error because the subresource is not recognized.
+
+## Functional Testing Approach
+
+Testing should include:
+
+1. unit tests for the VMI subresource API handler validation and routing
+2. unit tests for client-go method wiring
+3. unit tests for `virt-handler` lifecycle request handling
+4. unit tests for launcher cmd-server and domain-manager dispatch
+5. unit tests for `virtctl inject-nmi` command parsing and client invocation
+6. End-to-end test
+  * Given a running VMI with a guest configured to log NMI events, execute `virtctl inject-nmi` and verify that the guest records evidence of the NMI (for example, NMI-related kernel log entries).
+
+## Implementation History
+
+## Graduation Requirements
+
+### Alpha
+
+- Implement the NMI injection end-to-end behind InjectNMI feature gate
+  - Add the `injectnmi` API subresource and backend wiring
+  - Add client support via `virtctl inject-nmi`
+  - Support for the default hypervisor path
+- Add unit tests and e2e test
+
+### Beta
+
+- Document guest prerequisites and non-guarantees
+
+### GA
+
+- Documentation exists
+- No known major bugs in default backend


veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets beta for version: v1.10
+- This VEP targets GA for version: v1.11
+
+The change adds a runtime operation comparable to existing VM lifecycle subresources and does not alter the default behavior of existing workloads.
+So, this VEP does not have graduation phases guarded by a feature gate.
```
0xFelix, line 11: Why no feature gate? Why not start at Alpha?

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Goals
+
+- Add a new VMI subresource under `subresources.kubevirt.io`.
+- Implement the operation end-to-end using the existing KubeVirt runtime subresource architecture.
+- Enable injecting an NMI to a VMI by virt-launcher through libvirt.
+- Provide client support via `virtctl inject-nmi <vmi>` for invoking the new subresource.
```
EdDev, line 46: All these points are too low level and are very specific to an implementation choice.

Please focus the goals to the functionality requested and less on how to implement it. That usually means the goals are abstract, translating the need to a practical functionality (without exposing details on how it will be resolved).

Reading the motivation, I am unclear why this requirement aims at the core project. If the user can access the pod and execute the command directly, and this is a debug/troubleshoot action, can an external tool do all this work?
Such a tool could be also a UI or other manager over Kubevirt.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+* `403 Forbidden` the user is not authorized to perform this operation on the VMI
+* `404 Not Found` if the VMI does not exist
+* `409 Conflict` the VMI object exists but is not in a state where a running domain can receive NMI.
+* `500 Internal Server Error` if the backend request cannot be delivered
+
+## Alternatives
+
+### Alternative 1: Exec into `virt-launcher` and use libvirt or host-side tooling directly
+
+Doesn't fit the user story.
```
EdDev, line 144: Alternatives are relevant only if they answer the requirements (including the user-story). You may add an appendix to describes these options for the sake of context, perhaps as a workaround to the lack of this option in Kubevirt.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+- This proposal does not replace or redesign the existing `memory-dump` feature.
+- This proposal does not standardize guest-side dump collection or post-processing tools.
+
+## Definition of Users
+
+* VM owner.
+
+### User Story 1: Hung a guest
+
+As a VM owner, when a guest becomes stuck and unresponsive but the VMI remains alive, I want to inject an NMI so that the guest can generate a crash dump for later kernel analysis.
```
EdDev, line 61: Please add details on how the user is expected to fetch the dump.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
```
EdDev, line 67: Most of the description below is about mechanics that exists today, but very little information is provided on what exactly will be done in from of the hypervisor (I guess it is libvirt).
Things like:
- What command it will use?
- Is it blocking?
- What is the result of the command? Are there artifacts placed in some location? How a user will know it is over?

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,191 @@
...
+1. unit tests for the VMI subresource API handler validation and routing
+2. unit tests for client-go method wiring
+3. unit tests for `virt-handler` lifecycle request handling
+4. unit tests for launcher cmd-server and domain-manager dispatch
+5. unit tests for `virtctl inject-nmi` command parsing and client invocation
+6. end-to-end tests that verify:
+   * the request is accepted for a running VMI
+   * the request is rejected for a non-running VMI
+   * the request flows through all KubeVirt layers
+   * an event is emitted on success/failure
```
EdDev, line 173: e2e is first and most about the happy path, i.e. given a running VM, send an imperative NMI command and consume the dump.

The scenarios described here are unlikely to be accepted as e2e/functional tests.

EdDev (changes requested):
Thank you for the proposal.

I tried to feedback on the points that need attention.
I would be interested in a practical alternative presented where this is not part of Kubevirt core so we can evaluate it compared to the subresource imperative solution.

If possible, please add some explanation about the relation between this NMI injection and the current dump feature. While it was mentioned, it is not clear to me if and how they interact, or in which scenarios each will be used.

kmatsuo3225:
> Thank you for the proposal.
> 
> I tried to feedback on the points that need attention. I would be interested in a practical alternative presented where this is not part of Kubevirt core so we can evaluate it compared to the subresource imperative solution.
> 
> If possible, please add some explanation about the relation between this NMI injection and the current dump feature. While it was mentioned, it is not clear to me if and how they interact, or in which scenarios each will be used.

Could you help us understand what information is missing or unclear in the current Motivation section, if you feel this is not sufficient?

The current Motivation section covers:
- NMI is a common mechanism for collecting crash dumps across operating systems and is natively supported by many virtualization platforms.
- Dumps collected via `virtctl memory-dump` cannot be analyzed using standard OS vendor debugging tools such as `crash` (Linux) or `WinDbg` (Windows).
- Since OS vendors do not support analysis of dumps collected via `virtctl memory-dump`, users cannot receive support from the OS vendor.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: > Reading the motivation, I am unclear why this requirement aims at the core project. If the user can access the pod and execute the command directly, and this is a debug/troubleshoot action, can an external tool do all this work? Such a tool could be also a UI or other manager over Kubevirt.

Although `virsh inject-nmi` can be executed through `kubectl exec`, this approach depends on libvirt and exposes backend-specific implementation details to VM users.
Crash dump collection is typically performed by VM users, rather than cluster administrators, at the request of the OS vendor and should not require knowledge of the underlying virtualization stack.

Furthermore, major virtualization platforms such as Microsoft Hyper-V and VMware vSphere/ESX provide NMI injection as a standard feature.
We believe NMI injection should be a first-class feature in KubeVirt, alongside `virtctl memory-dump`.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: This VEP does not change the behavior of any existing function. It simply adds a new API, so users who do not use it are unaffected.
The PR for the `memory-dump` feature did not introduce a feature gate, so I followed the same approach here.
The implementation is also relatively small, so I was considering targeting Beta directly instead of Alpha. Do you see any issues with that?

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Good point.

This option does not satisfy the user story, so I removed it from the Alternatives section.
I also decided not to include it as a workaround for now, since exec'ing into a pod is not a recommended KubeVirt workflow.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: > All these points are too low level and are very specific to an implementation choice.
>
> Please focus the goals to the functionality requested and less on how to implement it. That usually means the goals are abstract, translating the need to a practical functionality (without exposing details on how it will be resolved).

Thank you. I revised the goals.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Dump collection is OS-specific.
I updated the user story to clarify that users must enable the OS-specific crash dump mechanism beforehand, following guidance from their OS vendor.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you. 

I added more details about the hypervisor backend in the Backend Implementation section.

veps/sig-compute/403-inject-nmi/vep.md
0xFelix: Adding a new API is a reason to gate this feature in my opinion. I don't see a strong reason to skip Alpha.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you. I updated the e2e test to cover the flow from an NMI injection request to the guest receiving the NMI.

As already mentioned in the VEP (for example, in the Non-Goals section), this VEP only covers NMI injection by the hypervisor. Therefore, I excluded crash dump collection from the e2e test.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,181 @@
...
+virtctl inject-nmi my-vmi -n test
+```
+
+### Expected responses
+
+* `202 Accepted` when the request is accepted and dispatched
+* `403 Forbidden` the user is not authorized to perform this operation on the VMI
+* `404 Not Found` if the VMI does not exist
+* `409 Conflict` the VMI object exists but is not in a state where a running domain can receive NMI.
+* `500 Internal Server Error` if the backend request cannot be delivered
```
EdDev: The operations through virt-api are async in most cases.
In order to serve all these responses, virt-api needs to contact the virt-handler, then wait for an expected event/status and only then respond. AFAIK think is not something we do.

In addition, the common use of validating VMI state at virt-api is an architectural bug. As the system is async, any status seen at a specific time does not assure success or even a correct failure.

IMO, the only feasible response here is to add a condition that expresses the success or failure to deliver the injected nmi.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,181 @@
...
+
+The command targets VirtualMachineInstance only.
+A VM-level alias is intentionally out of scope; it can be added later if the community wants parity with pause/unpause VM resolution behavior.
+
+The subresource request is handled through the same architecture already used by runtime VMI subresources:
+
+1. `virt-api` authenticates and authorizes the request, validates the VMI state, and forwards the request to `virt-handler`.
+2. `virt-handler` resolves the target VMI launcher socket and forwards the request over the existing cmd/gRPC channel.
+3. `virt-launcher` delegates the operation to the domain manager.
+4. The domain manager invokes the hypervisor backend to inject an NMI.
```
EdDev: Kubevirt, prefers expressing requests through a declarative change, over imperative. The current imperative commands are IMO also problematic and suffer from the same problems.

This flow is full of races (see my comment on the expected responses).

On the other hand, I understand why this one-time operation is considered a fit for being imperative.
It is not trivial to imagine setting a spec field that express the desire to inject the nmi.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you. I updated the description to include a feature gate and clarify that teh feature will not to skip Alpha phase.

veps/sig-compute/403-inject-nmi/vep.md
EdDev: @kmatsuo3225 , it is challenging for me to evaluate if such a dump is retrievable.
Please share a practical example on how one would be able to consume this artifact.

The OS may support it, but orchestration may still be needed to retrieve it from a Kubevirt VM. Same as the support of NMI by the guest OS is not sufficient to trigger it without wiring everything.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Repos
+
+* [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### High-level design
+
+This proposal introduces a new VMI subresource:
```
EdDev, line 69: Kubevirt already supports one method of memory-dump, with a similar orchestration, but also with an practical e2e use case (PVC is used to retrieve the dump itself) that is lacking in this solution.

1. I am interested to understand how a user will be able to fetch the dump from a VM.
2. Have you consider adding an option with the existing command to use a different method?
3. Have you consider providing a simple `kubectl exec` to execute the hypervisor command from within the VM pod? A simple script tool could be created that will resolve a VM name to its pod and run this command.

veps/sig-compute/403-inject-nmi/vep.md
```diff
@@ -0,0 +1,183 @@
...
+
+KubeVirt already provides a VM memory dump feature, but the earlier KubeVirt design proposal for this feature explicitly frames the solution around later inspection with Volatility3. [\[github.com\]](https://github.com/kubevirt/community/blob/main/design-proposals/vm-memory-dump.md)
+That workflow is useful for generic memory inspection and forensic analysis, but it does not address the main operational requirement for Linux kernel hang analysis. In practice, kernel hang analysis is typically performed on kdump-generated `vmcore` files using kernel-oriented tooling such as `crash` together with matching kernel debuginfo. Red Hat documentation for `crash` explains that it operates on kdump-generated dump files, and Linux kdump documentation explicitly lists NMI as one of the events that can trigger crash dump capture. [\[docs.redhat.com\]](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/analyzing-a-core-dump_assembly_managing-kernel-command-line-parameters-with-uki#running-and-exiting-the-crash-utility_analyzing-a-core-dump), [\[kernel.org\]](https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html)
```
EdDev, line 35: I tried to dig a bit on this subject, but I am not close to being an expert and my dependency on web searches and AI agents is a sign of warning. 

Based on what I managed to understand/learn, I don't think this holds as stated, at least for Linux. Looking at KubeVirt's actual implementation ([`memoryDump.go`](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/storage/memoryDump.go)), `memory-dump` calls libvirt's `CoreDumpWithFormat` with `DOMAIN_CORE_DUMP_FORMAT_RAW`, which is the same as `virsh dump --dump-format elf` [\[libvirt dump-format docs\]](https://libvir-list.redhat.narkive.com/pTJp6Xd1/libvirt-patch-v6-0-4-support-dumping-guest-memory-in-compressed-format), a plain ELF core produced by QEMU's `dump-guest-memory`. `crash` can load that format directly, `crash vmcore.img vmlinux` is a documented workflow [\[Ubuntu wiki: DebuggingKernelWithQEMU\]](https://wiki.ubuntu.com/DebuggingKernelWithQEMU). What's actually missing is that KubeVirt never sets libvirt's `<vmcoreinfo state='on'/>` domain XML feature, so QEMU's dump lacks the `VMCOREINFO` note `crash` needs to bootstrap symbol resolution, without it "`crash` will spend a good deal of time trying to open it prior to failing" [\[Oracle Linux blog: What's inside a Linux kernel core dump\]](https://blogs.oracle.com/linux/whats-inside-a-linux-kernel-core-dump). That's a configuration gap in `memory-dump`, not a fundamentally different or unconvertible format. Volatility3 can also read this same ELF artifact [\[Volatility3 ELF support discussion\]](https://github.com/volatilityfoundation/volatility3/issues/97), [\[Volatility3 tutorial\]](https://www.wirsingsecurity.com/tutorials/memory-forensics-with-volatility-3/), so the same dump file is plausibly usable by both tool ecosystems once that flag is set.

There is a real, narrower gap worth keeping in the Motivation: because the dump is captured externally, QEMU has to fake the per-vCPU register/task-state notes rather than using the crash kernel's own authoritative context, so backtrace fidelity would likely be worse than a true in-guest kexec-generated `vmcore` even once loadable [\[Oracle Linux blog\]](https://blogs.oracle.com/linux/whats-inside-a-linux-kernel-core-dump). That's a legitimate argument for NMI injection, but it's a fidelity/trigger-mechanism argument, not "different, unconvertible format."

Windows is a separate case where the current wording holds up better: QEMU's ELF `dump-guest-memory` output isn't compatible with the native Windows crash-dump (`MEMORY.DMP`) format, and libvirt doesn't expose QEMU's `win-dmp` output option, so there's a real format gap there, not just a missing toggle [\[Extracting Windows VM crash dumps\]](https://ladipro.wordpress.com/2017/01/06/extracting-windows-vm-crash-dumps/).

Given this, could the Motivation be revised to:
1. Make the Linux claim more precise (missing `vmcoreinfo` + lower dump fidelity, rather than incompatible formats), and
2. Explicitly separate the Linux and Windows cases, since the underlying gap is different for each?

It's also worth noting in Alternatives whether enabling `vmcoreinfo` in `memory-dump` was considered as a lighter-weight fix for the Linux case before introducing a new subresource.

<details>
  <summary>References:</summary>

- [kernel.org: kdump documentation](https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html)
- [Red Hat: Analyzing a core dump with `crash`](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/analyzing-a-core-dump_assembly_managing-kernel-command-line-parameters-with-uki)
- [KubeVirt `memoryDump.go` source](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/storage/memoryDump.go)
- [libvirt dump-guest-memory format options](https://libvir-list.redhat.narkive.com/pTJp6Xd1/libvirt-patch-v6-0-4-support-dumping-guest-memory-in-compressed-format)
- [Ubuntu wiki: Debugging Kernel With QEMU (loading ELF dumps in `crash`)](https://wiki.ubuntu.com/DebuggingKernelWithQEMU)
- [Oracle Linux blog: What's inside a Linux kernel core dump (`vmcoreinfo` requirement)](https://blogs.oracle.com/linux/whats-inside-a-linux-kernel-core-dump)
- [Volatility3 ELF dump support (issue #97)](https://github.com/volatilityfoundation/volatility3/issues/97)
- [Volatility3 memory forensics tutorial](https://www.wirsingsecurity.com/tutorials/memory-forensics-with-volatility-3/)
- [Extracting Windows VM crash dumps (format gap for Windows)](https://ladipro.wordpress.com/2017/01/06/extracting-windows-vm-crash-dumps/)
- [KubeVirt design proposal: VM memory dump](https://github.com/kubevirt/community/blob/main/design-proposals/vm-memory-dump.md)
- [KubeVirt user-guide: Virtual machine memory dump](https://kubevirt.io/user-guide/compute/memory_dump/)

</details>

EdDev (commented):
The VEP content is clear and the format good now, thank you!

After learning about this subject a bit more, I have raised some points inline.
The framing of the goal seems to have limited alternatives, so I think it can be reconsidered to be soften. E.g.: Is having a formal API a must?

veps/sig-compute/403-inject-nmi/vep.md
michalskrivanek: @EdDev I don't think retrieval needs to be part of this VEP.
I can think of one trivial example - the dump is persisted in the guest, so on subsequent boot you can retrieve it any way you want. Or even just keep it there and analyze entirely inside the guest.

0xFelix:
WDYT about making this an additional command in virtctl, that runs the inject nmi command on the Pod without another subresource API? Keeping changes to a minimum.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you. Based on your feedback, I removed `409` and `500` from the proposed status codes.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you. I agree that this has the potential issue you described. However, as you mentioned, I followed the existing approach used by one-time operations such as `soft-reboot` for the sake of simplicity.

I also removed the VMI state validation from `virt-api`.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you, @michalskrivanek. This is exactly my understanding as well.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: Thank you for the suggestion. I'll investigate the points you raised and consider revising the Motivation accordingly.

veps/sig-compute/403-inject-nmi/vep.md
kmatsuo3225: > I am interested to understand how a user will be able to fetch the dump from a VM.
 
As @michalskrivanek commented, a user can retrieve the dump by logging into the VM where it was persisted.
Also, with proper configuration, the dump can be automatically transferred to a remote server when it is generated.

> Have you consider adding an option with the existing command to use a different method?

Yes, I considered that option, and I think it could be an alternative to this VEP.

However, I think these features serve different purposes. NMI injection is a one-shot operation, while `memory-dump` is a workflow that involves preparing storage for the dump, retrieving the generated dump, monitoring its progress, and cleaning up the storage afterward.

In addition, the state of the VMI after NMI injection differs from that of a `memory-dump` operation and depends on the guest's configuration.

For these reasons, I think it is better to implement NMI injection as a separate VM control operation rather than combining it with the `memory-dump` workflow.

> Have you consider providing a simple kubectl exec to execute the hypervisor command from within the VM pod? A simple script tool could be created that will resolve a VM name to its pod and run this command.

I think `pod/exec` is a much broader privilege than required for NMI injection. As a cluster administrator, I would like to avoid granting pod-level access to a VM owner solely for the ability to inject an NMI.

In production environments, VM owners are typically expected to interact with VMs through KubeVirt APIs rather than directly accessing `virt-launcher` pods. Therefore, I believe NMI injection is better exposed as a dedicated API than as a workflow based on `pod/exec` access.

kmatsuo3225:
> WDYT about making this an additional command in virtctl, that runs the inject nmi command on the Pod without another subresource API? Keeping changes to a minimum.

@0xFelix 

Yes, I considered these two approaches, but I think it is better to implement NMI injection as described in this VEP.

- Adding an option to the existing command

  As I mentioned above, the existing command and NMI injection serve different purposes. The former is a workflow, whereas the latter is a one-shot operation.

- Adding a spec field that expresses the desire to inject an NMI

  When a VMI resource is created, it does not seem natural for it to contain a field such as `injectNMI: true`. NMI injection is an action requested at a specific point in time rather than a desired state that should be expressed in the VMI specification.

0xFelix:
I'm proposing to make this a virtctl only command. Why does it not suffice?

veps/sig-compute/403-inject-nmi/vep.md
EdDev: > I think `pod/exec` is a much broader privilege than required for NMI injection. As a cluster administrator, I would like to avoid granting pod-level access to a VM owner solely for the ability to inject an NMI.

I think this is the default. Please correct me if I am wrong.
The virt-launcher pod is an unprivileged one, because it is assumed the VM owner is able to access it.

> In production environments, VM owners are typically expected to interact with VMs through KubeVirt APIs rather than directly accessing `virt-launcher` pods. Therefore, I believe NMI injection is better exposed as a dedicated API than as a workflow based on `pod/exec` access.

I think this assumes that Kubevirt core should provide/expose all functionalities possible for VMs directly.
Several solutions are made available and exposed through satellite projects that integrate with Kubevirt.

I think we prefer to avoid imperative commands as much as possible and in this case, a declarative API is not a fit. Considering the orchestration needed after executing this command, I think the simplest option is the preferred one.

kmatsuo3225: I understand your points, but I'm still unclear about the criteria used to determine whether a use case should be exposed through a subresource API.

NMI injection is a general use case regardless of the underlying hypervisor. Existing operations such as `pause` and `soft-reboot` are implemented through subresource APIs, which led me to believe that NMI injection would be a reasonable fit as well.

Although I understand that there may be reasons I'm not aware of, could you help me understand the criteria used when deciding whether functionality should be exposed through a subresource API versus other mechanisms? If there are established guidelines or historical context behind these decisions, I'd be happy to follow them.

Issue #405, opened 2026-07-22
VEP 405: Resolve the virt-launcher pod IP in the portforward subresource


lllamnyp:
**Primary contact (assignee)**:

/assign @lllamnyp

**Current Feature Stage**:

New

**Feature Gate**:

`TunnelToLauncherPod` (proposed; naming open to SIG preference)

**Responsible SIGs**:

Primary SIG:
/sig network

**Enhancement link**:

VEP PR: https://github.com/kubevirt/enhancements/pull/406 (will update to the merged enhancement path once merged)

Originating discussion: https://github.com/kubevirt/kubevirt/issues/18399

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/406
  * [ ] Code PRs: https://github.com/kubevirt/kubevirt/pull/18564 (draft reference implementation)
  * [ ] Docs PRs:

**Additional context**:

virt-api's `portforward` subresource dials `vmi.status.interfaces[0].IP`, which
is not routable from the virt-api pod when the guest's primary interface is on
an isolated network (VPC, primary user-defined network). The enhancement makes
the subresource resolve and dial the current virt-launcher pod IP instead
(falling back to the interface IP), so `virtctl ssh`/`port-forward` work on
such networks. Direction accepted by SIG-network in
kubevirt/kubevirt#18399.

frenzyfriday:
Removing the VEP from 1.10 milestone as it was decided to be implemented without a VEP.

PR #406, opened 2026-07-22
VEP-405: Resolve the virt-launcher pod IP in the portforward subresource

lllamnyp:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/405
**SIG label**:

/sig network

### What this PR does

Adds VEP 405: fix target resolution in virt-api's `portforward` subresource by
dialing the current virt-launcher pod IP (falling back to the first interface
IP), so the subresource works for multihomed launcher pods whose guest primary
interface is on a network not routable from the pod network (VPC, primary
user-defined network). `virtctl ssh`/`port-forward`, as clients of the
subresource, start working on such networks as a consequence. No VMI API
change.

Direction discussed and accepted by SIG-network in kubevirt/kubevirt#18399.
A prototype of the proposed design (unit-tested) is available for reference:
https://github.com/kubevirt/kubevirt/compare/main...lllamnyp:kubevirt:vmi-pod-ip-resolver

- The per-binding forwarding matrix in the Design section records which
  bindings are expected to forward pod-IP traffic to the guest, with the basis
  for each expectation; none were independently verified as part of drafting
  the VEP.
- The desired behavior of in-flight tunnels during live migration is posed as
  an open question in the Functional Testing section.
- Targeting alpha in v1.10 (VEP freeze 2026-09-02).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/veps/sig-network/405-portforward-launcher-pod-ip/vep.md
+++ b/veps/sig-network/405-portforward-launcher-pod-ip/vep.md
@@ -1,0 +1,326 @@
+# VEP #405: Resolve the virt-launcher pod IP in the `portforward` subresource
+
+<!--
+Tracking issue: https://github.com/kubevirt/enhancements/issues/405
+Originating discussion: kubevirt/kubevirt#18399
+-->
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+The `portforward` subresource of `virt-api` tunnels TCP traffic to a VM by
+dialing the first IP reported in `vmi.status.interfaces[0].IP`. That address is
+the guest's own interface IP. For multihomed virt-launcher pods — where the
+guest's primary interface is attached to a network other than the cluster pod
+network (a VPC, a primary user-defined network) — the guest IP is not routable
+from `virt-api`, and the subresource fails even though the launcher pod itself
+remains reachable on the cluster network.
+
+This VEP fixes the subresource's target resolution: `virt-api` dials the
+**current `virt-launcher` pod IP**, falling back to the interface IP when the
+pod IP cannot be determined. Clients of the subresource — `virtctl
+port-forward` and `virtctl ssh` — start working on such networks as a
+consequence, but the primary target of this VEP is the subresource API itself.
+
+Whether traffic that lands in the launcher pod then reaches the guest depends
+on the network binding/provider forwarding pod-IP traffic to the guest — a
+property that varies by binding and is the main scope boundary of this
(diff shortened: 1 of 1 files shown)


kvaps (approved):
From the Cozystack side: this is the motivating deployment behind #18399. A VPC-style CNI where the guest owns its VPC address on the primary `bridge` interface and pod-IP traffic is DNAT'ed into the guest.

+1 on resolving the launcher pod IP in the `portforward` subresource instead of publishing it on the VMI status. The cluster network is the only one `virt-api` and `virt-launcher` are guaranteed to share, and dial-time resolution avoids the post-migration staleness the `status.podIP` variant would carry.

On the reachability table, provider-DNAT row: in our setup pod-IP traffic reaches the guest at the network level. We haven't confirmed the full `virtctl ssh` / `port-forward` path end to end yet, and we're happy to run and contribute that alpha e2e.

orelmisan (changes requested):
Thank you @lllamnyp for the thorough writeup and the prototypes!

I shared my position on the original issue (kubevirt/kubevirt#18399).

In short: I agree with dialing from virt-api to the virt-launcher pod IP for VMs connected to the pod network. If that solves the problem, we can avoid adding API surface entirely, and I don't think this needs a VEP. For existing core bindings used with pod network, the pod IP resolves to the same reachable guest so there is no observable behavior change to graduate.

veps/sig-network/405-portforward-launcher-pod-ip/vep.md
```diff
@@ -0,0 +1,326 @@
...
+forward it to the guest. This is the substantive scope boundary of the VEP. The
+following expectations are drawn from the SIG-network discussion in #18399 and
+from architectural reasoning; **none were independently verified as part of
+this VEP**:
+
+| Binding / provider | Pod IP expected to reach guest? | Basis |
+|---|---|---|
+| `passt` | Yes | Reported working by SIG-network (@nirdothan, @0xFelix) in #18399, via a small `virt-api` change. Not verified here. |
+| `masquerade` | Likely, with changes | SIG-network in #18399 expects tweaks (e.g. forwarding the tunnel port). Not verified here. |
+| `bridge`, guest holds the pod IP | Yes | The pod IP is the guest's own address, so dialing it is equivalent to today's behavior. Reasoning, not verified here. |
```
nirdothan, line 151: This row is somewhat confusing, because if the bridge is on the cluster network, this proposal changes nothing from current functionality and I believe that the same goes for `masquerade` and obviously for `SR-IOV` . The whole point here is the impact on bindings not connected to cluster network and I believe that only `passtBinding` will work OOB here.

nirdothan (approved):
Thank you for this good proposal.
I fully agree with @orelmisan that we should go with option 1.

In an offline conversation with SIG-Network maintainers we agreed that we can process an implementation PR without needing an approved VEP, as this proposal does not introduce new features or a significant architecture change but rather a bug fix.
However, the implementation would not be implemented in SIG-Network code area, therefore I recommend to confirm with @0xFelix who maintains this area.

0xFelix:
If there's no change to the public API, I think we can handle this as a fix without a dedicated VEP.

IIUC the change basically would be to determine the Pod's current IP in the Pod network instead of going with the IP address of the first interface on the VMI, right?

orelmisan:
> If there's no change to the public API, I think we can handle this as a fix without a dedicated VEP.
> 
> IIUC the change basically would be to determine the Pod's current IP in the Pod network instead of going with the IP address of the first interface on the VMI, right?

Right.

0xFelix:
Thanks, then let's go ahead without a dedicated VEP.

xpivarc:
+1 Please just link the work that follow this

PR #407, opened 2026-07-23
VEP 190: Alpha 2 (v1.10) - PreBoot, PreMigrationSource launcher hooks

iholder101:
### VEP Metadata

**SIG label**:
/sig compute

### What this PR does
Extend VEP 190 with the Alpha 2 scope targeting v1.10:
- Rename domainHooks to launcherHooks with discriminated union supporting DomainMutation, PreBoot, and PreMigrationSource hook types.
- PreBoot launcher hook for modifying generated artifacts (cloud-init ISO, NVRAM, etc.) before VM start via shared volume.
- PreMigrationSource launcher hook for customizing migration flags and parameters before MigrateToURI3.
- ~NodeReconcile node hook point firing every reconcile iteration with NodeContext (node info, VMI list, KubeVirt CR).~

### DIFF
--- a/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
+++ b/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
@@ -9,6 +9,7 @@ Owners:
 ### Target releases
 
 - This VEP targets alpha for version: v1.9. 
+- This VEP targets alpha 2 for version: v1.10.
 - This VEP targets beta for version:
 - This VEP targets GA for version:
 
@@ -41,6 +42,12 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
     - [Admission options](#admission-options)
     - [API example:](#api-example)
   - [Plugin metadata](#plugin-metadata)
+  - [Alpha 2 (v1.10): Launcher Hooks](#alpha-2-v110-launcher-hooks)
+    - [Renaming Domain Hooks to Launcher Hooks](#renaming-domain-hooks-to-launcher-hooks)
+    - [PreBoot Hook](#preboot-hook)
+    - [PreMigrationSource Hook](#premigationsource-hook)
+  - [Alpha 2 (v1.10): NodeReconcile](#alpha-2-v110-nodereconcile)
+  - [Alpha 2 (v1.10): API Examples](#alpha-2-v110-api-examples)
 - [API Examples (full version)](#api-examples-full-version)
 - [Alternatives](#alternatives)
 - [Possible future enhancements](#possible-future-enhancements)
@@ -310,6 +317,210 @@ These are the basic fields every plugin would need to populate:
 The plugin’s name comes from the Kubernetes object metadata (`metadata.name`).
 Versioning and upgrade path are deferred to [future enhancements](#possible-future-enhancements).
 
+### Alpha 2 (v1.10): Launcher Hooks
+
+Alpha 1 introduced **domain hooks** for modifying libvirt domain XML via CEL expressions or sidecar containers.
+In practice, the plugin system needs to support hook points inside `virt-launcher` that go beyond domain XML mutation -
+for example, modifying guest initialization artifacts or tuning live migration parameters.
+
+To reflect this broader scope, Alpha 2 renames `domainHooks` to `launcherHooks` and introduces a discriminated union
+that supports multiple hook types under one array. Each entry in `launcherHooks` is exactly one of:
+- **GuestDefinition**: The existing domain XML hook (CEL or sidecar), renamed for hypervisor-agnostic clarity. Same semantics as Alpha 1.
+- **PreBoot**: A new hook point for modifying generated artifacts before VM start.
+- **PreMigrationSource**: A new hook point for customizing live migration behavior.
+
+Shared fields (`condition`, `failureStrategy`, `timeout`) remain at the hook entry level.
+Hooks are applied in declaration order within each plugin, and in alphabetical order by plugin name across plugins -
+same as Alpha 1.
+
+#### Renaming Domain Hooks to Launcher Hooks
+
+Alpha 1 named these hooks `domainHooks`, but not all launcher-side hooks are related to domain
+mutation. `launcherHooks` better reflects that this category encompasses any hook that runs
+inside the `virt-launcher` pod - whether it modifies the guest definition, touches filesystem
+artifacts, or tunes migration parameters.
+
+Within `launcherHooks`, the existing domain XML hook becomes `guestDefinition` - a
+hypervisor-agnostic name that describes what the hook does (modify the guest's definition)
+without leaking libvirt-specific terminology.
+
+#### PreBoot Hook
+
+Some plugin use cases require modifying guest initialization artifacts that are generated at runtime -
+for example, cloud-init configuration that depends on CNI data (Multus network-status annotations)
+only available after pod creation. The existing domain hook cannot address this because it operates on
+domain XML, not guest initialization media.
+
+The `preBoot` hook fires inside `startDomain()`, after all artifacts (cloud-init ISO, NVRAM, firmware, etc.)
+have been generated but before the VM is started via `CreateWithFlags`. This placement gives the sidecar
+access to the finalized artifacts while still allowing modifications before the guest sees them.
+
+**Design:**
+- Sidecar-only (not CEL) - these hooks perform filesystem side effects, not structured data transformations.
+- Socket-based signaling: `virt-launcher` calls the sidecar's socket to signal that artifacts are ready.
+  The sidecar performs its modifications and responds when done.
+- Filesystem access via shared volume: the plugin author's Mutating Admission Policy (MAP) sets up an
+  `emptyDir` volume mounted in both the compute container and the sidecar container. The sidecar can
+  read and modify generated artifacts directly on the shared filesystem.
+- KubeVirt does not need artifact-specific code - the sidecar knows which files it needs to modify.
+- The [kubevirt/plugins](https://github.com/kubevirt/plugins) SDK will provide helpers for common operations
+  (e.g., cloud-init ISO extraction, modification, and repacking).
+
+**Ordering with multiple plugins:** PreBoot hooks fire sequentially in the standard plugin ordering
+(alphabetical by plugin name, declaration order within each plugin). Each plugin's sidecar sees the
+filesystem state left by the previous plugin's sidecar. Plugin authors should be aware of this
+and design their modifications to be composable.
+
+**Note on hook ordering:** Guest definition hooks fire during domain definition (before `startDomain()`).
+PreBoot hooks fire later, during `startDomain()`. A domain mutation hook could add a device that
+affects cloud-init configuration - plugin authors should be aware of this ordering when designing
+plugins that combine both hook types.
+
+**Example use case:**
+A networking plugin that configures guest cloud-init network settings based on runtime CNI data
+(see [VEP 337](https://github.com/kubevirt/enhancements/issues/337)). The plugin's sidecar reads
+Multus network-status annotations via the Downward API, extracts the cloud-init ISO from the shared
+volume, injects the appropriate network configuration, and repacks the ISO.
+
+#### PreMigrationSource Hook
+
+Live migration behavior in KubeVirt is currently controlled entirely by internal logic.
+Plugins that manage specialized hardware, networking, or storage may need to customize migration
+parameters on a per-workload basis - for example, enabling multifd parallel migration and tuning
+thread counts for high-bandwidth workloads, or selecting specific compression methods.
+
+The `preMigrationSource` hook fires in `virt-launcher` on the migration source, just before
+the `MigrateToURI3` libvirt API call. The sidecar receives the generated migration flags
+and parameters, and can modify them before the migration begins.
+
+**Design:**
+- Sidecar-only (not CEL) - migration parameter tuning requires logic that goes beyond simple expressions.
+- The sidecar receives the full set of migration flags (e.g., `VIR_MIGRATE_PARALLEL`,
+  `VIR_MIGRATE_COMPRESSED`) and parameters (e.g., `Bandwidth`, `ParallelConnections`,
+  `Compression` method, `MigrateDisks` list) via a new gRPC message.
+- The sidecar can enable/disable flags and set parameter values, giving full control over
+  migration behavior. This allows plugins to both enable features (e.g., multifd) and
+  configure them (e.g., number of parallel connections).
+- A new protobuf service and message types will be defined for this hook point.
+- The sidecar can also perform other pre-migration preparation (e.g., signaling external systems,
+  preparing source-side resources).
+
+**Naming:** Named `PreMigrationSource` to match the existing node hook convention and to
+leave room for a future `PreMigrationTarget` launcher hook on the destination side.
+
+**Example use case:**
+A storage plugin that enables multifd migration with a tuned thread count for VMs using
+high-throughput network-attached storage, while keeping the default single-stream migration
+for VMs with local disks.
+
+### Alpha 2 (v1.10): NodeReconcile
+
+Alpha 1's node hooks are lifecycle-event-driven - they fire at specific points in a VM's lifecycle
+(PreVMStart, PostVMStop, migration events, etc.). Some node-level plugin operations need continuous
+reconciliation rather than reacting to discrete events. For example, managing a node's hugepage pool requires knowing the total hugepage demand
+across all running VMs on a node, not just reacting to individual VM start/stop events.
+
+`NodeReconcile` is a new node hook point that fires on every `virt-handler` reconcile iteration.
+Unlike the existing lifecycle hook points which receive a single VMI, `NodeReconcile` provides
+a broader node-scoped context.
+
+**Design:**
+- Fires on every `sync()` call in the virt-handler VMI controller. Since virt-handler reconciles
+  per-VMI, this means `NodeReconcile` fires once per VMI reconcile. Plugin implementations must
+  be idempotent and should short-circuit when the relevant state hasn't changed.
+- The hook receives a `NodeContext` containing:
+  - Node information: labels, annotations, capacity, allocatable resources.
+  - A list of VirtualMachineInstances currently running on the node with key characteristics.
+  - The KubeVirt CR - providing the cluster-wide KubeVirt configuration (feature gates, migration
+    defaults, developer configuration, etc.) to the hook.
+- CEL conditions for `NodeReconcile` hooks evaluate against the same `NodeContext`, using `node`,
+  `vmis`, and `kubevirt` variables. This allows efficient filtering - for example, a plugin can
+  use a CEL condition to only fire when specific node labels are present or when the VMI count
+  changes.
+- The same `NodeContext` data is sent both to the CEL evaluator (for condition filtering) and
+  to the plugin's gRPC socket (for hook execution).
+
+**Performance considerations:** Since this hook fires on every VMI reconcile, plugins should
+use CEL conditions aggressively to filter unnecessary calls. The hook implementation itself
+should be lightweight - check whether relevant state has changed before performing expensive
+operations. The `failureStrategy` and `timeout` fields apply as usual.
+
+**Example use case:**
+A hugepage pool management plugin that tracks total hugepage demand across all VMs on the
+node and adjusts the node's hugepage allocation accordingly. The plugin uses a CEL condition
+to fire only when the node has specific labels, sums the hugepage requests from the VMI list,
+and rebalances the pool when the total demand changes.
+
+### Alpha 2 (v1.10): API Examples
+
+```yaml
+apiVersion: plugin.kubevirt.io/v1alpha1
+kind: Plugin
+metadata:
+  name: network-customizer
+spec:
+  condition: "vmi.metadata.labels['network-plugin'] == 'custom'"
+  failureStrategy: Fail
+
+  launcherHooks:
+    # Guest definition mutation via CEL (same as Alpha 1, under new structure)
+    - guestDefinition:
+        cel:
+          expression: |
+            Domain{CPU: DomainCPU{Mode: "host-passthrough"}}
+
+    # Guest definition mutation via sidecar
+    - guestDefinition:
+        sidecar:
+          socketPath: "/var/run/kubevirt-plugin/network-customizer/domain.sock"
+
+    # PreBoot: modify cloud-init for custom networking
+    - preBoot:
+        socketPath: "/var/run/kubevirt-plugin/network-customizer/preboot.sock"
+      timeout: 30s
+
+    # PreMigrationSource: tune migration for high-bandwidth workloads
+    - preMigrationSource:
+        socketPath: "/var/run/kubevirt-plugin/network-customizer/premigrate.sock"
+      condition: "vmi.metadata.labels['high-bandwidth'] == 'true'"
+      timeout: 60s
+
+  nodeHooks:
+    - socket: /var/run/kubevirt-plugins/network-customizer/node.sock
+      permittedHooks:
+        - PreVMStart
+        - PostVMStop
+        - NodeReconcile
+      condition: "node.metadata.labels['network-plugin-enabled'] == 'true'"
+```
+
+#### Launcher hook type structure
+
+Each entry in `launcherHooks` is a discriminated union - exactly one of the following must be set:
+
+| Hook Type | Purpose | Modes | Input/Output |
+|-----------|---------|-------|--------------|
+| `guestDefinition` | Guest definition mutation | CEL or sidecar | Domain XML in, mutated domain XML out |
+| `preBoot` | Modify generated artifacts before VM start | Sidecar only | Socket signal, filesystem access |
+| `preMigrationSource` | Customize migration flags and parameters | Sidecar only | Migration config in, modified config out |
+
+#### Node hook points (updated)
+
+All hook points from Alpha 1 remain unchanged. Alpha 2 adds `NodeReconcile`:
+
+| Hook Point | Scope | Context | When |
+|-----------|-------|---------|------|
+| PreVMStart | VMI | VMI + node name | Before VM launch |
+| PostVMStart | VMI | VMI + node name | After VM is running |
+| PreVMStop | VMI | VMI + node name | Before VM shutdown |
+| PostVMStop | VMI | VMI + node name | After VM stops |
+| PreMigrationSource | VMI | VMI + node name | Before migration from source |
+| PreMigrationTarget | VMI | VMI + node name | Before migration to target |
+| PostMigrationTarget | VMI | VMI + node name | After migration completes on target |
+| NodeReconcile | Node | NodeContext (node + VMIs + KubeVirt CR) | Every reconcile iteration |
+
+All hook points except `NodeReconcile` were introduced in Alpha 1.
+
 ## API Examples (full version)
 
 ```yaml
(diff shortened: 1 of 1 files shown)


veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -471,13 +682,24 @@ Refer to https://github.com/kubevirt/community/blob/main/design-proposals/featur
...
-- [ ] Simple domain hooks (CEL) are functional.
-- [ ] Advanced domain hooks (plugin sidecar via socket) are functional for a single hook point.
-- [ ] Node hooks are functional for at least PreVMStart and PreVMStop hook points.
-- [ ] failureStrategy (Fail/Ignore) and timeout are respected.
-- [ ] Basic functional tests covering domain hooks and node hooks.
+- [x] Feature gate (`StructuredPlugins`) guards all code changes.
+- [x] Plugin CRD is defined and registered.
+- [x] Simple domain hooks (JsonPatch) are functional.
+- [x] Advanced domain hooks (plugin sidecar via socket) are functional for a single hook point.
+- [x] Node hooks are functional for at least PreVMStart and OnVMStop hook points.
```
Acedus, line 689: Why was this changed to `OnVMStop`?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -310,6 +317,210 @@ These are the basic fields every plugin would need to populate:
...
+
+**Naming:** Named `PreMigrationSource` to match the existing node hook convention and to
+leave room for a future `PreMigrationTarget` launcher hook on the destination side.
+
+**Example use case:**
+A storage plugin that enables multifd migration with a tuned thread count for VMs using
+high-throughput network-attached storage, while keeping the default single-stream migration
+for VMs with local disks.
+
+### Alpha 2 (v1.10): NodeReconcile
```
Acedus, line 416: I'm conflicted on adding this.

For one, why can't the example provided be achieved by using the existing `PreVMStart` to gate VM startup until hugepages are allocated and `PostVMStop` to release hugepages when it stops? 

I think the Plugin DS should be responsible for fetching whatever it specifically needs to drive the synchronous state of the VMs under its purview, simple as that. Furthermore, `NodeReconcile` differs from every other hook in that it doesn't even gate a VM lifecycle transition. If it fails, no operation is blocked (is it?), and `failureStrategy` has no meaningful effect because there's nothing to fail, which IMO suggests that it's solving a different problem (i.e., background state reconciliation, and therefore it doesn't really belong in a lifecycle hook framework). 

Finally, I think it actually gives the Plugin author _less_ control over their own reconciliation cadence that they'd have internally.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -471,13 +682,24 @@ Refer to https://github.com/kubevirt/community/blob/main/design-proposals/featur
...
 Alpha
 
-- [ ] Feature gate (`Plugins`) guards all code changes.
-- [ ] Plugin CRD is defined and registered.
-- [ ] Simple domain hooks (CEL) are functional.
-- [ ] Advanced domain hooks (plugin sidecar via socket) are functional for a single hook point.
-- [ ] Node hooks are functional for at least PreVMStart and PreVMStop hook points.
-- [ ] failureStrategy (Fail/Ignore) and timeout are respected.
-- [ ] Basic functional tests covering domain hooks and node hooks.
+- [x] Feature gate (`StructuredPlugins`) guards all code changes.
```
Acedus, line 685: Changed the FG name?

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -471,13 +682,24 @@ Refer to https://github.com/kubevirt/community/blob/main/design-proposals/featur
...
+- [x] Plugin CRD is defined and registered.
+- [x] Simple domain hooks (JsonPatch) are functional.
```
Acedus, line 687: s/JsonPatch/CEL

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -41,6 +42,12 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
     - [Admission options](#admission-options)
     - [API example:](#api-example)
   - [Plugin metadata](#plugin-metadata)
+  - [Alpha 2 (v1.10): Launcher Hooks](#alpha-2-v110-launcher-hooks)
+    - [Renaming Domain Hooks to Launcher Hooks](#renaming-domain-hooks-to-launcher-hooks)
+    - [PreBoot Hook](#preboot-hook)
+    - [PreMigrationSource Hook](#premigationsource-hook)
```
Acedus, line 48: s/premigation/premigration

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -310,6 +317,210 @@ These are the basic fields every plugin would need to populate:
...
+affects cloud-init configuration - plugin authors should be aware of this ordering when designing
+plugins that combine both hook types.
+
+**Example use case:**
+A networking plugin that configures guest cloud-init network settings based on runtime CNI data
+(see [VEP 337](https://github.com/kubevirt/enhancements/issues/337)). The plugin's sidecar reads
+Multus network-status annotations via the Downward API, extracts the cloud-init ISO from the shared
+volume, injects the appropriate network configuration, and repacks the ISO.
+
+#### PreMigrationSource Hook
```
Acedus, line 385: While I think this is generally a good idea, I can think of several other libvirt operations that accept flags which may benefit from nearly identical hook points (e.g., `CreateFlags`), shouldn't they be included as well?

Acedus, line 385: nit: I know the difference between our existing `NodeHookPreMigrationSource` and this new `PreMigrationSource`, but it's probably worth documenting the it.

Acedus (commented):
Thanks @iholder101! I like _most_ of the design changes made for Alpha 2, left a few comments. On that note, now that Alpha 2 makes some substantial changes to the API, I think the VEP as a whole could use a bit of cleanup to align on the most "up-to-date" version (maybe have an Alpha 1 section similarly to how we have Alpha 2 to document the design process? Although I tend to think that's what the git commits are for).

lyarwood:
**Use case: alternative virt-launcher image selection via plugins**

cc @iholder101 @vladikr

The AIE project (GPU passthrough on NVIDIA AI-Factory platforms) currently uses an external MutatingWebhook to swap the virt-launcher image for a CentOS Stream 10 build with IOMMUFD-capable libvirt/QEMU. This webhook exists because plugins cannot influence which virt-launcher image virt-controller selects — `GetLauncherImage()` in `TemplateService` returns a fixed image from KubeVirt config.

With Alpha 2 broadening `launcherHooks` beyond domain XML into pre-boot artifacts and migration parameters, the Plugin CRD is clearly becoming the unified extension point for everything launcher-related. The one remaining gap is **launcher image selection itself**.

Would you consider a `spec.launcherImage` field (or similar) on the Plugin CRD for v1.10 scope? When a Plugin's condition matches a VMI, virt-controller would use the plugin's launcher image instead of the default. This would:

- Let plugins fully replace external webhooks that exist solely for image swapping
- Subsume the core functionality of the deferred Worker Pools VEP (#200) within the existing plugin framework
- Keep image selection declarative and CEL-condition-gated, consistent with the rest of the Plugin API

If this is out of scope for Alpha 2, we'd be happy to propose it as a follow-up VEP extension or a separate Alpha 2 deliverable.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: I think that this name is a bit more accurate. While `PreVMStart` continuously fires until the VM starts (not only whenever virt-handler actually starts it), the current `PreVMStop` fires only at the time virt-handler decides to stop/destroy the VM (but not before that).

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Honestly I'm conflicted on that as well.

The question is - do we want the Plugin framework to be able to manage things like KSM? We've been talking about extracting KSM from core KubeVirt for a while now, which simplistically means to enable KSM according to certain conditions, e.g. the node is low on memory. But such operations aren't related to a single VM's life-cycle.

On the other hand, virt-handler's `sync()` call is inherently bound to VM lifecycle. In a case where no VM needs attention from virt-handler this hook will simply not fire, which means it's inherently problematic for use-cases such as KSM management. Maybe we need to manage such things with an additional out-of-tree controller that'll deploy a DaemonSet, and not rely on Plugins.

I think I'll remove this hook for now.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: Yup! I definitely agree we probably need more hook points in this direction. One small thing is that we need to be careful not to be tied to "libvirt operations" but for general VM life-cycle hook points, but a "GuestCreation" hook point sounds perfectly reasonable to me. Two questions:
1. Do you have a specific use-case in mind regarding "CreateFlags"?
2. Do you think there are additional hook points we need to consider at this point?

We can also obviously add more hook points as we go, don't need to add them all now (unless we have concrete use-cases in mind).

iholder101:
> Thanks @iholder101! I like _most_ of the design changes made for Alpha 2, left a few comments. On that note, now that Alpha 2 makes some substantial changes to the API, I think the VEP as a whole could use a bit of cleanup to align on the most "up-to-date" version (maybe have an Alpha 1 section similarly to how we have Alpha 2 to document the design process? Although I tend to think that's what the git commits are for).

Thanks for the review @Acedus!
I think git commits are harder to read when we're talking about plain text rather than code changes. Maybe I'll just add another file to document the "history" that'll serve as a change-log. Will try to clean everything up :+1:

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
Acedus: I guess we can start with that and extend according to feedback :)

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
Acedus: I think the direction of extracting it is correct, that doesn't necessarily mean it fits into a plugin, but I don't rule out the option to have a plugin rely on a different entity that's responsible for managing KSM.

iholder101:
> With Alpha 2 broadening `launcherHooks` beyond domain XML into pre-boot artifacts and migration parameters, the Plugin CRD is clearly becoming the unified extension point for everything launcher-related. The one remaining gap is **launcher image selection itself**.

Hey @lyarwood!

Thanks for bringing this up. This idea was raised in other forums, but I'm not sure if we've ever discussed it publicly.

Currently the Plugin framework does not consist of hook points that are used by virt-controller, or that are used for virt-launcher pod generation. Currently, the way to alter the virt-launcher spec is to use mutating webhook/policy, that can be listed as part of the Plugin CR.

Generally speaking, the reasoning for avoiding hook points that are used by virt-controller are:
* It is unclear if there's a strong use-case for it. A mutating webhook/policy can do the same job without introducing more hook points.
* It is unclear in which way virt-controller will call this hook point - a service? a sidecar container?
* It would probably demand a non-negligible refactoring around `RenderLaunchManifest()`, as the full context around the pod generation would need to be handed into the hook. I'm thinking about things like volume rendering, resources, and more.

I wonder - would using mutating webhooks/policies (and posibly wrapping them up in a Plugin CR) do the job here?

lyarwood:
Thanks @iholder101 — a quick clarification on the hook point concern, because I think the framing differs from what I had in mind.

The three objections you raised (call mechanism, `RenderLaunchManifest()` refactoring, unclear use case) all apply to hook points that require an **outbound call** — reaching a sidecar socket, a service, an external process. Those concerns are completely valid for domain XML hooks.

But launcher image selection doesn't need any of that. What I'm proposing isn't a new hook point in the traditional sense — it's a new field, `spec.launcherImage`, on the Plugin CR. The implementation in virt-controller would be entirely internal:

1. List Plugin CRs (already done for existing hook machinery)
2. Evaluate each Plugin's `spec.condition` CEL expression against the VMI (already wired up)
3. If a match has `spec.launcherImage` set, use that instead of the default — no call, no socket, no IPC

The Plugin CR *is* the hook. The CEL condition evaluates inside the controller against the VMI it already holds, and the result is just a string to use as the image. There's no call mechanism question because nothing external is invoked.

On the mutating webhook/policy path: we've looked at this and the Plugin CR's admission reference fields are registration-only — they don't deploy or lifecycle-manage the webhook or policy. So wrapping our webhook in a Plugin CR doesn't reduce operational burden, and a CEL `MutatingAdmissionPolicy` still requires HCO to deploy and manage separate cluster-scoped objects. The image-in-Plugin-CR approach is simpler end-to-end for operators.

Happy to write up a concrete proposal for `spec.launcherImage` if this framing changes the picture at all.

iholder101:
> > Thanks @iholder101! I like _most_ of the design changes made for Alpha 2, left a few comments. On that note, now that Alpha 2 makes some substantial changes to the API, I think the VEP as a whole could use a bit of cleanup to align on the most "up-to-date" version (maybe have an Alpha 1 section similarly to how we have Alpha 2 to document the design process? Although I tend to think that's what the git commits are for).
> 
> Thanks for the review @Acedus! I think git commits are harder to read when we're talking about plain text rather than code changes. Maybe I'll just add another file to document the "history" that'll serve as a change-log. Will try to clean everything up 👍

@Acedus Took your advise and got rid of the "alpha1/alpha2" separation. Hopefully it's cleaner now! :pray:

iholder101:
@lyarwood adding an API specifically for changing the image seems a bit odd to me, especially since AFAICT you can achieve the exact same goal with a simple VAP. Isn't a VAP enough here?

lyarwood:
> @lyarwood adding an API specifically for changing the image seems a bit odd to me, especially since AFAICT you can achieve the exact same goal with a simple VAP. Isn't a VAP enough here?

@iholder101 For straight forward flat virt-launcher pod image replacement sure but the use case we have in AIE that's currently handled by the https://github.com/kubevirt/kubevirt-aie-webhook project is to replace the virt-launcher pod image depending on VMI properties, something that VAPs don't allow but plugins could. This would in turn allow us to drop the deployment through HCO and overall maintenance of the AIE webhook.

iholder101:
> > @lyarwood adding an API specifically for changing the image seems a bit odd to me, especially since AFAICT you can achieve the exact same goal with a simple VAP. Isn't a VAP enough here?
> 
> @iholder101 For straight forward flat virt-launcher pod image replacement sure but the use case we have in AIE that's currently handled by the [kubevirt/kubevirt-aie-webhook](https://github.com/kubevirt/kubevirt-aie-webhook?rgh-link-date=2026-08-19T14%3A58%3A12Z) project is to replace the virt-launcher pod image depending on VMI properties, something that VAPs don't allow but plugins could. This would in turn allow us to drop the deployment through HCO and overall maintenance of the AIE webhook.

I think if we go this route we need a more generalized mechanism. The use-case of mutating the pod according to VMI attributes is currently handled via webhooks/VAPs, but it makes sense to think about how to ease this process.

As I see it, there are two routes that make sense (maybe even both?):
1. Deploy code that will alter the pod (by e.g. a service, gRPC to a sidecar container, etc).
2. Via a CEL expression, with both the VMI and Domain variables being injected in to the CEL environment.

WDYT @Acedus @vladikr?
Should we go into the business of easing modifying the pod according to VMI conditions?
Should we start with a CEL-only approach?

Acedus:
> > > @lyarwood adding an API specifically for changing the image seems a bit odd to me, especially since AFAICT you can achieve the exact same goal with a simple VAP. Isn't a VAP enough here?
> > 
> > 
> > @iholder101 For straight forward flat virt-launcher pod image replacement sure but the use case we have in AIE that's currently handled by the [kubevirt/kubevirt-aie-webhook](https://github.com/kubevirt/kubevirt-aie-webhook?rgh-link-date=2026-08-19T14%3A58%3A12Z) project is to replace the virt-launcher pod image depending on VMI properties, something that VAPs don't allow but plugins could. This would in turn allow us to drop the deployment through HCO and overall maintenance of the AIE webhook.
> 
> I think if we go this route we need a more generalized mechanism. The use-case of mutating the pod according to VMI attributes is currently handled via webhooks/VAPs, but it makes sense to think about how to ease this process.
> 
> As I see it, there are two routes that make sense (maybe even both?):
> 
> 1. Deploy code that will alter the pod (by e.g. a service, gRPC to a sidecar container, etc).
> 2. Via a CEL expression, with both the VMI and Domain variables being injected in to the CEL environment.
> 
> WDYT @Acedus @vladikr? Should we go into the business of easing modifying the pod according to VMI conditions? Should we start with a CEL-only approach?

I'm sorry, but this is a non-starter for me. The Plugin framework (despite what the unfortunately relatively ambiguous motivation might lead you to believe) exists to enable granular control over VM lifecycle (e.g., domain XML, node hooks, migration hooks, etc), IOW concepts that are intrinsic and exclusive to KubeVirt. Pod mutation isn't, it's a k8s-generic concern and k8s already provides the machinery for it (i.e., admission policies and webhooks), and the fact that the condition triggering the mutation is VMI-aware doesn't change this either IMO. A VMI-aware condition with a generic action is _still_ a generic action, the Plugin CR already references admission webhooks exactly for this reason. 

TL;DR my 0.02$ is that if an extension we're proposing would be equally applicable to any other operator that wraps pods, it doesn't belong in the Plugin CRD.

lyarwood:
> > > > @lyarwood adding an API specifically for changing the image seems a bit odd to me, especially since AFAICT you can achieve the exact same goal with a simple VAP. Isn't a VAP enough here?
> > > 
> > > 
> > > @iholder101 For straight forward flat virt-launcher pod image replacement sure but the use case we have in AIE that's currently handled by the [kubevirt/kubevirt-aie-webhook](https://github.com/kubevirt/kubevirt-aie-webhook?rgh-link-date=2026-08-19T14%3A58%3A12Z) project is to replace the virt-launcher pod image depending on VMI properties, something that VAPs don't allow but plugins could. This would in turn allow us to drop the deployment through HCO and overall maintenance of the AIE webhook.
> > 
> > 
> > I think if we go this route we need a more generalized mechanism. The use-case of mutating the pod according to VMI attributes is currently handled via webhooks/VAPs, but it makes sense to think about how to ease this process.
> > As I see it, there are two routes that make sense (maybe even both?):
> > 
> > 1. Deploy code that will alter the pod (by e.g. a service, gRPC to a sidecar container, etc).
> > 2. Via a CEL expression, with both the VMI and Domain variables being injected in to the CEL environment.
> > 
> > WDYT @Acedus @vladikr? Should we go into the business of easing modifying the pod according to VMI conditions? Should we start with a CEL-only approach?
> 
> I'm sorry, but this is a non-starter for me. The Plugin framework (despite what the unfortunately relatively ambiguous motivation might lead you to believe) exists to enable granular control over VM lifecycle (e.g., domain XML, node hooks, migration hooks, etc), IOW concepts that are intrinsic and exclusive to KubeVirt. Pod mutation isn't, it's a k8s-generic concern and k8s already provides the machinery for it (i.e., admission policies and webhooks), and the fact that the condition triggering the mutation is VMI-aware doesn't change this either IMO. A VMI-aware condition with a generic action is _still_ a generic action, the Plugin CR already references admission webhooks exactly for this reason.

Thanks @Acedus, the AIE use case I've been pushing is specific to the launcher pod, would limiting the action (as I think I suggested earlier) to that specific pod help at all here?

Acedus:
> Thanks @Acedus, the AIE use case I've been pushing is specific to the launcher pod, would limiting the action (as I think I suggested earlier) to that specific pod help at all here?

It would still remain out-of-scope IMO. Is the webhook unsustainable on your end? Perhaps we can work something outside of Plugins or we can examine it as an extension in the future once we gain more confidence in the feature and its direction.

WDYT @iholder101 @lyarwood?

iholder101:
I tend to agree with @Acedus that at least currently this should remain out-of-scope. Managing webhooks / MAPs / VAPs / DaemonSets or even injecting the sidecar container as a launcher hook is currently the responsibility of the plugin author - not KubeVirt core.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
```diff
@@ -418,6 +532,34 @@ override each other's data.
 Only if a plugin is dependent on another plugin, both can edit the same fields, because now a clear ordering
 is defined between them. This way one plugin can do some work that another plugin would finalize.
 
+### NodeReconcile Hook
```
Acedus, line 535: I don't think this is necessary TBH, the discussion of it is pretty clear from the PR comments.

Acedus (approved):
Thanks @iholder101, small nit.

lyarwood:
@iholder101 @Acedus - Thanks both, I still think this is entirely reasonable for Plugins to express but appreciate I'm unlikely to get anywhere this cycle now. Comments below.

> It would still remain out-of-scope IMO. Is the webhook unsustainable on your end? 

I wouldn't call it unsustainable but the maintenance cost isn't exactly zero either.

FWIW I hate the fact that kubevirt-aie needs such a mechanism at the moment, I'd much rather provide a complete fork of kubevirt for each release but my hands are tied and we are being forced into supporting this awkward optional launcher image mutation hackaround.

> Perhaps we can work something outside of Plugins or we can examine it as an extension in the future once we gain more confidence in the feature and its direction.

I've already tried and failed to extend virt-operator to support this use case before being asked to try the plugin route, happy to try again but I'm also tired of going around in circles.

> I tend to agree with @Acedus that at least currently this should remain out-of-scope. Managing webhooks / MAPs / VAPs / DaemonSets or even injecting the sidecar container as a launcher hook is currently the responsibility of the plugin author - not KubeVirt core.

I really think the fixation on a generic solution here isn't helpful. As above the use case of launcher image modification at render time within the controller based on attributes of the VMI though a Plugin is very simple and I still believe is something Plugin authors  outside of the kubevirt-aie use case would benefit from.

veps/sig-compute/190-kubevirt-structured-plugins/vep.md
iholder101: I understand your point, however especially because this VEP is pretty heavy and we're likely to have additional multiple VEP PRs I'd rather have our ideas listed in a central place which we can refer to later in follow-up VEPs. In addition, I think it gives a sense of why we've designed the VEP the way it's designed. So if you don't insist I'd rather keep it there for now.

Maybe if this section becomes too long I can move it to a different file moving forward.

iholder101:
@lyarwood I completely understand that there's a maintenance cost to deploying external objects that interact with KubeVirt. That being said, this is not a problem Plugins aim to solve, and TBH I'm not even sure KubeVirt needs to solve in general.

Plugins are supposed to give you relevant hook points that can be used to intervene with Kubevirt's operations in key points. However, it deliberately avoids deploying the necessary components that are executing the relevant logic in these hook points. For example, these objects are expected to be deployed separately by the Plugin author:
* DaemonSets that are serving the Node hook points.
* virt-launcher sidecar containers (that are expected to be injected to the pod via a MAP/webhook).
* MAPs/VAPs/Webhooks that modify VMIs / virt-launcher pods.

The main reasoning for this design is that on the one hand Kubevirt doesn't know how to deploy these resources in the context of a specific plugin (how many resources does a DS need? Which capabilities? etc.), and on the other hand we want the responsibility of managing and defining these objects to be on the plugin author. This philosophy is aligned with other plugin mechanisms, e.g. NRI/CNI/CRI etc.

But even when thinking about Kubevirt in general, not Plugins necessarily, I think this philosophy is the right direction forward. I'm not convinced that we should get into the business of deploying stuff for external use, but tend to think that since **you** know your needs and the context of the problem you're trying to solve, **you** are the best person to manage and deploy these resources.

There's a maintenance cost to deploying these objects - whether Kubevirt does it or an external entity does that - the question is who pays this price. I think that Kubevirt should not pay this price for external components, but should focus on giving them the possibility to interact with Kubevirt (and take the necessary responsibility for it).

0xFelix:
> you know your needs and the context of the problem you're trying to solve, you are the best person to manage and deploy these resources

I think that is the point. A MAP doesn't know the context since it cannot access the VM or VMI to select what virt-launcher image to use when it is targeted to a Pod. KubeVirt knows more in that case and a plugin would allow to easily make a selection based on its knowledge.

Also, in that case it is not about deploying external resources, but just a single virt-launcher. There's a 1:1 relation between VM/VMI and the Pod and KubeVirt does already manage it. We're just looking for a way to tell KubeVirt which image it should deploy.

iholder101:
> I think that is the point. A MAP doesn't know the context since it cannot access the VM or VMI to select what virt-launcher image to use when it is targeted to a Pod. KubeVirt knows more in that case and a plugin would allow to easily make a selection based on its knowledge.

It's true it's not possible via a MAP, but it is possible with a webhook.
Regarding the "Kubevirt knows more" argument, if you have the VMI at hand, what does Kubevirt know that you don't?

> Also, in that case it is not about deploying external resources, but just a single virt-launcher.

By "external resources" I mean the mutating webhook that mutates the pod to use an alternative image.

> We're just looking for a way to tell KubeVirt which image it should deploy.

I still fail to understand why this use-case of image selection is fundamentally different than wanting to change resources / capabilities / volumes or anything else on the virt-launcher level depending on the VMI definition (or any other object).

lyarwood:
> > We're just looking for a way to tell KubeVirt which image it should deploy.
> 
> I still fail to understand why this use-case of image selection is fundamentally different than wanting to change resources / capabilities / volumes or anything else on the virt-launcher level depending on the VMI definition (or any other object).

I don't think they are fundamentally different, I would argue however that launcher image mutations will be more common and as such would benefit from being natively and declaratively handled by Plugins. The cost is tiny for KubeVirt but the saving for Plugin authors and users compared to the on-going cost of the MAPs/Webhook alternative is huge IMHO.

iholder101:
> I would argue however that launcher image mutations will be more common

Maybe you're right, maybe not. I'm honestly not sure at all that resource management for example would be less common. TBH selecting a custom image for virt-launcher sounds like a pretty niche use-case to me.

> The cost is tiny for KubeVirt but the saving for Plugin authors and users compared to the on-going cost of the MAPs/Webhook alternative is huge IMHO.

I agree that it's smaller for Kubevirt, but if Kubevirt has to pay this smaller price for many use-cases it's no longer small. Therefore, to me the main question is the one above - whether image selection is truly a special use-case.

BTW - I suggest moving this discussion to the tracking issue, as it's a bit unrelated to the scope in this PR :)

lyarwood:
> BTW - I suggest moving this discussion to the tracking issue, as it's a bit unrelated to the scope in this PR :)

Yeahh no issues, I've drafted https://github.com/kubevirt/enhancements/pull/436 to spare the VEP issue the same fate as this PR :D

vladikr:
Looks good!
Thank you.
/approve

michalskrivanek:
@iholder101 (anc cc @dankenigsberg since you're interesting in pluggability), FWIW I still find it very difficult to imagine a use case where the PreMigrationSource hook would be actually sensible. To me the most problematic aspect is introduction of fairly complex mechanism into an already time- and logic-sensitive flow.
Even the examples you cited need corresponding support elsewhere in the code (e.g. in monitoring).
I think it might be a little bit more capable if the sidecar had access to the whole MigrationOptions, and even more pluggable if the migration,stall handling, etc. would all move out. But that capability would come with such a high price in complexity and brittleness that it is absolutely not worth it.

iholder101:
Hey @michalskrivanek, thanks for these thoughts.

> FWIW I still find it very difficult to imagine a use case where the PreMigrationSource hook would be actually sensible.

Tuning the migration config - which isn't possible today since migration parameters aren't part of the domain XML. Multifd is a good example: it's currently a hardcoded amount of threads (8) with no user-facing knob. A PreMigrationSource hook would be the first way to tune it up for throughput or disable it (e.g. for more reliable functional tests) per-migration - the flags/params are fully built right before `MigrateToURI3`, so a hook can adjust them cleanly.

> To me the most problematic aspect is introduction of fairly complex mechanism into an already time- and logic-sensitive flow. Even the examples you cited need corresponding support elsewhere in the code (e.g. in monitoring).

I think the right answer here is to carefully identify more useful hook points in the future which will lead to the plugin author having a higher amount of control. I agree that monitoring could be an area to further explore.

> I think it might be a little bit more capable if the sidecar had access to the whole MigrationOptions, and even more pluggable if the migration, stall handling, etc. would all move out.

AFAICT exposing the whole `MigrationOptions` would give the plugin less control, not more, as it's a higher-level struct that gets translated into the libvirt flags/params, so a plugin mutating it is one abstraction removed from the actual knobs. PreMigrationSource deliberately exposes the low-level flags/params passed to `MigrateToURI3` instead, which is where the real control is. What am I missing here?

> But that capability would come with such a high price in complexity and brittleness that it is absolutely not worth it

This is a recurring discussion we've been having for a while now :)
I'd just say that it is always easier for a feature developer to push the logic into the core. But as a project, the price of accepting these all into the core is high as well. Balancing between those is indeed not a straight forward task, but my personal feeling is that Kubevirt's current tendency is accepting too much into the core TBH.

michalskrivanek:
> Hey @michalskrivanek, thanks for these thoughts.
> 
> > FWIW I still find it very difficult to imagine a use case where the PreMigrationSource hook would be actually sensible.
> 
> Tuning the migration config - which isn't possible today since migration parameters aren't part of the domain XML. Multifd is a good example: it's currently a hardcoded amount of threads (8) with no user-facing knob. A PreMigrationSource hook would be the first way to tune it up for throughput or disable it (e.g. for more reliable functional tests) per-migration - the flags/params are fully built right before `MigrateToURI3`, so a hook can adjust them cleanly.

yes, but a clean API knob would have been a lot easier to use. There's logic behind the selection of multifd on or off, a useful one, so the plugin would need to implement it too. unnecessarily. 

> > To me the most problematic aspect is introduction of fairly complex mechanism into an already time- and logic-sensitive flow. Even the examples you cited need corresponding support elsewhere in the code (e.g. in monitoring).
> 
> I think the right answer here is to carefully identify more useful hook points in the future which will lead to the plugin author having a higher amount of control. I agree that monitoring could be an area to further explore.

+1
though I still believe core features would greatly benefit from proper API

> > I think it might be a little bit more capable if the sidecar had access to the whole MigrationOptions, and even more pluggable if the migration, stall handling, etc. would all move out.
> 
> AFAICT exposing the whole `MigrationOptions` would give the plugin less control, not more, as it's a higher-level struct that gets translated into the libvirt flags/params, so a plugin mutating it is one abstraction removed from the actual knobs. PreMigrationSource deliberately exposes the low-level flags/params passed to `MigrateToURI3` instead, which is where the real control is. What am I missing here?

it would allow the plugin to control the other non-xml parameters that affect the monitoring/convergence logic behavior.
 
> > But that capability would come with such a high price in complexity and brittleness that it is absolutely not worth it
> 
> This is a recurring discussion we've been having for a while now :) I'd just say that it is always easier for a feature developer to push the logic into the core. But as a project, the price of accepting these all into the core is high as well. Balancing between those is indeed not a straight forward task, but my personal feeling is that Kubevirt's current tendency is accepting too much into the core TBH.

I still believe live migration is a very much core feature of a virtualization project. It's also been discussed that kubevirt is pretty bad at that compared to competition. Hiding things from API and/or making them awkward or unsupported to use  makes it more difficult for the out-of-tree/out-of-core implementation.
I do get the dilemma you're facing though, I guess that's the price of being popular...:)

and apologies for commenting on closed issue!

Issue #416, opened 2026-08-10
VEP 416 Auto provision backup volumes PVC (follow up of VEP 25)


awels:
**Primary contact (assignee)**:

/assign awels

**Current Feature Stage**:
alpha

**Feature Gate**:
None

**Responsible SIGs**:

Primary SIG:
/sig storage

**Enhancement link**:

**Timeline**:

* Alpha release target (1.10):
* Beta release target (x.y):
* Stable release target (x.y):

* [X] Alpha
  * [X] VEP PRs:
    * https://github.com/kubevirt/enhancements/pull/417
  * [X] Code PRs:
    * https://github.com/kubevirt/kubevirt/pull/18750
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

awels:
@iholder101 I updated the title to match your suggestion.

PR #417, opened 2026-08-10
Auto-provision backup target PVCs when omitted

awels:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/416
**SIG label**: /sig-storage

### What this PR does
This VEP describes the extra field in the VirtualMachineBackup status that shows which PVC was used for the backup.

### DIFF
--- a/veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
+++ b/veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
@@ -1,0 +1,482 @@
+# VEP #25 Addendum: Auto-provision VirtualMachineBackup target PVCs
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements]: https://github.com/kubevirt/enhancements/issues/416
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This addendum extends [VEP #25](./vep.md) so that `VirtualMachineBackup.spec.pvcName` is optional.
+When omitted, the VMBackup controller creates a filesystem RWO target PVC and records the resolved
+name on `VirtualMachineBackup.status.pvcName`. The same PVC name may also be recorded on the
+associated `VirtualMachineBackupTracker` checkpoint for vendor metadata—primarily when the caller
+supplied the target PVC and it outlives the backup CR.
+
+## Motivation
+
+VEP #25 requires callers to pre-create a filesystem PVC and pass its name via `spec.pvcName` for
+both push and pull mode. That forces backup vendors and operators to:
+
+1. Discover CBT-eligible disks and estimate required capacity before creating the backup CR.
+2. Choose a storage class compatible with hot-plugging into the virt-launcher pod.
+3. Create a PVC using the storage class and size to hold the backup.
+
+The parent VEP already states that for push mode "the PVC name containing the backup is provided in
+the `VirtualMachineBackup` CR status", but the Status API never defined that field, and `spec.pvcName`
+remained required. Making the PVC optional and exposing the resolved name in status closes that gap
+and removes boilerplate for the common case where KubeVirt can size and provision the target itself.
+
+## Goals
+
+- Make `VirtualMachineBackup.spec.pvcName` optional.
+- When `spec.pvcName` is omitted, auto-create a filesystem RWO PVC owned by the backup.
+- Optional `targetPvcSizePercent` on the backup spec to tune auto-provisioned target size (default 120).
+- Record the PVC actually used for the backup on `VirtualMachineBackup.status.pvcName`.
+- Persist the target PVC name on `VirtualMachineBackupTracker.status.latestCheckpoint.pvcName`
+  when a tracker is updated after a successful backup (see [Tracker checkpoint association](#tracker-checkpoint-association)).
+- Preserve the existing path where callers supply their own PVC via `spec.pvcName`.
+
+## Non Goals
+
+- Changing backup modes (push/pull), quiescing, export endpoints, or restore APIs.
+- Auto-provisioning source (guest) disks or VM state/CBT overlay PVCs.
+- Choosing arbitrary storage classes unrelated to the VM's disks (the class is copied from
+  CBT-eligible source PVCs or their bound PVs; callers that need a different class supply
+  `spec.pvcName`).
+- Garbage-collecting user-provided PVCs after backup completion (ownership and lifecycle of
+  caller-supplied PVCs remain the caller's responsibility).
+- Introducing a new feature gate; this rides on the existing incremental backup / CBT feature.
+
+## Definition of Users
+
+* Backup vendors integrating with `VirtualMachineBackup`
+* Cluster admins scripting cluster-wide backups
+* VM owners taking ad-hoc backups without managing scratch/target PVCs
+
+## User Stories
+
+* As a backup vendor, I want to create a `VirtualMachineBackup` without pre-creating a target PVC,
+  and then read `status.pvcName` to find where the backup data landed.
+* As a backup vendor using a tracker with my own target PVC, I want each checkpoint to record which
+  PVC held that backup so my catalog can correlate checkpoint metadata with durable storage even
+  after the `VirtualMachineBackup` CR is deleted.
+* As a cluster admin, I still want to supply my own PVC via `spec.pvcName` when I need a specific
+  storage class, size, or retention policy.
+
+## Repos
+
+[KubeVirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### Resolving the target PVC name
+
+The controller resolves the backup target PVC as follows:
+
+1. If `spec.pvcName` is set and non-empty, use that name (user-provided; no auto-create).
+2. Else if `status.pvcName` is already set, reuse that name (idempotent reconcile after create).
+3. Else derive a deterministic name from the backup name with suffix `backup-target-pvc`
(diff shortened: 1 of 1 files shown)


veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Auto-created PVC shape
+
+When auto-creating, the controller creates a PVC with:
+
+| Property | Value |
+| --- | --- |
+| `volumeMode` | Filesystem |
+| `accessModes` | `ReadWriteOnce` |
+| Size | Sum of CBT-eligible source PVC sizes (using CDI's `min(request, capacity)` rule), plus 20% filesystem overhead |
```
Acedus: What about size overrides? 20% filesystem overhead might be a bit too aggressive. I know that we're keeping the ability to specify an existing PVC, but I wonder if we could unify the fields into a single "target pvc" configuration spec (i.e., can specify PVC by name or its specification like size, SC, etc.).

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
```diff
@@ -0,0 +1,314 @@
...
+
+When `spec.pvcName` is set, behavior matches VEP #25 today: the controller waits for the PVC to
+exist, rejects block-mode PVCs, and does not create or own the PVC. `status.pvcName` is still set
+to the resolved (spec) name so consumers have a single place to read it.
+
+### Tracker checkpoint association
+
+On successful backup completion, when updating
+`VirtualMachineBackupTracker.status.latestCheckpoint`, the controller sets `pvcName` from
+`backup.spec.pvcName` if present, otherwise from `backup.status.pvcName`. This keeps the PVC
```
Acedus: Why is this necessary?

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
```diff
@@ -0,0 +1,314 @@
...
+   user-provided PVCs and can serve the default path directly.
+
+2. **Mutate `spec.pvcName` after create**  
+   Would make the resolved name visible on spec, but the backup spec is immutable
+   (`self == oldSelf`). Status is the correct place for controller-resolved values.
+
+3. **Separate CRD for backup target storage**  
+   Overkill for a single PVC per backup; ownership from the backup CR is sufficient.
+
+4. **Always require a StorageClass / size on the backup spec**
```
Acedus: I don't think it duplicates the PVC API surface, we can simply embed the PVC spec there (or parts of it). I'd say it's a pretty viable alternative all things considered.

Acedus (commented):
Thanks @awels, looks good, just a few questions.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
awels: Basically if you specify the name of the PVC, we will use that one. If not we will create it based on what is described here. I don't want to give more knobs to be honest. Maybe the percentage of the combined size? Default it to 120% (which would match what is described here), but allow people to override it. I don't want to do much more than that.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
awels: This is for incremental backups, the tracker is durable but we are expecting the VMBackup to be removed after completion, and we want to keep track of the PVCs containing the data.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
awels: I don't want to give more knobs than absolutely needed. I think the current approach (with the extra sizing knob if needed) should provide enough flexibility without providing a surface to create PVCs at will.

Also if we embed the PVC or parts of it, we will have to deal with mutability of those fields in a webhook or something similar.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
Acedus: Yes but what reason is there for tracking PVCs containing the data? For push mode I'd argue that once the backup completes it's the responsibility of the consumer to read the VMBackup status and interact with the PVC that was provisioned for the backup before deleting it, and for pull mode there's no need for any of that (it's only there for scratch space to facilitate the backup).

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
Acedus: I think most consumers would rather have more granular control over the PVC that we provision internally rather than creating that PVC with the options they need themselves, but we can keep the API lean for now and extend it as demand grows.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
```diff
@@ -0,0 +1,314 @@
...
+
+When auto-creating, the controller creates a PVC with:
+
+| Property | Value |
+| --- | --- |
+| `volumeMode` | Filesystem |
+| `accessModes` | `ReadWriteOnce` |
+| Size | Sum of CBT-eligible source PVC sizes (using CDI's `min(request, capacity)` rule), plus 20% filesystem overhead |
+| Storage class | Storage class of the boot-candidate CBT disk (lowest `bootOrder` among CBT disks, else first CBT-eligible volume). If the class cannot be determined, the backup fails and the caller must supply `spec.pvcName`. |
+| Labels | `backup.kubevirt.io/auto-created: "true"` |
```
alromeros, line 106: Nit: If we don't want to support size overrides it should also include `"cdi.kubevirt.io/applyStorageProfile"` to avoid minimum size provisioning errors.

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
alromeros: I don't have strong preferences here, but automatic size and sc inference has proven faulty in the past. For a temp pvc this might not be a big deal, but I'm fine with embedding the spec here.

awels: If we are going to make the user specify the PVC spec here, what is the point of this VEP? The entire ask is to not burden the user with the task of creating the backup PVC. But if we require them to create the PVC spec instead that doesn't help them any except that they don't have to manage the life cycle of the PVC, but they are still required to provide everything else. If we go that route might as well not do this VEP at all. So I am going to be pretty adamant that we don't ask the user to provide a PVC spec. 

I am okay with an optional size percentage based on what we would normally create. We have a good algorithm for deciding the storage class already which is based on what we learned from previous attempts to determine the right storage class.

alromeros (approved):

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
Acedus: > If we are going to make the user specify the PVC spec here, what is the point of this VEP? The entire ask is to not burden the user with the task of creating the backup PVC. 

Right, but I don't think this is us asking the user to specify the PVC spec, it'd be optional and they would have more granular control over that PVC should they require it. At that point one might say that they should just provision their own in advance and specify the name as it is now, but I think it's inconvenient for the same reasons that got this VEP conceived in the first place. 

Having said that, I don't mind deferring a proper on this to beta should we receive feedback that people want it.

alromeros:
@awels VEP looks good to me, but there are a couple of changes from another VEP update in progress that should be cleaned up. I'll lgtm once that's done.

alromeros:
Looks good!
/lgtm

awels:
@iholder101 Could you review/approve? I have lgtms from dedicated reviewers

vladikr:
Hey,  Thanks for this work.. I was trying to follow, but maybe I didn't fully grasp everything...
Before merging this I would like the VEP to specify:


Can you please explain the default targetPvcSizePercent in the pull and push cases? 
Reading the VEP, it looks to me that a VM with two 100 Gi disks gets a 240 Gi target, every time, whether this backup is a full push, an incremental push, or pull-mode.
Do I understand this correctly?
what happens if the target is too small?
why not to distinguish between an incremental and full backups? I mean, in that case the target would grow only by that incremental change with some overhead, no?
and in the pull mode, do you need all of that space at all?!


I have some difficulties around the storage class inference. 
What happens if we didn't find a Storage Class that supports Filesystem? does it fail with that reason?
Also, how do we make sure that the backup PVC can actually be attached on the node where the virt-launcher is running? I mean, the VEP says nothing about this. 

For example, the VEP says: create a PVC, then wait until observed and usable, then attach as a utility volume  and run backup.
But this “Usable” is undefined. What if a PVC can be Bound and still not be unusable as a utility volume?
Do we write a terminal reason?

vladikr:
I also don't fully understand the retry logic...
even if can the user pvc object from the status.pvcName - how do we know what's on it already is useable? 
is there an intent to wipe it? or reuse the files?

awels:
@vladikr I believe I addressed all your concerns, although I admit I am not 100% sure I understood them all

veps/sig-storage/25-incremental-backup/auto_provision_pvc.md
```diff
@@ -0,0 +1,470 @@
...
+
+The controller resolves the target name from `spec.pvcName`, else reuses `status.pvcName`, else
+derives a new name from the backup name. That idempotency is for reconcile safety, not for sharing
+PVCs across different backup CRs (each backup gets a unique derived name).
+
+If a PVC with the resolved name already exists and is owned by this backup, the controller reuses
+it **without wiping** existing content. A new auto-provisioned PVC is empty; retries on the same
+backup CR after a partial failure may reuse a PVC that already contains data from the failed run.
+Callers should not assume a clean filesystem on retry unless they delete the backup CR (which
+deletes an auto-created PVC via owner reference) and create a new one.
```
vladikr: This still sounds like a Failed backup can be started again on the same CR. 
Offline you said Failed is terminal, the auto-created PVC contents are discarded, and the recovery would be something like the following (IIUC) vendor deletes the VirtualMachineBackup and creates a new CR (new PVC).
I just want some statement to be added that `Failed` is terminal and the controller will not start again on that CR, so leftover files are only a problem if someone expects an in-place retry.

vladikr:
Looks good to me now. Thanks @awels !
/approve

Issue #419, opened 2026-08-14
VEP 419: Configurable external certificate strategy


awels:
**Primary contact (assignee)**:

/assign awels

**Current Feature Stage**:

VMExport is GA, this is an extension of VMExport

The feature it

**Feature Gate**:

None, this VEP adds a field to KubeVirt CR to indicate to the export controller if we want to use the kube root CA or not.

**Responsible SIGs**:

Primary SIG:
/sig storage

Additional SIGs (optional):
/sig compute

**Enhancement link**:

**Timeline**:

* [X] GA
  * [X] VEP PRs:
    * https://github.com/kubevirt/enhancements/pull/420
  * [X] Code PRs:
    * https://github.com/kubevirt/kubevirt/pull/18652
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #420, opened 2026-08-14
VMExport externalink certificate strategy

awels:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/pull/419
**SIG label**: /sig storage

### What this PR does
This VEP describes the new field in the KubeVirt CR that allows the user to configure the strategy for the certificate used for the external links in the export proxy.

The default strategy is to use the same certificate as currently used. The other options will be
- CustomRootCA: A custom root CA certificate that will be used to sign the certificates for the
external links.
- SystemTrust: Leaves the certificate empty having the clients use the system trust store.
- ClusterRootCA: The default and current behavior.

### DIFF
--- a/veps/sig-storage/419-configurable-cert-strategy/vep.md
+++ b/veps/sig-storage/419-configurable-cert-strategy/vep.md
@@ -1,0 +1,442 @@
+# VEP #419: Configurable VMExport External Certification Strategy
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
+- This VEP targets beta for version:
+- This VEP targets GA for version: v1.10
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR): https://github.com/kubevirt/enhancements/issues/419
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+`VirtualMachineExport` publishes external download links in
+`status.links.external`. Each link includes a PEM-encoded CA certificate in the
+`cert` field so clients can validate TLS connections to the export endpoint
+through the cluster ingress or route.
+
+Today, the export controller always populates that certificate from cluster
+trust material (`kube-root-ca.crt` for OpenShift Routes, or the Ingress TLS
+secret for Kubernetes Ingress). That works on many clusters, but it is not
+always correct: `kube-root-ca` makes no guarantee that the ingress or route
+actually presents a certificate signed by that CA. On hosted platforms where
+the edge certificate is signed by a public CA, publishing the cluster root CA
+is misleading and can cause client verification failures.
+
+This VEP adds cluster-level configuration on the `KubeVirt` CR so
+administrators can choose how the `cert` field in external export links is
+populated.
+
+## Motivation
+
+VMExport consumers (for example `virtctl vmexport download` and backup
+integrations) rely on `status.links.external[].cert` to build a trust chain when
+connecting to the export proxy through the cluster edge.
+
+The current behavior assumes the edge certificate chains to the cluster root CA
+(or can be read directly from the Ingress TLS secret). That assumption breaks
+down in common deployment models:
+
+- **Hosted / managed Kubernetes** where ingress certificates are signed by a
+  public CA and clients should use their system trust store instead of a
+  cluster-local root.
+- **Custom private CAs** where the ingress or route certificate is signed by a
+  CA that is not present in `kube-root-ca.crt` or the Ingress TLS secret, but
+  is known to the platform administrator.
+
+There is no reliable way for the export controller to infer the correct mode
+from in-cluster state alone. Intermediate and leaf CAs, split-horizon DNS, and
+platform-specific ingress configuration all make automatic detection unsafe.
+Administrators therefore need an explicit, cluster-level knob.
+
+## Goals
+
+- Add `exportConfiguration.externalCertificationStrategy` to the `KubeVirt` CR
+  with three supported values:
+  - `ClusterRootCA` (default): preserve today's behavior.
+  - `CustomRootCA`: publish certificates from a dedicated ConfigMap.
+  - `SystemTrust`: leave the external `cert` field empty so clients use their
+    system trust store (see [Naming: `SystemTrust` vs `None`](#naming-systemtrust-vs-none)).
+- Apply the selected strategy consistently for both OpenShift Route and
+  Kubernetes Ingress external link generation.
+- Surface a `KubeVirt` status condition when `CustomRootCA` is selected but
+  the required ConfigMap is missing or empty.
+- Reconcile existing `VirtualMachineExport` objects when the strategy or
+  backing ConfigMaps change.
+
+## Non Goals
+
+- Automatically detecting the correct strategy by comparing hostnames and CAs.
+  This does not account for intermediate or leaf CAs and can select the wrong
+  trust material.
+- Probing external endpoints from the export controller to discover presented
+  certificates. The controller must not reach out to unknown external endpoints.
+- Implicit strategy selection based on ConfigMap presence. Presence-based
+  behavior is hard to discover and troubleshoot; strategy selection must be
+  explicit.
+- OpenShift-specific APIs or Route status fields. The design must work on plain
+  Kubernetes with Ingress as well as OpenShift with Routes.
+- Changing internal export link certificates. This VEP only affects external
+  links exposed through the cluster ingress or route.
+
+## Definition of Users
+
+- **Cluster administrators**: Configure how external export link certificates
+  are published for their ingress or route setup.
+- **Meta-operator developers** (for example HCO on OpenShift Virtualization):
+  Expose the strategy through the platform operator and populate the KubeVirt CR
+  on behalf of administrators.
+- **VM owners and backup vendors**: Consume `status.links.external` and depend
+  on accurate CA material when downloading exports through the cluster edge.
+
+## User Stories
+
+### Story 1: Public-CA signed ingress on a hosted cluster
+
+As a cluster administrator running KubeVirt on a hosted platform whose ingress
+is signed by a public CA, I want external export links to omit the `cert` field
+so clients verify the endpoint using their system trust store.
+
+### Story 2: Private CA not in kube-root-ca
+
+As a cluster administrator whose ingress or route uses a private CA that is not
+present in `kube-root-ca.crt`, I want to provide that CA in a ConfigMap and
+have VMExport external links publish it.
+
+### Story 3: Preserve existing behavior
+
+As a cluster administrator on a cluster where `kube-root-ca` or the Ingress TLS
+secret already provides the correct trust material, I want the default behavior
+to remain unchanged when I do not configure anything.
+
+## Repos
+
+- [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt)
+
+## Design
+
+### API Change
+
+Add `ExportConfiguration` to `KubeVirtConfiguration`:
+
+```go
+type ExportConfiguration struct {
+    // ExternalCertificationStrategy controls how VirtualMachineExport external
+    // link certificates are populated.
+    // Defaults to ClusterRootCA.
+    // +optional
+    // +kubebuilder:validation:Enum=ClusterRootCA;CustomRootCA;SystemTrust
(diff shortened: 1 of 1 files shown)


veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+
+- This VEP targets GA for version: 1.10
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+approved
+- [ ] (R) GA target version is explicitly mentioned and approved
```
0xFelix: Keep placeholders from the template?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+Administrators therefore need an explicit, cluster-level knob.
+
+## Goals
+
+- Add `exportConfiguration.externalCertificationStrategy` to the `KubeVirt` CR
+  with three supported values:
+  - `ClusterRootCA` (default): preserve today's behavior.
+  - `CustomRootCA`: publish certificates from a dedicated ConfigMap.
+  - `SystemTrust`: leave the external `cert` field empty so clients use their
+    system trust store.
```
0xFelix: Would naming this `None` make it more obvious that no CA is going to be published?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+- Apply the selected strategy consistently for both OpenShift Route and
+  Kubernetes Ingress external link generation.
+- Surface a `KubeVirt` status condition when `CustomRootCA` is selected but
+  the required ConfigMap is missing or empty.
```
0xFelix, line 72: Would that also lead to the KV CR becoming not available? Or would it just lead to newly created VMEs to fail immediately?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+
+The field lives under `spec.configuration.exportConfiguration` so additional
+export-related cluster settings can be grouped here in the future.
+
+### Strategy Semantics
+
+| Strategy | Route source | Ingress source | `status.links.external.cert` |
+|----------|--------------|----------------|------------------------------|
+| `ClusterRootCA` (default) | `kube-root-ca.crt` ConfigMap (`ca.crt` key) | Ingress TLS secret (`tls.crt`) | Matching PEM chain, or best-effort chain from available certs |
+| `CustomRootCA` | `kubevirt-export-external-ca` ConfigMap (`ca.crt` key) | same ConfigMap | PEM chain from ConfigMap; empty if ConfigMap missing or empty |
```
0xFelix, line 158: Empty chain would imply that `SystemTrust` was set?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+`kubevirt-export-external-ca` ConfigMap is missing or its `ca.crt` entry is
+empty, virt-controller sets a `KubeVirt` condition:
+
+- **Type**: `ExportCustomExternalCA`
+- **Status**: `False`
+- **Reason**: `ExportExternalCAConfigMapMissing`
+
+The condition is removed when the ConfigMap becomes available or when the
+strategy is changed away from `CustomRootCA`. While misconfigured, external
+export link certificates remain blank.
```
0xFelix: Would it be worth to surface this error condition also on the exports?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Controller Reconciliation
+
+The VMExport controller in virt-controller:
+
+1. Reads the active strategy from cluster configuration via the KubeVirt CR.
+2. Watches the KubeVirt CR, `kube-root-ca.crt`, and
+   `kubevirt-export-external-ca` for changes.
+3. Reconciles the `ExportCustomExternalCA` condition on KubeVirt status.
```
0xFelix: Will the condition be present and set to `True` if everything is fine, or will it be missing instead?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,390 @@
+# VEP #419: Configurable VMExport External Certification Strategy
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets GA for version: 1.10
```
0xFelix: The question is, can we GA it immediately?

awels: There is no feature gate for this, so we do it and it is GA immediately or we don't do it.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: None would be clearer about not publishing a CA, but SystemTrust describes the intended client behavior. An empty cert also happens with misconfigured CustomRootCA, so None as a strategy name would still be ambiguous.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Fixed

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Neither. KubeVirt stays fully operational. Exports still reconcile and external links are published, they just have an empty cert until the ConfigMap is fixed. The condition is informational for cluster admins, not a blocker.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: No, CustomRootCA with a missing ConfigMap also produces an empty cert. Added a note that consumers should check the KubeVirt CR strategy field or the ExportCustomExternalCA condition to distinguish the two.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Intentionally only on the KubeVirt CR. This is a cluster-level config, not per-export. Duplicating it on every VMExport would be noisy. Exports already reflect the effect via empty cert on external links.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Omitted when healthy. Only present as False when CustomRootCA is selected but the ConfigMap is missing/empty. No True variant.

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,436 @@
...
+    -----END CERTIFICATE-----
+```
+
+The `ca.crt` key matches the key used by `kube-root-ca.crt`.
+
+### Status Condition
+
+When `externalCertificationStrategy` is `CustomRootCA` and the
+`kubevirt-export-external-ca` ConfigMap is missing or its `ca.crt` entry is
+empty, virt-controller sets a `KubeVirt` condition:
```
0xFelix: `virt-controller` does not manage the KubeVirt CR. `virt-operator` is responsible for it?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,436 @@
+# VEP #419: Configurable VMExport External Certification Strategy
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
```
0xFelix: Drop version here

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,436 @@
...
+
+## Implementation History
+
+- 2026-07-30: Initial implementation. PR: [kubevirt/kubevirt#18652](https://github.com/kubevirt/kubevirt/pull/18652)
+
+## Graduation Requirements
+
+This enhancement is a small, backward-compatible API addition: the default
+strategy preserves today's behavior, no feature gate is required, and callers
+opt in by setting the field. Alpha and beta may be abbreviated or skipped if
```
0xFelix, line 416: That means it goes straight to GA?

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,436 @@
...
+`None` would make it obvious that no CA is published in the link. `SystemTrust`
+was chosen instead because it describes the intended client behavior (verify
+using the system trust store) rather than only the absence of embedded CA
+material. An empty `cert` is also produced by misconfigured `CustomRootCA`, so
+`None` would be ambiguous as a strategy name. The active strategy is always
+read from the KubeVirt CR configuration field.
+
+### Match hostname and CA automatically
+
+Inspect available CAs and only publish a certificate when it matches the export
```
0xFelix, line 331: It still matches hostnames in strategies `ClusterRootCA` and `CustomRootCA`, right? See L161

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Right, there is no real template for something that doesn't really have alpha/beta/GA. So I am not sure how to represent this. I tried to shoehorn it into the existing process, but it feels awkward.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Good point, let me fix that.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: So basically L161 applies to the 'current' implementation, when using the kube-root-ca. The alternative described here is using just that to determine if we should publish a CA at all or not. Right now if there is no match, it will publish the entire CA bundle (as describe on L161). The thought was, what if we don't find a match, don't publish any CA. This works in the case of a cloud provider (which uses a public CA), but it fails for all other situations (where we self sign the Route/Ingress). This is calling out that was considered and rejected because it could break existing clusters that are not using public CAs

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: Done

veps/sig-storage/419-configurable-cert-strategy/vep.md
0xFelix: ACK

veps/sig-storage/419-configurable-cert-strategy/vep.md
```diff
@@ -0,0 +1,439 @@
...
+unchanged aside from the empty `cert` on external links.
+
+### Controller Reconciliation
+
+The VMExport controller in virt-controller:
+
+1. Reads the active strategy from cluster configuration via the KubeVirt CR.
+2. Watches the KubeVirt CR, `kube-root-ca.crt`, and
+   `kubevirt-export-external-ca` for changes.
+3. Reconciles the `ExportCustomExternalCA` condition on KubeVirt status.
```
0xFelix: This still sits underneath `virt-controller`, but has to be handled by `virt-handler` too.

veps/sig-storage/419-configurable-cert-strategy/vep.md
0xFelix: Can we make it more obvious to consumers without having to check the KubeVirt CR too? I'm thinking of virtctl for example. Go differentiates between empty pool and `nil` pool.

veps/sig-storage/419-configurable-cert-strategy/vep.md
awels: I am assuming you mean virt-operator, in which case you are correct, let me fix that.

0xFelix: Yeah, `virt-operator` of course.

veps/sig-storage/419-configurable-cert-strategy/vep.md
0xFelix: On the other hand, an empty string should be enough to infer from that there's nothing else to trust apart from the system trust store, so let's go ahead.

0xFelix (approved):
/lgtm

Thanks!

awels:
@iholder101 @vladikr could one of you approve?

awels:
@xpivarc could you review/approve? It has lgtm from the dedicated reviewer already

0xFelix:
@xpivarc @vladikr @iholder101 Ping

PR #421, opened 2026-08-18
Update Vep 172: Planned work for 1.10

jschintag:
### VEP Metadata

**Tracking issue**: #172 
**SIG label**: /sig compute

### What this PR does
Drop software emulation binary distribution, as it is not possible with preview qemu binaries.

Add new setting to allow users to choose emulation tier.

Enhance the e2e test strategy.

### DIFF
--- a/veps/sig-compute/172-cross-arch-virtualization/vep.md
+++ b/veps/sig-compute/172-cross-arch-virtualization/vep.md
@@ -77,6 +77,9 @@ Cross-architecture emulation would enable:
 - Set VMI conditions that distinguish between hardware-accelerated and
   software-emulated cross-architecture execution
 - Maintain backward compatibility with existing emulation behavior
+- Allow cluster administrators and users to configure a maximum emulation tier
+  so that emulation remains explicitly opt-in and workloads
+  are never silently placed on a slower emulation tier than intended
 
 ## Non Goals
 
@@ -122,6 +125,16 @@ This feature is intended for:
    distinguish between hardware-accelerated and software-emulated
    cross-architecture VMs so I can monitor my cluster's performance
    characteristics.
+
+7. As a cluster administrator, I want to set a cluster-wide policy that
+   cross-architecture VMs **must** use hardware acceleration, so that no VM
+   silently falls back to slow software emulation without my knowledge.
+
+8. As a user, I do not want my VM to use emulation at all, since I need maximum
+   performance.
+
+9. As a user, I want to test my new app on a different architecture, for which I
+   need software emulation, even if the cluster policy is set to hardware-only.
 
 ## Repos
 
@@ -553,25 +566,74 @@ scenarios.
 
 The feature gate behavior is as follows:
 
-| Feature Gate Enabled | useEmulation | Host Arch | Guest Arch | KVM Cross-Arch | Result |
-|---------------------|--------------|-----------|------------|----------------|--------|
-| No | false | amd64 | amd64 | N/A | Native KVM execution |
-| No | true | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
-| No | false | amd64 | arm64 | N/A | **ERROR** - cross-arch not allowed |
-| Yes | false | amd64 | arm64 | No | Cross-arch software emulation (TCG) |
-| Yes | false | s390x | arm64 | Yes | Cross-arch KVM-accelerated virtualization (SAE) |
-| Yes | false | s390x | arm64 | No | Cross-arch software emulation (TCG fallback) |
-| Yes | false | amd64 | amd64 | N/A | Native KVM execution |
-| Yes | true | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
+| Feature Gate Enabled | useEmulation | emulationPolicy | Host Arch | Guest Arch | KVM Cross-Arch | Result |
+|---------------------|--------------|-----------------|-----------|------------|----------------|--------|
+| No | false | — | amd64 | amd64 | N/A | Native KVM execution |
+| No | true | — | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
+| No | false | — | amd64 | arm64 | N/A | **ERROR** - cross-arch not allowed |
+| Yes | false | None (default) | amd64 | arm64 | No | **ERROR** - emulation not permitted |
+| Yes | false | Software | amd64 | arm64 | No | Cross-arch software emulation (TCG) |
+| Yes | false | Hardware | amd64 | arm64 | No | **ERROR** - no hardware cross-arch KVM available |
+| Yes | false | Hardware | s390x | arm64 | Yes | Cross-arch KVM-accelerated virtualization (SAE) |
+| Yes | false | Software | s390x | arm64 | Yes | Cross-arch KVM-accelerated virtualization (SAE, preferred) |
+| Yes | false | Software | s390x | arm64 | No | Cross-arch software emulation (TCG fallback) |
+| Yes | false | None (default) | amd64 | amd64 | N/A | Native KVM execution |
+| Yes | true | — | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
 
 **Key Points**:
 
 - The feature gate alone is sufficient for cross-arch emulation — `useEmulation` is **not** required
 - `useEmulation` only controls same-arch emulation when KVM is unavailable
+- `emulationPolicy` controls the **maximum fallback depth** for cross-arch emulation and defaults to `None` (native KVM only)
 - Native KVM VMs and cross-arch emulated VMs can coexist on the same cluster
 - When the feature gate is disabled, attempting to run a cross-arch VM returns an error at domain conversion time
-- When the feature gate is enabled, the scheduler **prefers** nodes matching the guest architecture (native execution with KVM) and only falls back to cross-arch emulation when no matching nodes are available
-- When hardware acceleration is available, it is automatically preferred over software emulation
+- When the feature gate is enabled but `emulationPolicy` is `None` (default), cross-arch VMs are rejected — emulation must be explicitly opted in
+- When hardware acceleration is available, it is automatically preferred over software emulation (when the policy permits both)
+
+### Emulation Policy
+
+To ensure that software emulation remains an explicit opt-in feature and is
+never silently imposed on a workload, a new `emulationPolicy` field is
+introduced in both the KubeVirt CR configuration and the VM/VMI spec.
+
+The field controls **which virtualization backends are permitted** for a
+cross-architecture VM. The available tiers, from most restricted to most
+permissive, are:
+
+```
+None (native KVM only)  <  Hardware (KVM only, no TCG)  <  Software (all tiers)
+```
+
+Each value defines the **maximum fallback depth** allowed:
+
+- **`None`** (default) — no emulation permitted. Only native KVM is allowed.
+  A cross-arch VM with this setting will not start via any form of emulation —
+  it must land on a node that natively runs the guest architecture. This is the
+  safe default: enabling the feature gate does not automatically cause VMs to
+  run under emulation unless explicitly opted in.
+- **`Hardware`** — hardware cross-arch KVM is permitted, but TCG software
+  emulation is not. The VM may use native KVM or hardware cross-arch KVM.
+  It will **not** fall back to software emulation and fails to start on nodes
+  where only software emulation is available.
+  > **Alpha restriction**: `Hardware` tier is not yet implemented. Setting it
+  > will cause the validation webhook to reject the change with:
+  > `Error: Hardware emulation is not yet implemented`.
+- **`Software`** — the full fallback chain is permitted: native KVM > hardware
+  cross-arch KVM > software emulation (TCG). KubeVirt selects the best
+  available tier automatically. Also used as a per-VM override to loosen a
+  cluster-wide `Hardware` policy.
+
+#### Precedence Resolution
+
+The effective emulation policy is resolved at domain conversion time in
+`KvmDomainConfigurator` using the following precedence (highest first):
+
+1. **Per-VM** — `vmi.Spec.EmulationPolicy` (if non-empty)
+2. **Global** — `clusterConfig.GetEmulationPolicy()` (if non-empty)
+3. **Default** — `None` (no emulation; native KVM only)
+
+This means a user can set `emulationPolicy: Software` on a VM to loosen a
+cluster-wide `Hardware` policy for that specific workload.
 
 ### VMI Conditions
 
@@ -695,8 +757,55 @@ Without this adjustment, `NodeSelectorRenderer` sets `kubernetes.io/arch` as a
 which would prevent scheduling on nodes with a different architecture —
 defeating the purpose of cross-architecture emulation.
 
-The implementation uses a combination of hard constraints and soft
-preferences:
+The `emulationPolicy` field determines which scheduling case applies:
+
+- **`None`** (default): The hard `kubernetes.io/arch` node selector is
+  preserved. The pod can only schedule on nodes natively running the guest
+  architecture. Emulation nodes are never selected.
+- **`Hardware`**: The hard `kubernetes.io/arch` node selector is removed.
+  A `requiredDuringSchedulingIgnoredDuringExecution` affinity permits nodes
+  that either natively match the guest architecture **or** carry the
+  `kubevirt.io/cross-arch-kvm-<arch>` label. Software-emulation-only nodes are
+  excluded. Preferred affinities order native (weight 100) over hardware
+  cross-arch (weight 50).
+- **`Software`**: The hard `kubernetes.io/arch` node selector is replaced by a
+  hard `kubevirt.io/vm-arch-<guest-arch>` node selector (permits native,
+  hardware cross-arch, and software nodes). Preferred affinities order native
+  (weight 100) > hardware cross-arch (weight 50) > software nodes.
+
+For the **`Hardware`** case, the resulting pod affinity looks like:
+
+```yaml
+spec:
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: kubernetes.io/arch
+            operator: In
+            values:
+            - arm64
+        - matchExpressions:
+          - key: kubevirt.io/cross-arch-kvm-arm64
+            operator: Exists
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        preference:
+          matchExpressions:
+          - key: kubernetes.io/arch
+            operator: In
+            values:
+            - arm64
+      - weight: 50
+        preference:
+          matchExpressions:
+          - key: kubevirt.io/cross-arch-kvm-arm64
+            operator: Exists
+```
+
+For the **`Software`** case, the implementation uses a combination of hard
+constraints and soft preferences:
 
 1. **Remove the hard `kubernetes.io/arch` node selector** so the pod is not
    restricted to nodes matching the guest architecture
@@ -718,12 +827,15 @@ Pending rather than CrashLooping.
 
 **`pkg/virt-controller/services/template.go`** (`newNodeSelectorRenderer`):
 
-When the feature gate is enabled, the `kubernetes.io/arch` node selector is
-**replaced** with a `kubevirt.io/vm-arch-<guest-arch>` node selector:
+When the effective `emulationPolicy` permits emulation (`Hardware` or `Software`),
+the hard `kubernetes.io/arch` node selector is removed. For `Software` policy, a
+`kubevirt.io/vm-arch-<guest-arch>` hard selector replaces it:
 
 ```go
 architecture := vmi.Spec.Architecture
-if t.clusterConfig.CrossArchitectureVirtualizationEnabled() {
+effectivePolicy := resolveEmulationPolicy(vmi, t.clusterConfig)
+if t.clusterConfig.CrossArchitectureVirtualizationEnabled() &&
+    effectivePolicy != emulationPolicyNone {
     architecture = ""
 }
 
@@ -746,10 +858,11 @@ if t.clusterConfig.CrossArchitectureVirtualizationEnabled() && vmi.Spec.Architec
 
 **`pkg/virt-controller/services/template.go`** (preferred affinities):
 
-Preferred node affinities are added after the pod is built:
+Preferred node affinities are added for both `Hardware` and `Software` policies:
 
 ```go
-if t.clusterConfig.CrossArchitectureVirtualizationEnabled() {
+if t.clusterConfig.CrossArchitectureVirtualizationEnabled() &&
+    effectivePolicy != emulationPolicyNone {
     // Weight 100: strongly prefer native architecture nodes
     setPreferredArchitectureAffinity(vmi.Spec.Architecture, &pod, 100)
 
@@ -915,7 +1028,7 @@ and cross-arch emulation is handled within the `KvmDomainConfigurator`.
 
 ## API Examples
 
-### Enable Cluster-Wide Cross-Architecture Emulation
+### Enable Cluster-Wide Cross-Architecture Emulation (Software Policy)
 
 ```yaml
 apiVersion: kubevirt.io/v1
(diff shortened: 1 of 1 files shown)


jschintag:
After discussion in yesterdays VEP meeting i will split of containerdisks and preference into their own VEP that depend on this one, like DRA.
Additionally i will look further into the containerdisk design to find a solution that does not involve resolving it during the reconcile loop.

jschintag:
I have split of the max emulation tier setting into a separate VEP #429, see PR #430.

Additionally i have dropped the containerdisk handling, as this needs further refinement and likely changes in kubernetes first.

This PR now focuses purely on the binary distribution and e2e test strategy.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -630,16 +630,55 @@ COPR repository. This is a known blocker for productization — the packages
...
+
+**`virt-launcher-emulation`** — an optional, additive image built `FROM
+virt-launcher` with cross-architecture QEMU system binaries layered on top
+(e.g. `qemu-system-aarch64`, `qemu-system-x86_64`).
+Operators who want cross-arch emulation pull and configure this image; the
+rest of the cluster is unaffected.
+
+Building `virt-launcher-emulation` is guarded by `KUBEVIRT_CROSS_ARCH_EMULATION`, but the rpms are cached normally using bazeldnf.
+The image is not built by default and not included in releases.
+This could only be considered for release should the binaries land in mainstream centos.
```
lyarwood: I don't think we can do any of this without the images in core CentOS Stream and AFAIK there are no plans to ship them there.

jschintag: What speaks against caching the rpms normally and having the image built logic?

As described here we would not ship the image, just have the logic for it and use it for e2e tests.

lyarwood: > What speaks against caching the rpms normally and having the image built logic?

We can't cache these RPMs in the WORKSPACE as the virt-preview RPMs are straight from COPR and rotate frequently, they also conflict with the CentOS Stream RPMs that confusingly have a higher EPOCH value. That's why I went with the env var based build time injection for now to keep things separate.

This image would also only truly be additive once the RPMs are present in CentOS  Stream, until then and as is the case now we'd be rebasing the entire virt stack on virt-preview within the image right?

> As described here we would not ship the image, just have the logic for it and use it for e2e tests.

I'm missing what's wrong with the current approach in that case?

```
KUBEVIRT_CROSS_ARCH_EMULATION=true make rpm-deps cluster-up cluster-sync functest
```

jschintag: I did not realize the problem with the differing qemu versions.

I guess it would make more sense to state this problem in the VEP and drop the binary distribution for software emulation entirely, to make it clear this is not the goal of the VEP.
Instead i will add the e2e test strategy, as well as the necessary qemu bump in the future for Hardware Emulation.

jschintag:
Updated PR after Feedback to focus more on specifiying the e2e test strategy, as well as dropping the requirements for software emulation binary distribution, as it is not viable with only preview binaries.

This makes the GA version of the VEP focus on supporting Hardware Based emulation using the newly announced Dual Architecture Processor for IBM Z, while using Software Emulation to provide hardware independent e2e test coverage.

jschintag:
After feedback from #430 , i have added back the emulation preference setting to the PR.

lyarwood (commented):
/lgtm

Apologies for not coming back to this! Looks good to me for continued alpha work in v1.10.0 thanks!

jschintag:
> /lgtm
> 
> Apologies for not coming back to this! Looks good to me for continued alpha work in v1.10.0 thanks!

No worries and thank you for looking after it. I know VEP freeze is a busy time.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -551,25 +565,75 @@ scenarios.
...
-| No | true | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
-| No | false | amd64 | arm64 | N/A | **ERROR** - cross-arch not allowed |
-| Yes | false | amd64 | arm64 | No | Cross-arch software emulation (TCG) |
-| Yes | false | s390x | arm64 | Yes | Cross-arch KVM-accelerated virtualization (SAE) |
-| Yes | false | s390x | arm64 | No | Cross-arch software emulation (TCG fallback) |
-| Yes | false | amd64 | amd64 | N/A | Native KVM execution |
-| Yes | true | amd64 | amd64 | No | Same-arch emulation (existing behavior) |
+| Feature Gate Enabled | useEmulation | emulationPolicy | Host Arch | Guest Arch | KVM Cross-Arch | Result |
+|---------------------|--------------|-----------------|-----------|------------|----------------|--------|
+| No | false | — | amd64 | amd64 | N/A | Native KVM execution |
```
0xFelix, line 571: How does it behave with `amd64` on `amd64`, but with no HW acceleration available? Reject?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -551,25 +565,75 @@ scenarios.
...
+  cluster-wide `Hardware` policy.
+
+#### Precedence Resolution
+
+The effective emulation policy is resolved at domain conversion time in
+`KvmDomainConfigurator` using the following precedence (highest first):
+
+1. **Per-VM** — `vmi.Spec.EmulationPolicy` (if non-empty)
+2. **Global** — `clusterConfig.GetEmulationPolicy()` (if non-empty)
+3. **Default** — `None` (no emulation; native KVM only)
```
0xFelix, line 633: And keeping the existing `useEmulation` setting in mind? I.e. with same arch on guest and host but no hardware emulation available.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -551,25 +565,75 @@ scenarios.
...
+
+This means a user can set `emulationPolicy: Software` on a VM to loosen a
+cluster-wide `Hardware` policy for that specific workload, but cannot exceed
+the cluster-wide `Software` ceiling.
```
0xFelix: Why make it a ceiling?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -725,12 +827,15 @@ Pending rather than CrashLooping.
 
 **`pkg/virt-controller/services/template.go`** (`newNodeSelectorRenderer`):
 
-When the feature gate is enabled, the `kubernetes.io/arch` node selector is
-**replaced** with a `kubevirt.io/vm-arch-<guest-arch>` node selector:
+When the effective `emulationPolicy` permits emulation (`Hardware` or `Software`),
+the hard `kubernetes.io/arch` node selector is removed. For `Software` policy, a
```
0xFelix, line 831: This says when `Hardware` is allowed, then `kubernetes.io/arch` is stripped, but in the example above it is still present?

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix, line 855: This is only the be added in case `Software` emulation is allowed?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -933,12 +1041,68 @@ spec:
...
 ```
 
 **Note**: `useEmulation` is **not** required for cross-architecture emulation.
 The feature gate alone is sufficient. `useEmulation` only controls same-arch
 emulation when KVM is unavailable and can remain disabled (the default).
 
+**Note on default policy**: Without an explicit `emulationPolicy`, the default
+is `None` — enabling the feature gate alone does **not** permit cross-arch
+emulation. Set `emulationPolicy: Software` to allow software emulation, or
+`emulationPolicy: Hardware` to allow hardware-accelerated emulation only
```
0xFelix, line 1054: Why make the cluster wide setting a ceiling? Is it an undesired use case to run a single VM with software acceleration while running others on hardware accel? I.e. setting `Software` explicitly on that VM, so it does not silently fall back to.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -933,12 +1041,68 @@ spec:
...
+    emulationPolicy: Hardware
+```
+
+This policy ensures no VM silently falls back to slow software emulation.
+VMs that land on a node without hardware cross-arch KVM support will fail to
+start rather than running under TCG.
+
+### Override Emulation Policy Per-VM
+
+A VM can override the cluster-wide emulation policy at the VM/VMI spec level.
```
0xFelix, line 1079: Ah, so it is possible. Reading the above it sounded like it wasn't.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -1076,9 +1254,11 @@ characteristics are TBD.
 
 ### Update Scenarios
 
-- **Enabling feature gate**: Existing VMs on matching architectures are unaffected; new cross-arch VMs can be created
+- **Enabling feature gate**: Existing VMs on matching architectures are unaffected; new cross-arch VMs can be created (subject to `emulationPolicy`)
```
0xFelix, line 1257: Existing VMs are unaffacted, but the node selectors on their virt-launcher would still be changed once it's replaced, right?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -1076,9 +1254,11 @@ characteristics are TBD.
...
 - **Disabling feature gate**: Existing running cross-arch VMs continue running; new cross-arch VMs fail to start
```
0xFelix, line 1258: Will it also break live migration for running cross-arch VMs? Will the node labeller stop adding the new labels?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -1076,9 +1254,11 @@ characteristics are TBD.
...
 - **Changing useEmulation**: Only affects same-arch emulation; cross-arch VMs are unaffected
+- **Changing emulationPolicy**: Affects only new VMs; existing running VMs are unaffected until they are rescheduled
+- **Downgrading to a previous version**: The `emulationPolicy` field is ignored; existing VMs continue unaffected, new VMs cannot use the policy
```
0xFelix: If the downgraded version does not ship the required QEMU anymore, the running VMs will also break on migrations, right?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -1132,6 +1318,7 @@ characteristics are TBD.
 2. **Rejection scenarios**:
    - Verify cross-arch VMI rejected when feature gate disabled
    - Verify error when QEMU binary missing
+   - Verify cross-arch VMI rejected when `emulationPolicy` is `None` (default)
```
0xFelix, line 1322: Would it really be rejected or simply never scheduled as there is no suitable node?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -77,6 +76,9 @@ Cross-architecture emulation would enable:
 - Set VMI conditions that distinguish between hardware-accelerated and
   software-emulated cross-architecture execution
 - Maintain backward compatibility with existing emulation behavior
+- Allow cluster administrators and users to configure a maximum emulation tier
+  via `emulationPolicy`, so that emulation remains explicitly opt-in and workloads
```
0xFelix: It's a goal but it's also already naming the solution.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -551,25 +565,75 @@ scenarios.
...
 - When the feature gate is disabled, attempting to run a cross-arch VM returns an error at domain conversion time
-- When the feature gate is enabled, the scheduler **prefers** nodes matching the guest architecture (native execution with KVM) and only falls back to cross-arch emulation when no matching nodes are available
-- When hardware acceleration is available, it is automatically preferred over software emulation
+- When the feature gate is enabled but `emulationPolicy` is `None` (default), cross-arch VMs are rejected — emulation must be explicitly opted in
+- When hardware acceleration is available, it is automatically preferred over software emulation (when the policy permits both)
+
+### Emulation Policy
+
+To ensure that software emulation remains an explicit opt-in feature and is
+never silently imposed on a workload, a new `emulationPolicy` field is
```
0xFelix, line 596: As naive user I might be inclined to think this also affects software emulation of an arch on the same arch, i.e. what `useEmulation` controlled up until now. Are there plans to retire `useEmulation` and to streamline everything under this new policy?

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: I'd assume you would need the `Software` `emulationPolicy` to then try to use TCG as a fallback? It is missing from the table however.

veps/sig-compute/172-cross-arch-virtualization/vep.md
lyarwood: There wasn't in the initial Alpha but now @jschintag has expanded the scope we might want to set that out for Beta in v1.11.0?

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: If we replace the `useEmulation` it should not schedule on that node without `Software` enabled

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Sorry, that sentence makes no sense this way, i think it slipped in through all the refactoring.

```
This means a user can set `emulationPolicy: Software` on a VM to loosen a
cluster-wide `Hardware` policy for that specific workload.
```

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: The nodeSelector is removed and it becomes an affinity instead.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Yes, it is not needed for HW, as that case already requires either HW or native virtualization.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: That depends on the cluster wide emulationPolicy. When the VM get's rescheduled, it would honor the policy, so if that was set to Software/Hardware the scheduling behaviour would change.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: It should not break live-migration for native VMs, at the moment i'm unsure if live-migration will be supported for cross-arch VMs at all.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: As long as they run, they should continue running as the virt-launcher image stays the same.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: The current default in kubevirt is to reject other architectures as long as the multiarchcluster FG is not enabled.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: I can set this as a goal as well.
Though i haven't gotten a beta target yet, since it depends on availability of new HW.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Removed "via `emulationPolicy`"

jschintag:
I fixed 2 small errors in phrasing.

As an open question on replacing `useEmulation` i would say since it is already in the exception window for 1.10 i would not add it here, but maybe do the consolidation in 1.11?

@lyarwood @0xFelix WDYT?

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: Yes, but trying to live migrate them would fail.

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: Is it outlined somewhere?

0xFelix:
> As an open question on replacing useEmulation i would say since it is already in the exception window for 1.10 i would not add it here, but maybe do the consolidation in 1.11?

You could mention that as a future goal / graduation criteria.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: It is an [open question](https://github.com/kubevirt/enhancements/blob/main/veps/sig-compute/172-cross-arch-virtualization/vep.md#open-questions), number 5:
```
Live migration: Can an ARM64 VM be migrated between two s390x SAE- capable hosts? What about migration between an s390x SAE host and a native ARM64 host?
```

jschintag:
> > As an open question on replacing useEmulation i would say since it is already in the exception window for 1.10 i would not add it here, but maybe do the consolidation in 1.11?
> 
> You could mention that as a future goal / graduation criteria.

I have added a Beta Criteria to decide on whether to replace it or not.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: I have added "until they are rescheduled" to the point

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: nit: Add v1.10.0 here

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: Multiarch is not the same as `CrossArchitectureVirtualization` if I'm not mistaken? So if multiarch is enabled, but no `CrossArchitectureVirtualization`, it will not be scheduled, right?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -67,8 +67,7 @@ Cross-architecture emulation would enable:
 
 ## Goals
 
-- Enable VMs to run on nodes with different architectures using either software
-  or hardware-accelerated virtualization
+- Enable VMs to run on nodes with different architectures using hardware-accelerated virtualization
```
0xFelix: What about the software emulation? Doesn't it still hold true?

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: I meant specifically in the case of disabling the feature gate after it was enabled first and VMs are already running.

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix, line 1268: Add note that has implications for new cross-arch VMs? They can't be migrated anymore.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -933,12 +1040,68 @@ spec:
...
+development VM, allowing it to also use software emulation:
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: arm64-dev-vm
+spec:
+  architecture: arm64
+  emulationPolicy: Software  # Allows software emulation, overrides cluster policy
```
0xFelix: These should sit underneath `spec.template.spec`.

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -933,12 +1040,68 @@ spec:
     developerConfiguration:
       featureGates:
       - CrossArchitectureVirtualization
+    emulationPolicy: Software
```
0xFelix, line 1044: What happens if this is changed to Hardware, but VMs we're already scheduled with software emulation? They will now fail to live-migrate?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -1291,19 +1498,21 @@ characteristics are TBD.
...
 - [ ] User feedback incorporated from Alpha release
-- [ ] Consider automated QEMU binary distribution approach
-- [ ] Consider per-VM emulation policy (e.g., a VMI-level field to disable
-      cross-arch emulation, prefer hardware acceleration, or require hardware
-      acceleration for specific workloads)
+- [ ] Emulation documentation updated (user-guide)
+- [ ] `emulationPolicy` field introduced and scheduling updated
+- [ ] Unit tests for policy precedence, enforcement, and webhook validation
+- [ ] `Hardware` emulation policy tier enabled
+- [ ] Decide if `emulationPolicy` should also replace `useEmulation` to consolidate emulation under a single setting
```
0xFelix: The changes you made are for Alpha, not Beta, right? So they should be moved to the Alpha section?

veps/sig-compute/172-cross-arch-virtualization/vep.md
```diff
@@ -77,6 +76,9 @@ Cross-architecture emulation would enable:
 - Set VMI conditions that distinguish between hardware-accelerated and
   software-emulated cross-architecture execution
 - Maintain backward compatibility with existing emulation behavior
+- Allow cluster administrators and users to configure a maximum emulation tier
```
0xFelix, line 80: `Maximum` still sounds like it cannot be overridden?

veps/sig-compute/172-cross-arch-virtualization/vep.md
0xFelix: Can we add it to the table?

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Live migration is not included here, since as stated it is as of yet unknown if it works at all.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Likely, though as live-migration is still an open question, the safest would be to set the emulationPolicy level to `Software` before live-migration.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: The VEP is already at alpha, these are improvements to the alpha stage, so i put it for beta.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: I left it out since the VEP is already alpha, this just adds improvements. Should i add it as an addition to 1.9 or replace it?

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: What is meant here is that the resulting config after resolving VM > Cluster > Default is the maximum allowed tier for the VM

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Isn't
```
| No | false | — | amd64 | amd64 | N/A | Native KVM execution |
```
already that case? This table only shows the results of where it would be scheduled, independent of what is available on the cluster. If kubernetes can't schedule the VM, it will just be stuck in Pending.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Done

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: I reversed the change.

veps/sig-compute/172-cross-arch-virtualization/vep.md
jschintag: Fixed

jschintag:
Updated according to offline discussion:
- Moved new goals to alpha instead of beta, as VEP stays alpha
- Added 1.10.0 to alpha targets
- Added new goal to clarify live-migration for beta stage

lyarwood (approved):
/hold

@0xFelix unhold if you're cool with this as discussed, lgtm.

0xFelix (approved):
Looking good, thanks!

/lgtm
/hold cancel

PR #414, opened 2026-08-05
Add design to support CPU DRA driver in KubeVirt

Sreeja1725:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/152
**SIG label**:  /sig compute

### What this PR does
VEP for supporting CPU DRA driver in KubeVirt

PR #424, opened 2026-08-18
VEP-115: DRA metadata topology and PCIe root complex grouping

johnahull:
## Summary

Amends VEP-115 (PCIe NUMA Topology Awareness) to:

1. **Address the PCIe topology mirroring open question**: group devices by PCIe root complex rather than NUMA node only, creating one `pxb-pcie` per root complex for accurate peer-to-peer DMA locality
2. **Extend VEP-115 for DRA**: read `resource.kubernetes.io/numaNode` and `resource.kubernetes.io/pcieRoot` from KEP-5304 device metadata
3. **Add a topology shim**: discovers topology from sysfs for devices without metadata (device plugin or DRA drivers that don't publish topology attributes), writes KEP-5304 metadata files so the converter has a single consumption path

No new feature gate, no API changes. Gated by existing `PCINUMAAwareTopology`.

## Motivation

VEP-115 groups devices by NUMA node — one `pxb-pcie` per NUMA node. This is inaccurate on hardware where a single NUMA node has multiple PCIe root complexes (NPS1, Intel with SNC off — 4+ root complexes per NUMA node). Devices on different root complexes get grouped together, and the guest cannot distinguish them for peer-to-peer DMA.

## Related

- VEP-115 original: #116
- VEP-152 (CPU DRA, defers NUMA passthrough to VEP-115): #414
- KubeVirt issue: https://github.com/kubevirt/kubevirt/issues/18817
- KEP-5304: https://github.com/kubernetes/enhancements/issues/5304

/cc @mresvanis @vladikr @alaypatel07 @enp0s3

🤖 Generated with [Claude Code](https://claude.com/claude-code)


veps/sig-compute/115-pcie-numa-topology-awareness/vep.md
```diff
@@ -190,6 +253,55 @@ DRA devices: the status controller must include the PCIe address in
...
+ensures all passthrough devices have metadata before the converter runs.
+
+#### Device Plugin Metadata Shim
+
+A topology shim in virt-launcher runs before domain conversion and writes
+KEP-5304-format metadata for any passthrough device that does not already
+have a metadata file. This covers:
+
+- Device-plugin-allocated devices (which never have KEP-5304 metadata)
+- DRA devices whose driver did not publish `numaNode` or `pcieRoot`
```
alaypatel07: We should leave this out of the scope, drivers that dont publish this information could have other issues, the right action would be to have a discussion with the driver why those attributes are not published.

veps/sig-compute/115-pcie-numa-topology-awareness/vep.md
```diff
@@ -190,6 +253,55 @@ DRA devices: the status controller must include the PCIe address in
...
+[`k8s.io/dynamic-resource-allocation/deviceattribute`](https://github.com/kubernetes/dynamic-resource-allocation)
+library to discover topology from sysfs:
+
+1. Identify passthrough devices without existing metadata files
+2. Call `deviceattribute.GetPCIBusIDAttribute` for pciBusID
+3. Call `deviceattribute.GetNUMANodeAttributeByPCIBusID` for numaNode
+4. Call `deviceattribute.GetPCIeRootAttributeByPCIBusID` for pcieRoot
+5. Write KEP-5304-format JSON to the metadata directory
+
+This ensures sysfs is read in exactly one place (the shim) and the
```
alaypatel07, line 305: There needs to be a note added to this section:

```
This works because it is assumed that dra drivers and device plugins will not be activated on the same node for the same devices. If this assumption is broken then the paths for dra devices could collide with device plugins.
```

That also begs a question, the metadata implementation has baked in assumptions that work on claim name and claim namespace. Since the claim name doesnt exist, what will be the file path of this information? What will be the placeholder claim name and metadata information in the metadata file?

How do metadata consumers know this is a topology information file, not to be considered for rendering the pcie/mdev UUID information in domxml?

alaypatel07 (changes requested):
Skimmed through this design, added some initial thoughts.

Still need to do a deeper review. Will follow it up later this week.

veps/sig-compute/115-pcie-numa-topology-awareness/vep.md
johnahull: Agreed. I've removed the DRA driver fallback from the shim's scope. The shim now covers only device-plugin-allocated devices. A DRA driver that does not publish `numaNode` or `pcieRoot` is treated as a driver-side gap to resolve with that driver upstream, not something KubeVirt backfills from sysfs. Updated in 24657502.

veps/sig-compute/115-pcie-numa-topology-awareness/vep.md
johnahull: Added your note verbatim about the assumption that DRA drivers and device plugins are not activated on the same node for the same devices, and the path collision risk if that assumption is broken (24657502).

johnahull: Here's the proposed layout for the claimless device-plugin case.

The shim writes topology-only metadata to a new subtree that kubelet never manages, alongside the two it does:

```
/var/run/kubernetes.io/dra-device-attributes/
  resourceclaims/...           (kubelet-owned)
  resourceclaimtemplates/...   (kubelet-owned)
  deviceplugin/<id>/topology/kubevirt-shim-metadata.json   (shim-owned)
```

`<id>` is `devplugin-<pool>-<bdf>`, deterministic and unique per node.

On disambiguation: the DRA host-device builders that emit `<hostdev>` entries into the domain XML (`CreateDRAHostDevices` / `CreateDRAGPUHostDevices`) never see these files. They only iterate spec entries that pass `IsHostDeviceDRA` / `IsGPUDRA` (`DeviceName == "" && ClaimRequest != nil`). Device-plugin devices have `DeviceName` set and no `ClaimRequest`, so they never reach a metadata lookup, and `resolveClaimMetadata` only builds paths from claim names in the spec so it can't construct a path into `deviceplugin/`. It's a structural guarantee, not a runtime check.

As a second safeguard, the shim files carry only `numaNode` and `pcieRoot`, never `pciBusID` or `mdevUUID`. If anything ever did point those builders at one, `createHostDeviceForGPU` requires one of those identity attributes and fails loudly instead of rendering a phantom device.

File contents: `DeviceMetadata` with `Namespace` empty and `PodClaimName` nil so it can't look like a real claim or template, request name `topology`, `Driver` set to a shim-specific identity, `Pool` = resource name, `Name` = BDF, `Attributes` limited to `numaNode` and `pcieRoot`.

Only the new metadata-enumerating reader this VEP adds walks all three subtrees. I'll write this up as a "Shim Metadata File Layout" section if the direction looks right.

johnahull: Added in b8fe31da as a "Shim Metadata File Layout" section.

vladikr:
@johnahull Thank you! 
I'll approve this in principle and let @alaypatel07 converge this. 
/approve

alaypatel07:
During the review of this design:
1. it was found that the memory dra driver is still not available for use. 
2. CPU being requested from VEP-152 could end up with memory from kubelet which is not numa aligned.

With that the biggest benefit from this VEP would be to get grouping of devices (gpu/nic) coming from same pcie to be configured in the same pxb-pcie.

With the above, there are two paths:
1. to modify this PR and 1.10 target to be focused on pcie aligned pxb-pcie grouping, have numa alignment with pcie grouping when cpu and memory dra drivers are available.
2. to differ this change until cpu/memory dra drivers are available.

My preference would be the 2nd approach, but if there are usecases out there that needs immediate solution doing down path 1 is also fine with me. @johnahull @vladikr thoughts?

johnahull:
> During the review of this design:
> 
> 1. it was found that the memory dra driver is still not available for use.
> 2. CPU being requested from VEP-152 could end up with memory from kubelet which is not numa aligned.
> 
> With that the biggest benefit from this VEP would be to get grouping of devices (gpu/nic) coming from same pcie to be configured in the same pxb-pcie.
> 
> With the above, there are two paths:
> 
> 1. to modify this PR and 1.10 target to be focused on pcie aligned pxb-pcie grouping, have numa alignment with pcie grouping when cpu and memory dra drivers are available.
> 2. to differ this change until cpu/memory dra drivers are available.
> 
> My preference would be the 2nd approach, but if there are usecases out there that needs immediate solution doing down path 1 is also fine with me. @johnahull @vladikr thoughts?

Today all devices on the same NUMA node share one pxb-pcie, so the guest can't tell which GPU pairs with which NIC for peer-to-peer DMA. The PCIe root grouping matters for multi-device passthrough on hardware with multiple root complexes per NUMA node to correctly places devices, such as for GPU Direct RDMA.  Not sure if anyone is waiting on this use case right now. I'm fine with deferring

johnahull:
Here is the updated plan for this VEP:

* For 1.10, we'll open a bugfix PR to fix the ordering for pcie devices to pcieroot first, then numa second. This will continue to read from sysfs. This will enable the use case for multi-device passthrough on hardware with multiple root complexes per NUMA node for device plugin. 
* We'll defer this implementation until 1.11 (or later) for DRA device passthrough. We need CPU and Memory DRA drivers to do DRA pcie passthrough and placement for pcieroot and numa node.

PR #425, opened 2026-08-18
Updated VEP 192 to explain the new data path for the decentralized migration proxy.

awels:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/192
**SIG label**: /sig compute

### What this PR does
This PR enhances the proxy VEP proposal to include a new datapath field that allows the admin to configure how the proxy behaves in various scenarios. Each scenario is described with an example of how to configure it, and why one would want to pick that particular combination.

In particular the interesting scenario is the proxy -> proxy communication over the pod network, which allows cross cluster live migration without a special migration network.


veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +77,338 @@ A further issue is operational: when nodes are added or removed, any mechanism t
...
+
+Traffic path with ingress routing and no dedicated networks (both unset):
+
+```
+source virt-handler → source sync controller
+  → target ingress/route hostname:port
+  → target sync controller → target virt-handler
+```
+
+The ingress/route must support gRPC over the synchronization port (for example, an NGINX Ingress with `backend-protocol: GRPC`, or an OpenShift Route with passthrough or re-encrypt TLS). The cluster admin creates the Service, Ingress, or Route; KubeVirt discovers the external endpoint and publishes it in `status.synchronizationAddresses`.
```
jean-edouard: > KubeVirt discovers the external endpoint and publishes it in `status.synchronizationAddresses`

How does it discover it? Where in https://github.com/kubevirt/kubevirt/pull/17922 does that happen?

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +77,338 @@ A further issue is operational: when nodes are added or removed, any mechanism t
...
+| `crosscluster0` | `migrations.crossClusterNetwork` | Peer sync-controller gRPC traffic between clusters |
+
+If `migrations.network` is set with the proxy datapath, synchronization controllers require the `migration0` interface at startup and will fail to start if it is missing. If `migrations.crossClusterNetwork` is set, synchronization controllers require the `crosscluster0` interface at startup and the sync gRPC port binds only on that interface; when omitted, peer traffic uses the pod network.
+
+The proxy datapath supports four network configurations:
+
+| `migrations.network` | `migrations.crossClusterNetwork` | virt-handler ↔ proxy | proxy ↔ proxy (peer gRPC) |
+|----------------------|----------------------------------|----------------------|---------------------------|
+| unset | unset | pod network | pod network (Service/Ingress/Route) |
+| set | unset | `migration0` (dedicated migration network) | pod network (Service/Ingress/Route) |
```
jean-edouard: FYI, AI flags this case as not correctly implemented (and not e2e-tested) in the code PR. It says `status.synchronizationAddresses` will wrongly get the IP of migration0.
I haven't verified it myself.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +77,338 @@ A further issue is operational: when nodes are added or removed, any mechanism t
...
+
+Cross-cluster live migration then works as long as every cluster can reach every other cluster's ingress/route endpoint.
+
+Traffic path with ingress routing and a dedicated in-cluster migration network (`network` set, `crossClusterNetwork` unset):
+
+```
+source virt-handler ──(migration0)──► source sync controller
+  ──(pod network / ingress/route)──► target sync controller
+  ──(migration0)──► target virt-handler
+```
```
jean-edouard: Nit: arrows messed up in this block

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +77,338 @@ A further issue is operational: when nodes are added or removed, any mechanism t
 
 ## Design
```
jean-edouard: Can you touch on how we handle the multiple sync-controller replicas? i.e make sure we never connect to the idle one

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -45,13 +45,16 @@ A further issue is operational: when nodes are added or removed, any mechanism t
 ## Goals
 
 * **Reduced addressing:** With the proxy in place, the cross-cluster live migration network needs only one IP per synchronization controller. Total addresses = (synchronization controllers per cluster) × (number of clusters). Example: 2 controllers per cluster × 10 clusters = 20 addresses.
+* **Datapath selection:** Allow admins to choose between direct virt-handler connectivity (`Direct`) and sync-controller proxying (`Proxy`) via `decentralizedLiveMigrationDatapath`.
```
jean-edouard: What if you have multiple clusters with a mix of `Direct` and `Proxy`? I assume that'll not end well, maybe the VEP should state that all clusters must have matching datapaths?

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +77,338 @@ A further issue is operational: when nodes are added or removed, any mechanism t
...
+
+* Existing decentralized live migration behavior.
+* Migration data flows directly between virt-handlers (via the migration network or pod network).
+* Synchronization controllers handle coordination only (gRPC status sync); they do not proxy migration data.
+
+**Proxy**
+
+* Synchronization controllers terminate virt-handler TLS and proxy migration traffic over a multiplexed gRPC tunnel.
+* Reduces IP requirements on the cross-cluster network (see motivation above).
+* Applies to decentralized live migrations generally (cross-namespace and cross-cluster).
```
jean-edouard: Why?

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: Actually kubevirt/kubevirt#17922 does NOT do that particular part, that will be a separate PR since it is outside of the scope of the proxy.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: leader election, if the controller is not the leader, it doesn't even listen on any ports. I can put a blurp in the VEP about this, but that hasn't changed since the initial implementation of decentralized live migration.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: Good point, I should make it clear that all clusters that will want to decentralized live migration should have the same datapath.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: Basically because we can't tell the difference between a cross cluster and cross namespace from the proxy perspective. Obviously there is no actual reason to use the proxy for cross namespace. But if you enabled the proxy, then everything goes through the proxy including the cross namespace, because we can't tell the difference. We just see the source or the target when processing so we can't determine if it is cross namespace or cross cluster without extra information.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: My AI agrees, I will update the actual PR to properly advertise the address.

awels: I fixed the PR to include a test and properly advertise the correct address

veps/sig-compute/191-decentralized-migration-proxy/vep.md
```diff
@@ -71,52 +78,360 @@ A further issue is operational: when nodes are added or removed, any mechanism t
...
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: virt-synchronization-controller
+  namespace: kubevirt
+spec:
+  selector:
+    kubevirt.io: virt-synchronization-controller
```
jean-edouard: Will that only match the leader?

jean-edouard (commented):
Shouldn't this be VEP 192? #191 is the PR for VEP 190...
Looking pretty good now otherwise. Since the network part is pretty intense, and I must admit I don't understand all of it, it would be great to have a review from a sig-network engineer.

veps/sig-compute/191-decentralized-migration-proxy/vep.md
awels: I will add a blurp to make that more explicit but basically a service only routes to pods that are ready, the inactive one will fail the readiness probe and thus not be part of the rotation.

jean-edouard:
I'm good merging this as Alpha so we don't slip another release.

Please add a short note that the Service/Ingress exposure still needs a sig-network pass (follow-up VEP update, not a blocker). Also rename the directory and document title from 191 to 192 so it matches the tracking issue.

Thanks!

awels:
> I'm good merging this as Alpha so we don't slip another release.
> 
> Please add a short note that the Service/Ingress exposure still needs a sig-network pass (follow-up VEP update, not a blocker). Also rename the directory and document title from 191 to 192 so it matches the tracking issue.
> 
> Thanks!

Done, added the note and renamed the directory

jean-edouard:
> > I'm good merging this as Alpha so we don't slip another release.
> > Please add a short note that the Service/Ingress exposure still needs a sig-network pass (follow-up VEP update, not a blocker). Also rename the directory and document title from 191 to 192 so it matches the tracking issue.
> > Thanks!
> 
> Done, added the note and renamed the directory

Something went wrong, the directory was copied instead of moved. Also the new file is a copy of an older version of the vep (missing sig-network note)

awels:
> > > I'm good merging this as Alpha so we don't slip another release.
> > > Please add a short note that the Service/Ingress exposure still needs a sig-network pass (follow-up VEP update, not a blocker). Also rename the directory and document title from 191 to 192 so it matches the tracking issue.
> > > Thanks!
> > 
> > 
> > Done, added the note and renamed the directory
> 
> Something went wrong, the directory was copied instead of moved. Also the new file is a copy of an older version of the vep (missing sig-network note)

Must have had a brain fart, it should be correct now.

veps/sig-compute/192-decentralized-migration-proxy/vep.md
```diff
@@ -0,0 +1,484 @@
+# VEP #191: Decentralized Live Migration LM network proxy support
```
jean-edouard:
```suggestion
# VEP #192: Decentralized Live Migration LM network proxy support 
```

jean-edouard (approved):

awels:
@iholder101 Could you review/approve?

vladikr:
Sorry for being late! 
@jean-edouard Thank you very much for shepherding this!

/approve

awels:
Thank you everyone, now for the fun part reviewing the matching PR :)

Issue #427, opened 2026-08-19
VEP 427: Source a Multus network's device from a DRA ResourceClaim


gunine:
**Primary contact (assignee)**:
@gunine

**Current Feature Stage**:
New

**Feature Gate**:
MultusNetworksWithDRA

**Responsible SIGs**:
Primary SIG:
/sig network
Additional SIGs (optional):
/sig compute

**Enhancement link**:
VEP PR: #428

**Timeline**:
* Alpha release target (x.y): 1.10
* Beta release target (x.y): 1.11 (tentative)
* Stable release target (x.y): 1.12 (tentative)
* [ ] Alpha
  * [ ] VEP PRs: #428
  * [ ] Code PRs: will follow VEP approval — design + PoC complete against v1.9.0 (API/admission/rendering/tests, ~4 small PRs; virt-launcher/virt-handler untouched)
  * [ ] Docs PRs: user-guide update, to be opened alongside the code PRs

**Additional context**:
Adds one optional field, `resourceClaim: {claimName, requestName}`, to `MultusNetwork`, so the device behind an existing Multus NetworkAttachmentDefinition (SR-IOV VF) is allocated from a DRA ResourceClaim request instead of a device-plugin extended resource. The network stays a plain Multus network at runtime (Multus runs the CNI and applies the NAD's VLAN/MAC/trust; virt-launcher is unchanged). Because the VF becomes a request in a user-authored claim, it can share one claim with a GPU under a `matchAttribute: resource.kubernetes.io/pcieRoot` constraint — scheduler-enforced GPU/NIC topology alignment for VMs, without migrating an existing Multus SR-IOV estate.

- Builds on the `spec.resourceClaims[]` / `ClaimRequest` infrastructure of VEP #10; complementary to and mutually exclusive with VEP #183 (DRA networks), behind a separate feature gate.
- Composes with what the network plumbing WG already merged: multus-cni DRA integration (k8snetworkplumbingwg/multus-cni#1492, merged 2026-07-08) and dra-driver-sriov in MULTUS mode.
- End-to-end validated on a physical testbed (2026-08-18): shared GPU+VF claim scheduled onto one PCIe root, NAD configuration applied by Multus.
- Presented at the KubeVirt DRA bi-weekly on 2026-08-19; review feedback (no-core-change alternatives incl. network binding plugins) analysed in the VEP's Alternatives section.

> [!IMPORTANT]
> Please keep this description up to date.

vladikr:
[...]

Issue #183, opened 2026-01-19
VEP 181: NetworkDevicesWithDRA


oshoval:
**Primary contact (assignee)**:
/assign @oshoval 

**Current Feature Stage**:

**Feature Gate**:
`NetworkDevicesWithDRA`

**Responsible SIGs**:
/sig network

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
  * [ ] Code PRs:
  * [ ] Docs PRs:

orelmisan:
@oshoval Can you please adjust the subject to VEP 183? Thanks.

oshoval:
Hi Valdik
I am working on it, in touch with sig-network about if it will be wrapped on 1.9,
once i conclude with sig-network if we plan it for 1.9 or afterwards will update.

VEP + implementation were already done basically, but since Kubevirt now depends on downward API, i need to update it.
https://github.com/kubevirt/enhancements/pull/185
https://github.com/kubevirt/kubevirt/pull/16939

Provider for 1.36 where i plan to test the new downward API
https://github.com/kubevirt/kubevirtci/pull/1659

oshoval:
Hi, seems we do hopefully
VEP is ready with mitigation plan, and i am working in parallel on the implementation of Kubevirt and the driver
Edy, Sebastian and sig-network would prioritize it as far as i understand

Relevant PRs (some are WIP, some are ready for review, marked the ones that are ready for review)

https://github.com/kubevirt/enhancements/pull/185 - VEP, **Ready for review**
https://github.com/kubevirt/kubevirtci/pull/1663 - provider
https://github.com/kubevirt/kubevirt/pull/17433 - VEP implementation
https://github.com/k8snetworkplumbingwg/dra-driver-sriov/pull/92 - support downward API, **Ready for review**
https://github.com/k8snetworkplumbingwg/dra-driver-sriov/pull/95 - support custom MAC annotation
https://github.com/kubevirt/kubevirt/pull/17490 - Fix multi SR-IOV / devices bug (per VM)

thanks

EdDev:
@oshoval , please update the issue description with all the PRs that handled it.
Can we consider it complete for v1.9 milestone?

oshoval:
yes
done

EdDev:
@oshoval , please update `Alpha release target` with the 1.9 target in which it got released.

[...]

PR #428, opened 2026-08-19
VEP-427: Source a Multus network's device from a DRA ResourceClaim

gunine:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/427
**SIG label**:

/sig network
/sig compute

### What this PR does

Introduces VEP-427: source a Multus network's device from a DRA ResourceClaim.

The proposal adds one optional field, `resourceClaim: {claimName, requestName}`, to
`MultusNetwork`. A network carrying it remains a plain Multus network at runtime —
Multus still invokes the CNI and applies the NAD's VLAN/MAC/trust configuration, and
virt-launcher builds the `<hostdev>` from network-status exactly as today. The field
only changes how the device behind the NAD is allocated: from a request in a
user-authored DRA ResourceClaim instead of a device-plugin extended resource.

Because the VF becomes a request in a user-authored claim, it can share that claim
with a GPU under a `matchAttribute: resource.kubernetes.io/pcieRoot` constraint —
**scheduler-enforced GPU/NIC topology alignment for VMs, without migrating an
existing Multus SR-IOV estate off its NADs**.

Scope: virt-api (admission) and virt-controller (pod rendering) only, behind a new
alpha feature gate `MultusNetworksWithDRA`. virt-launcher and virt-handler are
unchanged. Targeting alpha in v1.10.

- Builds on VEP #10 infrastructure (`spec.resourceClaims[]`, the shared
  `ClaimRequest` type). Complementary to — and mutually exclusive per VMI with —
  VEP #183's DRA networks; it does not reopen the "SR-IOV in DRA networks"
  question. Separate feature gates keep graduation independent.
- Composes with what the network plumbing WG already merged:
  multus-cni DRA integration (k8snetworkplumbingwg/multus-cni#1492, 2026-07-08)
  and dra-driver-sriov in MULTUS mode. A plain pod can use this shape today;
  this VEP lets virt-controller render the equivalent pod.
- Design + PoC are complete against v1.9.0 (~4 small PRs), and the flow was
  validated end-to-end on a physical testbed on 2026-08-18: a shared GPU+VF claim
  scheduled onto one PCIe root, with the NAD configuration applied by Multus.
- Presented at the KubeVirt DRA bi-weekly on 2026-08-19. Review feedback — whether
  the goal is reachable without modifying KubeVirt core, e.g. via network binding
  plugins — is analysed in Alternatives 2 and 3: an out-of-tree pod-mutating
  webhook can reproduce the rendering today but relies on untyped annotations and
  non-contractual internals; binding plugins cannot reach the allocation side.
- Known edges are recorded as Open Issues (claim→NAD mapping key, directly
  referenced claims, non-SR-IOV bindings) — none block alpha.
- Aiming for the v1.10 VEP freeze (2026-09-02).

### DIFF
--- a/veps/sig-network/427-multus-networks-with-dra/vep.md
+++ b/veps/sig-network/427-multus-networks-with-dra/vep.md
@@ -1,0 +1,494 @@
+# VEP #427: Source a Multus network's device from a DRA ResourceClaim
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+This proposal adds one optional field, `resourceClaim`, to `MultusNetwork`:
+
+```yaml
+networks:
+- name: sriov-net
+  multus:
+    networkName: vf-nad          # the existing NetworkAttachmentDefinition, unchanged
+    resourceClaim:               # NEW
+      claimName: gpu-vf
+      requestName: vf
+```
+
+A network carrying it remains a plain Multus network at runtime — Multus still
+invokes the CNI and applies the NAD configuration (VLAN, MAC, trust), the
+network-status annotation is still produced, and virt-launcher builds the
+`<hostdev>` from network-status exactly as today. The field only changes *how
+the device behind the NAD is allocated*: from a DRA `ResourceClaim` request
+named in `spec.resourceClaims[]`, instead of a device-plugin extended resource
+derived from the NAD's `k8s.v1.cni.cncf.io/resourceName` annotation.
+
+Because the device is now a request in a user-authored claim, it can share that
+claim with other DRA devices — in particular a GPU — and a
+[`matchAttribute` constraint](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/)
+(e.g. `resource.kubernetes.io/pcieRoot`) between the two requests is enforced
+by the Kubernetes scheduler. This is the primary use case: **scheduler-enforced
+GPU/NIC topology alignment for VMs, without migrating an existing Multus
+SR-IOV estate off its NADs.**
+
+This VEP builds on the core DRA infrastructure of VEP #10
+(`spec.resourceClaims[]`, the shared `ClaimRequest` type) and is complementary
+to VEP #183; see [Relationship to VEP #183](#relationship-to-vep-183).
+
+## Motivation
+
+Today a Multus-backed SR-IOV network is allocated through a device plugin:
+virt-controller reads the NAD's `k8s.v1.cni.cncf.io/resourceName` annotation
+and injects an opaque extended-resource request into the virt-launcher pod.
+The scheduler sees an integer count. It cannot express "this VF and this GPU
+must sit under the same PCIe root complex" — a constraint that directly
+determines P2P/GPUDirect-class throughput for VMs doing GPU networking.
+
+DRA can express exactly that: two requests in one claim under a
+`matchAttribute` constraint. But in current KubeVirt the two worlds cannot
+meet:
+
+- **DRA networks (VEP #183, `networks[].resourceClaim`)** replace the Multus
+  flow. A NAD estate written for Multus — relying on the `CNIDeviceInfoFile`
+  capability, Multus-applied VLAN/MAC/trust, the
+  `k8s.v1.cni.cncf.io/network-status` annotation that tooling reads — cannot
+  move to that model by configuration alone; it is a rewrite of the CNI
+  contract. VEP #183 accordingly scoped SR-IOV out of DRA network devices, and
+  currently requires a network binding plugin.
+- **DRA extended resources (`DeviceClass.extendedResourceName`)** let the
+  existing Multus flow allocate from a DRA pool without any KubeVirt change,
+  but the scheduler then creates an implicit single-request claim per device.
+  DRA has no cross-claim constraint, so the GPU (in the user's claim) and the
+  VF (in the implicit claim) cannot be constrained against each other. The
+  operator is left pinning PCIe root complexes by hand in DeviceClass CEL
+  selectors — and if a VF pool spans two root complexes, the pair is silently
+  split with no error, only degraded throughput.
+
+Meanwhile the ecosystem plumbing for "DRA allocates, Multus configures" has
+converged: multus-cni merged DRA `resource.k8s.io/v1` integration
+([multus-cni PR #1492](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1492),
+2026-07-08), which maps a pod's allocated claim devices to NADs via the
+`k8s.cni.cncf.io/resourceName` device attribute, and
+[dra-driver-sriov](https://github.com/k8snetworkplumbingwg/dra-driver-sriov)
+ships a MULTUS mode that prepares the VF (vfio binding, CDI injection,
+device-info file) while leaving CNI invocation to Multus. A plain pod can use
+this today. **Only KubeVirt cannot produce the equivalent pod shape** — this
+VEP closes that gap with a control-plane-only change.
+
+## Goals
+
+- Let a Multus-backed VMI network source its device (SR-IOV VF) from a DRA
+  ResourceClaim request, while Multus continues to run the CNI against the
+  existing NAD.
+- Thereby allow one claim to carry a GPU request and a VF request under a
+  `matchAttribute` constraint, enforced by the scheduler.
+- Keep the change confined to virt-api (admission) and virt-controller (pod
+  rendering). virt-launcher and virt-handler are unchanged.
+- Preserve full backward compatibility: without the new field, and with the
+  feature gate off, nothing changes.
+
+## Non Goals
+
+- Changing the DRA network path of VEP #183 (`networks[].resourceClaim`), or
+  taking any position on how *DRA networks* should model SR-IOV.
+- Mixing pure DRA networks and Multus networks in one VMI (the existing
+  rejection stays).
+- Deploying or managing the external components (DRA drivers, multus-cni);
+  they are cluster prerequisites.
+- Hot-plugging interfaces whose device comes from a claim (a pod's
+  `resourceClaims` are immutable).
+- Standardizing MAC/interface-name configuration through DRA — not needed
+  here, since the NAD/Multus path keeps doing both.
+
+## Definition of Users
+
+- **Admin**: manages the cluster, deploys the DRA drivers and multus-cni,
+  owns DeviceClasses and the existing NAD estate.
+- **VM owner**: authors ResourceClaims/Templates and VMs, and wants a VM's
+  GPU and SR-IOV VF to be topology-aligned.
+
+## User Stories
+
+- As a VM owner running GPU workloads with SR-IOV networking, I want the
+  scheduler to place my VM's GPU and VF under the same PCIe root complex, so
+  that P2P DMA between them does not cross the inter-socket interconnect.
+- As an admin with an existing Multus SR-IOV estate (NADs carrying VLAN/MAC
+  configuration and the `CNIDeviceInfoFile` contract), I want to adopt DRA
+  allocation for those VFs without rewriting the NADs or changing how the CNI
+  is invoked.
+- As an admin, I want VF allocation policy expressed in standard DRA objects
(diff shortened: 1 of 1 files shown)


gunine:
/cc @EdDev @lyarwood
As discussed at the DRA bi-weekly (2026-08-19): the feedback on exploring
no-core-change options is addressed in Alternatives 2 (out-of-tree pod-mutating
webhook — works today, verified against v1.9.0, but relies on untyped annotations
and non-contractual internals) and 3 (network binding plugins — no access to pod
resources.claims / allocation). Happy to walk through either in the next call.

veps/sig-network/427-multus-networks-with-dra/vep.md
```diff
@@ -0,0 +1,494 @@
...
+  but the scheduler then creates an implicit single-request claim per device.
+  DRA has no cross-claim constraint, so the GPU (in the user's claim) and the
+  VF (in the implicit claim) cannot be constrained against each other. The
+  operator is left pinning PCIe root complexes by hand in DeviceClass CEL
+  selectors — and if a VF pool spans two root complexes, the pair is silently
+  split with no error, only degraded throughput.
+
+Meanwhile the ecosystem plumbing for "DRA allocates, Multus configures" has
+converged: multus-cni merged DRA `resource.k8s.io/v1` integration
+([multus-cni PR #1492](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1492),
```
nirdothan, line 83: Hey @SchSeba in [VEP 183](https://github.com/kubevirt/enhancements/blob/7ee388aa7cf42e89be317d405d9b1c14d91a4012/veps/sig-network/183-dra-network/vep.md?plain=1#L207) it was decided to defer the implementation of SR-IOV and wait for an industry decision about the trajectory of DRA/Multus.
Can you please confirm the the Multus PR above represents a decision that is generally agreed upon?

gunine: Good point — +1, @SchSeba's confirmation on the NPWG direction would be very helpful here.

From this VEP's side: it deliberately doesn't reopen what #183 deferred — the network stays a plain Multus network and DRA only allocates, so this composes with what the NPWG has merged (#1492 + dra-driver-sriov MULTUS mode) rather than picking a winner for DRA networks. And the coupling is guarded: #1492 is merged but not yet released, and beta here already requires a released multus-cni with the integration — so if the industry direction shifts, this feature simply can't graduate.

aojea: My 2 cents based on my limited knowledge of the Kubevirt project so some things may be incorrect.

> so if the industry direction shifts, this feature simply can't graduate.

The motivation here makes total sense, and independently of the industry decision, the main goal for DRA was to provide common, standard APIs so different drivers and implementations can coexist and remain interchangeable, without leaking low-level mechanics into higher-level abstractions, so you should be able to switch providers without risk.

> the network stays a plain Multus network and DRA only allocates

The problem I see is in this API proposal are:

- It leaks the multus DRA implementation to the Kubevirt API
- This mixes device allocation with network configuration
- Creates two parallel, competing ways to declare DRA claims on networks with #183

Ideally, users should be able to provide their desired `ResourceClaim` for networks and devices cleanly without mutating existing network abstractions. Here are two possible solutions I think can solve this.

Solution 1: Zero API Changes to `MultusNetwork`.

1. The user defines the standard `MultusNetwork`:
   ```yaml
   spec:
     resourceClaims:
     - name: gpu-vf
       resourceClaimTemplateName: gpu-vf-same-root
     domain:
       devices:
         gpus:
         - name: gpu0
           claimName: gpu-vf
           requestName: gpu
         interfaces:
         - name: sriov-net
           sriov: {}
     networks:
     - name: sriov-net
       multus:
         networkName: vf-nad     # Standard MultusNetwork, unchanged
   ```
2. `virt-controller` includes the VMI's claim requests on the `compute` container's `resources.claims`.
4. To avoid double allocation (extended resource + DRA), `virt-controller` suppresses extended resource injection when DRA claims are present for that resource (or via a NAD annotation / cluster configuration, similar to the existing `ExternalNetResourceInjection` model).
5. Multus and virt-launcher work out-of-the-box via the standard `network-status` path.

Solution 2. Unify under `NetworkSource.ResourceClaim` (VEP #183)
If an explicit link between a network and a claim is required in the VMI spec, `NetworkSource.ResourceClaim` should be the canonical source for all DRA-backed networks:

```yaml
networks:
- name: sriov-net
  resourceClaim:                 # Single source of truth for DRA networks
    claimName: gpu-vf
    requestName: vf
```

If NAD-based CNI configuration (VLAN, MAC, trust) is needed for that DRA device, that configuration belongs to the **interface binding layer** in `domain.devices.interfaces`:

```yaml
domain:
  devices:
    interfaces:
    - name: sriov-net
      binding:
        name: sriov-cni          # Or sriov: { networkAttachmentDefinitionName: "vf-nad" }
```

gunine: Thanks @aojea — these are exactly the right API-review questions to ask, and the "two parallel ways" concern in particular deserves a direct answer. Taking the three concerns and the two solutions in turn.

### "Creates two parallel, competing ways to declare DRA claims on networks"

This is the concern I think carries the most weight, and the VEP already treats it as a first-class problem rather than an accident:

- The two forms are **mutually exclusive within one VMI** (the existing pure-DRA/Multus mixing rejection stays), so a user can never hold both semantics in one spec.
- They sit behind **separate feature gates** (`MultusNetworksWithDRA` vs `NetworkDevicesWithDRA`), so their maturity and graduation are independent.
- **Beta graduation explicitly includes the consolidation decision** with VEP #183 ("Decision on gate consolidation with `NetworkDevicesWithDRA`").
- Both fields are the **same shared `ClaimRequest` type** from VEP #10, so a later unification is a mechanical move of one field, not a semantic migration.

So the intended reading is not two competing endpoints, but coexistence with a defined convergence point — the same posture VEP #183 itself takes toward the deferred SR-IOV question.

### "It leaks the multus DRA implementation to the KubeVirt API"

The new field is `{claimName, requestName}` — the same provider-neutral `ClaimRequest` shape VEP #10 already uses for `gpus[]` and `hostDevices[]`. Nothing in it encodes anything Multus-DRA-specific: no driver name, no resourceName attribute, no mapping mechanics. Any DRA driver that publishes the `k8s.cni.cncf.io/resourceName` / `deviceID` device attributes works, so drivers remain interchangeable — which is precisely the DRA goal you describe. The Multus coupling lives in the cluster prerequisites (multus-cni with DRA integration), not in the API — and that coupling is the point: the target user has an existing NAD estate they need to keep.

### "This mixes device allocation with network configuration"

I'd argue it's the *current* device-plugin path that mixes the two: the NAD's `k8s.v1.cni.cncf.io/resourceName` annotation (configuration object) is what drives allocation today. This proposal separates them — allocation is a typed claim reference, configuration stays in the NAD. And note that VEP #183's `networks[].resourceClaim` also places a claim reference on the network object; the two designs are identical in that respect, so this can't be a differentiator between them.

### Solution 1 (zero API changes)

The blocker is timing plus a missing linkage:

- The claim↔NAD correspondence key (`k8s.cni.cncf.io/resourceName`) is a **ResourceSlice device attribute, known only after allocation** — that is exactly how multus's draclient resolves it (`pod.status.resourceClaimStatuses` → ResourceClaim → ResourceSlice). virt-controller renders the pod *before scheduling*, and deliberately does not watch ResourceClaims or ResourceSlices (a VEP #10 alpha design principle this VEP keeps). So "suppress the extended resource when DRA claims are present *for that resource*" has no reliable predicate at render time; it degrades to "suppress whenever any claim exists", which breaks VMIs mixing device-plugin-backed and claim-backed Multus networks — a coexistence the alpha explicitly supports.
- Without an explicit network→request link, virt-controller would have to place **unreferenced** claim requests on the compute container. That is precisely the non-contractual behavior Alternative 2 documents (the out-of-tree webhook exploits it today: one-directional admitter cross-check, wholesale claim copy). Baking it in as implicit behavior removes the possibility of admission-time validation — double-booked requests, unresolved claims, or default-network misuse would surface only late, as Pending pods or missing `/dev/vfio/*`.
- The NAD-annotation variant moves the mapping into an untyped annotation — a stronger form of the leak you're concerned about — and an `ExternalNetResourceInjection`-style cluster toggle is cluster-global where the decision is per-network.

So Solution 1 doesn't remove the information; it relocates it from a typed, validated field into heuristics and annotations.

### Solution 2 (unify under `NetworkSource.ResourceClaim`, NAD via the binding layer)

- The moment a network uses `networks[].resourceClaim` it becomes a DRA network (`IsDRANetwork == true`): virt-controller does not generate a Multus selection element for it, so **Multus never invokes the CNI, and the NAD's VLAN/MAC/trust configuration is never applied** — which is the entire purpose of this VEP. Moving the NAD reference to the interface/binding layer doesn't change that: binding plugins act at the domain-XML/launcher level and do not trigger pod networking. Making Multus run for DRA networks would require changes to annotation generation plus an exclusivity audit of the launcher's two hostdev builders — the blast radius Alternative 4 was rejected for, and larger.
- It also reopens exactly the question VEP #183 deferred (SR-IOV in DRA networks, "competing industry approaches") — this VEP is deliberately scoped not to touch it.
- It isn't API-neutral either: `sriov: {networkAttachmentDefinitionName: ...}` (or a binding extension) is also a new field — on the interface instead of the network — and it splits the allocation reference and the NAD reference across two places, which is more mixing, not less.
- The allocation-side limits of binding plugins (no access to the pod's `resources.claims[]`, resource requests, or scheduling) are analysed in Alternative 3.

Again, thanks for taking the time to look at this from the DRA API angle — this is exactly the kind of scrutiny the proposal needs, and the "two parallel ways" point in particular is one the SIG should weigh explicitly. Happy to go deeper on any of the above.

kzks3001:
Apologies for missing the review window. I'm commenting as the operator side this requirement originated from (GPU over APN Testbed — covered in detail in our [KubeCon Japan session](https://events.linuxfoundation.org/kubecon-cloudnativecon-japan/program/schedule/?id=1194272)). I'm not in a position to judge the API-shape questions raised in this thread — that's for the SIG. What I can offer is what this looks like from the operating side, since several premises in the discussion are claims about operators like us.

Our environment: H200 8-GPU nodes with CX7, KubeVirt carving out one VM per tenant, GPUs and VFs passed through. Since the whole point of the testbed is measuring performance, GPU/NIC PIX is a precondition rather than a nice-to-have. Off-PIX, GPUDirect RDMA detours through the CPU and the numbers stop being trustworthy.

The GPU and NIC counts are decided by the tenant. Requests are not uniform — 8 GPU / 8 NIC, 4 GPU / 4 NIC, 4 GPU / 1 NIC. Meanwhile the VirtualMachine resource is authored by us, the operator; tenants never touch the Kubernetes API. So unpredictable requests come in, and the responsibility for satisfying them at PIX lands on our side.

Today we handle this with Jammer Pods ([#17174](https://github.com/kubevirt/kubevirt/issues/17174)): place a Jammer Pod on every free device so they're all held, match up root complexes to pick a PIX set that satisfies the request, then delete only the Jammers on that set to release it. The VM can bind nothing but what's left, so PIX is forced. On our hardware each root complex holds one GPU and one NIC, so a matching root complex settles PIX — which is exactly what `matchAttribute: resource.kubernetes.io/pcieRoot` expresses. That part of the design lines up with what we verify on real hardware.

It works, but it amounts to us steering allocation from the outside as a side effect of Pods, when allocation belongs to the platform. The painful part is that it isn't idempotent: when the node reboots or the virt-launcher pod restarts, the devices we were holding return to the pool and the whole sequence has to run again. Something that inevitably happens in the normal lifecycle calls an operator every single time. And if we pick the wrong GPU/CX7 combination, the symptom shows up late — the VM boots fine, but the performance isn't there. That's the worst case, because if we get it wrong we don't notice until we hand it over to the tenant.

So, concretely, here is what we need from whatever shape this lands in:

1. **A way to express which NIC pairs with which GPU.** When a VM takes every device on the node (8 GPU / 8 NIC) the pairing falls out on its own. But for partial allocations like 4 GPU / 4 NIC or 4 GPU / 1 NIC, the pairing is a choice, and we need to be able to state it. That choice is what we make by hand today.

2. **Mistakes caught at apply time.** We hand-write specs with different counts every time, so mistakes are unavoidable. Finding out at hand-over is far too late, for the reason above. We need to be told when we've written something wrong, not discover it from a performance number.

3. **Per-VM granularity.** Our existing VMs run on the device plugin path and we want to move new VMs over gradually. Both reference the same NAD, because it's the same VLAN. So we need existing and new VMs to coexist on one cluster, sharing NADs, with the choice made per VM. That alpha supports both paths coexisting is useful for our migration plan.

4. **NADs left as they are.** VLAN / MAC / trust live on our NADs. Moving that configuration to another layer would mean rewriting our existing setup, which isn't something we can take on.

**One request:** it would help if the design stated explicitly whether the alignment survives a virt-launcher pod restart (whether the ResourceClaim's lifetime is tied to the VMI or to the Pod). For an environment like ours, where restarts are where this hurts most, that makes it much easier to evaluate.

Thank you for driving this.

Issue #429, opened 2026-08-21
VEP 429: Add emulation preference setting


jschintag:
**Primary contact (assignee)**:

/assign @jschintag 

**Current Feature Stage**: New

**Feature Gate**:
Re-using the same FG as VEP #172
`CrossArchitectureVirtualization`

**Responsible SIGs**: sig-compute

Primary SIG:
/sig compute

Additional SIGs (optional):
none

**Enhancement link**: #430 

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs:
    * [ ] https://github.com/kubevirt/enhancements/pull/430
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:
This VEP refines VEP #172 

> [!IMPORTANT]
> Please keep this description up to date.

jschintag:
Moved back into VEP #172

PR #430, opened 2026-08-21
VEP #429: Add emulation preference setting

jschintag:
Expand VEP #172 with a new setting option to allow users to choose whether to opt-in to emulation.

### VEP Metadata

**Tracking issue**: #429 
**SIG label**: /sig compute

### What this PR does

Propose new VEP for emulation preference setting

This VEP expands the existing VEP #172

### DIFF
--- a/veps/sig-compute/429-emulation-preference/vep.md
+++ b/veps/sig-compute/429-emulation-preference/vep.md
@@ -1,0 +1,234 @@
+# VEP #429: Add emulation preference setting
+
+## VEP Status Metadata
+
+### Target releases
+
+<!--
+A PR must update this section during the planning phase of a given release in order to track it.
+PRs that will not update the VEP during the planning phase will not be able to graduate the
+VEP by creating a code PR to kubevirt/kubevirt to bump the phase in-code.
+
+Please avoid targeting future releases in this section. Only capture the upcoming release.
+For example, during the planning phase for version v1.123, do **not** target beta for v.124 in advance.
+-->
+
+- This VEP targets alpha for version: 1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+To ensure that software emulation remains an opt-in feature, introduce a new setting for maximum allowed emulation tier.
+
+## Motivation
+
+Currently VEP #172 introduced Emulation options to enable hardware architecture independent VMs.
+However since emulation is slower than real hardware, it should always be an opt-in feature.
+To ensure this remains this way when VEP #172 FG becomes active with the beta stage, a new configuration option for emulation is needed.
+
+This also fullfills one of the Beta goals VEP #172 "Consider per-VM emulation policy".
+
+## Goals
+
+- Allow cluster administrators and users to choose their maximum emulation tier
+- Provide sensible defaults
+
+## Non Goals
+
+- Change the current default behaviour of non-emulation VMs
+
+## Definition of Users
+
+**Cluster administrator**: The person configuring the global kubevirt CR
+
+**User**: The VM owner
+
+## User Stories
+
+1. As a cluster administrator, I want to set a cluster-wide policy that
+   cross-architecture VMs **must** use hardware acceleration, so that no VM
+   silently falls back to slow software emulation without my knowledge.
+
+2. As a user, I do not want my VM to use emulation at all, since I need maximum performance
+
+3. As a user, I want to test my new app on other architecture, for which I need emulation.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+A new field `emulationPolicy` is introduced to kubevirt and VM/VMI Spec.
+
+The field controls **which virtualization backends are permitted**
+for a cross-architecture VM. The available tiers, from most restricted to most
+permissive, are:
+
+```
+None (native KVM only)  <  Hardware (KVM only, no TCG)  <  Software (all tiers)
+```
+At Alpha stage, Hardware Emulation is not yet implemented.
+Should it be set anyway the Validation Webhook will reject the change with `Error: Hardware emulation is not yet implemented`.
+
+Each value defines the **maximum fallback depth** allowed:
+
+- **`None`** (default) — no emulation permitted. Only native KVM is allowed.
+  A cross-arch VM with this setting will not start via any form of emulation —
+  it must land on a node that natively runs the guest architecture. This is the
+  safe default: enabling the feature gate does not automatically cause VMs to
+  run under emulation unless explicitly opted in.
+- **`Hardware`** — hardware cross-arch KVM is permitted, but TCG software
+  emulation is not. The VM may use native KVM or hardware cross-arch KVM.
+  It will **not** fall back to software emulation and fails to start on nodes
+  where only software emulation is available.
+- **`Software`** — the full fallback chain is permitted: native KVM > hardware
+  cross-arch KVM > software emulation (TCG). KubeVirt selects the best
+  available tier automatically. Also used as a per-VM override to loosen a
+  cluster-wide `Hardware` policy.
+
+The effective emulation tier is resolved at domain conversion time in
+`KvmDomainConfigurator` using the following precedence (highest first):
+
+1. **Per-VM** — `vmi.Spec.emulationPolicy` (if non-empty)
+2. **Global** — `clusterConfig.GetEmulationPolicy()` (if non-empty)
+3. **Default** — `None` (no emulation; native KVM only)
+
+### Scheduling
+
+Currently there are already 2 different scheduling methods, corresponding to `None` and `Software` tiers.
+
+**None**: Simple node selector using `kubernetes.io/arch`.
+
+**Software**: Introduced with VEP #172 during alpha, removes hard `kubernetes.io/arch` Node selector in
+              favour of affinities with priorities Native > Hardware > Software.
+
+To accomodate the case of allowing hardware emulation but not allowing software emulation, a new scheduling case is introduced:
+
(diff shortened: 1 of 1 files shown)


veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+To ensure that software emulation remains an opt-in feature, introduce a new setting for maximum allowed emulation tier.
+
+## Motivation
+
+Currently VEP #172 introduced Emulation options to enable hardware architecture independent VMs.
```
lyarwood, line 35: Motivation should reference the existing emulation configurable directly. KubeVirt already has a software-emulation opt-in: `developerConfiguration.useEmulation` (default `false`), which controls same-arch TCG fallback when `/dev/kvm` is unavailable. Frame this change as *consolidating* emulation opt-in rather than adding a third knob. It also realizes the deferred item already in #172's Beta goals: "Consider per-VM emulation policy … to disable cross-arch emulation, prefer/require hardware acceleration" — worth citing as evidence of need.

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+
+**User**: The VM owner
+
+## User Stories
+
+1. As a cluster administrator, I want to set a cluster-wide policy that
+   cross-architecture VMs **must** use hardware acceleration, so that no VM
+   silently falls back to slow software emulation without my knowledge.
+
+2. As a user, i do not want my VM to use emulation at all, since i need maximum performance
```
lyarwood: Nit: capitalize "I" here and on line 62.

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+
+3. As a user, i want to test my new app on other architecture, for which i need emulation.
+
+## Repos
+
+https://github.com/kubevirt/kubevirt
+
+## Design
+
+A new field `crossArchEmulation` is introducev to kubevirt and VM/VMI Spec.
```
lyarwood: Two things here:

1. **Rename `crossArchEmulation` → `emulationPolicy`.** The field holds a tiered policy, not a boolean, so an `...Emulation` name reads as a flag. A unified `emulationPolicy` should also **supersede and deprecate `useEmulation`** (the existing same-arch software-emulation opt-in), so we don't entrench two overlapping "allow software emulation" axes. Please specify the `useEmulation` → `emulationPolicy` migration: precedence when both are set, and the deprecation timeline.
2. Typo: "introducev" → "introduced".

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+
+The field controls **which virtualization backends are permitted**
+for a cross-architecture VM. The available tiers, from most restricted to most
+permissive, are:
+
+```
+None (native KVM only)  <  Hardware (KVM only, no TCG)  <  Software (all tiers)
```
lyarwood, line 79: `Hardware` isn't implementable at Alpha. The current `KvmDomainConfigurator` unconditionally sets `domain.Spec.Type = "qemu"` for cross-arch; hardware acceleration (SAE) is not implemented and is Beta/future in #172. Please state that only `None` and `Software` are meaningful at Alpha, and define what happens if a user sets `Hardware` before it exists (reject vs. treat-as-`Software`).

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+- **`Software`** — the full fallback chain is permitted: native KVM > hardware
+  cross-arch KVM > software emulation (TCG). KubeVirt selects the best
+  available tier automatically. Also used as a per-VM override to loosen a
+  cluster-wide `Hardware` policy.
+
+The effective emulation tier is resolved at domain conversion time in
+`KvmDomainConfigurator` using the following precedence (highest first):
+
+1. **Per-VM** — `vmi.Spec.crossArchEmulation` (if non-empty)
+2. **Global** — `clusterConfig.GetcrossArchEmulation()` (if non-empty)
```
lyarwood: `clusterConfig.GetcrossArchEmulation()` doesn't exist (and the casing is off). The established accessors are `ClusterConfig.AllowEmulation()` and `GetDeveloperConfigurationUseEmulation()` in `pkg/virt-config/virt-config.go` — a new accessor should follow that pattern. Also, resolving "at domain conversion time in `KvmDomainConfigurator`" needs the plumbing spelled out: the per-VM value is on the VMI passed to `Configure()`, but the global value must be threaded through `ConverterContext` the same way `AllowCrossArchEmulation` is today (see `converter-context.go` and how it's wired in `converter.go`).

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+VM configuration:
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachine
+metadata:
+  name: arm64-dev-vm
+spec:
+  architecture: arm64
+  crossArchEmulation: Software # Allows software emulation, overrides cluster policy
+  running: true
```
lyarwood: Nit: `running: true` is deprecated — use `runStrategy` in the example.

veps/sig-compute/429-emulation-preference/vep.md
```diff
@@ -0,0 +1,186 @@
...
+
+Downgrading to a previous version does not affect existing VMs, new VMs can't use the feature.
+
+Upgrading does not affect existing or new VMs, unless the setting is changed.
+
+## Functional Testing Approach
+
+The design should be tested in unit-test.
+
+Additionally e2e should cover both Software and None settings in e2e of VEP #172 to ensure the resulting pod affinities work as intended.
```
lyarwood, line 212: This is the only mention of scheduling, but `None` fundamentally changes #172's scheduling model: #172 relaxes the `kubernetes.io/arch` selector and adds a `kubevirt.io/vm-arch-<arch>` hard selector that *includes* TCG-only nodes. Under `None`, the pod must be constrained to natively-capable (or hardware-accel) nodes and **exclude** software-emulation-only nodes. Please add a Scheduling subsection to the Design describing the node-selector/affinity impact per policy value, rather than only the domain-conversion-time resolution.

lyarwood (commented):
Thanks for pushing this — making cross-arch emulation an explicit opt-in is the right instinct, and I agree it needs to be a first-class setting rather than an implicit consequence of the feature gate. My main recommendation is structural; the specifics are inline.

## Fold this into VEP #172 rather than making it a standalone VEP

The PR body says this "expands the existing VEP #172," and I don't think the content can stand on its own: the emulation policy only means anything *because of* cross-arch virtualization, and it reshapes #172's own design (scheduling — see inline). Per the new [VEP writing guide (`docs/vep-writing-guide.md`)](https://github.com/kubevirt/enhancements/pull/431):

> `vep.md` is the single source of truth ... Never split a proposal across multiple content-bearing files, and never redirect a reader elsewhere to understand what the feature is.
> As the VEP progresses (Alpha → Beta → GA), edit `vep.md` in place.

A separate VEP creates exactly the "redirect a reader elsewhere" problem — #172 would be incomplete without this to explain how you opt in. A supporting/appendix doc in the #172 directory isn't right either; the guide scopes extra files to non-essential material (research, POC, benchmarks), whereas this is the primary user-facing API.

#172 is still in Alpha (beta/GA targets blank), so the guide's "edit in place + record in Implementation History" path fits. I'd suggest repurposing this PR to edit `veps/sig-compute/172-cross-arch-virtualization/vep.md` and folding the #429 tracking issue into #172's tracker.

The inline comments still apply once the content lives in #172.

veps/sig-compute/429-emulation-preference/vep.md
jschintag: I have added a statement describing what should happen during alpha when `Hardware` is selected.

veps/sig-compute/429-emulation-preference/vep.md
jschintag: You are right, i have renamed the field.

veps/sig-compute/429-emulation-preference/vep.md
jschintag: Done

veps/sig-compute/429-emulation-preference/vep.md
jschintag: I added the reference to the Beta goal.

However i have to admit i'm no expert in the no kvm TCG Fallback. Does it make sense to bundle it with cross-arch software emulation? I only know that in VEP #172 you purposefully kept the code path separate.

veps/sig-compute/429-emulation-preference/vep.md
jschintag: Fixed

veps/sig-compute/429-emulation-preference/vep.md
jschintag: Added a section for scheduling.

jschintag:
Addressed most review comments.

>Fold this into VEP https://github.com/kubevirt/enhancements/issues/172 rather than making it a standalone VEP

@lyarwood I moved this out of VEP 172 after the feedback i received in the VEP Meeting last Tuesday that the VEP is getting to long for reviewer and it might be better to split it into more VEP like DRA is. I can move it back as well, both work for me. What would be the preference here?

lyarwood:
> Addressed most review comments.
> 
> > Fold this into VEP #172 rather than making it a standalone VEP
> 
> @lyarwood I moved this out of VEP 172 after the feedback i received in the VEP Meeting last Tuesday that the VEP is getting to long for reviewer and it might be better to split it into more VEP like DRA is. I can move it back as well, both work for me. What would be the preference here?

This is still core to the feature so I'd prefer to keep it together under the  original VEP, keep the VEP itself concise as possible and break any supporting material out into documents outside of the core vep.md. Apologies for the back and forth here but it's easier to review and reason against this way IMHO.

jschintag:
@lyarwood No worries, i can put it back into the original PR and just close the tracking issue.

Should i put it into https://github.com/kubevirt/enhancements/pull/421 or open a separate PR?

lyarwood:
> @lyarwood No worries, i can put it back into the original PR and just close the tracking issue.
> 
> Should i put it into #421 or open a separate PR?

Yeah feel free to combine into one under #421.

jschintag:
Closed in favour of #421

PR #431, opened 2026-08-21
docs: Add a VEP writing guide

EdDev:
## What this PR does

Adds `docs/vep-writing-guide.md`, a section-by-section guide on how to write a good VEP. The template (`veps/NNNN-vep-template/vep.md`) defines which sections a VEP needs; nothing documented how to write each one well, so the same feedback (document structure, framing goals as outcomes not mechanisms, testing rigor, upgrade/rollback reasoning, etc.) kept getting re-explained PR after PR.

The guide is grounded in recurring review feedback from recent VEP PRs and cross-checked against Kubernetes' KEP writing conventions, adapted to KubeVirt's existing VEP template (the template's section skeleton is unchanged).

- Linked from the VEP Creation step in `README.md`.
- Added to CodeRabbit's knowledge base (`.coderabbit.yaml`) so it's applied automatically during VEP review, alongside the existing `README.md`, `vep.md`, and `docs/feature-lifecycle.md` entries.

## For Reviewers

This is expected to evolve — treat gaps as "not yet captured" rather than exhaustive, and propose additions via PR when a review surfaces a recurring issue not yet in the guide.

### DIFF
--- a/docs/vep-writing-guide.md
+++ b/docs/vep-writing-guide.md
@@ -1,0 +1,172 @@
+# Writing a Good VEP
+
+This guide complements [`vep.md`](../veps/NNNN-vep-template/vep.md), the VEP template. The
+template defines which sections a VEP must have; this document is about how to write each one
+well.
+
+This guide is expected to evolve. Propose changes via PR when a review surfaces a recurring
+issue not yet captured here.
+
+## Document Structure
+
+- Keep the template's section set and order intact. Every section must be addressed — write
+  "N/A" with a one-line reason if it doesn't apply.
+- `vep.md` is the single source of truth: all feature-level content (motivation, design, API,
+  graduation) lives in it. Never split a proposal across multiple content-bearing files, and
+  never redirect a reader elsewhere to understand what the feature is.
+- As the VEP progresses (Alpha → Beta → GA), edit `vep.md` in place. Do not create a
+  phase-specific file (`foo-beta.md`, `foo-part2.md`); record the change in Implementation
+  History instead.
+- Extra files in the VEP directory are for material not required to understand the feature —
+  raw research, a POC report, benchmark data, an appendix. If a reader needs the extra file to
+  understand the feature or its design, it belongs in `vep.md` instead.
+
+## Overview
+
+- A short summary that lets a reader grasp the VEP's core topic and what problem it resolves —
+  not a repeat of Motivation or Design.
+- Avoid describing a specific mechanism or solution (the "how"); that belongs in Design.
+
+## Motivation
+
+- Explain why this matters now: the cost of the status quo and the benefit to users — not just
+  "we need X."
+- Back the claim with evidence where possible (bug reports, user requests, precedent) rather
+  than asserting need.
+- Don't restate Goals as prose.
+
+## Goals
+
+- State an outcome (what/why) only, never a mechanism (how) — the mechanism belongs in Design.
+- Not phased — don't write "Alpha: X, Beta: Y" here; stage scope belongs in Graduation
+  Requirements.
+- Each goal should be measurable: it must be possible to say whether it was met.
+
+## Non Goals
+
+- Justify each exclusion; "less work" isn't a reason on its own. A smaller scope is fine if
+  what's left is still viable and answers the real user need (an MVP); it's not fine if what's
+  left only works once the missing pieces are filled in later — whether by a follow-up VEP or
+  by updating this same VEP toward Beta/GA. Goals and Non Goals describe the feature's end
+  result, not an intermediate step. For example, excluding live-migration support needs strong
+  justification on its own — it's core to KubeVirt, and a feature that doesn't account for it
+  isn't viable just because it's smaller in scope.
+- A Non Goal usually just means out of scope for this VEP, with no claim about the future
+  either way. If you're actively rejecting an idea rather than merely leaving it out, say why
+  explicitly.
+
+## Definition of Users
+
+- Identify who is affected — a role, not a scenario. Save the concrete story for User Stories.
+
+## User Stories
+
+- Concrete, scenario-based ("As an admin managing X, I need Y so that Z"), not abstract
+  capability lists.
+- Multiple stories are fine when they show genuinely different needs or personas; don't pad
+  with near-identical variations of the same story.
+- Stop at what the user needs/experiences — implementation detail belongs in Design.
+
+## Design
+
+- This is where "how" lives; anything mechanism-level found elsewhere belongs here.
+- Describe what components do and how they interact — boundaries, interfaces, data — not the
+  code that implements it. A precise sequence of calls or a wire format that resolves real
+  ambiguity is still design; a step-by-step code walkthrough is not.
+- Avoid file and function names unless one is literally an API field or value — the code
+  changes constantly, and the VEP should describe a high-level abstraction that survives
+  refactoring.
+- When naming an API field or value, consider whether a generic term would future-proof it
+  against other backing mechanisms (e.g. `podStatus` instead of `multusStatus`, if the field
+  could plausibly apply beyond Multus). This doesn't mean avoiding the feature's own subject
+  technology — a VEP about `passt` still says `passt`.
+- Before introducing a new API mechanism, check whether an existing one already covers the same
+  need — reuse or extend it instead of adding a parallel path. If some duplication is genuinely
+  unavoidable (e.g. mirroring a field between the VM and VMI to resolve a lifecycle constraint),
+  justify it explicitly.
+- State the reasoning behind a choice when a reviewer could reasonably ask "why this way?";
+  skip it for trivial choices.
+- Check prior art (Kubernetes extension mechanisms, related KEPs) before designing something
+  novel.
+- Cross-reference instead of repeating content from another section.
+- Include risks and mitigations covering security, UX, and operational impact. Name the
+  specific risk and how it's mitigated; don't write a generic "there are risks" statement.
+
+## API Examples
+
+- Concrete request/response or CR examples, not prose description of fields.
+- An edge/error case is encouraged when it clarifies the contract, but not mandatory — the
+  normal case is the minimum bar.
+
+## Alternatives
+
+- An option must address the same goals as the chosen approach to count as an alternative; one
+  that only serves something already declared a Non Goal is out of scope, not an alternative.
+  For an option that does qualify, don't dismiss it by citing a Non Goal — that defines what
+  this VEP isn't attempting, not why a working alternative is worse. Reject it on its own
+  merits (cost, risk, complexity, fit).
+- Trading off a secondary goal doesn't disqualify an alternative; judge it against the primary
+  goal(s) it does address.
+- Ground the comparison in reference data/implementations from comparable projects where
+  possible.
+
+## Scalability
+
+- Quantify: call type, estimated frequency/throughput, originating component, for each new or
+  changed API interaction.
+- Call out any new list/watch behavior against existing resources explicitly. Avoiding this is
+  a common source of unreviewed scale regressions.
+
+## Update/Rollback Compatibility
+
+- Upgrades happen via live-migration: VMs launched under an older version keep running while
+  components upgrade underneath them. Address what happens to a VM that predates the change —
+  must it be migrated/restarted to pick up new behavior, or must the new code correctly handle
+  state/domain XML produced by the older version?
+- Cover version skew between coexisting components during a rolling upgrade, not just an
+  idealized atomic upgrade.
+- KubeVirt doesn't support rollback (downgrading a cluster to an older version) today — say so
+  plainly rather than describing theoretical support.
+- Still worth reasoning through what a full rollback would require if it were supported: would
+  it leave state or behavior an older component couldn't handle? This can surface real risks
+  even absent formal rollback support.
+- Separately — and distinct from a cluster-version rollback — address feature discontinuation
+  while still Alpha/Beta: if the feature or its Feature Gate is dropped, what happens to
+  VMs/resources already created under it? This matters most at Beta, where the Feature Gate
+  defaults to enabled, so users may be affected without having explicitly opted in.
+
+## Functional Testing Approach
+
+- Be precise about test tiers used (unit, integration, functional/e2e) and which claim each
+  one covers; don't loosely call something "e2e."
+- Unit tests are a baseline requirement for new code — call out and justify any exception
+  explicitly; don't rely on higher-level tests to cover what a unit test should.
+- For functional/e2e coverage, assert user-facing/API-level behavior, not internals (e.g.
+  domain XML) — internals violate layering.
+- Lead with the real happy-path scenario before edge cases.
+- This section verifies stated functionality; it isn't where new functionality gets defined.
+- All testing must show a sustained non-flaky signal over time, not just "tests exist."
+
+## Implementation History
+
+`vep.md` is edited in place as the feature moves through Alpha/Beta/GA rather than rewritten
+from scratch — this section is what keeps that evolution visible over time.
+
+- Terse, dated log of actual milestones (PR merged, alpha shipped, graduation approved).
+- Not a forward plan, and not narrative about why the document itself changed shape.
+
+## Graduation Requirements
+
+- Each stage's criteria must be objective and independently verifiable (a metric, test, or
+  documented behavior) — no "stable enough."
+- Functional tests: encouraged, not required, for Alpha; required for Beta and GA. Alpha's
+  relaxed bar applies only inside the feature's own Feature-Gated path; it must not introduce
+  risk to existing, already-stable functionality.
+- Skipping Alpha, or omitting a Feature Gate, needs strong explicit justification.
+- Beta must resolve all known functional/security/monitoring gaps — nothing user-facing left
+  open.
+- `On-By-Default Readiness` (Beta) exists because Beta Feature Gates default to on: state
+  concretely what must be true for that to be safe (known gaps closed, monitoring in place,
+  a way to disable it if it misbehaves) — not just "it's ready."
+- GA must leave no functional/security/testing item outstanding, including full removal of the
+  Feature Gate.
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -16,4 +16,5 @@ knowledge_base:
     filePatterns:
       - README.md
       - veps/NNNN-vep-template/vep.md
+      - docs/vep-writing-guide.md
       - docs/feature-lifecycle.md
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ a common process to ensure synchronization between SIGs and uniformity of the pr
 
 2. **VEP Creation**: VEP authors will initiate proposals via PRs to the `kubevirt/enhancements` repository.
    See the [VEP template](https://github.com/kubevirt/enhancements/blob/main/veps/NNNN-vep-template/vep.md),
-   each section needs to be filled even if the section doesn't apply to the VEP. The Issue's ID is used to for the VEP ID, title of the PR should start with "VEP-<ID>"
+   each section needs to be filled even if the section doesn't apply to the VEP. The Issue's ID is used to for the VEP ID, title of the PR should start with "VEP-<ID>".
+   See the [writing guide](https://github.com/kubevirt/enhancements/blob/main/docs/vep-writing-guide.md) for guidance on what makes each section good.
 
 3. **SIG Review and Collaboration**: Although each VEP can have multiple target SIGs, it needs to have a single owning SIG.
    In case of a cross SIG feature, the most relevant SIG should be chosen.


docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Explain why this matters now: the cost of the status quo and the benefit to users — not just
+  "we need X."
+- Back the claim with evidence where possible (bug reports, user requests, precedent) rather
+  than asserting need.
+- Don't restate Goals as prose.
+
+## Goals
+
+- State outcome (what/why) only, never mechanism (how) — mechanism belongs in Design.
```
dankenigsberg:
```suggestion
- State an outcome (what/why) only, never a mechanism (how) — the mechanism belongs in Design.
```

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Not phased — don't write "Alpha: X, Beta: Y" here; stage scope belongs in Graduation
+  Requirements.
+- Each goal should be measurable: it must be possible to say whether it was met.
+
+## Non Goals
+
+- Justify each exclusion; "less work" isn't a reason.
```
dankenigsberg: Why isn't scope limitation a good reason? I like delivering an MVP.

Maybe you can provide examples for good reasons?

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- This is where "how" lives; anything mechanism-level found elsewhere belongs here.
+- Avoid file and function names unless one is literally an API field or value — the code
+  changes constantly, and the VEP should describe a high-level abstraction that survives
+  refactoring.
+- When naming an API field or value, consider whether a generic term would future-proof it
+  against other backing mechanisms (e.g. `podStatus` instead of `multusStatus`, if the field
+  could plausibly apply beyond Multus). This doesn't mean avoiding the feature's own subject
+  technology — a VEP about `passt` still says `passt`.
+- If the same need could be met by two parallel API mechanisms, pick one or state the plan to
+  unify them — don't leave both unresolved.
```
dankenigsberg: "unify" suggests that the two mechanisms preexist. but vep is about introducing a new mechanism to perform a functionality. I don't get this item.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- State the reasoning behind non-obvious choices explicitly.
+- Cross-reference instead of repeating content from another section.
+- Include risks and mitigations: security, UX, and operational — not just "what could break."
```
dankenigsberg: no idea about the meaning of the end of your sentence ` — not just "what could break."`

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- An edge/error case is encouraged when it clarifies the contract, but not mandatory — the
+  normal case is the minimum bar.
+
+## Alternatives
+
+- Must address the same goals as the chosen approach. An option that only serves something
+  already declared a Non Goal isn't an alternative — it's out of scope.
+- Trading off a secondary goal doesn't disqualify an alternative; judge it against the primary
+  goal(s) it does address.
+- Don't reject an alternative just by pointing to a Non Goal — Non Goals define what this VEP
```
dankenigsberg: isn't this a repetition of the first list item (in clearer words, imo)

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+  isn't attempting, not a reason a working alternative is worse. Reject it on its own merits
+  (cost, risk, complexity, fit) instead.
+- Check prior art (Kubernetes extension mechanisms, related KEPs) before proposing something
+  novel.
```
dankenigsberg: this belongs to the Design section, imo. The chosen mechanism should learn from prior art, even more than the alternatives.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Ground the comparison in reference data/implementations from comparable projects where
+  possible.
+
+## Scalability
+
+- Quantify: call type, estimated frequency/throughput, originating component, for each new or
+  changed API interaction.
+- Call out any new list/watch behavior against existing resources explicitly — a common source
```
dankenigsberg:
```suggestion
- Call out any new list/watch behavior against existing resources explicitly. Avoiding this is a common source
```

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Be precise about test tiers used (unit, integration, functional/e2e) and which claim each
+  one covers; don't loosely call something "e2e."
+- Unit tests are a baseline requirement for new code — call out and justify any exception
+  explicitly; don't rely on higher-level tests to cover what a unit test should.
+- For functional/e2e coverage, assert user-facing/API-level behavior, not internals (e.g.
+  domain XML) — internals violate layering.
+- Lead with the real happy-path scenario before edge cases.
+- This section verifies stated functionality; it isn't where new functionality gets defined.
+- GA testing must show sustained non-flaky signal over time, not just "tests exist."
```
dankenigsberg:
```suggestion
- All testing must show a sustained non-flaky signal over time, not just "tests exist."
```

dankenigsberg (changes requested):
This is a very important addition. Please pull it through. But why is it a separate document? Can we add the explanation of a section as an

> indented text

of the template itself?

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+## Document Structure
+
+- Keep the template's section set and order intact. Every section must be addressed — write
+  "N/A" with a one-line reason if it doesn't apply.
+- `vep.md` is the single source of truth: all feature-level content (motivation, design, API,
+  graduation) lives in it. Never split a proposal across multiple content-bearing files, and
+  never redirect a reader elsewhere to understand what the feature is.
+- As the VEP progresses (Alpha → Beta → GA), edit `vep.md` in place. Do not create a
+  phase-specific file (`foo-beta.md`, `foo-part2.md`); record the change in Implementation
+  History instead.
```
0xFelix, line 19: It sometimes proves difficul  to keep track of an evolving feature and the history of all of its changes in a single file.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Each goal should be measurable: it must be possible to say whether it was met.
+
+## Non Goals
+
+- Justify each exclusion; "less work" isn't a reason.
+- Distinguish deferred (a future VEP will cover it) from rejected (explain why, or point to
+  Alternatives).
+
+## Definition of Users / User Stories
+
```
0xFelix, line 63: Add item along the lines of: "Do not repeat user stories in the definition of users; clearly separate definition of users and their intents"

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Avoid file and function names unless one is literally an API field or value — the code
+  changes constantly, and the VEP should describe a high-level abstraction that survives
+  refactoring.
+- When naming an API field or value, consider whether a generic term would future-proof it
+  against other backing mechanisms (e.g. `podStatus` instead of `multusStatus`, if the field
+  could plausibly apply beyond Multus). This doesn't mean avoiding the feature's own subject
+  technology — a VEP about `passt` still says `passt`.
+- If the same need could be met by two parallel API mechanisms, pick one or state the plan to
+  unify them — don't leave both unresolved.
+- State the reasoning behind non-obvious choices explicitly.
```
0xFelix: How do you define non-obvious vs obvious choice, IMO that's very subjective.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+## API Examples
+
+- Concrete request/response or CR examples, not prose description of fields.
+- An edge/error case is encouraged when it clarifies the contract, but not mandatory — the
+  normal case is the minimum bar.
+
+## Alternatives
+
+- Must address the same goals as the chosen approach. An option that only serves something
+  already declared a Non Goal isn't an alternative — it's out of scope.
```
0xFelix: Can't this section be used to explain why a certain goal was rejected, i.e. explaining it is not feasible to do so? Although the author may have like to address this goal as well.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+## Update/Rollback Compatibility
+
+- Upgrades happen via live-migration: VMs launched under an older version keep running while
+  components upgrade underneath them. Address what happens to a VM that predates the change —
+  must it be migrated/restarted to pick up new behavior, or must the new code correctly handle
+  state/domain XML produced by the older version?
+- Cover version skew between coexisting components during a rolling upgrade, not just an
+  idealized atomic upgrade.
+- KubeVirt doesn't support rollback (downgrading a cluster to an older version) today — say so
+  plainly rather than describing theoretical support.
```
0xFelix, line 129: Why add this information at all then, if rollback is not supported?

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+## Scalability
+
+- Quantify: call type, estimated frequency/throughput, originating component, for each new or
+  changed API interaction.
+- Call out any new list/watch behavior against existing resources explicitly — a common source
+  of unreviewed scale regressions.
+
+## Update/Rollback Compatibility
+
+- Upgrades happen via live-migration: VMs launched under an older version keep running while
```
0xFelix, line 122: This reads very VM centric, what about KubeVirt's components?

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Terse, dated log of actual milestones (PR merged, alpha shipped, graduation approved) — not a
+  forward plan, and not narrative about why the document itself changed shape.
+
+## Graduation Requirements
+
+- Each stage's criteria must be objective and independently verifiable (a metric, test, or
+  documented behavior) — no "stable enough."
+- Functional tests: encouraged, not required, for Alpha; required for Beta and GA.
+- Skipping Alpha, or omitting a Feature Gate, needs strong explicit justification.
```
0xFelix, line 165: ... and agreement from the community? Preferably discussed beforehand?

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+- Beta must resolve all known functional/security/monitoring gaps — nothing user-facing left
```
0xFelix, line 166: We should strive to make all features / code secure, even if it is Alpha. This reads as if it's okay to ship insecure code in Alpha.

docs/vep-writing-guide.md
EdDev: Extra files options is described below.
Core feature details need IMO to live in one central document and be up to day until GA.
I see no other what to manage it without loosing context and seeing the feature changes as a whole.

Once GA, its tracker is closed and new changes, if substantial (e.g. API evolving) will come through a new VEP.

This is pretty much how Kubernetes handles it per my understanding.
If you see other cases, have other ideas, or have KEPs examples please share so we can evaluate.

I tried to make it clearer in the history section.

docs/vep-writing-guide.md
EdDev: The point of the bullet is to stop Non Goals being used to shrink the feature into something that doesn't stand on its own, more PoC than real solution. An MVP that still answers the real user need at a smaller scope is completely fine. What's not fine is carving out a partial slice that only becomes useful once some other, unwritten VEP fills the gap.

This is actually one of the repeating patterns that I have seen, several VEPs aim at each stage (Alpha, Beta) instead of focusing on the end goal, making it hard to reason and understand if the feature is feasible in the first place.

I'll clarify this.

> Maybe you can provide examples for good reasons?

Added.

docs/vep-writing-guide.md
EdDev: The intent was to avoid cases where a new API is suggested when an existing one already exists which is closely related.
I would prefer such duplication to be avoided, or at least, reasoned in detail on why no other choice is available.

The last case I saw is duplicating a configuration between the VM and VMI, to resolve a lifecycle challenge.

I rephrased to try and express this better.

docs/vep-writing-guide.md
EdDev: I wanted to be less strict and allow trivial things to be suggested without too much text.
But it is a fair point to say it is subjective.

I can remove "non-obvious" if it helps.
Do you have a suggestion?

For now I rewarded a bit, but it is not bullet proof from the subjective concern.

docs/vep-writing-guide.md
EdDev: I tried to make it clearer.
The intention was to avoid using such generic sentences in VEPs and be explicit.

docs/vep-writing-guide.md
EdDev: Alternatives are for solutions at the design/implementation section. Goals are part of the requirements that the feature is aiming to solve.

If  one wants to explain why something is not a goal, that belongs to the requirements part. It can be explained in the non-goal section.

IMO the implementation/solution needs to answer the goals and the goals should not (formally) depend on the implementation. I agree that while brainstorming ideas, we negotiate the requirements, but once a decision is taken, I find the goals-solution relation uni-directional.

docs/vep-writing-guide.md
EdDev: Yes, they are related. I merged them.

docs/vep-writing-guide.md
EdDev: I think the more general case is expressed in the next bullet.

In practice, the VMs are one of the most effected components in the system while upgrading.
Most of the challenges (and misses( are with them.

I can change the order of the items and move this down a bit.
Do you have a suggestion on what to add or change in this regard?

docs/vep-writing-guide.md
EdDev: The two bullets below cover it: reasoning about a hypothetical rollback can surface real risks even if we never ship it, and feature discontinuation during Alpha/Beta (dropping the FG) is a separate, real case that does happen.

docs/vep-writing-guide.md
EdDev: The VEP is a proposal to the Kubevirt community and an open discussion. I do not think we need other medias or bureaucracy in this regard.

docs/vep-writing-guide.md
EdDev: On one hand, Alpha is expected to be protected by a FG and users who decide to enable it should consider that it is under experimentation. That means it is not bullet proof in any regards, it is not ready for production.
On the other hand, I agree that for security reasons, we should be careful not to expose the project to risks. This may happen if the code change is not protected by the FG.

I think that the overall security should be examined in general and if an Alpha feature exposes degradation of any king to other features, it should not be accepted. That is evaluated by existing coverage.
However, I find it hard to require the same from Alpha features. Just because we accept they can be added without functional tests, therefore we acknowledge they are not safe. We also state that they are not production ready.

Do you have a suggestion on what to add/change to make it safer per your feedback?
For now, I rephrased the functional tests bullet to make it clear.

docs/vep-writing-guide.md
EdDev: Done

docs/vep-writing-guide.md
EdDev: I split this one to make it clearer and added the suggested points.
I also reworded the "One read use case..." to clarify what was my intention.
Done

docs/vep-writing-guide.md
iholder101: (EDIT: wrote this before @EdDev comment appeared.)

I think what @EdDev means is: if you want to capture the history of the feature as an additional material for someone who wants a deep-dive into the feature - that's fine. But if this information is **essential** in order to understand the current design, it has to be inside `vep.md`.

docs/vep-writing-guide.md
EdDev: Done

docs/vep-writing-guide.md
EdDev: Done

docs/vep-writing-guide.md
EdDev: Done

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,148 @@
...
+  capability lists.
+- One real use case per point is enough; don't multiply near-identical variations.
+- Stop at what the user needs/experiences — implementation detail belongs in Design.
+
+## Design
+
+- This is where "how" lives; anything mechanism-level found elsewhere belongs here.
+- Avoid file and function names unless one is literally an API field or value — the code
+  changes constantly, and the VEP should describe a high-level abstraction that survives
+  refactoring.
```
iholder101, line 78: I agree.

In the spirit of this we can consider adding another bullet that says something like "avoid writing the implementation here. Focus on outlining the key design characteristics, interface APIs, component boundaries, etc". WDYT?

iholder101 (commented):
Overall looks very good. Thanks @EdDev !

EdDev (commented):
Thank you for the reviews @dankenigsberg , @0xFelix .

> This is a very important addition. Please pull it through. But why is it a separate document? Can we add the explanation of a section as an
> 
> > indented text
> 
> of the template itself?

I think the original intention was for the template to stay a clean, minimal skeleton people copy as-is. Once you start putting paragraphs of guidance under every heading, it becomes a wall of text to scroll past (and delete) every time someone starts a new VEP. It also has nowhere to hold the cross-cutting stuff, like keeping everything in one vep.md and not splitting into phase-specific files, since that isn't tied to any single section.

A separate doc can evolve on its own without touching the template every time we learn something new, and it's still one click away, linked from the README 
and fed to CodeRabbit the same way the template is.

If you have a concrete case where inline guidance worked better elsewhere (a KEP, say), happy to reconsider. For now I'd rather keep the template lean.

BTW, I also considered to place it close to the template, but it does not fit well with VEPs content IMO.

EdDev:
[change](https://github.com/kubevirt/enhancements/compare/84deeac0b98536e505c0a7aa5a322516428cf33b..48eed78ba4d08137c15f82a51f31445fcf633235): Responded to @dankenigsberg and @0xFelix reviews.

docs/vep-writing-guide.md
EdDev: I added a bullet to express this, thanks.
I tried to balance it in order not to hint that detail logic is not to be described based on how Kubevirt operates, hope I found the right balance.

EdDev:
[change](https://github.com/kubevirt/enhancements/compare/48eed78ba4d08137c15f82a51f31445fcf633235..f723f383062e36065f013c62c10b296273bdaba2): Answered @iholder101 [comment](https://github.com/kubevirt/enhancements/pull/431#discussion_r3852598289).

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,172 @@
...
+  code that implements it. A precise sequence of calls or a wire format that resolves real
+  ambiguity is still design; a step-by-step code walkthrough is not.
+- Avoid file and function names unless one is literally an API field or value — the code
+  changes constantly, and the VEP should describe a high-level abstraction that survives
+  refactoring.
+- When naming an API field or value, consider whether a generic term would future-proof it
+  against other backing mechanisms (e.g. `podStatus` instead of `multusStatus`, if the field
+  could plausibly apply beyond Multus). This doesn't mean avoiding the feature's own subject
+  technology — a VEP about `passt` still says `passt`.
+- Before introducing a new API mechanism, check whether an existing one already covers the same
```
dankenigsberg, line 83:
```suggestion
- Before introducing a new API mechanism, check whether an existing one already covers a similar
```

docs/vep-writing-guide.md
dankenigsberg: I, too, find the text about hypothetical downgrade a bit confusing. Disabling a feature, or an ancient node coming up live after an upgrade are more realistic. I agree that thinking about what would happen on a downgrade can help us envisage new->old communication that is important for stability, but the text here should be rewritten. (If I had a good idea I would have given it. Maybe you can start with the following bullets?)

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Graduation Requirements
+
+- Each stage's criteria must be objective and independently verifiable (a metric, test, or
+  documented behavior) — no "stable enough."
+- Functional tests: encouraged, not required, for Alpha; required for Beta and GA. Alpha's
+  relaxed bar applies only inside the feature's own Feature-Gated path; it must not introduce
+  risk to existing, already-stable functionality.
+- Skipping Alpha, or omitting a Feature Gate, needs strong explicit justification.
+- Beta must resolve all known functional/security/monitoring gaps — nothing user-facing left
```
dankenigsberg, line 166: how about

```suggestion
- Beta must resolve all functional/security/monitoring gaps that have become known during Alpha — nothing user-facing left
```

dankenigsberg (changes requested):

docs/vep-writing-guide.md
0xFelix: Got it! Ack

docs/vep-writing-guide.md
0xFelix: I think you are right in saying that a VM (user workload) is the most important thing to focus on during an update or a rollback. Reading again, I think it is fine.

docs/vep-writing-guide.md
```diff
@@ -0,0 +1,172 @@
...
+  VMs/resources already created under it? This matters most at Beta, where the Feature Gate
+  defaults to enabled, so users may be affected without having explicitly opted in.
+
+## Functional Testing Approach
+
+- Be precise about test tiers used (unit, integration, functional/e2e) and which claim each
+  one covers; don't loosely call something "e2e."
+- Unit tests are a baseline requirement for new code — call out and justify any exception
+  explicitly; don't rely on higher-level tests to cover what a unit test should.
+- For functional/e2e coverage, assert user-facing/API-level behavior, not internals (e.g.
```
0xFelix, line 144: Also try to conserve resources, e2e tests come at a cost. Try to cover as many aspects as reasonably possible in unit tests first.

docs/vep-writing-guide.md
0xFelix: Nevermind about the second part of my comment. But I wanted to express that justification alone is not sufficient. The community still needs to agree.

docs/vep-writing-guide.md
0xFelix: I like Dan's suggestion!

Issue #300, opened 2026-04-29
VEP 300: Device manager like policies for DRA resources


alaypatel07:
With #10, #183, and #152, KubeVirt now has enough DRA-backed device types to justify a KubeVirt-level DRA claim manager with policy-based claim generation.

The goal of this VEP is to improve the user experience by allowing users to express higher-level workload intent in the VMI spec, while KubeVirt translates that intent into concrete `ResourceClaim` / `ResourceClaimTemplate` objects. This would be similar in spirit to kubelet’s CPU manager or device manager: users describe what the workload needs, and the platform applies policy to produce the lower-level resource allocation shape.

For KubeVirt, the interesting part is that the policy may span multiple VMI-level resources at once: GPU, network devices, CPU, host devices, and potentially storage-related lifecycle needs. Examples include placing a GPU and NIC on the same PCIe root, aligning CPU and device topology, or eventually expressing storage/resource-claim lifecycle behavior once Kubernetes has the required primitives.

This VEP should explore a future `managedClaim` / `autoClaim` style API that lets KubeVirt own the generated claim lifecycle and policy translation, while preserving the stable consumer contract introduced by the DRA VEPs: individual VMI resource consumers continue to reference allocated resources through `(claimName, requestName)`.

**Primary contact (assignee)**:

/assign @alaypatel07 

**Current Feature Stage**:

**Feature Gate**:

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
/sig network
/sig storage

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/432
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

p.s. even if the VEP is currently marked as "Tracked", a "VEP Reviewer" is a requirement. Without it, the VEP will be not remain tracked for v1.10.

[...]

PR #432, opened 2026-08-21
VEP 300: Managed DRA Claims

johnahull:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/300
**SIG label**: /sig compute

### What this PR does

Adds VEP-300, a design for automated DRA (Dynamic Resource Allocation) ResourceClaim generation in KubeVirt.

- Admins define a `ManagedClaimProvisioner` CRD encoding DeviceClass mappings and the provisioner controller responsible for claim generation.
- Users reference the provisioner by name and declare their devices (`gpus[]`, `hostDevices[]`, `networks[]`, `cpu.dra`) as they do today, without hand-authoring `ResourceClaim` objects or knowing DeviceClass names.
- A reusable managed-claim provisioning framework generates the `ResourceClaim`, with a built-in topology-aligning provisioner (`policy.kubevirt.io/aligner`) and support for independent, third-party provisioner controllers.
- Claim names are derived deterministically from the VMI, so pod creation is decoupled from claim reconciliation (no status handshake).

- This proposal incorporates the managed-claim provisioning framework redesign contributed by Alay Patel (co-author).
- Scope boundaries: device-type support builds on VEP-152 (CPUs with DRA) and coordinates with VEP-183 (network DRA); those VEPs own their respective device semantics.
- The design relies on upstream KEP-6072 (standard `numaNode` attribute) for topology alignment and notes KEP-5491 (list-typed attributes) as a follow-on.

### DIFF
--- a/veps/sig-compute/300-managed-dra-claims/vep.md
+++ b/veps/sig-compute/300-managed-dra-claims/vep.md
@@ -451,8 +451,20 @@ The framework provides this common behavior to provisioner controllers.
 
 **Claim naming:** the generated claim is named
 `<vmi-name>-<claim-name>`. If this exceeds the 253-character DNS subdomain
-limit, the name is truncated and a short hash suffix is appended to
-preserve uniqueness.
+limit, the name is built from a fixed-length prefix of both the VMI name
+and the claim name plus a short hash suffix (up to 123 characters of each
+name joined with a 5-character SHA-256 digest of the full untruncated
+name). Taking a prefix of both names keeps each recognizable even when one
+is long, and the digest keeps names that share a long prefix from
+colliding onto the same ResourceClaim.
+
+**Labels and ownership:** every generated ResourceClaim carries an owner
+reference to its VMI with `controller: true`, so owner-reference garbage
+collection removes it when the VMI is deleted. It also carries labels
+identifying the managed claim entry, the ManagedClaimProvisioner that
+configured it, and the VMI name. Recording the VMI name both on the owner
+reference and as a label lets claims be listed by VMI without walking
+owner references.
 
 **Idempotency:** provisioner reconciliation is naturally idempotent. If
 the ResourceClaim already exists with the correct owner reference, the
@@ -510,9 +522,15 @@ reconciliation. It clears the condition after successfully reconciling
 the claim. The launcher pod remains pending until the claim exists and
 can be allocated.
 
-**ResourceClaim deleted externally:** if a managed claim's ResourceClaim
-is deleted while the VMI is running, its provisioner controller detects
-this through the informer and re-creates it on the next reconciliation.
+**ResourceClaim deleted externally:** every managed ResourceClaim carries
+a KubeVirt finalizer. If an external actor deletes the claim while the
+owning VMI is still running, the finalizer holds the object in the API
+server (terminating but not yet removed) so the in-use DRA allocation is
+not torn down in the gap. The provisioner controller detects the pending
+deletion, removes its finalizer, and re-creates the ResourceClaim so the
+VMI is returned to a healthy, non-terminating state. When the owning VMI
+itself is being deleted, the controller removes the finalizer without
+re-creating, allowing owner-reference garbage collection to complete.
 
 **No provisioner controller:** if no controller serves the configured
 provisioner name, no matching ResourceClaim appears and the launcher pod
--- a/veps/sig-compute/300-managed-dra-claims/vep.md
+++ b/veps/sig-compute/300-managed-dra-claims/vep.md
@@ -142,6 +142,14 @@ VEP-300's managed claims, which handle CPUs as part of cross-device claims.
   [dra-driver-cpu#114](https://github.com/kubernetes-sigs/dra-driver-cpu/issues/114).
 - [dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu)
   v0.2.0+: publishes `pcieRoot` list attribute in grouped mode (PR #163).
+- [KEP-5075](https://github.com/kubernetes/enhancements/issues/5075) (DRA
+  consumable capacity), building on partitionable devices
+  [KEP-4815](https://github.com/kubernetes/enhancements/issues/4815):
+  `DRAConsumableCapacity` gate, beta and on by default in Kubernetes 1.36.
+  CPU claims use grouped mode: the driver advertises one device per
+  socket or NUMA node with `allowMultipleAllocations: true` and a
+  `dra.cpu/cpu` capacity, and each claim consumes a quantity of it.
+  Individual mode (one device per CPU) is deprecated upstream and not used.
 
 ### API Changes
 
@@ -165,7 +173,11 @@ spec:
   # Each name is one of cpu, gpu, hostDevice, or network.
   deviceTypes:
   - name: cpu
-    deviceClassName: cpu.dra.k8s.io
+    deviceClassName: dra.cpu
+    # Capacity key the CPU DRA driver advertises (dra-driver-cpu uses
+    # dra.cpu/cpu). The generated claim consumes the derived vCPU count
+    # from this capacity. Required for the cpu type; ignored otherwise.
+    capacityName: dra.cpu/cpu
   - name: gpu
     deviceClassName: gpu.example.com
     opaque:
@@ -192,7 +204,12 @@ pairing policy in `ManagedClaimProvisioner.spec`.
 `deviceTypes` is a named list. The name maps VMI device declarations to
 the provisioner configuration: `cpu` maps to `domain.cpu.dra`, `gpu` to
 `domain.devices.gpus[]`, `hostDevice` to `domain.devices.hostDevices[]`,
-and `network` to `spec.networks[].resourceClaim`.
+and `network` to `spec.networks[].resourceClaim`. `deviceClassName` names
+the admin-installed DeviceClass. `capacityName` is required for the `cpu`
+type and ignored for the whole-device types (`gpu`, `hostDevice`,
+`network`): CPUs allocate by consumable capacity rather than by whole
+device, so the generated request must name the capacity key the driver
+advertises (see the CPU generation section below).
 
 `opaque` is optional driver-specific configuration. When present, the
 provisioner renders a Kubernetes `DeviceClaimConfiguration` in
@@ -281,12 +298,27 @@ type CPUDRASource struct {
 ```
 
 VEP-300 scans `cpu.dra` during claim generation alongside the other
-device types. The DeviceClassName is resolved from the provisioner
-CRD's `cpu.deviceClassName` field. The CPU count in the
-generated claim is derived from VEP-152's accounting formula
-(`cores x sockets x threads + emulatorThreadCPUs + supplementalPoolThreadCount`).
-See [VEP-152 (PR #414)](https://github.com/kubevirt/enhancements/pull/414)
-for details.
+device types. The DeviceClassName is resolved from the provisioner CRD's
+`cpu.deviceClassName` field.
+
+CPUs allocate through DRA consumable capacity in grouped mode, not as
+whole devices. The CPU DRA driver advertises one device per socket or NUMA
+node with a consumable CPU capacity (dra-driver-cpu publishes it under the
+key `dra.cpu/cpu`); a claim consumes a quantity of that capacity, and
+several claims can share one device until it is exhausted. The generated
+CPU request therefore leaves `count` at its default of 1 (one grouped
+device) and places the consumed quantity in `capacity.requests`, keyed by
+the provisioner's `cpu.capacityName`. Unlike the whole-device types, the
+CPU request is driven by capacity rather than count, so `count` is omitted.
+The quantity is derived from VEP-152's accounting formula
+(`cores x sockets x threads + emulatorThreadCPUs + supplementalPoolThreadCount`),
+so the user expresses guest topology and never writes a DRA quantity
+directly. The quantity is always populated: an unset `capacity` would
+default to consuming the entire grouped device (all CPUs on the socket or
+NUMA node). Individual mode (one device per CPU) is deprecated upstream and
+is not used. See
+[VEP-152 (PR #414)](https://github.com/kubevirt/enhancements/pull/414) for
+the accounting formula.
 
 ### DeviceClassName Resolution
 
@@ -317,12 +349,13 @@ GenerateClaim(managedClaimContext) -> ResourceClaimSpec:
         DeviceClassName from deviceTypes[name=network],
         create a DeviceRequest.
      d. Scan domain.cpu.dra (VEP-152) - if claimName matches,
-        look up DeviceClassName from deviceTypes[name=cpu],
-        create a DeviceRequest with CPU count derived from
-        VEP-152's accounting formula
+        look up DeviceClassName and capacityName from
+        deviceTypes[name=cpu], and create one DeviceRequest with
+        Capacity.Requests[capacityName]=quantity (Count defaults
+        to 1), where quantity is VEP-152's accounting formula
         (cores x sockets x threads + emulator + IOThreads).
-        The claim shape (capacity vs count) is determined by
-        the DeviceClass and driver mode, not by the user.
+        CPUs use DRA consumable capacity (grouped mode); the
+        whole-device types above keep Count=1 with no capacity.
 
   2. Validate:
      - At least one device must reference the claim.
@@ -718,7 +751,8 @@ spec:
   provisioner: policy.kubevirt.io/aligner
   deviceTypes:
   - name: cpu
-    deviceClassName: cpu.dra.k8s.io
+    deviceClassName: dra.cpu
+    capacityName: dra.cpu/cpu
   - name: gpu
     deviceClassName: gpu.example.com
   - name: network
@@ -789,8 +823,10 @@ spec:
         count: 1
     - name: cpus
       exactly:
-        deviceClassName: cpu.dra.k8s.io
-        count: 16
+        deviceClassName: dra.cpu
+        capacity:
+          requests:
+            dra.cpu/cpu: "16"
     constraints:
     - matchAttribute: resource.kubernetes.io/pcieRoot
       requests: [gpu0, gpu1, nic]
@@ -803,6 +839,12 @@ pairing and constraint policy is implementation behavior, not part of
 the `ManagedClaimProvisioner` API. The CPU DRA driver publishes
 `pcieRoot` as a list attribute (KEP-5491); `matchAttribute` uses set
 intersection.
+
+The `cpus` request omits `count` (it defaults to 1) and consumes 16 units of
+the grouped device's capacity through `capacity.requests` (DRA consumable
+capacity), rather than allocating 16 devices. The 16 is the guest vCPU count
+derived from `domain.cpu` (here `cores: 16`), so the user never writes the
+DRA quantity.
 
 ## Alternatives
 
--- a/veps/sig-compute/300-managed-dra-claims/vep.md
+++ b/veps/sig-compute/300-managed-dra-claims/vep.md
@@ -471,10 +471,18 @@ the ResourceClaim already exists with the correct owner reference, the
 framework converges it to the provisioner's desired spec.
 
 **RBAC:** the built-in provisioner controller needs `create`, `get`,
-`list`, `watch`, `update`, and `delete` permissions on `resourceclaims`
-in the `resource.k8s.io` API group, plus `patch` permission on
-`virtualmachineinstances/status`. Third-party provisioner controllers
-need equivalent permissions in their own RBAC configuration.
+`list`, `watch`, and `update` permissions on `resourceclaims` in the
+`resource.k8s.io` API group. The `update` verb covers adding and
+releasing the protection finalizer; the controller does not need
+`delete`, because owner-reference garbage collection removes a claim once
+its VMI is gone and the finalizer has been released. It also needs `get`,
+`list`, and `watch` on `virtualmachineinstances` and
+`managedclaimprovisioners` in the `kubevirt.io` API group, plus
+permission to emit events. It does not write VMI status: it reports
+failures as events, and virt-controller owns the aggregated VMI
+condition, preserving the single-writer invariant on VMI status.
+Third-party provisioner controllers need equivalent permissions in their
+own RBAC configuration.
 
 **Multiple managed claims per VMI:** a VMI can have multiple
 `resourceClaims[]` entries, each independently using
@@ -506,21 +514,22 @@ The validating webhook enforces:
    a managed claim.
 ### Error Handling
 
-Provisioner controllers report their own failures through the existing
-VMI status conditions list. The condition type
-`ManagedClaimProvisioningFailed` includes the managed claim entry and
-provisioner name in its reason and message; it is diagnostic only and
-is not a pod-creation readiness handshake.
+Provisioner controllers report their own failures as Kubernetes events
+on the VMI; they do not write VMI status. virt-controller aggregates the
+state of the generated ResourceClaims into a single `ManagedClaimsReady`
+condition on the VMI, which is diagnostic and not a pod-creation
+readiness handshake.
 
 **Claim generation failure:** if a provisioner controller cannot
 generate a claim (for example, no DeviceClassName is resolvable), it
-sets a `ManagedClaimProvisioningFailed` condition on the VMI. The
-condition reason is `FailedCreateResourceClaim`; its message identifies
-the managed claim entry, provisioner name, and error. The controller
-also emits a `FailedCreateResourceClaim` event and retries on the next
-reconciliation. It clears the condition after successfully reconciling
-the claim. The launcher pod remains pending until the claim exists and
-can be allocated.
+emits a `FailedManagedClaimProvisioning` warning event on the VMI whose
+message identifies the managed claim entry, provisioner name, and error,
+and retries on the next reconciliation. While a managed claim is still
+missing, virt-controller reports the VMI `ManagedClaimsReady` condition
+as false with reason `NotAllManagedClaimsReady`; the reason becomes
+`AllManagedClaimsReady` once every managed claim has been generated. The
+launcher pod remains pending until the claim exists and can be
+allocated.
 
 **ResourceClaim deleted externally:** every managed ResourceClaim carries
 a KubeVirt finalizer. If an external actor deletes the claim while the
@@ -878,8 +887,9 @@ scalability model. See
   invalid device type configuration, empty claims, duplicate request
   names, and provisioner existence
 - Integration tests with fake ManagedClaimProvisioner objects (envtest)
-- Integration tests: provisioning failure sets and successful
-  reconciliation clears `ManagedClaimProvisioningFailed`
+- Integration tests: provisioning failure emits a
+  `FailedManagedClaimProvisioning` event and holds `ManagedClaimsReady`
+  false, and successful reconciliation flips it to `AllManagedClaimsReady`
 - E2E: VMI with managed claim for GPU (requires DRA driver in CI)
 
 ## Graduation Requirements
--- a/veps/sig-compute/300-managed-dra-claims/vep.md
+++ b/veps/sig-compute/300-managed-dra-claims/vep.md
@@ -395,6 +395,15 @@ controller for its own provisioner name. Provisioner controllers are
 not registered in virt-controller and do not need to be compiled into
 KubeVirt.
 
+The Go package that implements this framework is a convenience for
+in-tree and third-party controllers; it is not part of KubeVirt's
+staged, versioned API surface and carries no semver or deprecation
+guarantee. The stable compatibility boundary that an external provisioner
+depends on is the wire contract: the `ManagedClaimProvisioner` and
+`ResourceClaim` API shapes, the generated-claim labels, and the
+deterministic claim-naming algorithm. A controller that reproduces those
+directly keeps working even if the helper package is refactored.
+
 ```go
 type ClaimProvisioner interface {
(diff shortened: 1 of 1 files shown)


veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,920 @@
...
+
+Each provisioner controller uses expectations to avoid reconciling
+before its ResourceClaim informer cache reflects a create or update.
+The framework provides this common behavior to provisioner controllers.
+
+### Implementation Details
+
+**Claim naming:** the generated claim is named
+`<vmi-name>-<claim-name>`. If this exceeds the 253-character DNS subdomain
+limit, the name is truncated and a short hash suffix is appended to
```
alaypatel07: two nits here:
1. in case 253 character limit is extended, the first 124 from both vmi-name and claim-name should be selected with 5 char hash
2. the vmi name should be present on the claim as owner reference and as a label

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,920 @@
...
+generate a claim (for example, no DeviceClassName is resolvable), it
+sets a `ManagedClaimProvisioningFailed` condition on the VMI. The
+condition reason is `FailedCreateResourceClaim`; its message identifies
+the managed claim entry, provisioner name, and error. The controller
+also emits a `FailedCreateResourceClaim` event and retries on the next
+reconciliation. It clears the condition after successfully reconciling
+the claim. The launcher pod remains pending until the claim exists and
+can be allocated.
+
+**ResourceClaim deleted externally:** if a managed claim's ResourceClaim
```
alaypatel07: This is catastrophic, I think we should have a finalizer that prevents external deletion.

alaypatel07 (commented):
Added a nit, but other then that IMO this is a very well drafted design. 

@rthallisey @lyarwood @vladikr @iholder101  PTAL

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Done. When the name exceeds 253 characters we now keep a fixed-length prefix of both the vmi-name and the claim-name joined with a 5-character SHA-256 digest of the full untruncated name, so both names stay recognizable and long shared prefixes cannot collide. Also added a "Labels and ownership" note: the generated ResourceClaim carries the VMI as an owner reference (`controller: true`) and records the VMI name as a label as well.

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Added a finalizer to managed ResourceClaims and rewrote the "ResourceClaim deleted externally" section: if an external delete lands while the VMI is running, the finalizer holds the object in the API server (terminating but not yet removed) so the in-use DRA allocation is not torn down in the gap; the controller then removes its finalizer and re-creates the claim to recover. During VMI teardown the controller removes the finalizer without re-creating, so owner-reference GC completes.

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,990 @@
+# VEP #300: Managed DRA Claims
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version: v1.11
```
lyarwood: Can you remove this, we only populate this per release.

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,990 @@
...
+controller that is temporarily unavailable.
+
+**VMI deletion during claim creation:** the controller checks
+`vmi.DeletionTimestamp` before creating claims. If the VMI is being
+deleted, the controller skips claim creation. Owner reference GC handles
+cleanup of any already-created claims.
+
+## API Examples
+
+### Single GPU
```
lyarwood: I appreciate this just demonstrates what's possible with this framework but I'd rather leave out the simple cases we don't expect users to actually use it for. Shouldn't this simple case be transparent to the end user through GPUsWithDRA?

johnahull:
Update for reviewers: two commits landed since the review ping, so the diff has moved. Quick guide to what changed and where to look.

**CPU claims modeled as DRA consumable capacity** (`7f773e5f`), design change. CPU managed claims move from individual mode (`count: N` whole devices) to grouped mode: dra-driver-cpu advertises one device per socket or NUMA node with a consumable `dra.cpu/cpu` capacity, and a claim consumes a quantity of it. The generated cpu request now omits `count` (defaults to 1, one grouped device) and carries `capacity.requests`, with the quantity from VEP-152's accounting formula. Adds an optional `capacityName` field on the cpu deviceType (defaults to `dra.cpu/cpu`) and documents the `DRAConsumableCapacity` dependency (KEP-5075). See the CPU narrative and the ResourceClaim examples.

**Provisioner extension contract clarified** (`596f64b0`), doc-only.
- Idempotency: convergence is a full-spec replace, so `GenerateClaim` must return the complete spec including topology constraints; a provisioner cannot defer constraint shaping to a later mutation without fighting the reconciler.
- Replaced the "Provisioner Expectations" section (which described a mechanism the controller does not implement) with "Reconcile Consistency" describing the live API-server read.
- Added the extension-point stability boundary: the helper Go package carries no semver guarantee; the wire contract (CRD shapes, labels, naming algorithm) is the stable boundary.
- Added a colliding-provisioner-names caveat and a passthrough / deferred-policy provisioner future extension.

vladikr:
Thank you everyone for pushing VEP152 and VEP300! 
I think these are the right building blocks.

I want to share a the view I was speaking about the on call today. 
This is just a proposal for discussion, not a request to throw either VEP away.

### What I am worried about

From the user’s point of view there is one intent: “this VM needs exclusive CPUs” (and sometimes “please place them with my GPU/NIC”).

`dedicatedCpuPlacement` was added years ago as that intent, closer to OpenStack Nova’s dedicated CPU model than to kubelet. 
The implementation later was based on the CPU Manager (since there was nothing else), so the field started to *sound* like a mechanism. 
If we now add `domain.cpu.dra` as the main way to request the same thing, we repeat that leak in the other direction. 
The user has to know whether the node runs CPU Manager or a DRA CPU driver, and rewrite the VM spec when that changes.

DRA is a good implementation. I just do not think it should become the name of the user facing API.

GPUs and host devices are different: they are discrete attachments, so `claimName` / `requestName` on the device is natural. 
CPUs requests are expressed via quantity, exclusivity and optional topology. That is why copying the GPU shape onto `domain.cpu` feels off to me.

### What I think each VEP should own

**VEP-152: keep it as plumbing.**  
virt-launcher consumes KEP-5304 CPU metadata and pins. 
virt-controller attaches whatever claims exist on the virt-launcher pod. 
Document the hostCPUs formula. Prefer grouped/consumable capacity, given where `dra-driver-cpu` is going. 
I would not make `domain.cpu.dra` the primary user API. If we need an explicit claim at all, it can stay an advanced escape hatch on `spec.resourceClaims`, not a new field on the CPU object.

**VEP-300: claim generation and policy.**  
A provisioner builds the `ResourceClaim`. 
That work should stay out of virt-controller. The gap I see in the current draft is only the *trigger*: it assumes `cpu.dra` on the VMI so the provisioner knows whom to treat. 
That makes 300 depend on a user-facing mechanism field that 152 does not actually need.

### A flow that does not put DRA on `domain.cpu`

1. The user writes intent (`cores`, `dedicatedCpuPlacement`, devices).
2. A provisioner (in-tree aligner or vendor controller) selects the VMI and creates a VMI-owned `ResourceClaim`.
3. The provisioner writes the claim name into **VMI status** and sets a condition (`ManagedClaimsReady`).  
     It does not edit `vmi.spec` and it does not edit the virt-launcher pod.
4. virt-controller waits on that condition, then copies the claim names onto the pod.
5. kubelet/DRA do the allocation; virt-launcher consumes metadata as it already does for GPUs.

That is the same idea as AAQ (an external controller gates progress) with one DRA-specific constraint: `pod.spec.resourceClaims` is fixed at create time, so the vendor cannot attach the claim after the pod exists. so our hook will be via the Status, virt-controller picks it and writes the pod spec.

RBAC follows from that: the provisioner can update `virtualmachineinstances/status` and `resourceclaims`. It should not update VMI spec or pods.

### Who says “how”: a MigrationPolicy-style object

Even `managedClaim` on every VM still asks the user to name a provisioner. 
I think the default path should look like `MigrationPolicy`: the admin/vendor targets a group of workloads, and the VM spec stays about the VM.

```yaml
apiVersion: kubevirt.io/v1alpha1
kind: ManagedClaimPolicy
metadata:
  name: aligned-gpu-workers
spec:
  selectors:
    virtualMachineInstanceSelector:
      matchLabels:
        kubevirt.io/claim-policy: aligned-gpu
  provisioner: policy.kubevirt.io/aligner
  params:
    align:
      attributes: [resource.kubernetes.io/numaNode]
```

The VM user writes:

```yaml
metadata:
  labels:
    kubevirt.io/claim-policy: aligned-gpu
spec:
  template:
    spec:
      domain:
        cpu:
          cores: 16
          dedicatedCpuPlacement: true
          isolateEmulatorThread: true
        devices:
          gpus:
          - name: gpu0
            claimName: topo
            requestName: gpu
```

If CPU Manager is all the cluster has, the same `dedicatedCpuPlacement: true` still works. 
If DRA is available and a policy matches, the provisioner satisfies that intent with a claim.

The winning policy should be visible on VMI status (same discoverability problem MigrationPolicy already solved - @EdDev fyi). 
Explicit `spec.resourceClaims[].resourceClaimName` we can consider for the opt-out, advanced path.

### Groups and “shared cpus”

As I mentioed on the call, a policy is a natural group. 
That is useful for “these VMs draw from the same oversubscribed CPU pool” or “keep them in the same alignment domain.”

I would not default to one `ResourceClaim` object reused by many VMs... 
DRA claims are usually one consumer each. 
I think the provisioner can still emit per-VMI claims against the same consumable-capacity pool / same topology constraints. 
`claimScope: Group` can be an advanced mode later if we truly need a single shared claim object.

Thinking about @alaypatel07 's questions, these are fair! “not dedicated” is not the same as “shared with other VMs,” 
and it does not say which vCPUs must be NUMA-aligned. 

I think those rules should live in the **provisioner contract** (or in policy params when two groups differ), like:

- `exclusive`: not shared with other pods  
- `shared`: oversubscribed pool  
- `unpinned`: that's today’s default non-dedicated behaviour  
- `align.include`: which of those classes sit next to the GPU  

Most users never see those knobs.... the policy would target their VMs via a set label or something else I couldn't think of 

### Happy path vs hybrid

Happy path (provisioner default = “all dedicated CPUs, align with GPUs in this group”):

```yaml
domain:
  cpu:
    cores: 16
    dedicatedCpuPlacement: true
    isolateEmulatorThread: true
devices:
  gpus:
  - name: gpu0
    claimName: topo
    requestName: gpu
```

Hybrid only when a group actually needs a split, as policy params

```yaml
# ManagedClaimPolicy params
cpu:
  exclusive: 8
  shared: 12
align:
  include:
    cpu: exclusive
    devices: [gpus]
```

`dedicatedCpuPlacement: true` together with a split is confusing, because today the boolean means *all* vCPUs are exclusive. 
not sure yet...

### Does a VM user ever have to specify a CPU claim?

Almost never, if VEP300 exists. 
Custom CEL, reusing an already-allocated claim, or an admin-owned shared pool are the exceptions .... and those already belong on `spec.resourceClaims`, the same place GPUs point today.

### How this is meant to land in the two PRs

I am not asking to block 152 on all of this.

- **152** can stay the low-level path: metadata, pinning, accounting, “claims that appear on the pod are consumed.” Please consider not adding `domain.cpu.dra` as the thing regular users write.
- **300** own claims, provisioners, finalizers. Please consider keying CPU inclusion off existing intent and a policy object, and handing claim names to virt-controller through status, not through a mechanism field on `domain.cpu`.

alaypatel07:
> Thank you everyone for pushing VEP152 and VEP300!
I think these are the right building blocks.

> I want to share a the view I was speaking about the on call today.
This is just a proposal for discussion, not a request to throw either VEP away.

Thanks for putting together the conclusion @vladikr, there are some things with the proposed API that doesnt sit well for me, and I will try to capture it inline.

> From the user’s point of view there is one intent: “this VM needs exclusive CPUs” (and sometimes “please place them with my GPU/NIC”).
> 
> dedicatedCpuPlacement was added years ago as that intent, closer to OpenStack Nova’s dedicated CPU model than to kubelet.
> The implementation later was based on the CPU Manager (since there was nothing else), so the field started to sound like a mechanism.
> If we now add domain.cpu.dra as the main way to request the same thing, we repeat that leak in the other direction.
> The user has to know whether the node runs CPU Manager or a DRA CPU driver, and rewrite the VM spec when that changes.
> 
> DRA is a good implementation. I just do not think it should become the name of the user facing API.
> 
> GPUs and host devices are different: they are discrete attachments, so claimName / requestName on the device is natural.
> CPUs requests are expressed via quantity, exclusivity and optional topology. That is why copying the GPU shape onto domain.cpu feels off to me.

The alternative to not having domain.cpu.dra misses a key point, how the virt-launcher gets its CPU is not captured in the vmi spec at all and instead it is described into the policy that the provisioner implements.  In my opinion this is a leaky abstraction. For example, a 4-vCPU topology can run on either two shared host CPUs with 2:1 oversubscription or four exclusive host CPUs (Libvirt API allows for this usecase). A provisioner can choose to implement an oversubscription policy or dedicated vCPU policy. I see this as the inverse of the dedicatedCpuPlacement problem. domain.cpu.dra risks exposing an implementation mechanism in the API, but removing host-allocation intent entirely pushes user-visible CPU semantics into the provisioner implementation.

Consider two VMIs, VMI1 and VMI2, each exposing four domain vCPUs.

- VMI1 requires four exclusive host CPUs, with each domain vCPU pinned one-to-one.
- VMI2 requires three host CPUs: vCPU 0 is pinned to one exclusive host CPU, while vCPUs 1-3 are oversubscribed across two shared host CPUs.

The provisioner could create a claim containing four exclusive CPU devices for VMI1. For VMI2, it could create a compound claim containing one exclusive and two shareable CPU devices. DRA can likely represent and allocate these resources.

The problem appears when virt-launcher creates the domain XML. Even if claim metadata distinguishes the exclusive and shared host CPUs, nothing in the VMI spec identifies which domain vCPU belongs to each allocation or how the shared vCPUs should be mapped. virt-launcher cannot determine that vCPU 0 requires an exclusive vcpupin, while vCPUs 1-3 should use the shared two-CPU set.

We therefore need a VMI API that defines the allocation of domain vCPUs and their mapping to host CPU allocation groups. The provisioner can choose how to satisfy those groups, but it should not invent the domain-to-host mapping.  

For VEP-152, if we dont want to leak DRA APIs, the there should be an API like this:
```yaml
domain:
  cpu:
    sockets: 1
    cores: 4
    threads: 1
    hostAllocation:
      groups:
      - name: exclusive
        targets:
          vcpus: "0"
        hostCPUs:
          count: 1
          # Exclusive prevents these host CPUs from being allocated
          # to another workload.
          accessMode: Exclusive
          # OneToOne pins each selected vCPU to a distinct host CPU.
          mappingPolicy: OneToOne
      - name: shared
        targets:
          vcpus: "1-3"
        hostCPUs:
          count: 2
          # Shared permits other workloads to use the same host CPUs.
          accessMode: Shared
          # SharedSet pins every selected vCPU to the complete allocated
          # host CPU set. Here, vCPUs 1-3 can run on either host CPU.
          mappingPolicy: SharedSet
```

This tells the virt launcher to generate the domxml like this:
```xml
<vcpu placement="static">4</vcpu>
<cpu>
  <topology sockets="1" cores="4" threads="1"/>
</cpu>
<cputune>
  <vcpupin vcpu="0" cpuset="2"/>
  <vcpupin vcpu="1" cpuset="6-7"/>
  <vcpupin vcpu="2" cpuset="6-7"/>
  <vcpupin vcpu="3" cpuset="6-7"/>
</cputune>
```

johnahull:
Agree with @alaypatel07: the vCPU-to-host-CPU mapping is intent and belongs in the spec. In this branch a CPU claim is a single grouped capacity request, so the claim structurally cannot carry that mapping, and it would be needed even under CPU Manager with no DRA. That makes it intent, not mechanism.

@vladikr's constraint still holds: keep DRA vocabulary out of the field. So I would trim the shape to the minimum virt-launcher needs:

```yaml
domain:
  cpu:
    cores: 4
    hostAllocation:
      groups:
      - vcpus: "0"
        exclusive: true
        hostCPUs: 1
      - vcpus: "1-3"
        exclusive: false
        hostCPUs: 2
```

Drop `mappingPolicy` (derivable) and use `exclusive: bool` instead of an `accessMode` enum. `dedicatedCpuPlacement: true` stays as the coarse all-dedicated knob and expands to a single all-exclusive group; `hostAllocation.groups` is the fine-grained form for mixed exclusive/shared. The two are mutually exclusive at admission.

EdDev:
I'll want to start with a general suggestion and then feedback on @vladikr and @alaypatel07 points.

I find the topic a bit overloaded. I wander if we can extract the topic of mapping between the VM specification requirement that reflects towards the guest from the wiring and allocation to the host.
If we can make this mapping generic enough, with all the needed orchestration, it will simplify the review and will make the solution cleaner IMO.

I think this mapping makes sense a lot, and not only for CPU.
It is simpler to place it a an adapting layer over creating conditions per functionality.
IMO, the main concept suggested for the separation fits all devices, even the ones with "1:1" or "static" mapping. Instead of ref directly a claim, always go through a translation layer that resolves it to the right backend which the cluster admin chose.

> GPUs and host devices are different: they are discrete attachments, so `claimName` / `requestName` on the device is natural.

I agree it is trivial to directly ref a claim, however, in order to preserve a unify and clear layering, I find it useful to consider the "translation" layer to serve all. Even if we cannot change existing API, we can deprecate it in favor of a recommended one.
While I do not see it as a must, I think it is a better path from an architectural point of view.

> vCPU-to-host-CPU mapping is intent and belongs in the spec. In this branch a CPU claim is a single grouped capacity request, so the claim structurally cannot carry that mapping, and it would be needed even under CPU Manager with no DRA. That makes it intent, not mechanism.

@alaypatel07 and @johnahull points are about providing a VM owner the ability to control and request mapping resources based on VM definition. IMO this is questionable and in many ways non trivial to consider per-VM topology management.
The scenarios presented are real (although I am not in the details of real users usage), however I am not sure if we need to expose low level mapping details through the VM spec.

The mapping is an intent, but we can consider it a backend intent, defined by a cluster admin.
I am not sure why that translation layer cannot express the same details suggested on the VM/VMI spec.
In the extreme case, it is possible to provide a 1:1 mapping between a VM to a translation/provider policy and in the best case scenario, all VMs point to the same translation policy.

I wanted to point out that we can use this separation to solve other needs that have been seen lately, like providing DRA networking in a mix with Multus.

alaypatel07:
> I find the topic a bit overloaded. I wander if we can extract the topic of mapping between the VM specification requirement that reflects towards the guest from the wiring and allocation to the host.
> If we can make this mapping generic enough, with all the needed orchestration, it will simplify the review and will make the solution cleaner IMO.
> 
> I think this mapping makes sense a lot, and not only for CPU.
> It is simpler to place it a an adapting layer over creating conditions per functionality.
> IMO, the main concept suggested for the separation fits all devices, even the ones with "1:1" or "static" mapping. Instead of ref directly a claim, always go through a translation layer that resolves it to the right backend which the cluster admin chose.

@EdDev honestly these concepts are very high level for us to discuss, can you please come up with a set of YAML specs to demo the translation layer? This is needed to understand if the high level concepts explained in the commentm will even work when put with implementation details of how DRA works in kubevirt and kubernetes

vladikr:
I don’t think mixed exclusive/shared guest vCPUs should be a first-class API.

The guest OS has no way to know which vCPUs are exclusive. 
There is no standard topology signal for that. 
Memory/NUMA alignment becomes incoherent if some vCPUs are pinned next to a GPU and the rest float. 

In practice people would have to pin workloads inside the guest onto vCPU numbers (sounds crazy to me...)
that is not a normal KubeVirt user.
If that's a requirement right now, I need to understand this better. I would push back on that at this point.

The split I do understand is the one we already have:
 - guest vCPUs are either all exclusive (dedicatedCpuPlacement: true) or not
 - emulator thread may be isolated (and all of these house keeping processes)
 - IOthreads may be extra exclusive CPUs or a shared pool

So I wouldn't add hostAllocation.groups. 
For me the dedicatedCpuPlacement, isolateEmulatorThread, IOthread settings already express the intent. Oversubscription and GPU alignment can stay in the provisioner/policy

At least for the first iteration. Let's then discuss other usecases.

alaypatel07:
> The guest OS has no way to know which vCPUs are exclusive.

The guest OS does not need to know if vCPUs are exclusive or not, it is the user intent and decision.

> There is no standard topology signal for that.
> Memory/NUMA alignment becomes incoherent if some vCPUs are pinned next to a GPU and the rest float.

The floating can happen for workloads that are running in the same NUMA not floating across NUMAs.

I can share this because we similar usecases running in production. Currently those are hacked around because of lack of nicer API from host allocation point of view. But with DRA, we dont need to hack it anymore.

These are the usecase we have:
1. floating host vcpus within a VMI guest domain.
2. shared host vcpus across VMIs and guest domains.

> I don’t think mixed exclusive/shared guest vCPUs should be a first-class API.

Why should this not be in the API? if not in the API, virt-launcher which renders the domxml (with no access to kube-api) cannot make an informed decision on how to generate the desired host allocation properties for the needed domxml.

> For me the dedicatedCpuPlacement, isolateEmulatorThread, IOthread settings already express the intent. Oversubscription and GPU alignment can stay in the provisioner/policy

GPU alignment can stay in provisioner/policy, oversubscription has to be on VMI API because it is information that has to be consumed by virt-launcher.

> At least for the first iteration. Let's then discuss other usecases.

I agree with this, for the first iteration, I am okay with:
1. VEP-152: virt-controller generating resource claim for cores, sockets, thread, emulator thread, iothread, dedicated placement: true
2. VEP-300; generating resource claim if there needs to be alignment

.. with a pre-condition that for next iteration, i.e. v1.11, there needs to be support for shared cpus across VMIs, and shared cpus within the domain (not explicit co-pining but targeted grouped sharing). For us to realistically make this work in 1.11, we should have a weekly call set up to discuss the API needed to get there.

I am planning to discuss this in further detail in sig-compute call tomorrow.

vladikr:
@alaypatel07 Thank you for the conversation today!
I think it was helpful to understand how we can proceed and address the use cases going forward.

> I can share this because we similar usecases running in production. Currently those are hacked around because of lack of nicer API from host allocation point of view. But with DRA, we dont need to hack it anymore.

Yes, +1

> 
> These are the usecase we have:
> 
> 1. floating host vcpus within a VMI guest domain.
> 2. shared host vcpus across VMIs and guest domains.

Yes. I think the message that I wanted to deliver is that we can't address everything in one go.
Let's build in stages.

I thought about the use cases in the comment.
I think we could build this as  cpu classes, something like:
```
domain:
  cpu:
    cores: 16
    classes:
    - name: dataplane
      count: 4
      isolation: exclusive
    - name: housekeeping
      count: 12
      isolation: shared
```
But I think we could come up with even a better UX and use a CPU profile:

```
cpu:
  cores: 16
  profile: nfv-dataplane   # means 4 exclusive + 12 shared, convention documented
```
These profiles would be known

Then the vendor policy would be something like
```
kind: ManagedClaimPolicy
spec:
  selectors: { matchLabels: { kubevirt.io/profile: nfv-dataplane } }
  provisioner: vendor.io/nfv
  params:
    dataplane:
      exclusive: true
      alignWith: [gpus]
    housekeeping:
      exclusive: false
      oversubscribe: "2:1"    
```

The we will need to publish the allocated cpus into the guest - i did something similar here https://github.com/kubevirt/kubevirt/pull/6946
But we will need to find a way to publish a stable convention into the guest.

EdDev (approved):
This VEP focuses only on the managed claims, i.e. generating claims per some logic given a VMI spec and a matching criteria.

I was looking into generalizing a provider to resolver the resource, for DRA and other (Multus) backends.
At some points is overaps and in others compliments.

I think it is fine to start the experiment and evaluate how this rolls out.
In parallel, I will try to introduce the other "provider" through a new VEP in the coming weeks.

I found the VEP text itself challenging to review, I think it is best if the size and wording would have been simplified, covering less content from other VEPs (e.g. 152).
As we are short in time, we could follow up on this later.

/lgtm

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,1101 @@
...
+## Design
+
+### Feature Gate
+
+All changes are gated behind `ManagedDRAClaims` (alpha, off by default).
+
+### Responsibility Boundary
+
+- **User owns:** device declarations (`gpus[]`, `hostDevices[]`,
+  `networks[]`, `cpu.dra`)
```
vladikr: should we remove this `cpu.dra` ?

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,1101 @@
...
+  launcher pod
+- **Claim provisioner owns:** desired ResourceClaim device requests and
+  constraints for one managed claim
+- **Scheduler owns:** device allocation, topology constraint satisfaction
+
+### Scope Boundary with VEP-152
+
+VEP-152 and VEP-300 are developed concurrently:
+
+- **VEP-152 owns:** `cpu.dra` struct (`CPUDRASource`), `deviceClassName`
```
vladikr: here `cpu.dra` too?

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,1101 @@
...
+type CPUDRASource struct {
+	// ClaimRequest references a specific request from a ResourceClaim
+	// listed in vmi.spec.resourceClaims[].
+	*ClaimRequest `json:",inline"`
+}
+```
+
+VEP-300 scans `cpu.dra` during claim generation alongside the other
+device types. The DeviceClassName is resolved from the provisioner CRD's
+`cpu.deviceClassName` field.
```
vladikr: is this still accurate?

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,1101 @@
...
+
+```yaml
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: gpu-vm
+spec:
+  resourceClaims:
+  - name: my-gpu
+    managedClaimProvisionerName: gpu-default
```
vladikr, line 614: I'm still unclear on how the user would know what to write here. 
Where and how can these be discovered?
I thought that the better way is to mimic the migration policy so that ManagedClaimProvisioner - but we can continue the discussion and be blocked by this.

veps/sig-compute/300-managed-dra-claims/vep.md
```diff
@@ -0,0 +1,1101 @@
+# VEP #300: Managed DRA Claims
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version: v1.11
```
iholder101: Can you please drop future release target?
You can re-add this in the v1.11 time-frame and we'll revisit this :)

Thanks!

veps/sig-compute/300-managed-dra-claims/vep.md
alaypatel07: https://github.com/johnahull/enhancements/blob/3eb5b8f84215e5999d8304014d8aeb30c0f77307/veps/sig-compute/300-managed-dra-claims/vep.md#alternative-4-label-selector-provisioner-with-implicit-claim-routing

See Alternative 4 in the later sections it has details.

> I'm still unclear on how the user would know what to write here.
Where and how can these be discovered?

Users can run the following, this is very similar to how users know what storageclass to populate when requesting to PVC.
```
kubectl get managedclaimprovisioners
```

vladikr:
I'll approve this in principle. 
there are some changes requested; let's address them before it goes in. 
/approve
/hold

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Done. Replaced with the user-owned CPU intent fields (`cores`, `dedicatedCpuPlacement`). Also scrubbed `cpu.dra` references throughout the VEP and replaced with pointers to VEP-152.

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Done. Collapsed the VEP-152 scope boundary to a short pointer. VEP-300 was carrying too much VEP-152 detail and it kept going stale.

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Trimmed this section to what VEP-300 owns. The CPU struct shape, field name, and accounting formula are VEP-152's to finalize. VEP-300 just consumes whatever struct VEP-152 defines.

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: +1 to Alay's answer. Added a discoverability note near the CRD description: `ManagedClaimProvisioner` objects are cluster-scoped and discoverable via `kubectl get managedclaimprovisioners`, same pattern as StorageClass for PVCs.

veps/sig-compute/300-managed-dra-claims/vep.md
johnahull: Done.

Issue #434, opened 2026-08-24
VEP 434: Mergeable Guest Memory


michalskrivanek:
**Primary contact (assignee)**:

/assign @mskrivan

**Current Feature Stage**:
New

**Feature Gate**:
-

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:
https://github.com/kubevirt/enhancements/pull/435

**Timeline**:

* Alpha release target (x.y): 1.10
* Beta release target (x.y):
* Stable release target (x.y):

* [X] Alpha
  * [X] VEP PRs: https://github.com/kubevirt/enhancements/pull/435
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

iholder101:
Hey!

Gentle reminder, this VEP doesn't have a dedicated "VEP Reviewer" in the [Github Enhancement Tracking Board for v1.10](https://github.com/orgs/kubevirt/projects/22).
Please note that as [explained here](https://groups.google.com/g/kubevirt-dev/c/ASosgJdymOQ), this is a requirement in order for the VEP to become "Tracked" for v1.10.

PR #435, opened 2026-08-24
VEP 434 - Mergeable Guest Memory

michalskrivanek:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/434
**SIG label**: /sig compute

### What this PR does
Adds a per-VM `spec.domain.memory.mergeable` field (default true). Setting it
to false opts the VM out of KSM by not marking guest RAM as MADV_MERGEABLE
(libvirt `<nosharepages/>`). Node-level `ksmConfiguration` is unchanged.

### DIFF
--- a/veps/sig-compute/434-mergeable-memory/vep.md
+++ b/veps/sig-compute/434-mergeable-memory/vep.md
@@ -1,0 +1,295 @@
+# VEP #434: Mergeable Guest Memory
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version: v1.10
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue [#434](https://github.com/kubevirt/enhancements/issues/434) created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [ ] (R) Alpha target version is explicitly mentioned and approved
+- [ ] (R) Beta target version is explicitly mentioned and approved
+- [ ] (R) GA target version is explicitly mentioned and approved
+
+## Overview
+
+KubeVirt can turn Kernel Same-page Merging (KSM) on for a node, or it can be managed
+externally, but every VM on that node is a merge candidate. Security- and
+latency-sensitive guests have no per-VM way to keep their RAM unshared. This VEP adds
+that control as a VMI annotation.
+
+## Motivation
+
+KSM is a density tool: identical pages across VMs are collapsed to one
+copy-on-write page. KubeVirt already exposes that at cluster scope via
+[`ksmConfiguration`](https://kubevirt.io/user-guide/cluster_admin/ksm/).
+QEMU, however, registers *every* guest's RAM as mergeable unless told
+otherwise. The result is all-or-nothing: enable KSM on the node and every VM
+participates; leave it off and nobody benefits.
+
+That is a poor fit for mixed clusters.
+
+- **Isolation.** Deduplication timing has been used as a cross-VM side
+  channel ([USENIX Security 2011](https://www.usenix.org/legacy/event/sec11/tech/full_papers/Suzaki.pdf),
+  later practical attacks). A tenant that must not leak through shared pages
+  currently has to disable KSM for the whole node.
+- **Latency.** A write to a merged page faults a private copy (`cow_ksm` in
+  [kernel KSM docs](https://docs.kernel.org/admin-guide/mm/ksm.html)). Guests
+  that care about tail latency pay that cost for a density feature they do
+  not want.
+- **Wasted scan CPU.** A VM with a unique working set still sits in ksmd's
+  candidate set. The scan helps nobody and still consumes host CPU.
+
+
+## Goals
+
+- A VM owner can opt a VM out of host page merging. After the VM is running,
+  its guest RAM is not a KSM candidate, including on nodes where KSM is
+  enabled.
+- VMs that do not opt out keep today's merging behavior.
+- The opt-out is a VMI annotation, so it travels with the VM object (create,
+  restart, migrate) and can be set from a preference.
+
+## Non Goals
+
+- **Cluster-wide KSM enablement and ksmd tunables** — rejected. That API
+  already exists (`ksmConfiguration`). This VEP only adds the missing per-VM
+  opt-out.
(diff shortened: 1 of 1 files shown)


veps/sig-compute/434-shareable-memory/vep.md
```diff
@@ -0,0 +1,299 @@
...
+
+Add `shareable` to `spec.domain.memory`. It is a `*bool` that defaults to
+`true` when omitted:
+
+```go
+// Shareable controls whether guest memory pages may be merged with other
+// processes via Kernel Same-page Merging (KSM).
+// Defaults to true. When false, guest RAM is not marked mergeable.
+// +optional
+Shareable *bool `json:"shareable,omitempty"`
```
jean-edouard:
```suggestion
Mergeable *bool `json:"shareable,omitempty"`
```

Maybe it's just me but I would not have guessed what `Shareable` guest memory means...
Naming is hard, `Mergeable` is just a suggestion, not perfect either... `KSMEnabled` maybe?

veps/sig-compute/434-shareable-memory/vep.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+The name is the user-visible property (may these pages be shared?), not the
+kernel feature (`ksm`) or the libvirt element (`nosharepages`). A generic
+name stays valid if the backing mechanism changes.
+
+| Spec | Effect |
+| --- | --- |
+| omitted / `nil` | Today's behavior: pages may be merged |
```
jean-edouard:
```suggestion
| omitted / `nil` | Today's behavior: pages may be merged (on compatible guests) |
```

veps/sig-compute/434-shareable-memory/vep.md
```diff
@@ -0,0 +1,299 @@
...
+
+`ksmConfiguration` still decides whether ksmd runs. This field only decides
+whether *this* VM's pages are candidates:
+
+- `shareable: false` on a KSM-enabled node → pages are not merged.
+- default on a KSM-disabled node → nothing is merged (same as today).
+
+### Feature gate and validation
+
+Feature gate: `ShareableMemory`.
```
jean-edouard: Does it really make sense to feature-gate the option to disable a feature?...
I know we tend to say "new API needs feature-gate", but this seems like a bit much to me...
@EdDev @vladikr WDYT?

jean-edouard (commented):
Looks good, just a few questions/suggestions in-line.

veps/sig-compute/434-shareable-memory/vep.md
michalskrivanek: It was bothering me a bit as well. I think `Mergeable` is best, it's not such an overloaded term (yet:) 
I've changed the naming

veps/sig-compute/434-shareable-memory/vep.md
iholder101: Some thoughts:
1. To me, the feature gate doesn't only guard from the feature itself from being enabled, but also from the entire logic that enables it. As a silly example, somewhere in the introduced logic there could be hidden bugs (nil pointer dereference, memory leaks, etc) that aren't directly related to the feature that's being implemented and that could cause regressions. For this reason, I do tend to think that a feature gate is beneficial.
2. In https://github.com/kubevirt/enhancements/pull/382 we've discussed KSM and the intent to extract it out of the project. In the spirit of these discussions, I'm not sure we want to add a core API that's KSM-facing.
3. Further to the above, since this is just a static addition of a libvirt-level XML field, isn't it a classic fit for CEL-based plugins?

michalskrivanek: > Some thoughts:
> 
>     1. To me, the feature gate doesn't only guard from the feature itself from being enabled, but also from the entire logic that enables it. As a silly example, somewhere in the introduced logic there could be hidden bugs (nil pointer dereference, memory leaks, etc) that aren't directly related to the feature that's being implemented and that could cause regressions. For this reason, I do tend to think that a feature gate is beneficial.

I do not mind either way, just pointing out that the amount of code is disproportional, and it's more likely to introduce such a problem in the FG code than the actual feature.

>     2. In [VEP 381: KSM Zero-Pages-Only Mode and Sleep Scaling Improvement #382](https://github.com/kubevirt/enhancements/pull/382) we've discussed KSM and the intent to extract it out of the project. In the spirit of these discussions, I'm not sure we want to add a core API that's KSM-facing.

This has nothing to do with kubevirt controlling KSM or not, this is useful regardless if there's a static or dynamic or unmanaged configuration. I think as long as kernel developers believe this is a thing, and all common distros likewise, there's value in clean configuration in kubevirt

>     3. Further to the above, since this is just a static addition of a libvirt-level XML field, isn't it a classic fit for CEL-based plugins?

it's a fit only if you believe this is not a core API. Otherwise it just adds complexity to both integrator and user.
This is not a libvirt-specific feature (the xml tag mentioned in description is just illustrative, the real feature is hypervisor deciding to do `madvise(MADV_MERGEABLE)` or not)

jean-edouard (approved):

vladikr:
Thank you for this work.

I don’t think `spec.domain.memory.mergeable` should be exposed like this. 

I'm on board with the converter change, but not the CRD field.
From my point of view, this is a host memory management hint not guest memory sizing or backing. 
Memory already has guest / maxGuest / hugepages. 
On the other hand, we already expose similar tunables as annotations, for example `kubevirt.io/memfd` , not as a Memory field. 

VEP 381 is also trying to get KSM out of core, so a new Memory field is the opposite direction.

I don't think we need a plugin for this, this is a valid case, but from my pov this is what the type of hints that we need to expose as annotations.
I would suggest something that opts out: `kubevirt.io/mergeable-memory: "false"` 
That way, a preference can carry it. or we can later promote it to be a Preference field later.

vladikr:
To be honest, I don't see much use for a featuregate. it's not something new, it's a hint to request an opt-out

michalskrivanek:
Ack. I believe that would work fine too!

michalskrivanek:
@vladikr @iholder101 @jean-edouard I updated the text to align with the annotation approach and drop the FG as per https://github.com/kubevirt/enhancements/pull/435#issuecomment-5516616722 

let me know if there's anything else to adjust. thanks!

vladikr:
Thank you @michalskrivanek !
I'm +1
/approve

iholder101:
I'm not a big fan of annotations, but this seems like a good compromise :) 

Thanks for the discussion!

/lgtm

PR #436, opened 2026-08-24
VEP 190: add spec.launcherImage for plugin-driven launcher image selection

lyarwood:
## Summary

Extends VEP 190 (Kubevirt Structured Plugins) with a \`spec.launcherImage\` field on \`PluginSpec\`, allowing plugin authors to declare an alternative virt-launcher image that virt-controller selects based on a CEL condition evaluated against the VMI.

- New user story covering plugin authors who need a custom virt-launcher (updated/debug stacks, vendor builds, distribution-specific builds)
- New **Launcher Image Selection** design section explaining the mechanism and distinguishing it from hook points that require IPC
- API example showing the minimal Plugin CR needed (condition + image, nothing else to deploy)
- Two new Alpha graduation criteria
- Future enhancement note on overlap/conflict detection

Related discussion: https://github.com/kubevirt/enhancements/pull/407

## Why not a webhook?

\`spec.launcherImage\` is the one pod mutation that reduces to a string lookup — evaluate a CEL condition against the VMI, return an image. Every other pod mutation (resources, capabilities, volumes) requires logic and a sidecar or webhook is the right home for that. Sending image selection back to a webhook means every plugin author maintains their own TLS, cert rotation, and admission infrastructure for the simplest possible mutation.

## Test plan

- [ ] Verify VMI matching a Plugin condition gets the alternative launcher image
- [ ] Verify VMI without a match gets the default launcher image
- [ ] Verify alphabetical ordering when multiple Plugins match

### DIFF
--- a/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
+++ b/veps/sig-compute/190-kubevirt-structured-plugins/vep.md
@@ -40,6 +40,8 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
   - [Admission Policies and Webhooks](#admission-policies-and-webhooks)
     - [Admission options](#admission-options)
     - [API example:](#api-example)
+  - [Launcher Image Selection](#launcher-image-selection)
+    - [Launcher image API example](#launcher-image-api-example)
   - [Plugin metadata](#plugin-metadata)
 - [API Examples (full version)](#api-examples-full-version)
 - [Alternatives](#alternatives)
@@ -131,6 +133,8 @@ For example, this includes:
   - A niche specialized behavior is needed that does not make sense upstream.
   - Too complex / dangerous / for most use-cases.
   - No expertise in this subject among the community members (no one to maintain it).
+
+- As a KubeVirt plugin author delivering specialised hardware support (e.g. GPU passthrough requiring a virt-launcher built against an alternative libvirt or QEMU), I want to declare an alternative launcher image in my Plugin CR so that virt-controller automatically selects it for matching VMIs — without deploying and maintaining a separate MutatingWebhookConfiguration, TLS certificates, or webhook service.
 
 ## Repos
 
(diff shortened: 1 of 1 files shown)


michalskrivanek:
after going through the parent discussion I like it. I think it could be very useful, it still looks and works like a plugin, and requires only minimal changes on kubvirt side, and the end result is nice and straightforward for the user.

harshitgupta1337:
/cc
The proposal to specify custom virt-launcher image is useful - and something we are proposing in #311 as well. 

Have you considered the fact that since the `node-labeler` component uses the default `virt-launcher` image based on the operator's version, the capabilities advertised by node labeler might not match the ones that the custom launcher could actually serve??

Furthermore, is it a hard requirement to customize `virt-launcher` image based on VMI spec? Because we are already able to customize the launcher image cluster-wide using the `VIRT_LAUNCHER_IMAGE` env-var of virt-operator.

lyarwood:
> /cc The proposal to specify custom virt-launcher image is useful - and something we are proposing in #311 as well.

ACK thanks for reminding me of this @harshitgupta1337 I'll definitely review this ahead of the discussion next week!
 
> Have you considered the fact that since the `node-labeler` component uses the default `virt-launcher` image based on the operator's version, the capabilities advertised by node labeler might not match the ones that the custom launcher could actually serve??

Yes there are numerous foot guns here but my understanding is that the assumption with plugins is that the author knows best and will be responsible for documenting or mitigating these somehow. In my specific kubevirt-aie use case the capabilities should match but the runtime behaviour is vastly different.

> Furthermore, is it a hard requirement to customize `virt-launcher` image based on VMI spec? Because we are already able to customize the launcher image cluster-wide using the `VIRT_LAUNCHER_IMAGE` env-var of virt-operator.

That's the use case with kubevirt-aie that we are aiming to replace with this suggestion while dropping the need for a seperate webhook yes. We explicitly don't want to use the image for all VMIs in the cluster, only VMIs with specific GPUs attached.

PR #437, opened 2026-08-24
VEP #165: ContainerPath Volumes to beta in 1.10

mhenriks:
### VEP Metadata

**Tracking issue**: 

https://github.com/kubevirt/enhancements/issues/165

**SIG label**: 

/sig storage

* ContainerPath to beta in 1.10
* Replace references to GKE with Azure (GKE doesn't need ContainerPath)
* Update doc to be consistent with upstream user guide: https://kubevirt.io/user-guide/storage/containerpath_volumes/
* Update graduation requirements completed/required

### DIFF
--- a/veps/sig-storage/165-containerpath-volumes/vep.md
+++ b/veps/sig-storage/165-containerpath-volumes/vep.md
@@ -5,16 +5,16 @@
 ### Target releases
 
 - This VEP targets alpha for version: v1.8
-- This VEP targets beta for version:
+- This VEP targets beta for version: v1.10
 - This VEP targets GA for version:
 
 ### Release Signoff Checklist
 
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [ ] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
+- [x] (R) Enhancement issue created, which links to VEP dir in [kubevirt/enhancements] (not the initial VEP PR)
 - [x] (R) Alpha target version is explicitly mentioned and approved
-- [ ] (R) Beta target version is explicitly mentioned and approved
+- [x] (R) Beta target version is explicitly mentioned and approved
 - [ ] (R) GA target version is explicitly mentioned and approved
 
 ## Overview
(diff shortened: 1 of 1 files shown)


awels (approved):
looks good to me

veps/sig-storage/165-containerpath-volumes/vep.md
```diff
@@ -431,11 +459,9 @@ _To be filled in as implementation progresses_
   - Service account token access
   - Multiple containerPath volumes per VM
   - Dynamic file updates visible in VM (token rotation scenarios)
-- [ ] Migration support validated
-- [ ] Performance testing shows acceptable overhead
```
xpivarc, line 435: Why do we drop this?

xpivarc (commented):
Please update the tracker with Alpha items first, as of now it only reference the VEP PR but does not reference implementation. That is required to check if Alpha was fully implement.

veps/sig-storage/165-containerpath-volumes/vep.md
mhenriks: Honestly shouldn't have been added in the first place. This is really just a new api on already existing virtiofs functionality that is GA. If the performance was sufficient for k8s secrets/serviceaccount tokens it should be fine for this use case as well

mhenriks:
@xpivarc thanks added a link to the original implementation PR to the tracker

veps/sig-storage/165-containerpath-volumes/vep.md
xpivarc: Was it maybe because of the additional webhook? It could be fine to defer if time is only concern.

xpivarc:
FYI I left some comments https://github.com/kubevirt/kubevirt/pull/16662 , I would appreciate if we can address them

mhenriks:
> FYI I left some comments [kubevirt/kubevirt#16662](https://github.com/kubevirt/kubevirt/pull/16662) , I would appreciate if we can address them

@xpivarc responded to you comments, can address for beta impl

veps/sig-storage/165-containerpath-volumes/vep.md
mhenriks: Good point, let's defer to GA to get more data

xpivarc: Sounds good to me :)

xpivarc (commented):
/approve
/hold
Please remove hold once you track follow ups on comments

PR #438, opened 2026-08-25
vep-template: add core KubeVirt justification section

iholder101:
### What this PR does
New VEPs should explicitly justify why the feature belongs in core KubeVirt rather than an external controller, plugin, or sibling repo.

### DIFF
--- a/veps/NNNN-vep-template/vep.md
+++ b/veps/NNNN-vep-template/vep.md
@@ -86,6 +86,16 @@ Tangible API examples used for discussion
 Outline any alternative designs that have been considered)
 -->
 
+## Does it belong to core KubeVirt?
+
+<!--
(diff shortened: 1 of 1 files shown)


lyarwood (approved):
lgtm but I'd recommend waiting until we are past the upcoming VEP freeze, this could be something we enforce for new and updated VEPs in v1.11.

iholder101:
> lgtm but I'd recommend waiting until we are past the upcoming VEP freeze, this could be something we enforce for new and updated VEPs in v1.11.

Sounds good to me :+1: 
Holding until VEP freeze

/hold

xpivarc (commented):
/approve

xpivarc:
Great addition

iholder101:
Freeze is behind us!

/unhold

PR #439, opened 2026-08-25
VEP #143: Add reference for per-VMI GetVMStats requests

machadovilaca:
### VEP Metadata

Tracking issue: https://github.com/kubevirt/enhancements/issues/143
SIG label: sig-observability

### What this PR does

allow per-VMI GetVMStats requests because sometimes Linux VMs don't require/have available the same agent commands as Windows VMs


iholder101:
Sounds good. Thanks!

/approve

Issue #440, opened 2026-08-26
VEP 440: Prune CI Lanes for Unsupported Kubernetes Versions on Release Branches


EdDev:
**Primary contact (assignee)**:

/assign @EdDev

**Current Feature Stage**:

New

**Feature Gate**:

N/A — this is a CI tooling change, not a code feature.

**Responsible SIGs**:

Primary SIG:
/sig release

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/441

**Timeline**:

* [ ] Phase 1: VEP
  * [ ] VEP accepted by SIG Release
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/441
* [ ] Phase 2: Implementation
  * [ ] `remove jobs` extended with the release-branch pruning flag and per-suite exception logic
  * [ ] Unit tests added covering prune / exception / no-op cases
  * [ ] Validated against real job configs, see the PoC report in the VEP directory
* [ ] Phase 3: Wire to the scheduled jobs
  * [ ] `periodic-kubevirt-job-remover` wiring updated to pass the new flag
  * [ ] First run reviewed manually in `--dry-run=true` mode
  * [ ] Periodic job runs unattended with `--dry-run=false`
  * [ ] Finalized once the next new release branch is observed to confirm the keep-latest-variant exception leaves it with a running lane per suite, and single-provider suites are left untouched

> [!IMPORTANT]
> Please keep this description up to date.

PR #441, opened 2026-08-26
VEP #440: Prune CI Lanes for Unsupported Kubernetes Versions on Release Branches

EdDev:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/440
**SIG label**: /sig release

### What this PR does

Extends the existing `robots/kubevirt remove jobs` automation (which already prunes stale k8s lanes from `main`) to also prune CI presubmit lanes for Kubernetes minor versions that are no longer supported upstream, on `kubevirt/kubevirt` release branches.

Pruning is scoped per test suite: only suites with more than one k8s-version provider variant are touched, and at least one variant is always kept per suite. Suites that only ever run against a single k8s version in a given branch are left untouched entirely.

This is a companion to the already-merged VEP #233 (branch-level EOL), which retires a release branch's CI entirely once the KubeVirt release itself ages out. This VEP instead prunes individual k8s-version lanes within a branch that is still active.

Validated against real, current job configs — see the included PoC report (90 job definitions removed across 11 of 12 release branches, with single-provider suites correctly surviving untouched).

None.

### DIFF
--- a/veps/sig-release/440-prune-unsupported-k8s-lanes/poc-report.md
+++ b/veps/sig-release/440-prune-unsupported-k8s-lanes/poc-report.md
@@ -1,0 +1,87 @@
+# PoC Report: Prune CI Lanes for Unsupported Kubernetes Versions on Release Branches
+
+Companion PoC results for [vep.md](./vep.md). This ran the extended
+`robots/kubevirt remove jobs` (new `--job-config-path-kubevirt-presubmits-release-branches`
+flag) against a scratch copy of the real `kubevirt/project-infra` job configs, using
+the live `kubernetes/kubernetes` GitHub releases API (unauthenticated). Nothing in the
+real repository was modified — the run was performed against throwaway temp copies of
+`github/ci/prow-deploy/files/jobs/kubevirt/kubevirt/kubevirt-presubmits*.yaml` and
+`kubevirt-periodics.yaml`.
+
+Upstream-supported window resolved at run time to **k8s 1.34 / 1.35 / 1.36**.
+
+## Executive summary
+
+- **90 presubmit job definitions removed** across 11 of 12 release branches
+  (319 → 229 versioned lanes, a 28% reduction).
+- Pruning operates **per test suite**, not per branch: a "suite" is a group of jobs
+  that are identical except for the k8s version they target (e.g. `sig-compute` run
+  against 1.31/1.32/1.33 in one file). Only suites with more than one provider variant
+  are ever touched. If every variant of a multi-provider suite is unsupported, its
+  single most recent variant is kept so the suite is never left without any lane.
+  Suites that only ever run against one k8s version in a given branch are **left
+  untouched entirely**, regardless of how old that version is.
+- This matches what the real job configs actually look like: in **11 of 12**
+  branches, the only suites with more than one provider variant are the canonical
+  `sig-compute`, `sig-network`, `sig-operator`, `sig-storage` lanes. A handful of
+  branches (`0.58`, `0.59`, `1.0`, `1.1`) additionally have 1-2 "special" suites that
+  happen to also carry more than one provider variant (`sig-compute-migrations`,
+  `sig-compute-cgroupsv2`, `sig-storage-cgroupsv2`, `ipv6-sig-network`) and were
+  pruned the same way.
+- `release-1.9` had **zero removals** — every multi-provider suite in that branch is
+  already within the current 3-release Kubernetes support window (1.34/1.35/1.36).
+- `kubevirt-presubmits.yaml` (main) and `kubevirt-periodics.yaml` were untouched.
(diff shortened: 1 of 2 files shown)


akalenyu (approved):
I think this makes sense. Release branches get less care wrt to test fixes and maintenance in general, which is fine, so from my experience the additional k8s versions testing would in most cases make it harder to get PRs in the release branches.

Consumers of upstream kubevirt who would like to maintain the entire testing matrix can always set up downstream lanes (in practice, most of them will care about a single version coupling)

lyarwood (commented):
Quick suggestion otherwise lgtm!

This nicely complements [VEP #233](https://github.com/kubevirt/enhancements/blob/main/veps/sig-release/233-release-branch-eol-policy/vep.md) (and its rework in [#341](https://github.com/kubevirt/enhancements/pull/341)) — while #233/#341 bounds the branch lifecycle and shifts ownership to dedicated maintainer groups, this VEP solves the orthogonal problem of CI provider/lane bloat within still-active branches. Reusing `robots/kubevirt remove jobs`, preserving single-provider suites, and guaranteeing that multi-provider suites retain at least one running lane are all solid design choices.

The main piece I'd like to see covered is a **lightweight exception / lane-retention process** for vendors, downstream distributions, or maintainers who care about keeping a specific k8s minor version lane alive on a given release branch (especially for branches in extended maintenance).

Specifically:

1. **Governance / Request Process**:
   - What is the process if a downstream consumer needs an older k8s lane retained on an active release branch?
   - It doesn't need to be heavy: a post to `kubevirt-dev` or an issue with `/sig release` declaring intent, along with a commitment to:
     - Actively triage and fix failures in that lane.
     - Sponsor/fund the CI resources if infra capacity is constrained.

2. **Technical Opt-Out**:
   - Because `periodic-kubevirt-job-remover` runs unattended and generates PRs against `project-infra`, we should specify how an excepted lane is protected from being repeatedly re-pruned.
   - For example: an explicit job annotation/label in the job YAML (e.g. `kubevirt.io/preserve-lane: "true"` or `# prow-job-remover: keep`) or a simple exclude list in the remover configuration.

Adding a short section to the VEP describing this exception path would make the policy complete while still keeping automated pruning as the default behavior.

EdDev:
> The main piece I'd like to see covered is a **lightweight exception / lane-retention process** for vendors, downstream distributions, or maintainers who care about keeping a specific k8s minor version lane alive on a given release branch (especially for branches in extended maintenance).

In general, I am in fine with exceptions.
However, I am not sure I am in favor of reasoning about downstream project in relation to Kubevirt CI.
A D/S project should probably take care of its own CI process and depend as little as possible on the U/S one.
At the point where Kubernetes is no longer under support, we cannot give any guaranties about it and the CI is mainly focus on preserving the integration with that old version, nothing more.

> 
> Specifically:
> 
> 1. **Governance / Request Process**:
>    
>    * What is the process if a downstream consumer needs an older k8s lane retained on an active release branch?

It is a D/S challenge to resolve.
The maintainer may approach the SIG-CI for an exception, but I do not think this needs to be recorded officially.

>    * It doesn't need to be heavy: a post to `kubevirt-dev` or an issue with `/sig release` declaring intent, along with a commitment to:
>      
>      * Actively triage and fix failures in that lane.
>      * Sponsor/fund the CI resources if infra capacity is constrained.

It makes sense, however, I find this too theoretical at this moment and see no strong motivation for us to be so dynamic. We do not have capacity to handle so much tweaks and if someone does have a very strong need, they can try and convince the SIG Release and SIG CI to give the exception.
Perhaps we should wait to see if this is will ever surface.

> 2. **Technical Opt-Out**:
>    
>    * Because `periodic-kubevirt-job-remover` runs unattended and generates PRs against `project-infra`, we should specify how an excepted lane is protected from being repeatedly re-pruned.
>    * For example: an explicit job annotation/label in the job YAML (e.g. `kubevirt.io/preserve-lane: "true"` or `# prow-job-remover: keep`) or a simple exclude list in the remover configuration.

In case there is a single job/lane, there will be no removal.
I.e. if one decides that the latest Kubernetes version at that time is not the right version, it can be changed and everything will continue as usual.

> Adding a short section to the VEP describing this exception path would make the policy complete while still keeping automated pruning as the default behavior.

My suggestion is not to open this option formally, we have no incentive to give such exceptions.
However, nothing is mentioned in the VEP about blocking it.
Can we defer this to the first case this will be needed? Then we can evaluate our policy.
Said that, I will comply with the SIG decision on the matter and update the VEP accordingly.

lyarwood:
> My suggestion is not to open this option formally, we have no incentive to give such exceptions. However, nothing is mentioned in the VEP about blocking it. Can we defer this to the first case this will be needed? Then we can evaluate our policy. Said that, I will comply with the SIG decision on the matter and update the VEP accordingly.

ACK makes sense thanks @EdDev!

/lgtm

PR #446, opened 2026-08-31
VEP #256: Correct OCI artifact annotations to match implementation

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/256
**SIG label**: /sig storage

### What this PR does

Corrects the OCI artifact annotations specified in VEP #256 so they describe
the format that `virt-exportserver` actually produces, and resolves an
ambiguity found while designing the import side (VEP #395).

Three commits:

1. **Align documented annotations with implementation.** The three
   `io.kubevirt.pvc.*` layer annotations (`capacity`, `volumeMode`,
   `accessModes`) were never implemented; `computeDiskDescriptor` sets a
   single `io.kubevirt.disk.size` instead, carrying the uncompressed image
   size as a binary SI quantity. The index sets only
   `org.opencontainers.image.created`, not `title` or `description`, and the
   manifest sets no annotations at all.

2. **Make the disk name layer annotation multi-valued.** Layers are keyed by
   PersistentVolumeClaim, not by volume, so two volumes referencing the same
   claim share one layer and a singular annotation cannot describe it.
   `io.kubevirt.disk.name` becomes `io.kubevirt.disk.names`, a comma-separated
   lexically sorted list.

3. **Target Alpha in v1.10.**

Commit 1 is documentation catching up to code. Commit 2 is a design decision
that needs review.

`io.kubevirt.disk.name` keeps its documented meaning of "volume name from the
VM spec" throughout. The implementation currently emits the sanitized PVC name
instead, which is tracked as a bug in kubevirt/kubevirt#18984 - this PR
deliberately does not write that behavior into the spec.

Commit 2 resolves the open question raised at the end of that issue
(singular vs. multi-valued) in favour of multi-valued. Two consequences worth
a look:

- The plural key is a deliberate break. A consumer reading the singular
  `io.kubevirt.disk.name` would take `vol1,vol2` as a single volume name and
  mis-correlate silently; renaming makes it fail at lookup instead. This is
  only free because `OCIExport` is still Alpha and disabled by default.
- `org.opencontainers.image.title` cannot hold a list, so it uses the first
  name in the sorted order and is now documented as a filename rather than a
  correlation key.

The comma separator relies on volume names being DNS labels. Note that
`Volume.Name` carries no kubebuilder pattern in the API type - the constraint
comes from the name flowing into the virt-launcher pod spec, not from KubeVirt
validation.

### DIFF
--- a/veps/sig-compute/256-oci-export-proposal/vep.md
+++ b/veps/sig-compute/256-oci-export-proposal/vep.md
@@ -145,8 +145,6 @@ since a snapshot represents VM state.
     }
   ],
   "annotations": {
-    "org.opencontainers.image.title": "my-vm",
-    "org.opencontainers.image.description": "Fedora 42 Server VM",
     "org.opencontainers.image.created": "2026-04-09T12:00:00Z"
   }
 }
@@ -154,6 +152,10 @@ since a snapshot represents VM state.
 
 The `platform.architecture` field maps to `spec.template.spec.architecture`,
 defaulting to `amd64` when unset. The `platform.os` field is always `linux`.
+
+The index carries a single annotation, `org.opencontainers.image.created`, an
+RFC 3339 UTC timestamp recording when the export server prepared the artifact.
+The resource name and the rest of the object metadata live in the config blob.
 
 #### Image Manifest
 
@@ -176,9 +178,7 @@ Each manifest in the index describes one architecture variant:
       "size": 2147483648,
       "annotations": {
         "io.kubevirt.disk.name": "rootdisk",
-        "io.kubevirt.pvc.capacity": "10Gi",
-        "io.kubevirt.pvc.volumeMode": "Block",
-        "io.kubevirt.pvc.accessModes": "ReadWriteOnce",
+        "io.kubevirt.disk.size": "10Gi",
         "org.opencontainers.image.title": "rootdisk.raw.zst"
       }
     },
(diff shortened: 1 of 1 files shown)


veps/sig-compute/256-oci-export-proposal/vep.md
```diff
@@ -4,7 +4,7 @@
 
 ### Target releases
 
-- This VEP targets alpha for version: v1.9
+- This VEP targets alpha for version: v1.9, v1.10
```
Acedus, line 7: nit: Do we need to add a version if it already reached alpha in 1.9?

veps/sig-compute/256-oci-export-proposal/vep.md
Acedus, line 1: Isn't that implicitly true for Alpha, though? (Alpha features carry no stability guarantees)

Acedus (approved):
Two small comments, looks good aside from that.

veps/sig-compute/256-oci-export-proposal/vep.md
0xFelix: I think so, in order to keep track of it in the current cycle.

veps/sig-compute/256-oci-export-proposal/vep.md
0xFelix: Let's get rid of it then.

Issue #447, opened 2026-08-31
VEP 447: Runtime Domain Reconciliation for Structured Plugins


keinsword:
**Primary contact (assignee)**: @keinsword 

/assign @keinsword 

**Current Feature Stage**: New

**Feature Gate**: Plugins

**Responsible SIGs**:

Primary SIG:
/sig compute

Additional SIGs (optional):
N/A

**Enhancement link**:

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [x] VEP PRs: https://github.com/kubevirt/enhancements/pull/448
  * [ ] Code PRs:
  * [ ] Docs PRs:

> [!IMPORTANT]
> Please keep this description up to date.

PR #448, opened 2026-08-31
VEP 447: Runtime Domain Reconciliation for Structured Plugins

keinsword:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/447
**SIG label**: /sig compute

### What this PR does

Introduces a VEP extending VEP 190 structured plugins with runtime domain reconciliation.

The proposal allows an advanced domain-hook sidecar to receive both desired and live domain XML while a VMI is running. The plugin can reconcile external resources and return desired XML containing plugin-owned vhost-user disks. KubeVirt validates the runtime delta and remains solely responsible for libvirt attach and detach operations.

The initial use case is hotplugging vhost-user block devices backed by endpoint-oriented storage such as NBS, without requiring a PVC, DataVolume, KubeVirt volume source, or hotplug attachment pod.

The proposal also defines:

- Stable alias-based ownership for plugin-managed disks.
- Idempotent endpoint preparation and orphan cleanup.
- Attach, asynchronous detach, restart, and failure semantics.
- Live migration for converged disk sets, while serializing migration and hotplug.
- An intentionally narrow alpha scope limited to advanced hooks and vhost-user disks.

This is a follow-up to VEP 190 rather than a replacement.

The proposal intentionally reuses the existing `MutateDomain` RPC by adding a `Runtime` invocation context and optional live-domain input.

Areas where early feedback would be particularly useful:

- Whether provider-resource reconciliation is appropriate inside an advanced domain hook.
- The alias-based device ownership and validation boundary.
- Failure behavior when the runtime hook is unavailable.
- The proposed live migration convergence and freeze requirements.

### DIFF
--- a/veps/sig-compute/447-structured-plugin-runtime-device-lifecycle/vep.md
+++ b/veps/sig-compute/447-structured-plugin-runtime-device-lifecycle/vep.md
@@ -1,0 +1,512 @@
+# VEP #447: Runtime Domain Reconciliation for Structured Plugins
+
+Owners:
+
+- TBD
+
+## VEP Status Metadata
+
+### Target releases
+
+- This VEP targets alpha for version:
+- This VEP targets beta for version:
+- This VEP targets GA for version:
+
+### Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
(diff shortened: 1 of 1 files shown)


keinsword:
/cc @iholder101 

Would love to get your review on this one 🙏

PR #459, opened 2026-09-09
VEP #256: Revert the disk name layer annotation to singular

0xFelix:
### VEP Metadata

**Tracking issue**: https://github.com/kubevirt/enhancements/issues/256
**SIG label**: /sig compute

### What this PR does

Reverts the disk name layer annotation to the singular `io.kubevirt.disk.name`, in VEP #256 and in VEP #395 which reads it.

#446 made it multi-valued because layers are keyed by PersistentVolumeClaim, so several volumes referencing one claim would share a layer and a single name could not describe it. sig-storage has since decided to reject such exports instead, kubevirt/kubevirt#19057 implements it. Every layer belongs to exactly one volume again.

Also carries the layer selector of VEP #395 becoming a `matchAnnotations` map, which was written before the revert and touches the same section.

Three commits, one VEP each. The selector change comes first, the annotation revert applies on top of it.

The implementation never followed #446, `pkg/storage/oci` has always emitted the singular annotation, and kubevirt/virt-import reads it singular too. So this restores agreement between the VEPs and the code rather than creating a new divergence. kubevirt/kubevirt#19066 makes the value the volume name from the VM as VEP #256 specifies.

`OCIExport` is Alpha and disabled by default, the same reason #446 gave for renaming the annotation in the first place.

### DIFF
--- a/veps/sig-compute/256-oci-export-proposal/vep.md
+++ b/veps/sig-compute/256-oci-export-proposal/vep.md
@@ -177,7 +177,7 @@ Each manifest in the index describes one architecture variant:
       "digest": "sha256:111...",
       "size": 2147483648,
       "annotations": {
-        "io.kubevirt.disk.names": "rootdisk",
+        "io.kubevirt.disk.name": "rootdisk",
         "io.kubevirt.disk.size": "10Gi",
         "org.opencontainers.image.title": "rootdisk.raw.zst"
       }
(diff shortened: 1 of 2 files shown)


0xFelix:
PTAL @Acedus @akalenyu

Acedus (approved):

Issue #457, opened 2026-09-09
VEP: Mutable Virtual Machine Display Name


sradco:
**Primary contact (assignee)**:

/assign @sradco

**Current Feature Stage**:

New

**Feature Gate**:

VMDisplayName (proposed)

**Responsible SIGs**:

Primary SIG:
/sig compute

**Enhancement link**:

https://github.com/kubevirt/enhancements/pull/461

**Timeline**:

* Alpha release target (x.y):
* Beta release target (x.y):
* Stable release target (x.y):

* [ ] Alpha
  * [ ] VEP PRs: https://github.com/kubevirt/enhancements/pull/461
  * [ ] Code PRs:
  * [ ] Docs PRs:

**Additional context**:

Propose an optional, mutable, non-unique `VirtualMachine.spec.displayName` so users can set a human-readable inventory name (including spaces, punctuation, and Unicode that labels cannot store) without changing `metadata.name`, UID, or associated resources.

No schema default or backfill from `metadata.name`. Presentation fallback is client-side (`displayName` else `metadata.name`). Exact discovery uses `--field-selector` on the **stored** field only.

This is the Kubernetes-compatible alternative to object rename. KubeVirt previously removed `virtctl rename` because renaming Kubernetes objects contradicts core Kubernetes identity (kubevirt/kubevirt#5564). A well-known label remains the preferred alternative if SIG-compute decides label-safe names are enough.

> [!IMPORTANT]
> Please keep this description up to date.

PR #461, opened 2026-09-09
VEP-457: Mutable Virtual Machine Display Name

sradco:
Propose an optional VM spec.displayName for
inventory names that Kubernetes labels cannot
store, without renaming objects.
Tracker: kubevirt/enhancements#457.

Signed-off-by: Shirly Radco <sradco@redhat.com>
Co-authored-by: AI Assistant <noreply@cursor.com>

Made with [Cursor](https://cursor.com)


veps/sig-compute/457-vm-display-name/vep.md
```diff
@@ -0,0 +1,321 @@
...
+
+Administrators may recognize a VM by a descriptive inventory name that differs from its provisioning identifier. Imported names can also contain spaces or non-ASCII characters. Preserving a recognizable name helps users identify the workload without changing the resource identity used by automation.
+
+KubeVirt already rejected renaming the object (`metadata.name`) as incompatible with Kubernetes identity (see [kubevirt/kubevirt#5564](https://github.com/kubevirt/kubevirt/pull/5564)). The remaining problem is storing presentation text on the existing VM without mutating identity or restarting the workload. Prior object-rename history is recorded under [Prior art](#prior-art-object-rename); it motivates preserving identity, not a particular storage mechanism.
+
+## Goals
+
+- Allow setting, changing, and clearing a human-readable VM inventory name—including spaces, punctuation, and Unicode that labels cannot store—on running and stopped VMs without a rename-induced restart or VMI recreation.
+- Allow duplicate display names, including within one namespace, while keeping `metadata.name` and `metadata.uid` as the operational identity.
```
mtessun: Not sure I understand this correctly, but having the same display name within a namespace might be confusing. So could we get this that the name is unique within a namespace? If this is not possible, then I think this is a restriction not a goal.

veps/sig-compute/457-vm-display-name/vep.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+A pointer is required so JSON Merge Patch `{"spec":{"displayName":null}}` clears the field. An unrelated merge patch that omits the key leaves it unchanged. Full-object replacement that omits the field removes it. Server-side apply follows normal field ownership. There is no schema default, admission default, or backfill.
+
+### Validation
+
+- **Presence**: Optional. Omission leaves the field absent (`nil`).
+- **Empty string**: Invalid when the field is present. Length is 1 to 253 Unicode code points.
+- **Characters**: CRD CEL plus admission reject leading/trailing whitespace, line breaks, and control characters. Printable Unicode letters, numbers, internal spaces, and punctuation are allowed. Matching is exact and case-sensitive.
+- **Uniqueness**: Not required, including within one namespace.
```
mtessun, line 112: I am not sure about this. I think it should be unique within one namespace - to avoid confusion.

veps/sig-compute/457-vm-display-name/vep.md
```diff
@@ -0,0 +1,346 @@
...
+
+Kubernetes has **no** native annotation selector. `kubectl get vm` cannot filter on an annotation without fetching the list and filtering client-side (or adding a kubectl feature this VEP does not propose). There is also no single validated contract: each product can pick a different key. An annotation is the right alternative only if native CLI discovery is dropped and the name is console-only.
+
+### Other options
+
+| Alternative | Trade-off |
+| --- | --- |
+| Unique display names | Addresses lookup, but needs a naming controller for a descriptive attribute. Rejected; duplicates remain an end-state API freeze. |
+| Separate VMInfo CR | Selectable data, but another API, controller, RBAC, and stale mapping. |
+| `virtctl`-only alias lookup | Improves one CLI; does not store human text on the VM API or help `kubectl get vm`. |
```
fabiand, line 247: I think the `alias` idea is not virtctl specific, the idea is:

1. Set a label `myalias=true`  on the VM
2. Use `kubectl get vm -l myalias`  or `virtctl stop -l myalias` to work with virtctl and kubectl.

However, this approach still requires kube limitations on the name of the label

veps/sig-compute/457-vm-display-name/vep.md
```diff
@@ -0,0 +1,346 @@
...
+
+**Caveat:** If SIG-compute decides that label-safe names are acceptable, stop this spec-field design and document a well-known label instead.
+
+### Annotation
+
+An annotation can store spaces and Unicode (large string, few charset limits). Consoles could read `kubevirt.io/display-name` today.
+
+Kubernetes has **no** native annotation selector. `kubectl get vm` cannot filter on an annotation without fetching the list and filtering client-side (or adding a kubectl feature this VEP does not propose). There is also no single validated contract: each product can pick a different key. An annotation is the right alternative only if native CLI discovery is dropped and the name is console-only.
+
+### Other options
```
fabiand, line 241: To me this section is missing a comparison with just using a displayName annotation or displayName/alias label.

To me it is a must to discuss this, because these are obvious options.

veps/sig-compute/457-vm-display-name/vep.md
```diff
@@ -0,0 +1,346 @@
...
+A pointer is required so JSON Merge Patch `{"spec":{"displayName":null}}` clears the field. An unrelated merge patch that omits the key leaves it unchanged. Full-object replacement that omits the field removes it. Server-side apply follows normal field ownership. There is no schema default, admission default, or backfill.
+
+### Validation
+
+- **Presence**: Optional. Omission leaves the field absent (`nil`).
+- **Empty string**: Invalid when the field is present. Length is 1 to 253 Unicode code points.
+- **Characters**: CRD CEL plus admission reject leading/trailing whitespace, line breaks, and control characters. Printable Unicode letters, numbers, internal spaces, and punctuation are allowed. Matching is exact and case-sensitive.
+- **Uniqueness**: Not required, including within one namespace.
+
+Accepting non-unique names is a permanent API decision. Uniqueness cannot be added later without breaking existing clusters.
```
mtessun, line 114: That's why we should carefully take that decision. Even with UI help, the customer/admin needs to assure to not have duplicates in one namespace as they might lead to confusion.

veps/sig-compute/457-vm-display-name/vep.md
sradco: +1. I will move to restrictions.
Uniqueness would introduce complexities we do not want to get into.
UI can do a best effort check before it creates/updated this field, but in CLI/gitops I would keep that up to the user. We can have an alert if vms share the same display name

veps/sig-compute/457-vm-display-name/vep.md
sradco: same as above, Uniqueness would introduce complexities we do not want to get into.
UI can do a best effort check before it creates/updated this field, but in CLI/gitops I would keep that up to the user. We can have an alert if vms share the same display name

veps/sig-compute/457-vm-display-name/vep.md
sradco: I agree this has to be an explicit, permanent choice.

We are accepting duplicates because a display name is descriptive text (imports, spaces/Unicode), not a second unique key. A uniqueness controller is out of scope and could not be added later without breaking clusters that already share a name.

Confusion is the real risk. The contract is:

Search by display name returns the list of matching VMs (zero, one, or many).
Start/stop/delete and other operations always use metadata.name + namespace, or UID.
UIs should show the object name (and UID for destructive actions) next to the display name, and must not silently take the first match.
Using display name as a unique selector for automation or policy is not supported. Listing by display name is fine. Binding a single mutating operation to it is not.

veps/sig-compute/457-vm-display-name/vep.md
sradco: Updated

PR #468, opened 2026-09-14
CodeRabbit config for VEP auto-review

iholder101:
### What this PR does

Set up configuration for [documentation-driven](https://github.com/kubevirt/community/pull/444) CodeRabbit reviews for VEPs.


iholder101:
/cc @EdDev @vladikr 

@vladikr FYI that "custom_checks" should provide a review in a table-form that will reflect the different checks separately.

.coderabbit.yaml
```diff
@@ -3,17 +3,105 @@ language: en-US
...
+          Return PASS, FAIL,
+          or INCONCLUSIVE with concise evidence and a follow-up action; do not
+          claim completeness without evidence.
+
   tools:
     trufflehog:
       enabled: true
 
 knowledge_base:
-  opt_out: true
```
EdDev, line 13: This implies that learning from comments is enabled, which we considered problematic in the past.

.coderabbit.yaml
```diff
@@ -3,17 +3,105 @@ language: en-US
 reviews:
   profile: chill
   request_changes_workflow: false
+  review_details: true
   auto_review:
     enabled: false
+  path_filters:
```
EdDev: We did this a bit different on k/kubevirt[1].
How about we start with something minimalist as that one, before we try new things?

[1] https://github.com/kubevirt/kubevirt/blob/main/hack/reviewcfg/.coderabbit.yaml#L10

iholder101: Yup, makes sense. Removed it :+1: 
These paths are already defined under `knowledge_base.code_guidelines.filePatterns` where it fits better

.coderabbit.yaml
iholder101: You're right.

I removed `opt_out: true` because guideline files are excluded when opt-out is enabled and I wanted `design_guidelines.md` loaded reliably.

I added repository-local scope and a 30-day approval delay as mitigation. It means that the learning hints wait for 30 days before they become part of the knowledge base, giving the admin a chance to reject them. I've reached @dhiller offline to try and understand who has access to the admin dashboard.

Unfortunately I'm not sure we can get around that while still referencing `design_guidelines.md` which is extremely important for reviews. WDYT?

docs/feature-lifecycle.md
```diff
@@ -137,6 +137,26 @@ All VEP updates and implementation work are tied to the [KubeVirt release cycle]
...
+
+A VEP may declare only one incomplete graduation target at a time. The target
+must be the current release cycle or the immediately upcoming release. A VEP
+must not pre-declare a sequence of future graduations, such as Alpha in one
+release, Beta in the next release, and GA in a later release.
+
+Before each graduation phase, the author must submit a VEP update during the
+design phase for that cycle. The update proposes the next stage and release,
+giving maintainers an opportunity to discuss whether the feature is ready for
+graduation. For example, a feature that reached Alpha in v1.8 may later use a
```
0xFelix, line 150: In this specific case, what about redeclaring Alpha in v1.9, for better context provided to the reader?

iholder101: What do you mean by "redeclaring"? Are you thinking about an "alpha2" kind of thing? If so, I think it still qualifies as there'll be a single version target added to the VEP PR.

But maybe you refer to the following:
* A feature is targeted as alpha for v1.7.
* The feature wasn't implemented in v1.7, say because of unconverging discussions that started at the implementation PR over something that wasn't anticipated in the design review.
* Then, until v1.8 the discussions are resolved, and it is finally being implemented in v1.8.
* In v1.9 the author wants to target Beta but also update the alpha version to reflect reality retroactively, so they also update alpha version to become v1.8 instead of v1.7.

Not sure this is what you meant, but it is a corner case.
Do you think we should throw another sentence to clarify this in the doc?

0xFelix: I meant, if it would be okay add the current development cycle version to Alpha like this:

`This VEP targets alpha for version: v1.8, v1.9`

